PR #5, opened 2017-01-24
Remove memory leak from epicsEnvSet() on systems that support setenv()

hdante:
Example valgrind log showing memory leak, fixed by this patch:

    ==11103== 22 bytes in 1 blocks are definitely lost in loss record 39 of 745
    ==11103==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    ==11103==    by 0x555886E: mallocMustSucceed (cantProceed.c:43)
    ==11103==    by 0x5562082: epicsEnvSet (osdEnv.c:39)
    ==11103==    by 0x52F4347: iocshRegisterCommon (iocshRegisterCommon.c:40)
    ==11103==    by 0x409462: Registration (ScalerLinux_registerRecordDeviceDriver.cpp:296)
    ==11103==    by 0x409462: __static_initialization_and_destruction_0 (ScalerLinux_registerRecordDeviceDriver.cpp:301)
    ==11103==    by 0x409462: _GLOBAL__sub_I_ScalerLinux_registerRecordDeviceDriver (ScalerLinux_registerRecordDeviceDriver.cpp:301)
    ==11103==    by 0x41123C: __libc_csu_init (in /home/ABTLUS/henrique.almeida/code/ScalerLinux/bin/linux-x86_64/ScalerLinux)
    ==11103==    by 0x617A7BE: (below main) (libc-start.c:247)

### DIFF
--- a/src/libCom/osi/os/default/osdEnv.c
+++ b/src/libCom/osi/os/default/osdEnv.c
@@ -34,25 +34,28 @@
  */
 epicsShareFunc void epicsShareAPI epicsEnvSet (const char *name, const char *value)
 {
-    char *cp;
-
+#if defined(_BSD_SOURCE) || (_POSIX_C_SOURCE >= 200112L) || (_XOPEN_SOURCE >= 600)
     iocshEnvClear(name);
-    
-	cp = mallocMustSucceed (strlen (name) + strlen (value) + 2, "epicsEnvSet");
-	strcpy (cp, name);
-	strcat (cp, "=");
-	strcat (cp, value);
-	if (putenv (cp) < 0) {
-		errPrintf(
-                -1L,
-                __FILE__,
-                __LINE__,
-                "Failed to set environment parameter \"%s\" to \"%s\": %s\n",
-                name,
-                value,
-                strerror (errno));
+    setenv (name, value, 1);
+#else
+    char *cp;
+    iocshEnvClear(name);
+    cp = mallocMustSucceed (strlen (name) + strlen (value) + 2, "epicsEnvSet");
+    strcpy (cp, name);
+    strcat (cp, "=");
+    strcat (cp, value);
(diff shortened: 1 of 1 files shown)


mdavidsaver:
I see no problem with this change.  However, it should be proposed at https://code.launchpad.net/epics-base  This repository is only a mirror.

ralphlange:
Well,
* testing for a BSD and POSIX macros in the `.../os/default` directory that contains implementations for all systems (including non BSD and non-posix) strikes me as somewhat counter-intuitive. The main idea of the `.../os/<OS_CLASS>` mechanism is to avoid `#ifdef`s.
(It does work as neither `.../os/posix` nor `.../os/freebsd` implement `osdEnv.c`.)
* has this change been tested on Windows (that also seems to use the default implementation)?

anjohnson:
I'm slightly concerned that this would link the low-level `epicsEnvSet()` routine to the higher-level iocsh subsystem, since every program that calls `epicsEnvSet()` would now have to link in the iocsh.cpp object file. This isn't a problem on workstation OSs that use shared libraries, but for statically-linked targets and some embedded systems it might be (VxWorks has its own osdEnv.c but RTEMS does not).

I would prefer that the call to `iocshEnvClear()` be put in the `epicsEnvSetCallFunc()` routine in src/libCom/iocsh/libComRegister.c and maybe also in the `iocshLoad()` function in iocsh.cpp — it might not remove quite as many memory leaks, but it won't add unnecessary links between different libCom subsystems.

Note that environment variables can be a common source of memory leaks and even race conditions, and the solutions aren't always portable. See [this blog entry](http://www.club.cc.cmu.edu/~cmccabe/blog_the_setenv_fiasco.html) for some of the issues.

PR #6, opened 2017-11-29
perl 5.8 compatibility fix

bhill-slac:
Many of our host machines are running perl 5.8.8 and throw an error for the possesive quantifiers in this regex.
Support for the ++ and *+ possesive quantifiers wasn't added till perl 5.10 and the epics docs still say perl 5.8.1 is the requirement.
They don't appear to be needed in this regex and removing them generated the same errSymTbl.c in my tests.


anjohnson:
Merged manually, thanks Bruce!

PR #7, opened 2017-11-30
Fixed .gitignore for git versions prior to 1.8.2

bhill-slac:
Changed **/O.* in submodule .gitignore files to O.*/ as it works for git versions before 1.8.2 as well as newer versions.


bhill-slac:
Not sure if I did this right.   Trying to avoid creating a pull request for each submodule.

anjohnson:
Hi Bruce, I don't think there *is* a right way to do this now with submodules, unfortunately. I currently have no way of seeing the changes you want in the submodules (the PR only gives the new sha of your local commits to those modules, but the commits themselves currently only exist on your machine), but I'm pretty sure they just involve replacing `**/O.*` with `O.*/` in the .gitignore files, which I will do. Please let me know if there was anything else.

bhill-slac:
That's the only change I made.
Looks like one pull request per submodule is needed in the future.
Thanks!
- Bruce

anjohnson:
Committing equivalent changes by another route.

PR #8, opened 2017-12-07
Debugging travis-prepare.sh

anjohnson:
I'm trying to work out why Travis isn't creating the `modules/RELEASE.$EPICS_HOST_ARCH.local` file in its `$HOME/.source/epics-base` tree.

This pull request is for triggering Travis builds only and will not be merged.

### DIFF
--- a/.ci/travis-prepare.sh
+++ b/.ci/travis-prepare.sh
@@ -35,7 +35,7 @@ fi
 install -d "$HOME/configure"
 cd "$HOME/configure"
 
-cat << EOF > RULES_USER
+cat <<'EOF' > RULES_USER
 SHOW_MAKEFILES = $(MAKEFILE_LIST:%=show-makefile.%)
 show-makefiles: $(SHOW_MAKEFILES)
 
--- a/.ci/travis-prepare.sh
+++ b/.ci/travis-prepare.sh
@@ -32,6 +32,24 @@ then
   fi
 fi
 
+install -d "$HOME/configure"
+cd "$HOME/configure"
+
+cat << EOF > RULES_USER
+SHOW_MAKEFILES = $(MAKEFILE_LIST:%=show-makefile.%)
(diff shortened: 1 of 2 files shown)


ralphlange:
Debugging Travis builds is frustrating - welcome to the club.
I would really like to try doing it locally in an original Travis-CI Docker container that you can log into. That should be more efficient.

anjohnson:
I created a local directory tree and bash script to try and duplicate the build environment (including setting HOME) which helped, but I didn't think to unset EPICS_HOST_ARCH, which turned out to be the issue — the submodules need it to include the right `../RELEASE.<host>.local` file which sets EPICS_BASE for them (which they would need to be able to set it for themselves; they could try harder and look for `$(TOP)/../../startup/EpicsHostArch.pl` but I don't like that solution much).

The fix I committed last night just exports the EPICS_HOST_ARCH variable from the core build so it's available for the submodules, which works but I'd like to go further. I'm considering changing all the `.ci/travis-prepare.sh` and `.ci/travis-build.sh` scripts to make it an environment variable there as well. I also want to add a check in the submodule CONFIG files to error out when it isn't set for them (because it's really hard to track down the problem).

We still aren't installing the setup directory when INSTALL_LOCATION is set. I found my implementation of that the other day, I will propose it for the next release.

Sorry, I'll end my rambling comments now…

ralphlange:
Re-doing the Travis-CI scripts is actually on my to-do-list, I have a few basic requirements somewhere on a piece of paper.
The main goal is to minimize the countless repetition of code in the scripts, by adding a ci module that contains all the "meaty" parts, and only leave stubs in the code modules that check out the ci module and call the scripts in there.
One of the general rules that @mdavidsaver put on that list: always use explicit arguments and don't convey stuff through the environment. Just wanted to mention this before you take off and do things through the environment.

PR #10, opened 2018-02-28
Bump NUM_TIME_EVENTS to 599 to allow more unused event numbers for TSE timing.

bhill-slac:
Allows Timing Pattern Receiver (TPR) channels to map to event numbers:
 TPR 0: 300-399, TPR 1: 400-499, TPR 2: 500-599


bhill-slac:
Hmmm!   This pull request is based against 3.16.   Not sure how it ended up as a pull request for core/master.

PR #11, opened 2018-02-28
Bump NUM_TIME_EVENTS to 599 to allow more unused event numbers for TSE timing

bhill-slac:
Allows Timing Pattern Receiver (TPR) channels to map to event numbers:
 TPR 0: 300-399, TPR 1: 400-499, TPR 2: 500-599

### DIFF
--- a/src/ioc/db/dbScan.c
+++ b/src/ioc/db/dbScan.c
@@ -25,6 +25,7 @@
 #include "dbDefs.h"
 #include "ellLib.h"
 #include "epicsEvent.h"
+#include "epicsGeneralTime.h"
 #include "epicsMutex.h"
 #include "epicsPrint.h"
 #include "epicsRingBytes.h"
@@ -112,7 +113,7 @@ typedef struct event_list {
     struct event_list *next;
     char                event_name[MAX_STRING_SIZE];
 } event_list;
-static event_list * volatile pevent_list[256];
+static event_list * volatile pevent_list[NUM_TIME_EVENTS];
 static epicsMutexId event_lock;
 
 /* IO_EVENT*/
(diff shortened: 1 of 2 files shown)


mdavidsaver:
@bhill-slac Please be aware that PR's aren't accepted.  This repo is a read-only mirror of https://code.launchpad.net/epics-base and changes need to be proposed through the launchpad site.  I can help with this if you're having problems doing so.

anjohnson:
Note that (as described in epicsGeneralTime.h) the largest legal time event number is `NUM_TIME_EVENTS-1` so with this change time event number 599 would be illegal and attempts to use it would cause either an error or a buffer overflow. You probably meant to make it 600.

However doing this will cause every future IOC anywhere to reserve an additional `(600-256)*8=2752` bytes for timestamps that will currently only be used at SLAC. I accept that currently most IOCs are allocating `256*8=2048` bytes that they never use because very few IOCs actually have a hardware event receiver, but that wasn't my design and I know we can do better.

It might be nice to see an API added by which an event receiver driver would request some range of event numbers from the generalTime code, and only then does the latter allocate storage for the timestamps. I'd be open to discussion about whether the request should specify the range or it just gets given one in the response, which affects whether you would know in advance what the actual numbers are going to be. Having generalTime allocate the numbers would be more conducive to using time event names in TSE, but would obviously mean more work and may be somewhat excessive.

As a simpler alternative, how about you make NUM_TIME_EVENTS and the `gtPvt.eventTime` array in epicsGeneralTime.c configurable at initialization time. Add an initialization routine that can be called by an event time provider _before_ it registers its event provider routine which sets the table size (which should default to 256 if not that routine is not called). That routine can't itself do the memory allocation since existing drivers won't call it, so the array should only be allocated when the first event provider gets registered, and `generalTimeGetEventPriority()` would need protection in case no event providers exist.

@mdavidsaver as owner of the MRF driver would you have any objections to something like this suggestion?

bhill-slac:
Closing this pull request in favor of a simpler alternative.   Please see:
https://code.launchpad.net/~bhill/epics-base/+git/epics-base/+merge/341067

PR #12, opened 2018-02-28
Replace some strcpy calls w/ strncpy.

bhill-slac:


ralphlange:
Please keep in mind that - while the GitHub mirror is often faster and more convenient to use - the upstream repository of Base is at LaunchPad, and pull requests will have to be re-done there as merge proposals.

bhill-slac:
Should I redo these as launchpad merge proposals?

ralphlange:
If I remember correctly, Andrew has been applying some very simple PRs by just doing the change upstream. But for anything else, and in general, merge proposals are the way to get a proper review and have things appear on the radar for inclusion in upcoming releases.

ralphlange:
I will add a sentence on the title page of the mirror to make it more clear.

anjohnson:
We might want to cherry-pick some of Bruce's changes onto an earlier branch anyway so it's probably a waste of time redoing them all as merge requests on Launchpad.

For this particular change, I note that line 95 of stringinRecord.c has
```C
    STATIC_ASSERT(sizeof(prec->oval)==sizeof(prec->val));
```
so it is already guaranteed that OVAL must be the same size as VAL and using `strncpy()` is unnecessary. I haven't checked the stringoutRecord.c file but I would be surprised if it doesn't have the same assertion.

anjohnson:
s/is/should be/. I guess if someone already overwrote the trailing nil byte this would prevent the error propagating, so I don't really have an objection to it.

bhill-slac:
I'm out on vacation the rest of this week.
Michael showed me the git URL for launchpad so I can create
merge requests for any of these that are still pending next week.

Cheers,
- Bruce

anjohnson:
To continue my previous discussion, we have in the past allowed all 40 characters of a `DBF_STRING` field to be used (with no trailing nil byte), so device support may fill up the VAL field. This change therefore protects the IOC in this (rare) situation. I am applying it manually to 3.15, and also replacing the other `strcpy()` in the files. Neither the SVAL nor IVOV fields hare that size check though, so I'm adding them too.

I will now close this PR, but since I haven't actually cherry-picked the commit it won't be marked as merged by Github, although I have effectively done so.

PS: `return` is a statement, not a function call, and it shouldn't be made to look like one by putting the return value in paren's. I've been fixing these in the EPICS code-base for years…

PR #13, opened 2018-02-28
Rework module dependency conflict error message for easier debugging.

bhill-slac:
Now a simply grep for asyn in the build log will show you line by
line which modules depend on which asyn versions.

### DIFF
--- a/src/tools/convertRelease.pl
+++ b/src/tools/convertRelease.pl
@@ -230,11 +230,11 @@ sub checkRelease {
             if (exists $macros{$parent} &&
                 AbsPath($macros{$parent}) ne AbsPath($ppath)) {
                 print "\n" unless ($status);
-                print "Definition of $parent conflicts with $app support.\n";
-                print "In this application a RELEASE file defines\n";
-                print "\t$parent = $macros{$parent}\n";
-                print "but $app at $path defines\n";
-                print "\t$parent = $ppath\n";
+                print "Definition of $parent conflict between this app and $app.\n";
+                print "In this application a RELEASE file\n";
+                print "conflicts with $app at $path\n";
+                print "  This App: $parent = $macros{$parent}\n";
+                print "  $app: $parent = $ppath\n";
                 $status = 1;
             }
         }


anjohnson:
Which are the important line(s) that you want to be able to grep for? I might want to adjust the messages so it would be good to know exactly what information you're wanting to see.

bhill-slac:
Here's what an ADCORE conflict looks like w/o this patch:
% egrep ADCORE make.log.old.convertRelease
Definition of ADCORE conflicts with TIMESTAMPFIFO support.
	ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.1.0
	ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.0.0
Definition of ADCORE_MODULE_VERSION conflicts with TIMESTAMPFIFO support.
	ADCORE_MODULE_VERSION = R3.1-2.1.0
	ADCORE_MODULE_VERSION = R3.1-2.0.0
Definition of ADCORE conflicts with ADSIMDETECTOR support.
	ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.1.0
	ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.0.0
Definition of ADCORE_MODULE_VERSION conflicts with ADSIMDETECTOR support.
	ADCORE_MODULE_VERSION = R3.1-2.1.0
	ADCORE_MODULE_VERSION = R3.1-2.0.0

The error msg shows that we have 2 different versions of ADCORE, but it's
not readily apparent which module uses R3.1-2.1.0 and which uses R3.1-2.0.0.

After this patch is applied, the error output looks like this:
% egrep ADCORE make.log.new.convertRelease
Definition of ADCORE conflict between this app and TIMESTAMPFIFO.
  This App: ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.1.0
  TIMESTAMPFIFO: ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.0.0
Definition of ADCORE_MODULE_VERSION conflict between this app and TIMESTAMPFIFO.
  This App: ADCORE_MODULE_VERSION = R3.1-2.1.0
  TIMESTAMPFIFO: ADCORE_MODULE_VERSION = R3.1-2.0.0
Definition of ADCORE conflict between this app and ADSIMDETECTOR.
  This App: ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.1.0
  ADSIMDETECTOR: ADCORE = /reg/g/pcds/epics/R3.15.5-2.0/modules/ADCore/R3.1-2.0.0
Definition of ADCORE_MODULE_VERSION conflict between this app and ADSIMDETECTOR.
  This App: ADCORE_MODULE_VERSION = R3.1-2.1.0
  ADSIMDETECTOR: ADCORE_MODULE_VERSION = R3.1-2.0.0

The main difference is pre-pending the MACRO definition w/ the name of the module which contains that value of the MACRO, or "This App" if it's the current ioc or module being built.

ralphlange:
Makes sense.
I would use "module" instead of "app" to minimize confusion.
(Yes, I know it's called "makeBaseApp". Still, I think we consistently call the result a "module". At least we should.)

bhill-slac:
"module" isn't always appropriate as these conflicts often arise when building ioc applications.
I don't think there's a good way for convertRelease.pl to determine if you're building a module or an ioc app.

ralphlange:
In my experience, the common terminology is that any compilation unit following the standard EPICS structure is called "module", with the differentiation happening between an "IOC application module" and a "support module".

bhill-slac:
I find many instances of EPICS modules which build/install executable applications under bin/T_A, but I've never seen an IOC application get used as a component module of another IOC application.
Granted there's nothing in the EPICS build system to preclude that, but if we're talking about minimizing confusion our IOC engineers might be confused to find their IOC referred to as a module.

bhill-slac:
Closed.  Please see:
https://code.launchpad.net/~bhill/epics-base/+git/epics-base/+merge/341086

ralphlange:
Real-life example of a setup I have worked with for a long time:

- Module A builds the IOC binaries, DBD files, and the generic parts of the st.cmd file. (In an embedded TOP, test IOCs are built that use the artifacts from the outer module.)
- Module B has the stream protocol file for a specific device, together with DB templates and matching generic display files. (Again, test IOC in an embedded TOP.)
- Module C creates an IOC instance by using the binary and DBD from A. It has an *App directory for every App that is running on the IOC. The App for the specific device uses the protocol file and template DB from B. It only contains the substitution file and two st.cmd snippets (1 line each) for configuring the device support and loading the database.

Note also that in this setup the SLAC concept and term of an "IOC engineer" does not make sense. Every IOC is shared between the Apps (*App directories like VacuumApp, PsApp, TimingApp) that it runs, usually between 2 and 5. Engineer responsibility is per App, not per IOC.

In my experience, any assumption about how things are used or done at a different place is usually wrong.

mdavidsaver:
> Closed. Please see:
> https://code.launchpad.net/~bhill/epics-base/+git/epics-base/+merge/341086

Can the LP request be closed as well?  (aka. change Status to Rejected?)

mdavidsaver:
> Can the LP request be closed as well? (aka. change Status to Rejected?)

Ignore this.  I was confused and didn't notice that the discussion here happened first.

PR #14, opened 2018-02-28
Fixed camonitor server side relative timestamps bug

bhill-slac:
tsFirst needs to get initialized from the first server side
timestamp instead of the client side tsNow.


bhill-slac:
Closed.  Please see
https://code.launchpad.net/~bhill/epics-base/+git/epics-base/+merge/341085

PR #17, opened 2018-02-28
Applied epics-thread-show-linux-pids.patch so we can see linux pid in…

bhill-slac:
… epicsThreadShow()


anjohnson:
Doesn't MCoreUtils already provide this? It's still on SourceForge which is down right now so I can't get to it to confirm…

ralphlange:
MCoreUtils in fact does provide this, and has been on GitHub since April 20, 2015.
https://github.com/epics-modules/MCoreUtils

anjohnson:
Thanks Ralph, I need to fix my links in the base pages of the website. Should we reject this PR?

anjohnson:
Actually I was wrong about pointing to MCoreUtils, the problem with this PR is that it's not modifying the libCom/osi/os/Linux/osdThreadExtra.c source file, which contains the version of `epicsThreadShowInfo()`  that Linux IOCs run. If there's another column needed in the `epicsThreadShowAll()` output it needs to be added to the Linux osdThreadExtra.c file, not the Posix one (and so shouldn't need the conditional code at all). However the output on Linux already contains a `LWP ID` column, and the ID of the `_main_` thread below corresponds to the pid of my IOC:
```
epics> epicsThreadShowAll
          _main_      0x21ad330     5506      0       0       OK
          errlog      0x22491a0     5509     10       0       OK
          taskwd      0x22495f0     5510     10       0       OK
      timerQueue      0x2249f30     5511     70       0       OK
```
Thus this PR is superfluous.

ralphlange:
You're right!
I remember I added this when doing the expert contract work at ITER in 2012/13, which - after discussing it with you - partly ended up in MCoreUtils and partly in Base.

Actually, I wanted to check tomorrow why this was not ending up in the Base part (where it belongs) - nice to realize we did the right thing back then.

PR #18, opened 2018-02-28
Fix seg fault when empty string is passed to caget or cainfo

bhill-slac:
Found this via a scripting error.  
From the cmd line you need to use quotes to pass in an empty string.


anjohnson:
I believe this has already been fixed, for both the 3.15 and 7.0 branches I get:
```
Hex$ caget ""
CA error Invalid string occurred while trying to create channel ''.
```

ralphlange:
Andrew fixed it, in a different way, and also on 27-Apr-2017 (!!) on the 3.14 branch. The fix was released as part of 3.14.12.7:
https://git.launchpad.net/epics-base/diff/?h=3.14&id=f2e54be965d14c62d879971807ee691f3a3c0c5d

PR #19, opened 2018-02-28
Sort DIRECTORY_TARGETS to avoid errors if 2 FILE_TYPEs share INSTALL dir

bhill-slac:
Also happens if 2 modules try to add support for the same FILE_TYPE.


anjohnson:
The sort happens too early in this fix since other build files (including the Makefile) can add to the `DIRECTORY_TARGETS` variable (and their directories won't get created at all with this change). I am putting a sort in the actual build rule instead, which saves creating the extra variable and fixes the original problem.

PR #20, opened 2018-02-28
Add diagnostic support for showing record values when TPRO is >= 2

bhill-slac:


anjohnson:
Not sure that I like overloading TPRO like this; Eric Norum would probably call it a Camel's Nose issue.

Does anyone else think this is a good idea?

ralphlange:
I know at least three Device Supports where TPRO > 1 is used to switch on debugging (TPRO being verbosity) for specific records. Which is very handy.

We have to be careful with defining things that were undefined for decades. This deserves at least a call on Tech Talk to warn people, and maybe a definition that allows both record support and device support to share the field.

bhill-slac:
I don't see a conflict here.   We're already sharing TPRO between record support and several device support modules, 
with larger values of TPRO enabling more verbose diagnostic output.

Note that this change does not affect how TPRO propagates. i.e.   Propagated TPRO via plockSet->trace is still 1, so to 
get more verbose diagnostics for a particular record you have to set TPRO >= 2 for that record.

bhill-slac:
Closing pull request.   Please see:
https://code.launchpad.net/~bhill/epics-base/+git/epics-base/+merge/341081

PR #21, opened 2018-09-23
Add epicsGetExecName()

mdavidsaver:
Let's try an experiment?  This code isn't ready to be merged (via a LP merge request), but is in need of some feedback.  So this PR is intended to assess interest, and track development progress (including CI results).

The branch adds epicsGetExecName() which is an OSI abstraction on the many OS specific mechanisms to find an (perhaps not "the") absolute path to the executable file from which the running process was loaded.

The goal of this is to enable runtime loading of file with paths relative to the executable.  My idea is to reduce the need to compile ```$(INSTALL_LOCATION)``` into executables (softIoc being my first target).  This is part of a larger plan to allow a built Base tree to be moved (relocated) without modification.

There are ports for WIN32, Linux, Darwin, solaris, and freebsd.  My primary concern is Linux, and to a lesser extent WIN32 where relocatable trees are possible.  For me the others are nice to have, but could be dropped for lack of interest in making them work.

Current state

* The Darwin, solaris, and freebsd implementation hasn't even been compile tested.
* The Darwin implementation doesn't follow symlinks.
* The WIN32 implementation has only been tested with mingw+wine
* Need manual testing of situations involving symlinks.
* Uniform header file (unixFileName.h vs copy+paste in osiFileName.h)?

### DIFF
--- a/modules/libcom/src/misc/unixFileName.h
+++ b/modules/libcom/src/misc/unixFileName.h
@@ -14,7 +14,29 @@
 #ifndef unixFileNameH
 #define unixFileNameH
 
+#include <shareLib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #define OSI_PATH_LIST_SEPARATOR ":"
 #define OSI_PATH_SEPARATOR "/"
+
+/** Return the absolute path of the current executable.
+ @returns NULL or the path.  Caller must free()
+ */
+epicsShareFunc
+char *epicsGetExecName(void);
+
+/** Return the absolute path of the directory containing the current executable.
+ @returns NULL or the path.  Caller must free()
+ */
+epicsShareFunc
+char *epicsGetExecDir(void);
+
+#ifdef __cplusplus
+}
+#endif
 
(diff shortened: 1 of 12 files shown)


anjohnson:
This approach looks reasonable to me. I'm not really sure what "not following symlinks" really means, so some testing is in order. Maybe the test program should attempt to open the path it was given to check that it resolves to an actual file, and maybe check the `epicsGetExecDir()` functionality by creating a file in the CWD and opening it through that path.

How to handle the embedded targets though?

mdavidsaver:
> I'm not really sure what "not following symlinks"

Meaning that `epicsGetExecName()` might for some targets return a path to a symlink instead instead of following it through to an actual file.  I'm not sure if this presents any practical problem.  I raise it more as a matter of consistency.

> How to handle the embedded targets though?

I don't think anything meaningful can be done with these targets.  So the default behavior of returning NULL seems reasonable to me.

mdavidsaver:
Moved to https://code.launchpad.net/~epics-core/epics-base/+git/make/+merge/359132

PR #24, opened 2019-03-09
Thread priorities rework

till-s:
I added recent modifications by Heinz Junkes to my original patch (in essence: remove the 'isEpicsThread' member as proposed earlier).

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -272,19 +272,25 @@ int           low, try;
      * by resource limitations (but that is ignored
      * by sched_get_priority_max() [linux]).
      */
-    low = min;
 
-    while ( low < max ) {
+    low = min + 1;
+
+    /*
+     * At this point, low-1 is known to be a good
+	 * priority. Keep that a loop invariant
+     */
+
+    while ( low <= max ) {
         try = (max+low)/2;
         if ( try_pri(try, prm->policy) ) {
-            max = try;
+            max = try - 1;
         } else {
             low = try + 1;
         }
     }
 
     prm->min_pri = min;
-    prm->max_pri = try_pri(max, prm->policy) ? max-1 : max;
+    prm->max_pri = low - 1;
     prm->ok = 1;
 
     return 0;
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -737,15 +737,19 @@ epicsShareFunc void epicsShareAPI epicsThreadSetPriority(epicsThreadId pthreadIn
 #endif /* _POSIX_THREAD_PRIORITY_SCHEDULING */
 }
 
-epicsShareFunc epicsThreadBooleanStatus epicsShareAPI epicsThreadHighestPriorityLevelBelow(
-    unsigned int priority, unsigned *pPriorityJustBelow)
+/* Step the OSI-priority in the coarser-grained domain: if there are more
+ * OSI than OSS levels then step in the OSS domain otherwise in the OSI domain.
+ */
+static epicsThreadBooleanStatus
+coarserGrainedPriorityStep(unsigned priority, int step, unsigned *pNewPriority)
 {
-    unsigned newPriority = priority - 1;
+    unsigned newPriority = priority + step;
+
 #if defined(_POSIX_THREAD_PRIORITY_SCHEDULING) && _POSIX_THREAD_PRIORITY_SCHEDULING > 0
-    if ( pcommonAttr->valid ) {
+    if ( pcommonAttr->valid && step != 0 ) {
         int newOss;
         if ( epicsThreadPriorityMax - epicsThreadPriorityMin > pcommonAttr->maxPriority - pcommonAttr->minPriority ) {
-            newOss = osi2posixPriority( priority ) - 1;
+            newOss = osi2posixPriority( priority ) + step;
             if ( newOss < pcommonAttr->minPriority ) {
                 return epicsThreadBooleanStatusFail;
             }
@@ -757,48 +761,37 @@ epicsShareFunc epicsThreadBooleanStatus epicsShareAPI epicsThreadHighestPriority
             newOss = osi2posixPriority( newPriority );
         }
         newPriority = posix2osiPriority( newOss );
-        if ( newPriority >= priority ) {
-            return epicsThreadBooleanStatusFail;
-        }
+		if ( step < 0 ) {
+			if ( newPriority >= priority ) {
+            	return epicsThreadBooleanStatusFail;
+			}
+        } else {
+			if ( newPriority <= priority ) {
+            	return epicsThreadBooleanStatusFail;
+			}
+		}
     }
 #endif /* _POSIX_THREAD_PRIORITY_SCHEDULING */
-    /* test against Max necessary because priorities are unsigned; this discovers wrap-around */
+    /* test against Max necessary (even for negative steps) because priorities are unsigned;
+	 * this discovers wrap-around
+	 */
     if ( newPriority <= epicsThreadPriorityMax && newPriority >= epicsThreadPriorityMin ) {
-        *pPriorityJustBelow = newPriority;
+        *pNewPriority = newPriority;
         return epicsThreadBooleanStatusSuccess;
     }
     return epicsThreadBooleanStatusFail;
 }
 
+epicsShareFunc epicsThreadBooleanStatus epicsShareAPI epicsThreadHighestPriorityLevelBelow(
+    unsigned int priority, unsigned *pPriorityJustBelow)
+{
+	return coarserGrainedPriorityStep( priority, -1, pPriorityJustBelow );
+}
+
 epicsShareFunc epicsThreadBooleanStatus epicsShareAPI epicsThreadLowestPriorityLevelAbove(
     unsigned int priority, unsigned *pPriorityJustAbove)
 {
-    unsigned newPriority = priority + 1;
-#if defined(_POSIX_THREAD_PRIORITY_SCHEDULING) && _POSIX_THREAD_PRIORITY_SCHEDULING > 0
-    if ( pcommonAttr->valid ) {
-        int newOss;
-        if ( epicsThreadPriorityMax - epicsThreadPriorityMin > pcommonAttr->maxPriority - pcommonAttr->minPriority ) {
-            newOss = osi2posixPriority( priority ) + 1;
-            if ( newOss > pcommonAttr->maxPriority ) {
-                return epicsThreadBooleanStatusFail;
-            }
-        } else {
-            if ( newPriority > epicsThreadPriorityMax ) {
-                return epicsThreadBooleanStatusFail;
-            }
-            newOss = osi2posixPriority( newPriority );
-        }
-        newPriority = posix2osiPriority( newOss );
-        if ( newPriority <= priority ) {
-            return epicsThreadBooleanStatusFail;
-        }
-    }
-#endif /* _POSIX_THREAD_PRIORITY_SCHEDULING */
-    if ( newPriority <= epicsThreadPriorityMax ) {
-        *pPriorityJustAbove = newPriority;
-        return epicsThreadBooleanStatusSuccess;
-    }
-    return epicsThreadBooleanStatusFail;
+	return coarserGrainedPriorityStep( priority, +1, pPriorityJustAbove );
 }
 
 epicsShareFunc int epicsShareAPI epicsThreadIsEqual(epicsThreadId p1, epicsThreadId p2)
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -129,7 +129,7 @@ epicsShareFunc int epicsThreadGetPosixPriority(epicsThreadId pthreadInfo)
 #endif /* _POSIX_THREAD_PRIORITY_SCHEDULING */
 }
 
-static void setSchedulingPolicy(epicsThreadOSD *pthreadInfo,int policy)
+static void setSchedulingPolicy(pthread_attr_t *pattr, epicsThreadOSD *pthreadInfo,int policy)
 {
 #if defined(_POSIX_THREAD_PRIORITY_SCHEDULING) && _POSIX_THREAD_PRIORITY_SCHEDULING > 0
     int status;
@@ -137,18 +137,18 @@ static void setSchedulingPolicy(epicsThreadOSD *pthreadInfo,int policy)
     if(!pcommonAttr->usePolicy) return;
 
     status = pthread_attr_getschedparam(
-        &pthreadInfo->attr,&pthreadInfo->schedParam);
+        pattr,&pthreadInfo->schedParam);
     checkStatusOnce(status,"pthread_attr_getschedparam");
     pthreadInfo->schedParam.sched_priority = epicsThreadGetPosixPriority(pthreadInfo);
     pthreadInfo->schedPolicy = policy;
     status = pthread_attr_setschedpolicy(
-        &pthreadInfo->attr,policy);
+        pattr,policy);
     checkStatusOnce(status,"pthread_attr_setschedpolicy");
     status = pthread_attr_setschedparam(
-        &pthreadInfo->attr,&pthreadInfo->schedParam);
+        pattr,&pthreadInfo->schedParam);
     checkStatusOnce(status,"pthread_attr_setschedparam");
     status = pthread_attr_setinheritsched(
-        &pthreadInfo->attr,PTHREAD_EXPLICIT_SCHED);
+        pattr,PTHREAD_EXPLICIT_SCHED);
     checkStatusOnce(status,"pthread_attr_setinheritsched");
 #endif /* _POSIX_THREAD_PRIORITY_SCHEDULING */
 }
@@ -170,32 +170,44 @@ static epicsThreadOSD * create_threadInfo(const char *name)
     return pthreadInfo;
 }
 
+static int attr_init(pthread_attr_t *pattr, size_t stackSize)
+{
+int status;
+
+    status = pthread_attr_init(pattr);
+    checkStatusOnce(status,"pthread_attr_init");
+    if(status) return status;
+    status = pthread_attr_setdetachstate(pattr, PTHREAD_CREATE_DETACHED);
+    checkStatusOnce(status,"pthread_attr_setdetachstate");
+#if defined (_POSIX_THREAD_ATTR_STACKSIZE)
+#if ! defined (OSITHREAD_USE_DEFAULT_STACK)
+    status = pthread_attr_setstacksize(pattr, stackSize);
+    checkStatusOnce(status,"pthread_attr_setstacksize");
+#endif /*OSITHREAD_USE_DEFAULT_STACK*/
+#endif /*_POSIX_THREAD_ATTR_STACKSIZE*/
+    status = pthread_attr_setscope(pattr,PTHREAD_SCOPE_PROCESS);
+    if(errVerbose) checkStatusOnce(status,"pthread_attr_setscope");
+	return status;
+}
+
+static void attr_destroy(pthread_attr_t *pattr)
+{
+int status;
+    status = pthread_attr_destroy(pattr);
+    checkStatusQuit(status,"pthread_attr_destroy","free_threadInfo");
+}
+
 static epicsThreadOSD * init_threadInfo(const char *name,
-    unsigned int priority, unsigned int stackSize,
+    unsigned int priority,
     EPICSTHREADFUNC funptr,void *parm)
 {
     epicsThreadOSD *pthreadInfo;
-    int status;
 
     pthreadInfo = create_threadInfo(name);
     if(!pthreadInfo)
         return NULL;
     pthreadInfo->createFunc = funptr;
     pthreadInfo->createArg = parm;
-    status = pthread_attr_init(&pthreadInfo->attr);
-    checkStatusOnce(status,"pthread_attr_init");
-    if(status) return 0;
-    status = pthread_attr_setdetachstate(
-        &pthreadInfo->attr, PTHREAD_CREATE_DETACHED);
-    checkStatusOnce(status,"pthread_attr_setdetachstate");
-#if defined (_POSIX_THREAD_ATTR_STACKSIZE)
-#if ! defined (OSITHREAD_USE_DEFAULT_STACK)
-    status = pthread_attr_setstacksize( &pthreadInfo->attr,(size_t)stackSize);
-    checkStatusOnce(status,"pthread_attr_setstacksize");
-#endif /*OSITHREAD_USE_DEFAULT_STACK*/
-#endif /*_POSIX_THREAD_ATTR_STACKSIZE*/
-    status = pthread_attr_setscope(&pthreadInfo->attr,PTHREAD_SCOPE_PROCESS);
-    if(errVerbose) checkStatusOnce(status,"pthread_attr_setscope");
     pthreadInfo->osiPriority = priority;
     return(pthreadInfo);
 }
@@ -210,8 +222,6 @@ static void free_threadInfo(epicsThreadOSD *pthreadInfo)
     status = pthread_mutex_unlock(&listLock);
     checkStatusQuit(status,"pthread_mutex_unlock","free_threadInfo");
     epicsEventDestroy(pthreadInfo->suspendEvent);
-    status = pthread_attr_destroy(&pthreadInfo->attr);
-    checkStatusQuit(status,"pthread_attr_destroy","free_threadInfo");
     free(pthreadInfo);
 }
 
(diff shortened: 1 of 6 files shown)


modules/libcom/test/Makefile
```diff
@@ -288,4 +288,10 @@ TESTPROD_HOST += cvtFastPerform
 cvtFastPerform_SRCS += cvtFastPerform.cpp
 testHarness_SRCS += cvtFastPerform.cpp
 
+TESTPROD_Linux += nonEpicsThreadPriorityTest
```
mdavidsaver: Some changes are needed to get the test to build and run correctly.

From the [Travis CI results](https://travis-ci.org/epics-base/epics-base/builds/503942069?utm_source=github_status&utm_medium=notification).  The RTEMS and cross-mingw builds are failing because `TESTS` is unconditional, but the executable isn't bult.

Unless @anjohnson has another idea, the best I can think of is to wrap the whole source file in an `#if` where the `#else` is a short `MAIN()` which calls `testSkip()`.

For that matter, is there a reason to limit this test to Linux?  Seems like it should be run on all pthread/POSIX targets.

Also nonEpicsThreadPriorityTest.cpp needs to be add to`testHarness_SRCS` or the RTEMS test harness for libCom will fail to link.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -111,44 +112,91 @@ if(status) { \
...
+    if ( oss < pcommonAttr->minPriority )
+        oss = pcommonAttr->minPriority;
+    if ( oss > pcommonAttr->maxPriority )
+        oss = pcommonAttr->maxPriority;
+    return oss;
+}
+
+static unsigned posix2osiPriority(int posixPriority)
+{
+double   slope, osid;
```
mdavidsaver: I don't get too annoyed about whitespace, except for indentation.  Please always indent the entire function body 4 spaces.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -111,44 +112,91 @@ if(status) { \
...
+unsigned osi;
+
+    if ( ! pcommonAttr->valid )
+        return epicsThreadPriorityMedium;
+
+    if ( pcommonAttr->maxPriority == pcommonAttr->minPriority )
+        return epicsThreadPriorityMedium;
+
+    slope =   (double)(epicsThreadPriorityMax   - epicsThreadPriorityMin)
+            / (double)(pcommonAttr->maxPriority - pcommonAttr->minPriority);
```
mdavidsaver, line 158: I've almost convinced myself that there is no chance that `valid` is true and `maxPriority==minPriority`.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -111,44 +112,91 @@ if(status) { \
...
+    return osi;
+}
+#endif
+
+

+epicsShareFunc int epicsThreadGetPosixPriority(epicsThreadId pthreadInfo)
+{
+#if defined(_POSIX_THREAD_PRIORITY_SCHEDULING) && _POSIX_THREAD_PRIORITY_SCHEDULING > 0
+    if (pcommonAttr->schedPolicy != pthreadInfo->schedPolicy)
```
mdavidsaver: Maybe a silly question, but why cache schedPolicy?  Is it so expensive to test each time.  it's at least possible for it to change.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -667,12 +726,28 @@ epicsShareFunc epicsThreadBooleanStatus epicsShareAPI epicsThreadHighestPriority
 {
     unsigned newPriority = priority - 1;
 #if defined(_POSIX_THREAD_PRIORITY_SCHEDULING) && _POSIX_THREAD_PRIORITY_SCHEDULING > 0
-    int diff;
-    diff = pcommonAttr->maxPriority - pcommonAttr->minPriority;
-    if(diff<0) diff = -diff;
-    if(diff>1 && diff <100) newPriority -=  100/(diff+1);
+    if ( pcommonAttr->valid ) {
+        int newOss;
+        if ( epicsThreadPriorityMax - epicsThreadPriorityMin > pcommonAttr->maxPriority - pcommonAttr->minPriority ) {
```
mdavidsaver, line 751: I'm confused by this condition.  Could this be simplified by working with `priority` as a real number here, as well as in `osi2posixPriority()` where rounding is already happening?

mdavidsaver (changes requested):
The new test case needs some fixing to build (or not) correctly.

Also.  I notice that some of the computed OSS priorities are slightly different with this change.  Looks like a rounding change.  Not sure if this is significant.

before

```
epics> epicsThreadShowAll
          _main_ 0x55db1e6f8f40      787      0       0       OK
          errlog 0x55db1e7a14a0      790     10      10       OK
          taskwd 0x55db1e7a1930      791     10      10       OK
      timerQueue 0x55db1e7a2290      792     70      69       OK
           cbLow 0x55db1e7a6450      793     59      58       OK
        cbMedium 0x55db1e7aa4d0      794     64      63       OK
          cbHigh 0x55db1e7ae680      795     71      70       OK
        dbCaLink 0x55db1e7aeb10      796     50      50       OK
        scanOnce 0x55db1e7b5ac0      797     67      66       OK
         scan-10 0x55db1e7b5d40      798     60      59       OK
          scan-5 0x55db1e7b5f90      799     61      60       OK
          scan-2 0x55db1e7b61e0      800     62      61       OK
          scan-1 0x55db1e7b6430      801     63      62       OK
        scan-0.5 0x55db1e7b6680      802     64      63       OK
        scan-0.2 0x55db1e7b68d0      803     65      64       OK
        scan-0.1 0x55db1e7b6b20      804     66      65       OK
         CAS-TCP 0x55db1e7bf4d0      805     16      16       OK
         CAS-UDP 0x55db1e7bf720      806     12      12       OK
      CAS-beacon 0x55db1e7bf970      807     14      14       OK
```

after

```
epics> epicsThreadShowAll
          _main_ 0x55c6540d0f40    11289      0       0       OK
          errlog 0x55c654177f00    11292     10      11       OK
          taskwd 0x55c6541787a0    11293     10      11       OK
      timerQueue 0x55c6541790a0    11294     70      70       OK
           cbLow 0x55c65417d210    11295     59      59       OK
        cbMedium 0x55c654181240    11296     64      64       OK
          cbHigh 0x55c6541853a0    11297     71      71       OK
        dbCaLink 0x55c6541857e0    11298     50      50       OK
        scanOnce 0x55c65418c750    11299     67      67       OK
         scan-10 0x55c65418c980    11300     60      60       OK
          scan-5 0x55c65418cb80    11301     61      61       OK
          scan-2 0x55c65418cd80    11302     62      62       OK
          scan-1 0x55c65418cf80    11303     63      63       OK
        scan-0.5 0x55c65418d180    11304     64      64       OK
        scan-0.2 0x55c65418d380    11305     65      65       OK
        scan-0.1 0x55c65418d580    11306     66      66       OK
         CAS-TCP 0x55c654195ee0    11307     18      19       OK
         CAS-UDP 0x55c6541960e0    11308     16      17       OK
      CAS-beacon 0x55c6541962e0    11309     17      18       OK
```

till-s:
Regarding the change of the computed OSS priorities: that's actually a bugfix. The old computation was:
`oss = (maxOss - minOss)/100 * osi + minOss
`
which is wrong. For the max. osi prioritiy (99) the computed oss priority was
`  99/100 maxOssPri + 1/100 minOssPri`
now it should be correct.

modules/libcom/src/osi/os/posix/osdThread.c
till-s: The way I understand epicsThreadLowestPriorityLevelAbove/Below() is that it should find the next priority level that is possible on the particular implementation. Rounding must be avoided (otherwise the result may be the same priority). The intention here is to take a one-sized step in the coarser domain (osi vs. oss). I.e., if there are  more OSI priority levels than OSS levels then we map the current OSI -> OSS, take one step in the OSS domain and map back. If OSS is finer grained than OSI then we can just take one OSI step.

A brute-force version could be

````
currentOSS = osi2PosixPriority( priority );
 do {
  if ( ++priority > epicsThreadPriorityMax)
    return epicsThreadBooleanStatusFail;
} while ( osi2PosixPriority( priority ) == currentOss );
  
 *pPriority = priority;
return epicsThreadBooleanStatusSuccess;
````

File: modules/libcom/src/osi/os/posix/osdThread.c (hunk 4 above)
Comment (till-s) [reply modules/libcom/src/osi/os/posix/osdThread.c]: True. But that requires ugly coding using pthread as well as epicsThread APIs at the same time.  I'm not opposed to eliminating the cached version at all - but I have no idea how expensive the test is and how often this routine is called (from base code).

File: modules/libcom/src/osi/os/posix/osdThread.c (hunk 2 above)
Comment (till-s) [reply modules/libcom/src/osi/os/posix/osdThread.c]: Ok, done.

File: modules/libcom/src/osi/os/posix/osdThread.c (hunk 3 above)
Comment (till-s) [reply modules/libcom/src/osi/os/posix/osdThread.c:158]: That's not known. The sysadmin can set a ceiling for the RT-priority the process may acquire (RLIMIT_RTPRIO). This is also why the available range of OSS priorities is computed dynamically ('findPriorityRange'). It is therefore possible that the process is allowed to use SCHED_FIFO but it may only use a single priority level. This is why this test is needed.

File: modules/libcom/test/Makefile (hunk 1 above)
Comment (anjohnson) [reply modules/libcom/test/Makefile]: The name `nonEpicsThreadPriorityTest` doesn't make it obvious that this test code is for Posix architectures only so a name-change would be nice. The suggestions below have not been tested.

This conditional in the `Makefile` should be sufficient to make it build only on Posix targets:
```makefile
ifeq ($(POSIX),YES)
TESTPROD_HOST += nonEpicsThreadPriorityTest
testHarness_SRCS += nonEpicsThreadPriorityTest.cpp
epicsRunLibComTests_CFLAGS += -DPOSIX_TARGET
TESTS += nonEpicsThreadPriorityTest
endif
```

The `epicsRunLibComTests.c` also needs a couple of `#ifdef` wrappers so it only runs the `nonEpicsThreadPriorityTest` routine on Posix targets, e.g:
```c
#ifdef POSIX_TARGET
int nonEpicsThreadPriorityTest(void);
#endif
// etc.
```
There might be other/better macros that could be used instead of defining our own here, since `epicsRunLibComTests.c` already includes `epicsThread.h` and hence `osdThread.h`.

till-s: @Andrew: I'll be happy to rename, suggestions welcome. The current name tries to express that the test exercises non-epics Threads (i.e., threads that were generated directly via the pthread API) that are 'promoted' to epics-threads.

How about `epicsThreadFromPthreadPriorityTest`?

modules/libcom/src/osi/os/posix/osdThread.c
mdavidsaver: Maybe the possibility of divide by zero needs to be handled then?

modules/libcom/src/osi/os/posix/osdThread.c
mdavidsaver: Ok, you've convinced me that this isn't a practical concern.

modules/libcom/src/osi/os/posix/osdThread.c
mdavidsaver: This block definitely needs a comment stating the "coarser domain" intent.  It makes sense to me when you describe it in those terms.

Also, seeing the added complexity in common between these two functions.  Would it be better to factor out this common code as a single private helper function into which is passed either +1 or -1 ?

modules/libcom/src/osi/os/posix/osdThread.c
till-s: Hmm  - I think it's pretty obvious that this *is* actually handled.

````
if ( a == b ) {
  return something;
}
/* a-b cannot be zero here */
y = x/(a-b);
````

File: modules/libcom/src/osi/os/posix/osdThread.c (hunk 5 above)
Comment (till-s) [reply modules/libcom/src/osi/os/posix/osdThread.c:751]: I'll look into that.

File: modules/libcom/src/osi/os/posix/osdThread.c (hunk 3 above)
Comment (mdavidsaver) [reply modules/libcom/src/osi/os/posix/osdThread.c:158]: Ok, my mistake.  I was focused too much on the logic in `once()`.

File: modules/libcom/test/Makefile (hunk 1 above)
Comment (till-s) [reply modules/libcom/test/Makefile]: Re. POSIX_TARGET macro alternative: I had tried the _POSIX_THREAD_PRIORITY_SCHEDULING macro which is available via osdThread.h on posix targets -- unfortunately that doesn't work for RTEMS because RTEMS does indeed define that macro (and support pthreads) but EPICS does not use the pthread API. It seems therefore necessary to use some epics-specific device (such as the one you suggested).

Comment (anjohnson) [reply modules/libcom/test/Makefile]: RTEMS 5.x defines that macro, but until you have Heinz' changes to support RTEMS-5 (which have not yet been merged either) the libCom routines use the old RTEMS APIs (and I didn't think they worked on RTEMS-5 any more; was I misled?). Thus this PR needs to assume that RTEMS will *not* be one of the targets running the test, but that will change when Heinz' stuff also gets merged. The RTEMS 4.9 and 4.10 builds that Travis does will never run that test.

**Comment by mdavidsaver** [comment]
I like the way this cleanup is going, and want to see it completed and merged.  In the context of an imminent [7.0.2.1 patch release](https://epics.anl.gov/core-talk/2019/msg00119.php) it is more extensive than I'd be comfortable including.  Is it feasible and worthwhile creating a simpler/smaller fix to only the [original problem](https://bugs.launchpad.net/epics-base/+bug/1816841)?

File: modules/libcom/test/Makefile (hunk 1 above)
Comment (till-s) [reply modules/libcom/test/Makefile]: Unfortunately, testing for the 'POSIX' Makefile variable is not viable. RTEMS says POSIX=YES but still does not use the pthread API. So we need to figure out a way of finding platforms which use os/posix/osdThread.c -- not that easy with the current build system since any platform is free to override with it's own os/xyz/osdThread.c.

I'm thinking - maybe we should just scrap this test. It exercises a fringe-feature of posix/osdThread (the possibility to 'promote' a native thread to an 'epicsThread')
and I'm starting to spend way more time trying to find a way to *prevent* the test
from being built/executed than on anything else related to this patch. Plus, I'm pretty sure
travis doesn't run with the sufficient privilege to engage the RT scheduler, so what's the point...

File: modules/libcom/src/osi/os/posix/osdThread.c (hunk 5 above)
Comment (till-s) [reply modules/libcom/src/osi/os/posix/osdThread.c:751]: Done

**Comment by mdavidsaver** [comment]
As mentioned in https://github.com/epics-base/epics-base/pull/25#issuecomment-473740910 I'm bother by the assumption that `sched!=SCHED_FIFO` implies priority zero.  This is demonstrably not true for `SCHED_RR`.  And should not be necessary for `SCHED_OTHER`, which always has OSS priority zero.

The core issue seems to me the use of only `SCHED_FIFO` when computing the priority mapping.  For me (Linux 4.19) `sched_get_priority_min(SCHED_FIFO)==1` and `sched_get_priority_max(SCHED_FIFO)==99`.  `SCHED_RR` also gives the `[1, 99]` range.  `SCHED_OTHER` gives `[0, 0]`.

http://man7.org/linux/man-pages/man7/sched.7.html

says:

```
Conceptually, the scheduler maintains a list of runnable threads for
each possible sched_priority value.  In order to determine which
thread runs next, the scheduler looks for the nonempty list with the
highest static priority and selects the thread at the head of this list.
```

Which I read as a clear statement that there is only one range of priorities.  eg. that priority 1 has the same meaning, and relative weight, with `SCHED_FIFO` or  `SCHED_RR`.

Surprisingly to me, I find a similar, though less clear, statement in POSIX-2001 http://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_08.html (see 2.8.4).

It seems to me that a more complete mapping would consider priority independently.  Map using the full support range of priorities (`[0, 99]` on Linux), then select `SCHED_*` based on desired priority.

The question would then becomes how to portable determine the full range of priorities and/or all allowed `SCHED_*`?

mdavidsaver:
It also seems to me that `main()` for IOCs should set `epicsThreadPriorityIocsh` on all targets.  Maybe this could be done in `iocsh(NULL)` to exclude things which aren't IOCs?

hjunkes:
I have now adopted the "new" osdThread.c, but get this in travis:
/usr/bin/gcc  -D_GNU_SOURCE -D_DEFAULT_SOURCE           -D_X86_64_  -DUNIX  -Dlinux     -O3 -g   -Wall      -mtune=generic      -m64 -fPIC -I. -I../O.Common -I. -I../../src/osi/compiler/gcc -I../../src/osi/compiler/default -I. -I../../src/osi/os/Linux -I../../src/osi/os/posix -I../../src/osi/os/default -I.. -I../../src/as -I../../src/bucketLib -I../../src/calc -I../../src/cvtFast -I../../src/cppStd -I../../src/cxxTemplates -I../../src/dbmf -I../../src/ellLib -I../../src/env -I../../src/error -I../../src/fdmgr -I../../src/flex -I../../src/freeList -I../../src/gpHash -I../../src/iocsh -I../../src/log -I../../src/macLib -I../../src/misc -I../../src/osi -I../../src/pool -I../../src/ring -I../../src/taskwd -I../../src/timer -I../../src/yacc -I../../src/yacc -I../../src/yajl -I/home/travis/build/epics-base/epics-base/include/compiler/gcc -I/home/travis/build/epics-base/epics-base/include/os/Linux -I/home/travis/build/epics-base/epics-base/include -I/home/travis/build/epics-base/epics-base/include/compiler/gcc -I/home/travis/build/epics-base/epics-base/include/os/Linux -I/home/travis/build/epics-base/epics-base/include        -c ../../src/osi/os/posix/osdThread.c
/usr/bin/gcc  -D_GNU_SOURCE -D_DEFAULT_SOURCE           -D_X86_64_  -DUNIX  -Dlinux     -O3 -g   -Wall      -mtune=generic      -m64 -fPIC -I. -I../O.Common -I. -I../../src/osi/compiler/gcc -I../../src/osi/compiler/default -I. -I../../src/osi/os/Linux -I../../src/osi/os/posix -I../../src/osi/os/default -I.. -I../../src/as -I../../src/bucketLib -I../../src/calc -I../../src/cvtFast -I../../src/cppStd -I../../src/cxxTemplates -I../../src/dbmf -I../../src/ellLib -I../../src/env -I../../src/error -I../../src/fdmgr -I../../src/flex -I../../src/freeList -I../../src/gpHash -I../../src/iocsh -I../../src/log -I../../src/macLib -I../../src/misc -I../../src/osi -I../../src/pool -I../../src/ring -I../../src/taskwd -I../../src/timer -I../../src/yacc -I../../src/yacc -I../../src/yajl -I/home/travis/build/epics-base/epics-base/include/compiler/gcc -I/home/travis/build/epics-base/epics-base/include/os/Linux -I/home/travis/build/epics-base/epics-base/include -I/home/travis/build/epics-base/epics-base/include/compiler/gcc -I/home/travis/build/epics-base/epics-base/include/os/Linux -I/home/travis/build/epics-base/epics-base/include        -c ../../src/osi/os/Linux/osdThreadExtra.c
../../src/osi/os/posix/osdThread.c: In function ‘createImplicit’:
../../src/osi/os/posix/osdThread.c:637:25: error: ‘commonAttr’ has no member named ‘usePolicy’
         if ( pcommonAttr->usePolicy && pthreadInfo->schedPolicy == pcommonAttr->schedPolicy ) {
                         ^
make[4]: *** [osdThread.o] Error 1
make[4]: Leaving directory `/home/travis/build/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
make[3]: *** [install.linux-x86_64] Error 2
make[3]: Leaving directory `/home/travis/build/epics-base/epics-base/modules/libcom/src'
make[2]: *** [src.install] Error 2
make[2]: Leaving directory `/home/travis/build/epics-base/epics-base/modules/libcom'
make[1]: *** [libcom.install] Error 2
make[1]: Leaving directory `/home/travis/build/epics-base/epics-base/modules'
make: *** [modules.install] Error 2
The command ".ci/travis-build.sh" exited with 2.

You can check: https://travis-ci.org/epics-base/epics-base/builds/507865382

hjunkes:
I'm just new here ;-(
I must have caught the wrong version of osdThread.c.

Sorry

till-s:
@mdavidsaver, re. mapping thread priorities: I had thoughts along similar lines but there are subtle problems:
- you don't know what scheduling policies the system implements
- you don't know what priority ranges these policies support (w/o knowing all the supported policies, that is)
- multiple policies' ranges may overlap
- the sysadmin is free to reduce the max. priority your process is allowed to use (and this is not reflected in sched_get_priority_max() - at least under linux).

IMHO - let's not lose the perspective. The entire problem only arises in `createImplicit()`. Without that the world is much simpler: the process runs (as far as epics is concerned) either with SCHED_OTHER or SCHED_FIFO, period. This is essentially the current approach.

Remains `createImplicit()`. In my opinion this should simply be scrapped. It should not be permissible to use the epicsThread API from a thread which was not created using that API (some special treatment of `main` nonwithstanding).

As it stands `createImplicit` is used by
 - `epicsThreadSuspendSelf()` -> disallow from non-epics thread or change implementation to use signals w/o requiring an epicsThread context (instead of using an 'epicsEvent' inside the epicsThread context for synchronization you can simply send STOP/CONT to the target thread [as this could confuse GDB other signals could be used])
 - `epicsThreadExitMain()` -> not really used here; can be eliminated anyways
 - `epicsThreadGetIdSelf()` -> return error from non-epics thread
 - `epicsThreadGetNameSelf()` -> return whatever
so eliminating `createImplicit()` wouldn't seem unreasonable.

anjohnson:
@till-s wrote:
> Remains `createImplicit()`. In my opinion this should simply be scrapped. It should not be permissible to use the epicsThread API from a thread which was not created using that API (some special treatment of main nonwithstanding).

That's not going to fly. There are too many places where the CA library gets initialized and used in a thread created by some other software framework (think CA client support for something like matlab or Qt).

till-s:
Well - then that leaves us with what my original proposal was: 
 - if the non-epics thread uses SCHED_FIFO AND epics also does then map OSS -> OSI and allow the non-epics thread to use epicsThreadSetPriority()
 - if the non-epics thread uses another policy or SCHED_FIFO is not used by epics then map OSS->OSI priority zero (same as when epics is built w/o RT support) and disallow epicsThreadSetPriority from non-epics thread.

Note that the traditional behavior (before any of the recent patches) has been:

 - disallow epicsThreadSetPriority for non-epics thread
 - if epics was built w/o RT support map non-epics thread priority to OSI priority zero
 - if epics was built with RT support map non-epics thread priority to OSI priority assuming that 
        - non-epics thread uses SCHED_FIFO
        - epics uses SCHED_FIFO as well

The latter is wrong and triggered this bug-report (for the particular case where epics uses SCHED_FIFO but the non-epics thread does not, IIRC).

mdavidsaver:
>     * you don't know what scheduling policies the system implements

Correct.  This will always carry either an assumption, or OS specific code.  I'm fine with either.  I think it would be enough to probe for the `SCHED_*` which are defined by POSIX.

> As it stands `createImplicit` is used by

Practically issues arise from `epicsThreadGetPriority()` and `epicsThreadGetPrioritySelf()` not being defined to signal an error, and returning unsigned.

With the above assumptions it will never be possible to make this API work from truly arbitrary threads.  So in practice I'm happy if these are well defined for `SCHED_FIFO` and `SCHED_OTHER`, with other case being OS specific/undefined (maybe print a warning).

till-s:
OK then, what you ask for is mostly what the PR24 implements:

a. epics built w/o RT support -> all non-epics thread priorities map to OSI priority zero (legacy behavior)
b. epics build with RT support AND epics has sufficient privilege to use SCHED_FIFO
    - non-epics thread uses SCHED_FIFO  -> non-epics thread's OSS priority gets mapped to OSI priority from the range allowed to use by SCHED_FIFO
    - non-epics thread uses SCHED_OTHER -> map to OSI priority zero
    - non-epics thread uses another policy -> map to OSI priority zero

epicsThreadSetPriority sets OSI priority to requested priority (legacy behavior) and changes the OSS priority *IF*

  -  epics built with RT support
  -  AND epics has sufficient privilege to use SCHED_FIFO
  -  AND thread is currently using SCHED_FIFO policy

The only additional modification that could be made (deviating from legacy) would be
 - map SCHED_FIFO priorities to 1..99 (instead of 0..99; reserving 0 for SCHED_OTHER)
 - let epicsThreadSetPriority select policy (if process has privilege) according to requested OSI
   priority: zero -> SCHED_OTHER 1..99 -> SCHED_FIFO

I'll be happy to add the last proposal but at some point this has to come to a conclusion. Since this thread has been going on for a while now and several amendments have been made I'd like to
get an idea of how many modifications you still want to have done. Therefore I suggest that I add that last change, cancel this PR and start a new one which hopefully can be approved with only minor change-requests remaining.

mdavidsaver:
Rather than my trying to describe it, here is a sketch of what I'm thinking of to find the OSI <-> OSS priority mappings.  [sched.cpp.txt](https://github.com/epics-base/epics-base/files/2986462/sched.cpp.txt)

While writing this I talked myself out of using it.  I find myself making assumptions I would rather avoid.  eg. that the various `SCHED_*` fill in a contiguous range of priorities (no gaps).  The "right way" is just too complicated to undertake w/o knowing that it is necessary.

So I'll switch gears and ask for tests to check the assumptions being made about runtime OS behavior.  Then anyone wishing to experiment with RT scheduling on a POSIX target other than Linux will have a starting point.

What I think I see so far are:

1. `sched_get_priority_min(SCHED_OTHER)==0`
1. `sched_get_priority_max(SCHED_OTHER)==0`
1. `sched_get_priority_min(SCHED_FIFO)==1`

mdavidsaver:
Agreed (I hope) algorithm for computing OSI <-> OSS priority mapping.

1. Find _min() and _max() for FIFO and OTHER.  Take the smaller of the two _min() and the larger of the two _max().  This is taken as the maximum OSS priority range.
2. If there are holes either map them or flag error.
3. Find the highest permitted OSS priority.  This bounds the permitted OSS range.
4. Compute a linear mapping from OSI [0,99] to this permitted OSS range.
5. priority range of unknown policies is capped/clipped.

anjohnson:
Too old, closing as abandoned.

PR #25, opened 2019-03-14
stripped-down fix for 1816841 only

till-s:
As requested - here's a (preliminary!) fix for bug 1816841 only

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -567,14 +567,13 @@ static epicsThreadOSD *createImplicit(void)
     pthreadInfo->osiPriority = 0;
 
 #if defined(_POSIX_THREAD_PRIORITY_SCHEDULING) && _POSIX_THREAD_PRIORITY_SCHEDULING > 0
-    {
-    struct sched_param param;
-    int policy;
-    if(pthread_getschedparam(tid,&policy,&param) == 0)
-        pthreadInfo->osiPriority =
-                 (param.sched_priority - pcommonAttr->minPriority) * 100.0 /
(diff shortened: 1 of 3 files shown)


till-s:
The travis build that failed seems completely unrelated to this pull-request...

mdavidsaver:
Looks good, and definitely meets my definition of "simpler/smaller fix".

I've manually triggered the build with the failing test to re-run.

mdavidsaver:
The appveyor queue is running about 3 days behind.  So my only immediate interest is in the travis-ci tests.

mdavidsaver:
Something about this change has been bothering me.  The current implementation is clearly wrong.  If I revert the fix, then the new fest fails with:

```
1..2
not ok  1 - 'createImplicit' assigned correct priority (-1) to non-EPICS thread
ok  2 - epicsEventWait returned 0
```

However, the fix seems to carry some Linux specific assumptions which I suspect will need to be addressed in #23 and or #24.

For the present (impending 7.0.2.1) this is good enough.

till-s:
Not sure I understand. The fix only attempts to compute an OSI priority if the non-epics thread which is implicitly being converted to an EPICS thread uses SCHED_FIFO. Otherwise the OSI priority is not computed, i.e., set to zero which is also what happens if the user disabled real-time scheduling in the configuration (DONT_USE_POSIX_THREAD_PRIORITY_SCHEDULING).

````
if ( pcommonAttr->usePolicy && pthreadInfo->schedPolicy == pcommonAttr->schedPolicy ) {
  /* compute OSI priority from OSS, assuming SCHED_FIFO */
````
`usePolicy` indicates whether pcommonAttr->schedPolicy is valid (which implies that it equals SCHED_FIFO)

PR #26, opened 2019-03-26
Timestamp before outlink

bhill-slac:
We've been using this patch in SLAC photon controls since 2014.
There was discussion re this in Core-Talk, Nov 26, 2014, subject "Incorrect Timestamp in aSubRecord."
This issue has to do with whether a record should update its timestamp before or after processing output links.
pro: Record processing chains can be created by output links and/or FLNK.  Users shouldn't need to know which method was used when using TSEL links to fetch the timestamp which initiated the chain.
con: Device driver code which relies on epicsTimeEventCurrentTime or epicsTimeEventBestTime for timestamping could see a slightly earlier timestamp depending on the time spent in the driver output routine and the resolution of their timestamp provider.

Note: Section 11.4.3 of the App. Dev. Guide has sample code in which recGblGetTimeStamp() is called after processing the output link.   If we make this change, the Guide should be updated as well.

### DIFF
--- a/modules/database/src/std/rec/calcoutRecord.c
+++ b/modules/database/src/std/rec/calcoutRecord.c
@@ -226,6 +226,7 @@ static long process(struct dbCommon *pcommon)
     struct calcoutRecord *prec = (struct calcoutRecord *)pcommon;
     rpvtStruct *prpvt = prec->rpvt;
     int doOutput;
+    unsigned char    pact=prec->pact;
 
     if (!prec->pact) {
         prec->pact = TRUE;
@@ -241,6 +242,12 @@ static long process(struct dbCommon *pcommon)
             }
         }
         checkAlarms(prec);
+
+        if ( !pact ) {
+            /* Update the timestamp before writing output values so it
+             * will be uptodate if any downstream records fetch it via TSEL */
+            recGblGetTimeStamp(prec);
+        }
         /* check for output link execution */
         switch (prec->oopt) {
         case calcoutOOPT_Every_Time:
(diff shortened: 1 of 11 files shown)


anjohnson:
Needs community discussion, and fixing for simm stuff. Core group doesn't have major objections to this idea...

bhill-slac:
Resolved merge conflict in documentation/RELEASE_NOTES.html

anjohnson:
Group 10/4: We think this is reasonable, and won't break too much stuff.

@keenanlang Would synApps users have any objections to these changes? Are there parallel changes in your modules that should also be done?

mdavidsaver (changes requested):
Meeting: Acceptable in principle.  Needs to be updated to account for SIMM link changes.  aka. `recGblGetTimeStampSimm()`

bhill-slac:
Updated to use recGblGetTimeStampSimm as appropriate.
Also applied change to int64outRecord.c.
A few tab characters got expanded by my text editor.   As we're trying to avoid tab characters I left them expanded, just adjusting indendation as needed.

keenanlang:
> @keenanlang Would synApps users have any objections to these changes? 

As far as I can tell, we don't have anything that relies on any specific behavior that would be changed here, so no objections.

>Are there parallel changes in your modules that should also be done?

At the least, the records in the calc module seem like they should also get this treatment.

modules/database/src/std/rec/calcoutRecord.c
```diff
@@ -226,6 +226,7 @@ static long process(struct dbCommon *pcommon)
     struct calcoutRecord *prec = (struct calcoutRecord *)pcommon;
     rpvtStruct *prpvt = prec->rpvt;
     int doOutput;
+    unsigned char    pact=prec->pact;
 
     if (!prec->pact) {
```
anjohnson, line 231: Since you've just read `prec->pact` into `pact` this line could become `if (!pact) {`. Not sure whether most compilers will do that optimization themselves or not.

anjohnson (approved):
I haven't actually tested this myself, but the Core group have already approved these changes in principle and the latest version seems to have fixed the simulation stuff that we asked for. I made one minor suggestion inline, not essential though.

anjohnson:
Core group 11/18: Approve

mdavidsaver (approved):

PR #27, opened 2019-05-17
DB include w/ macros

mdavidsaver:
An idea I'm trying out.  Partly an excuse for cleanup in the DBD parser code.  Enables a .db files include w/o additional macro (re)definitions.  A sort of hierarchical alternative to MSI.

```
include "other.db"   # current syntax
include("other.db")
include("other.db", "MAC=one")
include("other.db", "MAC=two,OTHER=foo")
include("other.db", "MAC=two", "OTHER=foo")
```

This PR also closes a corner case bug in the parser which requires that an include be at the end of a line (through re-use of line buffer and parser state).  eg.

```
include "other.db" record(ai, "x") {
```

Currently fails to parse with the rather contradictory error

```
Error: syntax error
 at or before "record" in path "."  file "other.db" line 1
```

Where the "record" in question is actually the remainder of the line in the enclosing .db file.

### DIFF
--- a/configure/RULES_BUILD
+++ b/configure/RULES_BUILD
@@ -273,11 +273,9 @@ YACCOPT ?= $($*_YACCOPT)
 # are creating it
 #
 %.c: %.y
-	@$(RM) $*.tab.c
-	@$(RM) $*.tab.h
+	@$(RM) $*.c
+	@$(RM) $*.h
 	$(YACC) -b$* $(YACCOPT) $<
-	$(MV) $*.tab.c $*.c
-	$(if $(findstring -d, $(YACCOPT)),$(MV) $*.tab.h $*.h,)
 
 # must be a separate rule since when not using '-d' the
 # prefix for .h will be different then .c
--- a/modules/database/src/ioc/dbStatic/RULES
+++ b/modules/database/src/ioc/dbStatic/RULES
@@ -11,3 +11,11 @@
 
 # dbLexRoutines.c is included in dbYacc.c
 dbYacc.c: dbLex.c $(IOCDIR)/dbStatic/dbLexRoutines.c
+
+# Parser debugging hints:
+# set dbStaticDebug global (in dbStaticLib.c)
+# Emit dbYacc.output with parser state info
+#dbYacc_YACCOPT = -v
+# Enable conditional debug prints.
+# at runtime set $YYDEBUG=9
+#dbYacc_CPPFLAGS = -DYYDEBUG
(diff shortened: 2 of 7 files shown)


anjohnson:
We already have a syntax defined for including other database files while defining macros, which has been in use for over 15 years, just not by the IOC. The syntax is [documented in Visual DCT](https://epics-extensions.github.io/VisualDCT/doc/DES-VisualDCT_EPICS_Databases_Hierarchy_Support.html) and is fully supported by that tool (and by a more capable but probably incomplete [dbFlatten Perl script](https://code.launchpad.net/~anj/epics-base/+git/db-hierarchy/+ref/dbh-2) that I've been tinkering with off and on for some time).

We can't easily implement the full hierarchical template expansion inside the IOC because doing so requires making at least 2 passes through the database, but we could support parts of it based on this code — I see no reason to add another syntax though. You have almost recreated the `expand()` statement, which the spec defines like this:

```
expand("slideMotor.vdb", slmot1) {
  macro(name, "sm1")
  macro(address, "4")
  macro(demand, "slide1:demand.VAL")
}
```

The second parameter to `expand()` is a name for this particular template instance, but is only used if the template file provides any port macros (a port is a macro that gets passed upwards to be expanded in the parent file using the syntax `$(<instance>.<port>)`). I would suggest adding parser support to the IOC for the `template()` syntax as well, but aborting if a template includes any `port()` statements, since expanding ports is what introduces the need for multiple passes during expansion. If you do this the IOC will then be able to load any `.vdb` files created by VisualDCT that don't rely on ports for their interfaces.

Please note the important descriptions towards the end of section 2 of the spec, one of which says "an `expand` statement with no `macro` definitions will not pass any macro values" down into the child template; you probably won't like that, but it's how we designed it back in 200x (this simplifies the context for the child and avoids the need to provide a syntax for deleting macros from the parent scope that shouldn't be visible in the child).

I haven't tried your code, but I will be happy to support adding this feature to the IOC if you are willing to change it to use the older syntax.

mdavidsaver:
Meeting: acceptable with syntax change to allow macros in multiple blocks

> dbinclude := INCLUDE STR
> dbinclude := INCLUDE STR COMMA maclist
> maclist :=
> maclist := STR
> maclist := maclist ',' STR

include "blah.db" , "A=B" , "X=Y,J=X"

mdavidsaver:
Rebase'd and implemented (almost) the agree syntax.  I added parentheses "()" to overcome what is either a limitation of antelope, or of my understanding of grammars.  Applying

```ebnf
include : tokenINCLUDE tokenSTRING include_macro_list
include_macro_list : /* empty */
                   | include_macro_list ',' tokenSTRING
{ rule };
```

to the pathological case of:

```
include "include2.db", "MAC=2"   record(longin, "three") {
```

Starts runs the rule with `tokenRECORD` from the include line already on the stack.  This triggers a syntax error when eg. the first token of "include2.db" is also `tokenRECORD`.

PR #30, opened 2019-07-15
redo blockingSockTest to check recv()/accept()/recvfrom()

mdavidsaver:
Expand blockingSockTest to cover interruption of accept() and recvfrom() as well as recv() to cover usage in pvAccessCPP.

This testing ensures that recv() is interrupted by shutdown() on all targets except WIN32.  It also tests that, on Linux and RTEMS, shutdown() interrupts accept() and recvfrom() on a UDP socket.

The assurance wrt. RTEMS is required, as simply close()ing a socket may leave a concurrent accept() or recvfrom() unrecoverable stuck.

The assurance wrt. Linux is useful as it prevents a (so far theoretical) race which would lead to a worker calling recv()/send() and a closed, or reused, file descriptor.

### DIFF
--- a/modules/libcom/test/blockingSockTest.cpp
+++ b/modules/libcom/test/blockingSockTest.cpp
@@ -7,292 +7,274 @@
 * in file LICENSE that is included with this distribution.
 \*************************************************************************/
 
+#include <stdexcept>
+
 #include <string.h>
 #include <stdio.h>
 
-#include "osiSock.h"
-#include "osiWireFormat.h"
-#include "epicsThread.h"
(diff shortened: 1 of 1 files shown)


mdavidsaver:
The current results of testing:

Linux and RTEMS: shutdown() interrupts recv(), accept(), and recvfrom().

WIN32: shutdown() interrupts recvfrom()

OSX: shutdown() interrupts recv()

I would like results for solaris and vxWorks.  Also interesting to know for cygwin, freebsd, and iOS (I suspect these will be the same as WIN32, RTEMS, and OSX respectively).

PR #31, opened 2019-09-04
Preparing the waveform record DBD-POD file; Creation of the menuFtype…

joaopaulosm:
First version of the new waveform record POD file; menuFtype POD file creation;

### DIFF
--- a/src/ioc/db/menuFtype.dbd.pod
+++ b/src/ioc/db/menuFtype.dbd.pod
@@ -7,6 +7,15 @@
 # and higher are distributed subject to a Software License Agreement found
 # in file LICENSE that is included with this distribution. 
 #*************************************************************************
+
+=head1 Menu menuFtype
+
+This menu is used for the C<FTVL> field of all record types.
+
+=menu menuFtype
+
+=cut
+
 menu(menuFtype) {
 	choice(menuFtypeSTRING,"STRING")
 	choice(menuFtypeCHAR,"CHAR")
--- a/src/ioc/db/menuFtype.dbd.pod
+++ b/src/ioc/db/menuFtype.dbd.pod
@@ -10,7 +10,7 @@
 
 =head1 Menu menuFtype
 
-This menu is used for the C<FTVL> field of all record types.
+This menu is used for the C<FTVL> and similar fields of many record types.
 
 =menu menuFtype
 
--- a/src/std/rec/waveformRecord.dbd.pod
+++ b/src/std/rec/waveformRecord.dbd.pod
@@ -1,34 +1,12 @@
-#*************************************************************************
-# Copyright (c) 2002 The University of Chicago, as Operator of Argonne
-#     National Laboratory.
-# Copyright (c) 2002 The Regents of the University of California, as
-#     Operator of Los Alamos National Laboratory.
-# EPICS BASE is distributed subject to a Software License Agreement found
-# in file LICENSE that is included with this distribution. 
-#*************************************************************************
+=pod
 
 =title Waveform Record (waveform)
 
-...
-
-=head2 Record-specific Menus
-
-=head3 Menu waveformPOST
-
-The MPST and APST fields use this menu to determine when to post new value
-and archive monitors respectively.
-
-=menu waveformPOST
-
-...
-
-=head2 Parameter Fields
-
-The record-specific fields are described below.
+The waveform record type is used to interface waveform digitizers. The record
+stores its data in arrays. The array can contain any of the supported data
+types.
 
 =recordtype waveform
-
-...
 
 =cut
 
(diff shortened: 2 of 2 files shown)


src/ioc/db/menuFtype.dbd.pod
```diff
@@ -7,6 +7,15 @@
 # and higher are distributed subject to a Software License Agreement found
 # in file LICENSE that is included with this distribution. 
 #*************************************************************************
+
+=head1 Menu menuFtype
+
+This menu is used for the C<FTVL> field of all record types.
```
anjohnson: s/field of all/and similar fields of many/

src/std/rec/waveformRecord.dbd.pod
```diff
@@ -1,45 +1,508 @@
-#*************************************************************************
-# Copyright (c) 2002 The University of Chicago, as Operator of Argonne
-#     National Laboratory.
-# Copyright (c) 2002 The Regents of the University of California, as
-#     Operator of Los Alamos National Laboratory.
-# EPICS BASE is distributed subject to a Software License Agreement found
-# in file LICENSE that is included with this distribution. 
-#*************************************************************************
```
anjohnson: Please restore the copyright notice header, it doesn't need to be removed and should be present in this file.

src/std/rec/waveformRecord.dbd.pod
```diff
@@ -1,45 +1,508 @@
...
+=pod
 
 =title Waveform Record (waveform)
```
anjohnson, line 10: Perl POD markup starts as soon as it sees any POD directive which can be the `=title` one, making the earlier `=pod` unnecessary. It's okay to leave this in, but it's actually redundant.

src/std/rec/waveformRecord.dbd.pod
```diff
@@ -1,45 +1,508 @@
...
+VAL. (If the INP link is a constant, elements can be placed in the array via
+dbPuts.) NELM specifies the number of elements that the array will hold, while
+FTVL specifies the data type of the elements.
+The RARM field causes the device to re-arm when this field is set to 1. 
+
+=head4 Possible data types for FTVL
+
+=begin html
+
+<blockquote><table border="1"><tr>
```
anjohnson: Is this table just a copy of the output from the `=menu menuFtvl` POD conversion? There's a better way to include that by actually including the menu file into the DBD part, then it's available for the `=menu` keyword:
```patch
--- a/src/std/rec/waveformRecord.dbd.pod
+++ b/src/std/rec/waveformRecord.dbd.pod
@@ -22,6 +22,10 @@ and archive monitors respectively.
 
 ...
 
+=head3 Menu FTVL
+
+=menu menuFtype
+
 =head2 Parameter Fields
 
 The record-specific fields are described below.
@@ -32,6 +36,8 @@ The record-specific fields are described below.
 
 =cut
 
+include "menuFtype.dbd"
+
 menu(waveformPOST) {
     choice(waveformPOST_Always,"Always")
     choice(waveformPOST_OnChange,"On Change")
```

If the waveformRecord.c file had a `#include "menuFtype.h"` it would be necessary to delete that include line to avoid the C compiler seeing duplicate definitions (in this case it doesn't, but other records might) but this approach does work and is better.

anjohnson (commented):
Looks great. Some additional changes requested below though...

src/std/rec/waveformRecord.dbd.pod
joaopaulosm: I don't know why I did that, will be included in the next commit.

src/std/rec/waveformRecord.dbd.pod
joaopaulosm: Thanks for the hint!

src/std/rec/waveformRecord.dbd.pod
joaopaulosm: Thanks for the hint again, this is exactly what I wanted to do but didn't know how.

src/std/rec/waveformRecord.dbd.pod
```diff
@@ -9,37 +9,452 @@
...
 
-=head2 Record-specific Menus
+=recordtype waveform
 
-=head3 Menu waveformPOST
+=cut
 
-The MPST and APST fields use this menu to determine when to post new value
-and archive monitors respectively.
+include "menuFtype.dbd"
```
mdavidsaver, line 20: Something, I think this line, is causing waveformRecord.h to contain duplicate definitions for `menuFtype*`.

```
In file included from ../demo.cpp:9:0:

/home/travis/.source/epics-base/include/menuFtype.h:7:5: error: redeclaration of ‘menuFtypeSTRING’

     menuFtypeSTRING                 /* STRING */,

     ^

In file included from ../demo.cpp:8:0:

/home/travis/.source/epics-base/include/waveformRecord.h:19:5: note: previous declaration ‘menuFtype menuFtypeSTRING’

     menuFtypeSTRING                 /* STRING */,

     ^
```

https://travis-ci.org/mdavidsaver/pva2pva/jobs/581491657#L3490

anjohnson: After discussion here I've added a guard around the typedef generated in the menu.h and record.h files so duplicates won't cause this problem and the dbd.pod files can include any menu.dbd files they want for documentation purposes (as was done in this PR). That change is now present on the 3.15 branch.

PR #33, opened 2019-09-04
Update compressRecord.dbd.pod based on Wiki + Content update

SaeedHaghtalab:
- Applied the compressRecord.pod to compressRecord.dbd.pod
- Change on src/std/rec/Makefile to include images
- Some update on the content


src/std/rec/Makefile
```diff
@@ -53,3 +53,9 @@ stdRecords_DBD = $(patsubst %,%.dbd,$(stdRecords))
 dbRecStd_SRCS += $(patsubst %,%.c,$(stdRecords))
 
 HTMLS += $(patsubst %.dbd.pod,%.html,$(notdir $(wildcard ../rec/*Record.dbd.pod)))
+
+vpath %.jpg $(USR_VPATH) .. $(SRC_DIRS) ../rec/image
+vpath %.gif $(USR_VPATH) .. $(SRC_DIRS) ../rec/image
```
anjohnson: Could you use a modern non-lossy image format instead of `.jpg` – either `.png` or `.svg`, and maybe also convert the `.gif` image into that same format?

Originally you were loading the image files from the image subdirectory; if you write
```Makefile
vpath %.png $(SRC_DIRS)
HTMLS += image/compress-1.png
HTMLS += image/compress-2.png
```
the files should be found and will be installed into the `html/image` subdirectory as well.

src/std/rec/compressRecord.dbd.pod
```diff
@@ -7,43 +7,387 @@
...
 =cut
 
 menu(compressALG) {
 	choice(compressALG_N_to_1_Low_Value,"N to 1 Low Value")
 	choice(compressALG_N_to_1_High_Value,"N to 1 High Value")
 	choice(compressALG_N_to_1_Average,"N to 1 Average")
+	choice(compressALG_N_to_1_Median,"N to 1 Median")
 	choice(compressALG_Average,"Average")
 	choice(compressALG_Circular_Buffer,"Circular Buffer")
-	choice(compressALG_N_to_1_Median,"N to 1 Median")
```
anjohnson: Did you move the `N to 1 Median` choice entry? Doing that changes the ENUM value of that choice, which could break some existing databases that might write a number to the ALG field so I would prefer to leave it where it was (compatibility has to win over giving in to OCD tendencies!).

anjohnson (commented):
Looks good. A few of changes I'd like to see are detailed below.

src/std/rec/Makefile
SaeedHaghtalab: Thanks! I will change it in next commit.

src/std/rec/compressRecord.dbd.pod
SaeedHaghtalab: Ops! sorry I didn't think that the order matters here. Will revert it in next commit.

PR #35, opened 2019-09-04
New seq record POD file

joaopaulosm:
Please review the generated html file. I've added sections to explain the SELL, SHFT and OFFS fields which probably need improvements/corrections.


anjohnson:
Merged changes by hand after fixing rm/add => git mv issue.

PR #36, opened 2019-09-05
Update selRecord.dbd.pod based on pod file from Wiki

SaeedHaghtalab:
- Merged selRecord.pod into selRecord.dbd.pod 
- Some pod formatting
- no content update yet


anjohnson:
Merged by hand after rm+add => git mv adjustment

PR #37, opened 2019-09-05
Update subRecord.dbd.pod based on pod from Wiki

SaeedHaghtalab:
- Merged subRecord.pod into subRecord.dbd.pod
- Some pod formatting
- Removed deprecated STYP field from documentation
- no content update yet


anjohnson:
Merged by hand after rm+del => git mv adjustment.

PR #38, opened 2019-09-05
First version of longin POD file

joaopaulosm:
First version, without cosmetics. Content needs verification.


anjohnson:
Merged by hand after rm+add => git mv correction.

PR #39, opened 2019-09-05
First commit of longout record POD file

joaopaulosm:
First version for the longout record, also without cosmetics.


anjohnson:
Merged by hand after del+add => git mv adjustment.

PR #40, opened 2019-09-05
Release notes

Jimbomonkey:
Replace left angle brackets with &lt; so they display properly in markdown, and add properly formatted table

Initial conversion from html to md using pandoc was done with option --wrap=none to avoid the word wrapping occurring in the wrong places


Jimbomonkey:
Superseded by #44

PR #41, opened 2019-09-05
Update subArrayRecord.dbd.pod based on Wiki+Added menuAlarmStat.dbd.pod

SaeedHaghtalab:
Applied the pod file without content check.


anjohnson:
Merged by hand after changing rm+add => git mv

PR #42, opened 2019-09-06
Adding POD to stringout record and menuYesNo, menuOmsl, menuIvoa menus

SaeedHaghtalab:
- The _menuIvoa.dbd_ is included in _stringout.dbd.pod_ but I didn't remove the `#include "menuIvoa.h"` from _stringout.h_ to test the redefinition guard.
- In original pod file from wiki, There are some device support routines described that are not defined in _devSoSoft.c_. I just left the `write_stringout` routine there which is the only routine defined for stringout record. I'm not sure if I should keep them or not.


anjohnson:
Merging by hand to fix minor issues.

PR #50, opened 2019-09-09
Update dbTest.c

krmpotic:
Fix MAX define.


anjohnson:
Thanks, merged. This error hasn't been noticed before because the `MAX(x, y)` macro isn't actually used in that source file, although the adjacent `MIN(x, y)` macro is.

PR #52, opened 2020-02-11
Fix event record device support with constant INP

gabrielfedel:
This fix apply to event record device with constant INP.
Now when the event record is processed the associated records with the
same SCAN setup get triggered correctly, it is not more necessary to set
VAL on event record.

Fixes lp: #1829770

### DIFF
--- a/src/std/rec/eventRecord.c
+++ b/src/std/rec/eventRecord.c
@@ -106,10 +106,11 @@ static long init_record(eventRecord *prec, int pass)
 	recGblInitConstantLink(&prec->siol,DBF_STRING,&prec->sval);
     }
 
-    prec->epvt = eventNameToHandle(prec->val);
-
     if( (pdset=(struct eventdset *)(prec->dset)) && (pdset->init_record) ) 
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Seems straightforward and correct.

gabrielfedel:
Is this error on travis expected ?

ralphlange:
Harmless.
It's a test failing on the MacOS builder, because the timer was calling back slightly later than the threshold of 5% allowed deviation from the requested time.
For a builder machine, these thresholds should probably be raised a bit.

gabrielfedel:
@ralphlange do you know if there is a way to check inside test code if it is running on a builder machine?

anjohnson:
@gabrielfedel This random failure has nothing to do with your code so don't worry about it, we'll fix it ourselves at some point (I already have other changes to this test file on my machine at work).

mdavidsaver:
Picked onto 3.15 branch as 983937a52ff9ffe04b3013b51dd240fe8da3f0c5

PR #53, opened 2020-02-11
Add "extern C" for many C headers

gabrielfedel:
I've tried to cover all C headers without extern C.
This allow that these files could be imported to C++ correctly.

### DIFF
--- a/modules/database/src/ioc/db/dbLockPvt.h
+++ b/modules/database/src/ioc/db/dbLockPvt.h
@@ -86,6 +86,10 @@ struct dbLocker {
     lockRecordRef refs[DBLOCKER_NALLOC]; /* actual length is maxrefs */
 };
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /* These are exported for testing only */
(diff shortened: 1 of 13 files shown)


gabrielfedel:
I haven't changed header files with inline functions, If I understood correctly these functions should work correctly on C++.

mdavidsaver:
The headers epicsMemFs.h epicsRtemsInitHooks.h osdVME.h osdSock.h and yajl_* are public, so an extern C is a needed.  The others are (I think) internal, where it does not harm.  All in all, looks fine to me.

> inline functions ... should work correctly on C++.

Yes for `static inline`.  Not for plain `inline`, but then we shouldn't have any of these in headers includeable by both C and C++ as it has different effects in these two languages.

gabrielfedel:
@mdavidsaver Is there some standard to know if a header is internal or not? 

If you or anyone else think is necessary I can change the commit to remove the extern C from internal code.

mdavidsaver:
>  if a header is internal or not?

Public headers are those which appear under `include/` after `make` has finished.  These will also have `INC += blah.h` entry in a Makefile.

mdavidsaver:
Applied as 6feb1c788d261376d3b14440d0ce2114b3f6acd6

PR #55, opened 2020-02-12
Error case for NULL arg causing segfault in iocshArgPersistentString

ghost:
As per [bug report](https://bugs.launchpad.net/epics-base/+bug/1824732), `iocshPersistentString` will check if the given arg is `NULL` before any C lib functions are used against it which can cause segfaults.

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -314,13 +314,18 @@ cvtArg (const char *filename, int lineno, char *arg, iocshArgBuf *argBuf,
         break;
 
     case iocshArgPersistentString:
-        argBuf->sval = (char *) malloc(strlen(arg) + 1);
-        if (argBuf->sval == NULL) {
-            showError(filename, lineno, "Out of memory");
+        if (arg != NULL) {
+            argBuf->sval = (char *) malloc(strlen(arg) + 1);
+            if (argBuf->sval == NULL) {
+                showError(filename, lineno, "Out of memory");
+                return 0;
+            }
+            strcpy(argBuf->sval, arg);
+            break;
+        } else {
+            showError(filename, lineno, "Unable to copy as provided arg was NULL");
             return 0;
         }
-        strcpy(argBuf->sval, arg);
-        break;
 
     case iocshArgPdbbase:
         /* Argument must be missing or 0 or pdbbase */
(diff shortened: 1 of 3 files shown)


modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -314,13 +314,18 @@ cvtArg (const char *filename, int lineno, char *arg, iocshArgBuf *argBuf,
...
+        if (arg != NULL) {
+            argBuf->sval = (char *) malloc(strlen(arg) + 1);
+            if (argBuf->sval == NULL) {
+                showError(filename, lineno, "Out of memory");
+                return 0;
+            }
+            strcpy(argBuf->sval, arg);
+            break;
+        } else {
+            showError(filename, lineno, "Unable to copy as provided arg was NULL");
```
mdavidsaver: I don't think that `!arg` is an error and rather should be passed through as `!argBuf->sval`.

ghost: I've updated this now so `argBuf->sval = NULL` in the event a null argument is passed.

Should we also test for `argBuff == NULL`? Or is it implicit that `argBuf` cannot be NULL?

Freddie also mentioned that we could potentially use the epics strdup?

ghost:
I've added a test case, but I've passed the argument in the cmd as `epicsThreadSleep \0`, I have a feeling this will be interpreted as an empty string. What route would you recommend in order to test a NULL input for the `iocshArgPersistentString`?

modules/libcom/test/iocshTestNullArgument.cmd
```diff
@@ -0,0 +1,4 @@
+on error break
+position before_error
+epicsThreadSleep \0
```
mdavidsaver: This line isn't doing what you think.  This is passing `"\\0"`.  Pass nothing to pass an NULL.  eg. just `epicsThreadSleep` with no number.  That said, this does doesn't cover the change to `iocsh.cpp` as argument type of epicsThreadSleep isn't iocshArgPersistentString.  (as mentioned on the LP bug, there are no internal users)

mdavidsaver:
The change to `iocsh.cpp` now looks fine.

modules/libcom/test/iocshTestNullArgument.cmd
ghost: Thanks. This clarifies my previous message.

mdavidsaver:
Applied as 45bbe274e9f13f935a5e31ba2e7cc46e26495e7f and 6e0706a3121a56ba395db1fcb28ba58c5c623e81

PR #58, opened 2020-02-12
Fix multiple-IOC start race condition on TCP listen

blondejamtart:
caservertask.c: Moved listen into rsrv_grab_tcp to allow retry if failed (race condition with multiple IOCs starting simultaneously). Fix for bug https://bugs.launchpad.net/epics-base/+bug/1862328

### DIFF
--- a/modules/database/src/ioc/rsrv/caservertask.c
+++ b/modules/database/src/ioc/rsrv/caservertask.c
@@ -69,15 +69,7 @@ static void req_server (void *pParm)
     IOC_sock = conf->tcp;
 
     /* listen and accept new connections */
-    if ( listen ( IOC_sock, 20 ) < 0 ) {
-        char sockErrBuf[64];
-        epicsSocketConvertErrnoToString (
-            sockErrBuf, sizeof ( sockErrBuf ) );
-        errlogPrintf ( "CAS: Listen error: %s\n",
-            sockErrBuf );
-        epicsSocketDestroy (IOC_sock);
-        epicsThreadSuspendSelf ();
-    }
+    /*    moved into rsrv_grab_tcp()     */
 
     epicsEventSignal(castcp_startStopEvent);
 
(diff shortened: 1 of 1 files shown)


modules/database/src/ioc/rsrv/caservertask.c
```diff
@@ -69,15 +69,7 @@ static void req_server (void *pParm)
...
-    if ( listen ( IOC_sock, 20 ) < 0 ) {
-        char sockErrBuf[64];
-        epicsSocketConvertErrnoToString (
-            sockErrBuf, sizeof ( sockErrBuf ) );
-        errlogPrintf ( "CAS: Listen error: %s\n",
-            sockErrBuf );
-        epicsSocketDestroy (IOC_sock);
-        epicsThreadSuspendSelf ();
-    }
+    /*    moved into rsrv_grab_tcp()     */
```
mdavidsaver: This isn't a useful comment.

modules/database/src/ioc/rsrv/caservertask.c
```diff
@@ -252,6 +244,21 @@ SOCKET* rsrv_grab_tcp(unsigned short *port)
                 ok = 0;
                 break;
             }
+            if ( listen ( tcpsock, 20 ) < 0 ) {
```
mdavidsaver: I find myself wondering if there is value in having separate error handling for `listen()`.  That is, why not have something like:

> if(bind(tcpsock, &scratch.sa, sizeof(scratch))==0 && listen ( tcpsock, 20 )==0) {

mdavidsaver:
Looks fine now.

@anjohnson Do you want to pick this onto the 3.15 branch?

mdavidsaver:
@btester271828 Can you confirm that your test case script been re-run with the present version of this PR?

blondejamtart:
Can confirm the I've run the test script against the latest version of the PR code; ran it for 20000 tries of launching 5 IOCs simultaneously overnight, no errors occurred (the same test against unmodified EPICS had 285 errors for the number of attempts).

mdavidsaver:
Ok.  To be picked onto 3.15 branch.

mdavidsaver:
Picked onto 3.15 branch as 4844fbbd8271e513d2baf7cd3726a10619c3fa9c

PR #60, opened 2020-02-13
Fix seq record on "Specified" mode when using DOL0

gabrielfedel:
Fix seq record on "Specified" mode when using DOL0 and add a test for it.
Fix https://bugs.launchpad.net/epics-base/+bug/1842661


gabrielfedel:
Some comments about this PR:
* I've just added one test for seq, related to this bug but I don't know if this is the best way to implement this kind of test. If it is, my idea is to improve the tests on seq
* This bug happens also on 3.15, Should I do another PR? or there is some approach to apply the same fix to both versions (7 and 3.15) ?

modules/database/test/std/rec/seqTest.c
```diff
@@ -0,0 +1,55 @@
...
+void testSeqSpecified(void){
+    int i;
+    for (i=0; i < 16; i++) {
+        testdbPutFieldOk("seq0.SELN", DBR_USHORT, i);
+
+        testdbPutFieldOk("ai0", DBR_LONG, 0);
+
+        testdbPutFieldOk("seq0.PROC", DBR_USHORT, 1);
+
+        epicsThreadSleep(0.1);
```
mdavidsaver: Is this sleep required?  It might not be.  If it is, then it will result in false positive failures during CI builds.

gabrielfedel: I've tried without this and it didn't work. My impression is it is needed to wait some time to ai record could be processed.  

Do you know some way to avoid this?

mdavidsaver: I think `testSyncCallback()` from dbUnitTest.h is the right tool.  The issue is that seqRecord is always asynchronous, even when `DLY*` is zero.  (cf. `processNextLink()`) So `dbProcess()` returns before the callbacks have run.  In this case (`SELM==Specified`) I think there will be only one `callbackRequest()`, so only one `testSyncCallback()` will be needed.

gabrielfedel: Thank you. I was looking for testSyncCallback before I saw your comment. I've just made a new commit using it.

mdavidsaver:
Ok.

mdavidsaver:
Applied as d82529058aea2adfca19e411139712c98c4bb251 and 90c0f5c48d133cfd0b394cd724eddbca6ff41a1b

PR #61, opened 2020-02-13
Added a null check in epicsWin32ThreadEntry to prevent dereferencing …

karlvestin:
…a null pointer in case fetchWin32ThreadGlobal fails.

LP: #1863118


mdavidsaver:
Applied as 8075b3c316c2673026b9708b32a9fb7a0afbba8b

PR #62, opened 2020-02-17
Covers work done at the 'Codathon' event at Diamond.

pheest:
I've made a point of making seeking to make one change at a time per file.
Noted: several files previously had a mixture of TAB and space indentation.
I've converted tha affected files to use space indentation only, before comitting the desired logic change.

### DIFF
--- a/modules/ca/src/client/caEventRate.cpp
+++ b/modules/ca/src/client/caEventRate.cpp
@@ -55,7 +55,7 @@ void caEventRate ( const char *pName, unsigned count )
         }
         epicsTime end = epicsTime::getCurrent ();
     
-        printf ( " done(%f sec).\n", end - begin );
+        printf ( " done(%f sec).\n", double(end - begin) );
     }
 
     {
@@ -74,7 +74,7 @@ void caEventRate ( const char *pName, unsigned count )
     
         epicsTime end = epicsTime::getCurrent ();
     
-        printf ( " done(%f sec).\n", end - begin );
+        printf ( " done(%f sec).\n", double(end - begin) );
     }
         
     {
@@ -91,7 +91,7 @@ void caEventRate ( const char *pName, unsigned count )
         }
         epicsTime end = epicsTime::getCurrent ();
     
-        printf ( " done(%f sec).\n", end - begin );
+        printf ( " done(%f sec).\n", double(end - begin) );
     }
 
     double samplePeriod = initialSamplePeriod;
--- a/modules/ca/src/client/test_event.cpp
+++ b/modules/ca/src/client/test_event.cpp
@@ -87,7 +87,7 @@ extern "C" void epicsShareAPI ca_dump_dbr (
         dbr_enum_t *pvalue = (dbr_enum_t *)pbuffer;
         for (i = 0; i < count; i++,pvalue++){
             if(count!=1 && (i%10 == 0)) printf("\n");
-            printf("%d ",*pvalue);
+            printf("%hd ",*pvalue);
         }
         break;
     }
@@ -134,7 +134,7 @@ extern "C" void epicsShareAPI ca_dump_dbr (
     {
         struct dbr_sts_string *pvalue 
           = (struct dbr_sts_string *) pbuffer;
-        printf("%2d %2d",pvalue->status,pvalue->severity);
+        printf("%2hd %2hd",pvalue->status,pvalue->severity);
         printf("\tValue: %s",pvalue->value);
         break;
     }
@@ -143,7 +143,7 @@ extern "C" void epicsShareAPI ca_dump_dbr (
         struct dbr_sts_enum *pvalue
           = (struct dbr_sts_enum *)pbuffer;
         dbr_enum_t *pEnum = &pvalue->value;
-        printf("%2d %2d",pvalue->status,pvalue->severity);
+        printf("%2hd %2hd",pvalue->status,pvalue->severity);
         if(count==1) printf("\tValue: ");
         for (i = 0; i < count; i++,pEnum++){
             if(count!=1 && (i%10 == 0)) printf("\n");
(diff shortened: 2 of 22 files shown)


modules/ca/src/client/caEventRate.cpp
```diff
@@ -55,7 +55,7 @@ void caEventRate ( const char *pName, unsigned count )
         }
         epicsTime end = epicsTime::getCurrent ();
     
-        printf ( " done(%f sec).\n", end - begin );
+        printf ( " done(%f sec).\n", double(end - begin) );
```
anjohnson, line 58: @pheest can you explain why this change and the similar ones below are being suggested (maybe quote the reasoning from the tool)? In this case both `end` and `begin` are `epicsTime` objects, and AFAIK the only way to parse the original code is into a call to the `epicsTime` method `double operator- ( const epicsTime & ) const`.

modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -394,12 +394,12 @@ static long processTarget(dbCommon *psrc, dbCommon *pdst)
 
     psrc->pact = TRUE;
 
-    if (psrc->ppn)
+    if (psrc && psrc->ppn)
```
anjohnson, line 397: I believe @mdavidsaver only just merged a change removing the `psrc &&` test of this line as unnecessary (given the uses above). We should be careful not to undo that.

modules/ca/src/client/test_event.cpp
```diff
@@ -87,7 +87,7 @@ extern "C" void epicsShareAPI ca_dump_dbr (
         dbr_enum_t *pvalue = (dbr_enum_t *)pbuffer;
         for (i = 0; i < count; i++,pvalue++){
             if(count!=1 && (i%10 == 0)) printf("\n");
-            printf("%d ",*pvalue);
+            printf("%hd ",*pvalue);
```
anjohnson, line 90: `dbr_enum_t` is a typedef for `unsigned short`. The C++ rules for variadic arguments cause `*pvalue` to be promoted to `int`, which is capable of holding the full range of values from an `unsigned short`, and a `%d` format will therefor print them all properly. If we're going to be picky about adding the `h` length-modifier we must also match the unsigned nature of the original value, so in this case we should use `%hu` instead of `%hd`, since if `*pvalue` has the top bit set I suspect the output generated by `%hd` would be wrong.

I don't see most of these length-modifier changes as particularly helpful myself, although others may disagree. There might be some that are correct and needed though, so if you have some way to distinguish them I'd be happy to see those.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -1010,7 +1010,7 @@ devSup* dbDTYPtoDevSup(dbRecordType *prdes, int dtyp) {
     return (devSup *)ellNth(&prdes->devList, dtyp+1);
 }
 
-devSup* dbDSETtoDevSup(dbRecordType *prdes, dset *pdset) {
+devSup* dbDSETtoDevSup(dbRecordType *prdes, struct dset *pdset) {
```
anjohnson, line 1013: We converted `struct dset` into `dset` here and in other places last November in commit 156b137af01576, so this change and similar ones below would be going backwards.

modules/database/src/std/rec/histogramRecord.c
```diff
@@ -140,9 +140,9 @@ static long wdogInit(histogramRecord *prec)
     if (!prec->wdog && prec->sdel > 0) {
         /* initialize a callback object */
         pcallback = calloc(1, sizeof(myCallback));
-        pcallback->prec = prec;
         if (!pcallback)
             return -1;
+        pcallback->prec = prec;
```
anjohnson, line 145: Definitely a bug, going back over 15 years. I am manually applying a version of this fix to the 3.15 branch (we fix bugs in the earliest supported branch that contains them).

modules/libcom/test/epicsThreadPoolTest.c
```diff
@@ -8,7 +8,7 @@
 #include "epicsThreadPool.h"
 
 /* included to allow tests to peek */
-#include "../../src/pool/poolPriv.h"
+#include "../src/pool/poolPriv.h"
```
anjohnson, line 11: Sorry, this change is wrong when using the EPICS build system. When the epicsThreadPoolTest.c file is compiled the working directory is always `modules/libcom/test/O.$(T_A)` so the relative path to `modules/libcom/src` must be `../../src`.

modules/libcom/test/fdmgrTest.c
```diff
@@ -119,7 +119,7 @@ int main (int argc, char **argv)
          SEVCHK(ca_search (argv[1], &chan), NULL);
     }
 
-    while (1) {
+    for(int loop=0; loop < 5; loop++)  {
```
anjohnson, line 122: Interesting… This performance measurement program has not been built by default since 2006. It's not obvious what it actually does, but it does need a PV name argument to exercise all its functionality. It is also not build-compatible with the 7.0 branch, because despite living in the module/libcom/test directory it uses the cadef.h header from the ca module (which isn't built until after libcom/test, so if we enabled it in the Makefile the build would fail).

Your change at least makes the program exit so I'd be happy for us to apply it, but I don't think we can enable it in the build until someone further evaluates its usefulness and maybe moves it into modules/ca/test (which doesn't exist yet).

anjohnson (commented):
Various inline comments and questions. I haven't reviewed the changes in the files with indentation tab/space changes as they're hard to see and I ran out of time.

I just pushed the histogramRecord.c change to the 3.15 branch.

modules/database/src/std/rec/histogramRecord.c
mdavidsaver: cf. 41f1b0ffb51b611f4290dfcc9454b8b72dedbc81

PR #63, opened 2020-02-17
Make longout record only write on change

joaopaulosm:
This pull request refers to bug 1398215 on Launchpad (https://bugs.launchpad.net/epics-base/+bug/1398215), which is **"Make output records only write on change"**. The first record to have this feature implemented is the longout record.

Two fields were added to the record: OOPT and OVAL. The first one is a menu field with usual options from other records, the second holds the last value written and its used for the comparison during the record process routine.

OOPT can assume the following options:
- Every Time (default)
- On Change
- When Zero
- When non-zero
- When Transition to Zero
- When Transition to non-zero
- Write Once then On Change (*)

If the OUT field is changed during runtime and OOPT is "On Change", the record special function will force OOPT to be "Write Once then On Change". This will cause output record write to new destination even if the value is equal the previous one. 

The second commit refers to test routines of the feature that was implemented.

### DIFF
--- a/modules/database/src/std/rec/longoutRecord.c
+++ b/modules/database/src/std/rec/longoutRecord.c
@@ -217,9 +217,9 @@ static long special(DBADDR *paddr, int after)
             return(0);
         }
 
-        /* Detect an output link re-direction (change)*/
+        /* Detect an output link re-direction (change) */
         if (dbGetFieldIndex(paddr) == longoutRecordOUT) {
-            if (!after)
+            if (after)
                 prec->outpvt = OUT_LINK_CHANGED;
             return(0);
         }
--- a/modules/database/src/std/rec/longoutRecord.c
+++ b/modules/database/src/std/rec/longoutRecord.c
@@ -81,6 +81,9 @@ rset longoutRSET={
 };
 epicsExportAddress(rset,longoutRSET);
 
+#define OUT_LINK_UNCHANGED   0
+#define OUT_LINK_CHANGED     1
+
 static void checkAlarms(longoutRecord *prec);
 static void monitor(longoutRecord *prec);
 static long writeValue(longoutRecord *prec);
@@ -120,7 +123,8 @@ static long init_record(struct dbCommon *pcommon, int pass)
     prec->mlst = prec->val;
     prec->alst = prec->val;
     prec->lalm = prec->val;
-    prec->oval = prec->val;
+    prec->pval = prec->val;
+    prec->outpvt = OUT_LINK_UNCHANGED;
     return 0;
 }
 
@@ -213,13 +217,12 @@ static long special(DBADDR *paddr, int after)
             return(0);
         }
 
-        /* If OOPT is "on change" we force a write operation */
+        /* Detect an output link re-direction (change)*/
         if (dbGetFieldIndex(paddr) == longoutRecordOUT) {
-            if ((!after) && (prec->oopt == longoutOOPT_On_Change))
-                prec->oopt = longoutOOPT_Write_Once_Then_On_Change;
-            return 0;
+            if (!after)
+                prec->outpvt = OUT_LINK_CHANGED;
+            return(0);
         }
-
 
     default:
         recGblDbaddrError(S_db_badChoice, paddr, "longout: special");
@@ -392,10 +395,7 @@ static void monitor(longoutRecord *prec)
 
 static long writeValue(longoutRecord *prec)
 {
-<<<<<<< HEAD
     longoutdset *pdset = (longoutdset *) prec->dset;
-=======
->>>>>>> 2b7ca9598 (Added OOPT to longout record)
     long status = 0;
 
     if (!prec->pact) {
@@ -435,7 +435,7 @@ static long writeValue(longoutRecord *prec)
 
 static void convert(longoutRecord *prec, epicsInt32 value)
 {
-        /* check drive limits */
+    /* check drive limits */
     if(prec->drvh > prec->drvl) {
         if (value > prec->drvh) value = prec->drvh;
         else if (value < prec->drvl) value = prec->drvl;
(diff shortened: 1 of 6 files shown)


anjohnson:
Group review 2020-03-25: Needs testing, approve in principle before merging.

anjohnson:
This PR now has conflicts in longoutRecord.c file, probably from 36a8b51d8. More importantly though the two additional fields and their behavior need to be documented in the longoutRecord.dbd.pod reference text, and it should also have a brief entry in the Release Notes. Maybe at the upcoming codeathon?

joaopaulosm:
Sounds good! Maybe expand to the remaining records. Looking forward to the codeathon!

bfrk:
I would prefer to add a separate field to store the "write once" bit, rather than adding an option to the menu. Rationale:
1. It is cleaner than fiddling with the nominally user defined OOPT.
2. Whether a change in OUT should trigger a write should be decided by the database developer, not by the record support.

The semantics of this field would be "consider a change in OUT as a value change wrt the OOPT 'on change' value", type could be menuYesNo.

joaopaulosm:
Fair enough, it also keeps OOPT consistent with the other records that already use it.

joaopaulosm:
Ok, I've done some modification on the implementation. 
Three fields were added: OOPT, PVAL and OOCH. The first one is a menu field with the same options as CALCOUT record, the second holds the previous value written and its used for the comparison during the record process routine. The third (OOCH) is the flag that tells the record support if it should consider a change in OUT as a value change (default to YES).

So, the default behaviour when OOPT is "On Change" and the OUT link is changed is to write to the output during the next time the record is processed even if VAL is the same. To avoid this behaviour, OOCH must be set to NO.

I've also added more test cases in [longoutTest.c](https://github.com/joaopaulosm/epics-base/blob/longout_oopt/modules/database/test/std/rec/longoutTest.c) to validate all cases. The tests are done using a simple longout record that writes to the B field of CALC records. These calc records are configured as counters (CALC = A + 1, with A being the calc record itself). This way, we can measure how many times the longout has written to the output link by checking the value of the calc fields.

If the implementation, nomenclature and documentation are OK, I can extend this feature to the remaining output records:
- bo
- ao
- mbbo
- mbboDirect
- int64out
- lso
- stringout
- ???

modules/database/src/std/rec/longoutRecord.dbd.pod
```diff
@@ -71,7 +80,45 @@ DTYP field must then specify the C<<< Soft Channel >>> device support routine.
...
+=item *
+C<Transition To Non-zero> -- when record is processed, write output only if
+VAL is non-zero and last value was zero.
+
+=back
+
+=head4 Changes in OUT field when OOPT = On Change
+
+The OOCH field determines if a change in OUT field should cause a write operation
+even when the value is the same and OOPT = On Change. By default, OOCH is set to YES.
```
anjohnson: It took me a moment to understand that this is referring to a change to the link address in the `OUT` field, could you clarify that in the wording of the field description here, maybe make it:

```
If OOCH is C<YES> (its default value) and the OOPT field is C<On Change>,
the record will write to the device support the first time the record gets
processed after its OUT link is modified, even when the output value has
not actually changed.
```

modules/database/src/std/rec/longoutRecord.c
```diff
@@ -210,6 +216,14 @@ static long special(DBADDR *paddr, int after)
                 recGblCheckSimm((dbCommon *)prec, &prec->sscn, prec->oldsimm, prec->simm);
             return(0);
         }
+
+        /* Detect an output link re-direction (change)*/
+        if (dbGetFieldIndex(paddr) == longoutRecordOUT) {
+            if (!after)
```
anjohnson: I think you probably want to make this conditional `if (after)` since it is possible for the change to be rejected by the put operation which happens between the "before" and "after" calls to your `special()` routine. If that happens the OUT field will stay the same and the second call to `special()` never happens – you probably don't want to set the flag until you know the new link value has been accepted.

You should be able to add a test for that, try writing the string `{bad:link}` into the OUT field. The put will be rejected and the `special(paddr, 1)` call skipped.

modules/database/src/std/rec/longoutRecord.c
```diff
@@ -421,10 +435,61 @@ static long writeValue(longoutRecord *prec)
...
+        break;
+    }
+
+    if (doDevSupWrite)
+        status = pdset->write_longout(prec);
+
+    prec->pval = prec->val;
+    prec->outpvt = OUT_LINK_UNCHANGED; /* reset status of OUT link */
+    return status;
+}
```
anjohnson: Can you add a newline to the end of this file please. Posix requires all text files to end with a newline and all the Unix editors would include one automatically; I think it was editors that came from the Windows world that started leaving it off, which can cause some surprising bad effects.

anjohnson (changes requested):
I _love_ all the tests, thank-you for writing them!

I'm requesting a few minor changes but I think this looks good otherwise.

joaopaulosm:
Thanks @anjohnson! All your requests were implemented.

joaopaulosm:
We may have a problem with the "On Change" OOPT mode: the first (initial) processing. Since the `PVAL` (previous val) is being initialised with the same value as `VAL` in the `init_record` [function](https://github.com/joaopaulosm/epics-base/blob/1d85bc7424576fb1f7eb6c890d6a42c8d54254f4/modules/database/src/std/rec/longoutRecord.c#L125-L126), the first time record is processed it won't write to its output (`OUT`) link.  

What should we do?
- Create another field so the database developer decides if the IOC should force an initial write when `OOPT = On Change`;
- Initialise `PVAL` with a different value than `VAL` when `OOPT = On Change`, assuming this behaviour;
- Nothing (?)

anjohnson:
There would seem to be overlap with the OOCH and OUTPVT field in that problem, could you maybe initialize OUTPVT to `OUT_LINK_CHANGED` in the `init_record()` method instead, or use the OOCH setting to determine the behavior?

joaopaulosm:
Done! Added a few tests cases for that scenario.

anjohnson:
Group meeting 7/28 Re-run tests against latest.
ANJ to do final approval.

anjohnson:
Merged after rebase and some minor edits, at 413f14e4.

Sorry for the long delay, at least it got in before 2023!

joaopaulosm:
Thank you Andrew! I should probably find some time to extend this feature to the other output records. In fact, I think I already did it for the stringout.

PR #66, opened 2020-03-24
Fix epicsFindSymbol() on OSX and WIN32

mdavidsaver:
I've found that `epicsFindSymbol()` seems not to work on OSX and WIN32.  The WIN32 is more complicated, and outside of my usual experience, so I'm going through a PR.

The fix on OSX is easy to explain.  `dlsym(0, name)` errors because `0` is not a valid handle.  On Linux (glibc anyway) `RTDL_GLOBAL` is zero, but on OSX this seems not to be the case, so the existing code probably never worked.

`GetProcAddress(0, name)`, while [undocumented](https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getprocaddress), seems to behave like `GetProcAddress(GetModuleHandle(NULL), name)` and only searches for symbols exported from the executable file.  This may be sufficient in some static linking situations, but isn't very useful in DLL builds.

There does not seem to be a direct analog to `RTDL_GLOBAL` for WIN32, but the same effect can be achieved with `EnumProcessModules()` to [enumerate all loaded modules](https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocessmodules), this includes the executable file, and all DLLs loaded implicitly and explicitly.

I've chosen to use the `K32EnumProcessModules()` variant, which doesn't require linking against `psapi.dll`, but requires Windows 7 or later.  This could be relaxed if it is desirable.

I'm also included a change so that the WIN32 version of `epicsLoadError()` uses thread local storage.

### DIFF
--- a/modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
+++ b/modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
@@ -5,8 +5,13 @@
 \*************************************************************************/
 /* osi/os/WIN32/osdFindSymbol.c */
 
+/* avoid need to link against psapi.dll
+ * requires windows 7 or later
+ */
+#define NTDDI_VERSION NTDDI_WIN7
 
 #include <windows.h>
+#include <psapi.h>
 
 #define epicsExportSharedSymbols
 #include "epicsFindSymbol.h"
@@ -51,8 +56,22 @@ epicsShareFunc const char *epicsLoadError(void)
 
 epicsShareFunc void * epicsShareAPI epicsFindSymbol(const char *name)
 {
-    void* ret = GetProcAddress(0, name);
+    HMODULE dlls[128];
+    DWORD ndlls=0u, i;
+    void* ret = NULL;
+
+    /* As a handle returned by LoadLibrary() isn't available to us,
+     * try all loaded modules in arbitrary order.
+     */
+    if(K32EnumProcessModules(GetCurrentProcess(), dlls, sizeof(dlls), &ndlls)) {
+        for(i=0; !ret && i<ndlls; i++) {
+            ret = GetProcAddress(dlls[i], name);
+        }
+    }
     if(!ret) {
+        /* only capturing the last error code,
+         * but what else to do?
+         */
         epicsLoadErrorCode = GetLastError();
     }
     return ret;
--- a/modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
+++ b/modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
@@ -11,7 +11,16 @@
 #define epicsExportSharedSymbols
 #include "epicsFindSymbol.h"
 
-static int epicsLoadErrorCode = 0;
+#ifdef _MSC_VER
+#  define STORE static __declspec( thread )
+#elif __GNUC__
+#  define STORE static __thread
+#else
+#  define STORE static
+#endif
+
+STORE
+int epicsLoadErrorCode = 0;
 
 epicsShareFunc void * epicsLoadLibrary(const char *name)
 {
@@ -28,7 +37,7 @@ epicsShareFunc void * epicsLoadLibrary(const char *name)
 
 epicsShareFunc const char *epicsLoadError(void)
 {
-    static char buffer[100];
+    STORE char buffer[100];
 
     FormatMessage(
         FORMAT_MESSAGE_FROM_SYSTEM,
@@ -42,5 +51,9 @@ epicsShareFunc const char *epicsLoadError(void)
 
 epicsShareFunc void * epicsShareAPI epicsFindSymbol(const char *name)
 {
-    return GetProcAddress(0, name);
+    void* ret = GetProcAddress(0, name);
+    if(!ret) {
+        epicsLoadErrorCode = GetLastError();
+    }
+    return ret;
 }
--- a/modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
+++ b/modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
@@ -19,6 +19,7 @@
 #include <psapi.h>
 
 #define epicsExportSharedSymbols
+#include "epicsStdio.h"
 #include "epicsFindSymbol.h"
 
 #ifdef _MSC_VER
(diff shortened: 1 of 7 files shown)


mdavidsaver:
The definition of `K32EnumProcessModules()` seems to be present in mingw-w64 until 6.0.0 (in Debian stable) but missing in 5.0.0 (ubuntu bionic).

modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
```diff
@@ -28,19 +47,46 @@ epicsShareFunc void * epicsLoadLibrary(const char *name)
...
+    void* ret = NULL;
+
+    /* As a handle returned by LoadLibrary() isn't available to us,
+     * try all loaded modules in arbitrary order.
+     */
+    if(epicsEnumProcessModules(GetCurrentProcess(), dlls, sizeof(dlls), &ndlls)) {
+        for(i=0; !ret && i<ndlls; i++) {
+            ret = GetProcAddress(dlls[i], name);
+        }
+    }
```
FreddieAkeroyd, line 111: The return argument from epicsEnumProcessModules() is not the number of DLLs but the number of bytes that would be required to store them, you would need to divide by sizeof(HMODULE) to get the real number of DLLs. It could also potentially return a number larger number than passed, so you would need to be compared to sizeof(dlls) and if it is bigger than this number reallocate dlls to a larger size and call again. It may thus be better to call initially with a size of zero and then malloc dlls based on the returned lpcbNeeded

modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
```diff
@@ -28,19 +47,46 @@ epicsShareFunc void * epicsLoadLibrary(const char *name)
...
+    DWORD n;
 
-    FormatMessage(
+    n = FormatMessage(
         FORMAT_MESSAGE_FROM_SYSTEM,
         NULL,
         epicsLoadErrorCode,
         0,
         buffer,
         sizeof(buffer)-1, NULL );
```
FreddieAkeroyd, line 57: You should also pass FORMAT_MESSAGE_IGNORE_INSERTS to the function in this situation as you are not supplying any arguments the message might require.

modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
```diff
@@ -28,19 +47,46 @@ epicsShareFunc void * epicsLoadLibrary(const char *name)
...
+
+    /* n - number of chars stored excluding nil.
+     *
+     * Some messages include a trailing newline, which we strip.
+     */
+    for(; n>=1 && (buffer[n-1]=='\n' || buffer[n-1]=='\r'); n--)
+        buffer[n-1] = '\0';
+
```
FreddieAkeroyd, line 73: If FormatMessage() fails n == 0 so a previous message written to buffer may be returned. Maybe something like    epicsSnprintf(buffer, sizeof(buffer), "epicsLoadError: Unable to format WIN32 error code %lu", epicsLoadErrorCode)   in that case

modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
```diff
@@ -5,13 +5,32 @@
...
+#ifdef _MSC_VER
+#  define STORE static __declspec( thread )
+#elif __GNUC__
+#  define STORE static __thread
+#else
+#  define STORE static
+#endif
+
+STORE
+int epicsLoadErrorCode = 0;
```
FreddieAkeroyd, line 35: Should this be a DWORD ?

modules/libcom/src/osi/os/WIN32/osdFindSymbol.c
mdavidsaver: yes... oops.

anjohnson:
Just recently epicsStackTraceTest.c has started failing on Appveyor's MinGW x86 (32-bit) builds of the 3.15 branch, e.g. [here](https://ci.appveyor.com/project/epics-base/epics-base/builds/32328965/job/nrl3ofh4bswo0ie5#L5258). Both dynamic and static builds fail, but the x64 (64-bit) builds succeed. Don't know if there can be any connection with what you're doing here (especially since you're not obviously working on the MinGW build and this is the 7.0 branch anyway); just wondering if this change should be on the 3.15 branch.

FreddieAkeroyd:
The 3.15 build does not upload tap files, so I can't see the exact error there, I would need to rebuild myself locally which I can do later. The 7.0 branch has a better version of osdFindAddr     https://github.com/epics-base/epics-base/blob/7.0/modules/libcom/src/osi/os/WIN32/osdFindAddr.c and it notes that static builds cannot print function names, but DLL ones can get exported names but only on x64. I also seem to recall the stack trace itself needs to jump through various hoops to try and stop the function being tested being optimised away ... so it may be gcc update at appveryor?

FreddieAkeroyd:
error looks to be
```
C:\projects\epics-base-test\src\libCom\test\O.win32-x86-mingw>
C:\projects\epics-base-test\src\libCom\test\O.win32-x86-mingw>epicsStackTraceTes
t.exe
1..5
ok  1 - epicsStackTraceGetFeatures() obtains features
# calling a few nested routines and eventually dump a stack trace
# now scan the result for what we expect
# found 0 x epicsStackTrace

# found 0 x epicsStackTraceRecurseGbl

# found 0 x epicsStackTraceRecurseLcl

ok  2 # SKIP no support for dumping library symbols on this platform
ok  3 # SKIP no support for dumping global symbols on this platform
ok  4 # SKIP no support for dumping local symbols on this platform
not ok  5 - dumping addresses
# Dumping a stack trace of thread 'win8b8':
# [  707A9C83]
# [  004013E3]
# [  772C6A14]
# [  77E3AD8F]
# [  77E3AD5A]
#

    Results
    =======
       Tests: 5
      Passed:   4 = 80.00%
      Failed:   1 = 20.00%
     Skipped:   3 = 60.00%

C:\projects\epics-base-test\src\libCom\test\O.win32-x86-mingw>
```

FreddieAkeroyd:
and epicsStackTraceRecurseGbl is 00401680

FreddieAkeroyd:
It is expecting epicsStackTraceRecurseGbl twice

mdavidsaver:
@anjohnson I think this change is ready, and could go into 7.0 immediately.  I see this as fairly low risk given the historic brokenness.

The part I might imagine causing problems is the need to link mingw (but not MSVC) builds with psapi.dll.  Though I think that adding to `OP_SYS_LDLIBS` should be sufficient.

FreddieAkeroyd:
The only thing I have a slight caution with is defining NTDDI_VERSION in a source file, I can't find something saying it may lead to issues but normally this macro would be defined at a library/project level. If defined at a file level, different files could pull in potentially different behaviour from windows.h  In this particular case it is probably fine, but an alternative here could be to look at PSAPI_VERSION which is now defined in psapi.h and if it is 2 or greater then K32EnumProcessModules is available to use.

mdavidsaver:
I can't recall why I used `NTDDI_VERSION` instead of `PSAPI_VERSION`.  Probably didn't read the documentation closely enough.  As windows specific definitions (eg. structs) are I think confided to this compilation unit, the worst that could result would be linking errors.  Still, I've switched to the more specific `PSAPI_VERSION` anyway.  We'll see if it survives CI.

mdavidsaver:
Merged with 514962724234c0777a42f6b051d9be3e448b43a7

PR #67, opened 2020-04-01
Add a getter function for field_type

EmilioPeJu:
Some time ago, the DCT  related functions were removed from `dbStaticLib`, leaving no function to retrieve a field type. IIRC the documentation suggests that the fields in DBENTRY and others shouldn't be accessed directly.
In this change, I would like to provide a getter function for a field dbf type, I have 2 comments about it:
- I use a different name for the function to avoid confusion with the deprecated one.
- I don't check pflddes for NULL, as every use of this functions is always done after calling any form of `dbNextField`, which makes access safe.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -1381,6 +1381,14 @@ char * dbGetFieldName(DBENTRY *pdbentry)
     return(pflddes->name);
 }
 
+int dbGetFieldDbfType(DBENTRY *pdbentry)
+{
+    dbFldDes  	*pflddes = pdbentry->pflddes;
+
+    if(!pflddes) return(-1);
+    return(pflddes->field_type);
+}
+
 char * dbGetDefault(DBENTRY *pdbentry)
 {
     dbFldDes  	*pflddes = pdbentry->pflddes;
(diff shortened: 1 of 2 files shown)


EmilioPeJu:
The reason why we would like this change is because we use an application/library, made in python, called iocbuilder which requires to know the fields type.
I implemented a workaround by accessing `field_type`, but that could potentially break if the layout of the related structures is changed (i.e. adding new fields, reordering, .. etc).

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -1381,6 +1381,11 @@ char * dbGetFieldName(DBENTRY *pdbentry)
     return(pflddes->name);
 }
 
+dbfType dbGetDbfFieldType(DBENTRY *pdbentry)
+{
+    return(pdbentry->pflddes->field_type);
```
anjohnson: This code could result in a segfault if it gets called when no field is selected and `pdbentry->pflddes` is NULL – see the check for that possibility in the routines either side of it. The question then is what to return in that case, which is a problem because there is no really appropriate `dbfType` enum value for that. If you change the return type of your routine to be an `int` you can then return -1 in that case. Your application should be able to handle that value too.

One other request, could you make the routine name `dbGetFieldDbfType()` so it's closer to the other `dbGetField...()` routine names.

With those changes I'd be happy to accept this, thanks!

EmilioPeJu: Thank you for your review, I amended the commit with your suggested changes.
I did not check for NULL before, because I thought selecting the field was always done when you call any `dbGetField*()`. I guess there are exceptions to this.

Araneidae:
The integration test failures look unrelated to this merge request.  I'd be sorry to see this not merged.

anjohnson:
@Araneidae The Core group are aware of those test failures at the moment. This PR will get merged as currently proposed, it may take some time to get to it but we are planning another 7.0 release in the next month or two where this PR will be included.

mdavidsaver:
Picked as 0f88c67473d246be4f721f6f058289eaffcd75b4

PR #70, opened 2020-04-21
No double-colon rules

anjohnson:
This is my preferred approach, which avoids the double-colon rules but should be just as flexible. It will probably need some refining, but includes the enhanced test-results rule that I was working on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /include/
 /templates/
 /configure/*.local
+/.tests-failed
 O.*/
 /QtC-*
 *.orig
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -77,7 +77,7 @@ build_script:
   - cmd: ci/appveyor-make.bat
 
 test_script:
-  - cmd: ci/appveyor-make.bat runtests
+  - cmd: ci/appveyor-make.bat tapfiles test-results
 
 #---------------------------------#
 #         notifications           #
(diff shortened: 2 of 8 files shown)


anjohnson:
See https://ci.appveyor.com/project/anjohnson/epics-base/builds/32325246 for my Appveyor build.

ralphlange:
Not having heard from you, I was also working on a branch ... will rebase on your if necessary.
Thanks!

ralphlange:
With that narrowing of the static pattern rules for %tap and %xml resulting from perl tests, this works fine for my Google Test integration.
Thanks a lot!!

ralphlange:
Don't worry about the force push - I was solely removing my last commit of yesterday which was heading off into the wrong direction. (And that was not in this branch.)

mdavidsaver:
I still find myself `clean::ing` up after the last such change, so my first concern is avoiding a repetition.  In this instance I think such a change is acceptable, if only because unittests are still so rare in EPICS modules.

The only other such rule left seems to be `show-makefiles::`.

ralphlange:
This branch by Andrew is reverting my change (introducing double::rules) that I did just a few days ago. No side effects on other base or user code.

anjohnson:
As Ralph said my change undoes one of his from a few days earlier; I don't know of anyone else who was making use of the original single-colon target but if they were this means they would no longer have to make any changes.

The `show-makefiles::` rule is an internal one that I wouldn't expect any other module to be implementing or for us to ever need to change.

I will be merging this branch hopefully sometime today, once I get to the bottom of my Inbox.

PR #71, opened 2020-04-23
Add alarm message and time tag to dbCommon

mdavidsaver:
Adding extra meta-data to be transported by PVA.

Starts by adding two (user) fields to dbCommon named: AMSG (`alarm.message` in NTScalar) and UTAG (`timeStamp.userTag`).

AMSG is an alarm message string to go along with SEVR and STAT.  A new function `recGblSetSevrMsg()` for (maybe) raising an alarm with a message string.  Some internal use is made, including in `dbGetLink()` to indicate which link field has caused an alarm (eg. "INPA" in calc).

```c
int recGblSetSevrMsg(void *precord, epicsEnum16 new_stat,
                     epicsEnum16 new_sevr,
                     const char *msg, ...)
```

This alarm message string is added to `db_field_log` to maintain consistency, although I'm not certain that this isn't the right approach.

UTAG is an integer (signed 32-bit atm.) which is treated as part of the timestamp.  It is also added to `db_field_log`, and moved by a DB_LINK to `.TIME` via the new `dbGetTimeStampTag()` link support accessor.  I've also added a new server side filter `utag` which can filter updates based on UTAG.  It applies a filter condition `UTAG&M==V` where `M` and `V` are client specified integers.

My idea is that a knowledgeable driver can fill in UTAG with a bit-mask describing the conditions under which this value was acquired.  For example, an intelligent beam measurement device might fill this in based on information received from a timing system.  The most obvious case is to have one bit for "beam on".  Say bit 0.  Then

> camonitor BPM0:X.VAL{"utag":{"M":1,"V":1}}

will only show updates when beam should be present.

Additionally, two new link support "methods" are added `lset::getAlarmMsg` and `lset::getTimestampTag` which like the existing getAlarm and getTimestamp.  `dbGetTimeStampTag()` is added with usage in `recGblGetTimeStampSimm()` to propagate UTAG along with TIME through supported link types (DB and calc in Base).

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -523,7 +523,7 @@ long dbProcess(dbCommon *precord)
             (precord->lcnt++ < MAX_LOCK) ||
             (precord->sevr >= INVALID_ALARM)) goto all_done;
 
-        recGblSetSevr(precord, SCAN_ALARM, INVALID_ALARM);
+        recGblSetSevrMsg(precord, SCAN_ALARM, INVALID_ALARM, "Async in progress");
         monitor_mask = recGblResetAlarms(precord);
         monitor_mask |= DBE_VALUE|DBE_LOG;
         pdbFldDes = pdbRecordType->papFldDes[pdbRecordType->indvalFlddes];
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -359,6 +359,14 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
             *pushort++ = pcommon->acks;
             *pushort++ = pcommon->ackt;
             pbuffer = (char *)pushort;
+            if (!pfl || pfl->type == dbfl_type_rec) {
+                STATIC_ASSERT(sizeof(pcommon->amsg)==sizeof(pfl->amsg));
+                strncpy(pbuffer, pcommon->amsg, sizeof(pcommon->amsg)-1);
+            } else {
+                strncpy(pbuffer, pfl->amsg,sizeof(pfl->amsg)-1);
+            }
+            pbuffer[sizeof(pcommon->amsg)-1] = '\0';
+            pbuffer += sizeof(pcommon->amsg);
         }
         if( (*options) & DBR_UNITS ) {
             memset(pbuffer,'\0',dbr_units_size);
@@ -386,10 +394,13 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
             if (!pfl) {
                 *ptime++ = pcommon->time.secPastEpoch;
                 *ptime++ = pcommon->time.nsec;
+                *ptime++ = pcommon->utag;
             } else {
                 *ptime++ = pfl->time.secPastEpoch;
                 *ptime++ = pfl->time.nsec;
+                *ptime++ = pfl->utag;
             }
+            *ptime++ = 0; /* padding */
             pbuffer = (char *)ptime;
         }
         if( (*options) & DBR_ENUM_STRS )
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -361,7 +361,7 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
             pbuffer = (char *)pushort;
         }
         if( (*options) & DBR_AMSG ) {
-            if (!pfl || pfl->type == dbfl_type_rec) {
+            if (!pfl) {
                 STATIC_ASSERT(sizeof(pcommon->amsg)==sizeof(pfl->amsg));
                 strncpy(pbuffer, pcommon->amsg, sizeof(pcommon->amsg)-1);
             } else {
@@ -404,7 +404,7 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
         }
         if( (*options) & DBR_UTAG ) {
             epicsUInt64 *ptag = (epicsUInt64*)pbuffer;
-            if (!pfl || pfl->type == dbfl_type_rec) {
+            if (!pfl) {
                 *ptag++ = pcommon->utag;
             } else {
                 *ptag++ = pfl->utag;
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -359,6 +359,8 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
             *pushort++ = pcommon->acks;
             *pushort++ = pcommon->ackt;
             pbuffer = (char *)pushort;
+        }
+        if( (*options) & DBR_AMSG ) {
             if (!pfl || pfl->type == dbfl_type_rec) {
                 STATIC_ASSERT(sizeof(pcommon->amsg)==sizeof(pfl->amsg));
                 strncpy(pbuffer, pcommon->amsg, sizeof(pcommon->amsg)-1);
@@ -390,7 +392,6 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
         }
         if( (*options) & DBR_TIME ) {
             epicsUInt32 *ptime = (epicsUInt32 *)pbuffer;
-            epicsUInt64 *ptime64;
 
             if (!pfl) {
                 *ptime++ = pcommon->time.secPastEpoch;
@@ -399,13 +400,16 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
                 *ptime++ = pfl->time.secPastEpoch;
                 *ptime++ = pfl->time.nsec;
             }
-            ptime64 = (epicsUInt64*)ptime;
+            pbuffer = (char *)ptime;
+        }
+        if( (*options) & DBR_UTAG ) {
+            epicsUInt64 *ptag = (epicsUInt64*)pbuffer;
             if (!pfl || pfl->type == dbfl_type_rec) {
-                *ptime64++ = pcommon->utag;
+                *ptag++ = pcommon->utag;
             } else {
-                *ptime64++ = pfl->utag;
+                *ptag++ = pfl->utag;
             }
-            pbuffer = (char *)ptime64;
+            pbuffer = (char *)ptag;
         }
         if( (*options) & DBR_ENUM_STRS )
             get_enum_strs(paddr, &pbuffer, prset, options);
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -390,18 +390,22 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
         }
         if( (*options) & DBR_TIME ) {
             epicsUInt32 *ptime = (epicsUInt32 *)pbuffer;
+            epicsUInt64 *ptime64;
 
             if (!pfl) {
                 *ptime++ = pcommon->time.secPastEpoch;
                 *ptime++ = pcommon->time.nsec;
-                *ptime++ = pcommon->utag;
             } else {
                 *ptime++ = pfl->time.secPastEpoch;
                 *ptime++ = pfl->time.nsec;
-                *ptime++ = pfl->utag;
             }
-            *ptime++ = 0; /* padding */
-            pbuffer = (char *)ptime;
+            ptime64 = (epicsUInt64*)ptime;
+            if (!pfl || pfl->type == dbfl_type_rec) {
+                *ptime64++ = pcommon->utag;
+            } else {
+                *ptime64++ = pfl->utag;
+            }
+            pbuffer = (char *)ptime64;
         }
         if( (*options) & DBR_ENUM_STRS )
             get_enum_strs(paddr, &pbuffer, prset, options);
--- a/modules/database/src/ioc/db/dbAccessDefs.h
+++ b/modules/database/src/ioc/db/dbAccessDefs.h
@@ -33,16 +33,18 @@ DBCORE_API extern int dbAccessDebugPUTF;
 /*  The database field and request types are defined in dbFldTypes.h*/
 /* Data Base Request Options    */
 #define DBR_STATUS      0x00000001
-#define DBR_UNITS       0x00000002
-#define DBR_PRECISION   0x00000004
-#define DBR_TIME        0x00000008
-#define DBR_ENUM_STRS   0x00000010
-#define DBR_GR_LONG     0x00000020
-#define DBR_GR_DOUBLE   0x00000040
-#define DBR_CTRL_LONG   0x00000080
-#define DBR_CTRL_DOUBLE 0x00000100
-#define DBR_AL_LONG     0x00000200
-#define DBR_AL_DOUBLE   0x00000400
+#define DBR_AMSG        0x00000002
+#define DBR_UNITS       0x00000004
+#define DBR_PRECISION   0x00000008
+#define DBR_TIME        0x00000010
+#define DBR_UTAG        0x00000020
+#define DBR_ENUM_STRS   0x00000040
+#define DBR_GR_LONG     0x00000080
+#define DBR_GR_DOUBLE   0x00000100
+#define DBR_CTRL_LONG   0x00000200
+#define DBR_CTRL_DOUBLE 0x00000400
+#define DBR_AL_LONG     0x00000800
+#define DBR_AL_DOUBLE   0x00001000
 
 /**********************************************************************
  * The next page contains macros for defining requests.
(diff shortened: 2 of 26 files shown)


mdavidsaver:
I encountered some complications with filters and QSRV described in https://github.com/epics-base/pva2pva/issues/32

modules/database/src/ioc/db/dbAccessDefs.h
```diff
@@ -106,7 +106,8 @@ epicsShareExtern int dbAccessDebugPUTF;
 	epicsUInt16	status;		/* alarm status */\
 	epicsUInt16	severity;	/* alarm severity*/\
 	epicsUInt16	acks;		/* alarm ack severity*/\
-	epicsUInt16	ackt;		/* Acknowledge transient alarms?*/
+    epicsUInt16	ackt;		/* Acknowledge transient alarms?*/\
+    char amsg[48];
```
anjohnson: Curious: Why is this 48? Making it larger than MAX_STRING_SIZE=40 seems likely to be occasionally annoying.

mdavidsaver: I recall having the alignment wrong at one point (must be a multiple of 8) and rounded up.  Reducing to 40 seems reasonable.

mdavidsaver:
TODO list from meeting

* recGblSetSevrMsg() use sprintf()
* compatibility recipe w/ `#define HAS_recGblSetSevrMsg`
* add lset::getAlarmMsg()

anjohnson:
+ Pedantically speaking the code should call `epicsVsnprintf()`
+ Might want to make that `#define HAS_` name something a bit more generic, since the same flag can also be used for the lset change, the presence of the `AMSG` and `UTAG` fields and anything else that comes along before this gets merged & released.

mdavidsaver:
Updated to address review comments.

TODO: test coverage for new link support calls.

mdavidsaver:
Draft Release Notes entry

### Add Alarm Message and Time Tag Fields

Two new fields have been added to `dbCommon` so will be present in all
records: `AMSG` and `UTAG`.

#### `AMSG`

`AMSG` can hold an arbitrary 40-character string, providing additional
information about the alarm condition indicated in `STAT` and `SEVR`. With no
alarm it will hold an empty string. The new `recGblSetSevrMsg()` function can
be used in place of `recGblSetSevr()` to signal an alarm while providing a
message.

For example, a device support's `read_bi()` routine for a hypothetical
multi-channel ethernet attached device might flag a communication error
between the IOC and controller, or an error involving a certain channel like
this:

```c
static long read_bi(biRecord* prec) {
    ...
    if (!priv->connected) {
        recGblSetSevrMsg(prec, COMM_ALARM, INVALID_ALARM,
            "No controller connected");
        return S_dev_noDevice;
    }
    if (!priv->err) {
        recGblSetSevrMsg(prec, READ_ALARM, INVALID_ALARM,
            "Channel %u disconnexted", priv->chan);
        return S_dev_noDevice;
    }
    return status;
}
```

#### `UTAG`

`UTAG` holds an `epicsUInt64` value which is semantically part of the record's
timestamp (`TIME`). The value defaults to zero if not explicitly set. Device
support or an event time provider which supports this feature may write a tag
value directly to the `dbCommon::utag` field.

`TSEL` links will copy both `TIME` and `UTAG` between records if the link type
supports this (CA links do not).

A `utag` server side channel filter has been added which can be configured to
filter out monitor updates which don't pass the test `(UTAG & M) == V` where
`M` and `V` are client specified integers. For example running the command
`camonitor BPM0:X.{utag:{M:1,V:1}}` will only show updates for which
`(UTAG & 1) == 1` i.e. the least significant bit of the `UTAG` field is set.

This feature is intended for use by intelligent devices which can provide
contextual information along with a value/alarm/time.  For example, a beam
diagnostic device which is aware of whether a beam signal should be present
(eg. from a global timing system).

#### Link Support

Two new optional methods have been added to the Link Support Entry Table
(`struct lset`): `lset::getAlarmMsg()` and `lset::getTimeStampTag()`. See
comments in dbLink.h for details on implementing these.

Two new accessor functions have also been added which call these methods:
`dbGetAlarmMsg()` and `dbGetTimeStampTag()`.

#### Compatibility

User code that could call these interfaces but must continue to build on older
versions of EPICS may add some of the following macro definitions, and ensure
that the variables referenced by output pointers are initialized.

```c
#ifndef HAS_ALARM_MESSAGE
#  recGblSetSevrMsg(REC, STAT, SEVR, ...) recGblSetSevr(REC, STAT, SEVR)
#endif
#ifndef dbGetAlarmMsg
#  define dbGetAlarmMsg(LINK, STAT, SEVR, BUF, BUFLEN) dbGetAlarm(LINK, STAT, SEVR)
#endif
#ifndef dbGetTimeStampTag
#  define dbGetTimeStampTag(LINK, STAMP, TAG) dbGetTimeStamp(LINK, STAMP)
#endif
```

mdavidsaver:
I think this change is now ready to go.

@anjohnson Let's try something new this time.  Please feel free to edit the "Draft changelog entry" in https://github.com/epics-base/epics-base/pull/71#issuecomment-720668967.

mdavidsaver:
As discussed this past Wednesday, I've changed UTAG to be `epicsUInt64` via a new typedef `epicsUTag`.

I will consider this PR to be approved unless I hear objections by Friday (11th).

mdavidsaver:
FYI.  While I think it is worthwhile going ahead with the change in this PR, I don't think it will be immediately practical to change `timeStamp.userTag` from int32 -> uint64 in QSRV.  Most java clients including cs-studio, and anything using normativeTypesCPP, will break for lack of client side type conversions...

modules/database/src/ioc/db/dbAccessDefs.h
```diff
@@ -107,7 +107,8 @@ epicsShareExtern int dbAccessDebugPUTF;
         epicsUInt16     status;         /* alarm status */\
         epicsUInt16     severity;       /* alarm severity*/\
         epicsUInt16     acks;           /* alarm ack severity*/\
-        epicsUInt16     ackt;           /* Acknowledge transient alarms?*/
+        epicsUInt16     ackt;           /* Acknowledge transient alarms?*/\
+        char amsg[40];
```
anjohnson: Please follow the other char array examples and define a macro `DB_AMSG_SIZE` for the string length, or use `MAX_STRING_SIZE` instead.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -359,6 +359,14 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
...
             *pushort++ = pcommon->ackt;
             pbuffer = (char *)pushort;
+            if (!pfl || pfl->type == dbfl_type_rec) {
+                STATIC_ASSERT(sizeof(pcommon->amsg)==sizeof(pfl->amsg));
+                strncpy(pbuffer, pcommon->amsg, sizeof(pcommon->amsg)-1);
+            } else {
+                strncpy(pbuffer, pfl->amsg,sizeof(pfl->amsg)-1);
+            }
+            pbuffer[sizeof(pcommon->amsg)-1] = '\0';
+            pbuffer += sizeof(pcommon->amsg);
```
anjohnson, line 371: I would much prefer that this code follow the example of the other `DBR_` options defined at the top of dbAccessDefs.h and make both amsg and utag optional fields that get copied if additional bits are set in the `*options` value. That means writing a bit more code, but it means the code style matches and we don't copy data that isn't being asked for; in particular db_access.c would never set the new bits. You can adjust the values of the existing bit constants when inserting `DBR_AMSG` if you want `DBRamsg` to be after `DBRstatus` (the comments in dbAccessDefs.h say the members must appear in the same order as the bits, although I'm not sure if that actually matters as long as the parts always appear in the same order).

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -382,6 +390,7 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
         }
         if( (*options) & DBR_TIME ) {
             epicsUInt32 *ptime = (epicsUInt32 *)pbuffer;
+            epicsUInt64 *ptime64;
```
anjohnson: See above, suggest using `DBR_UTAG`, `DBRutag` and `struct dbr_utag`. This pointer would be called `putag`.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -523,7 +538,7 @@ long dbProcess(dbCommon *precord)
             (precord->lcnt++ < MAX_LOCK) ||
             (precord->sevr >= INVALID_ALARM)) goto all_done;
 
-        recGblSetSevr(precord, SCAN_ALARM, INVALID_ALARM);
+        recGblSetSevrMsg(precord, SCAN_ALARM, INVALID_ALARM, "Async in progress");
```
anjohnson, line 545: There are other reasons why PACT might be true, some device support `init_record()` routines leave it TRUE if the address is illegal. I would suggest something like "Record still active" instead.

modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -338,26 +338,43 @@ static long dbDbGetUnits(const struct link *plink, char *units, int unitsSize)
...
         *severity = precord->sevr;
+    if (msgbuf && msgbuflen) {
+        strncpy(msgbuf, precord->amsg, msgbuflen-1);
+        msgbuf[msgbuflen-1] = '\0';
+    }
     return 0;
 }
 
-static long dbDbGetTimeStamp(const struct link *plink, epicsTimeStamp *pstamp)
+static long dbDbGetAlarm(const struct link *plink, epicsEnum16 *status,
```
anjohnson: If a link type implements the `getAlarmMsg()` method it shouldn't need to provide a `getAlarm()` method as well. Your code in dbLink.c already handles that fall-back, so this method should be deleted.

modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -338,26 +338,43 @@ static long dbDbGetUnits(const struct link *plink, char *units, int unitsSize)
...
 {
     dbChannel *chan = linkChannel(plink);
     dbCommon *precord = dbChannelRecord(chan);
     *pstamp = precord->time;
+    if(ptag)
+        *ptag = precord->utag;
     return 0;
 }
 
+static long dbDbGetTimeStamp(const struct link *plink, epicsTimeStamp *pstamp)
```
anjohnson: As above, this method no longer needs to be provided.

modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -404,7 +421,9 @@ static lset dbDb_lset = {
     dbDbGetPrecision, dbDbGetUnits,
     dbDbGetAlarm, dbDbGetTimeStamp,
```
anjohnson: These two will then become `NULL`

modules/database/src/ioc/db/dbEvent.c
```diff
@@ -682,11 +682,15 @@ db_field_log* db_create_event_log (struct evSubscrip *pevent)
         struct dbChannel *chan = pevent->chan;
         struct dbCommon  *prec = dbChannelRecord(chan);
         pLog->ctx = dbfl_context_event;
+        pLog->mask = pevent->select;
```
anjohnson: I see what you did there (not that I'm objecting)...

modules/database/src/ioc/db/dbLink.c
```diff
@@ -314,10 +314,27 @@ long dbTryGetLink(struct link *plink, short dbrType, void *pbuffer,
...
 
+static
+void setLinkAlarm(struct link* plink)
+{
+    struct dbCommon *precord = plink->precord;
+    dbRecordType *rdes = precord->rdes;
+    const char* amsg = NULL;
+    short i;
+
+    for(i=0; i<rdes->no_links; i++) {
```
anjohnson: This name lookup is already implemented in the routine `dbLinkFieldName()` at the top of this source file...

modules/database/src/ioc/db/dbLink.h
```diff
@@ -282,6 +282,8 @@ typedef struct lset {
      * @param   status      where to put the alarm status (or NULL)
      * @param   severity    where to put the severity (or NULL)
      * @returns status value
+     *
+     * @note Link types which provide getAlarm should also provided getAlarmMsg().
```
anjohnson: Why? Link types that don't need to build against older Base versions don't need to provide the older one at all, and the new method is preferred for all internal link types.

modules/database/src/ioc/db/dbLink.h
```diff
@@ -361,6 +363,29 @@ typedef struct lset {
      * @returns status value
      */
     long (*doLocked)(struct link *plink, dbLinkUserCallback rtn, void *priv);
+
+    /** @brief Extended version of getAlarm
+     *
+     * Equivalent of getAlarm() and also copy out alarm message string.
+     * The msgbuf argument may be NULL and/or msgbuflen==0, in which case
+     * the call must be the same as a call to getAlarm().
```
anjohnson: s/the call must/the result must/

modules/database/src/std/filters/filters.dbd.pod
```diff
@@ -285,3 +285,5 @@ once every minute:
  ...
 
 =cut
+
+registrar(utagInitialize)
```
anjohnson, line 291: Missing documentation belongs here...

modules/database/src/std/link/lnkCalc.c
```diff
@@ -728,23 +740,40 @@ static long lnkCalc_getAlarm(const struct link *plink, epicsEnum16 *status,
...
+    if (msgbuf && msgbuflen) {
+        strncpy(msgbuf, clink->amsg, msgbuflen-1);
+        msgbuf[msgbuflen-1] = '\0';
+    }
 
     return 0;
 }
 
-static long lnkCalc_getTimestamp(const struct link *plink, epicsTimeStamp *pstamp)
+static long lnkCalc_getAlarm(const struct link *plink, epicsEnum16 *status,
```
anjohnson: Not needed.

modules/database/src/std/link/lnkCalc.c
```diff
@@ -728,23 +740,40 @@ static long lnkCalc_getAlarm(const struct link *plink, epicsEnum16 *status,
...
         *pstamp = clink->time;
+        if(ptag)
+            *ptag = clink->utag;
         return 0;
     }
 
     return -1;
 }
 
+static long lnkCalc_getTimestamp(const struct link *plink, epicsTimeStamp *pstamp)
```
anjohnson: Not needed.

anjohnson (changes requested):
Sorry I didn't get to review this in detail sooner, I know I had promised it for last week.

My main issue is described in the second and subsequent inline comments, please add 2 more getOptions bits for the new metadata members that can be fetched by `dbGetField()`.

Whether you accept that or not, there is some code missing from dbTest.c's `printBuffer()` since the `dbtgf()` routine calls `dbGetField()` with all `getOptions` bits set and prints the results.

The other issue is that the link types provided with Base only need to implement the two new `lset` routines and can remove the routines they replace.

anjohnson:
Also needs a few updates to the Release Notes text, and they need adding as an entry to the file.

mdavidsaver:
> Also needs a few updates to the Release Notes text, and they need adding as an entry to the file.

Feel free to edit, I'll copy and commit the contents of https://github.com/epics-base/epics-base/pull/71#issuecomment-720668967 after merging.

modules/database/src/std/filters/filters.dbd.pod
mdavidsaver: I've at least started something.

Also, fyi the links at the top of filters.dbd.pod don't work for me.  The difference appears to be that the anchors below have quotes (or quote like characters) while the link targets don't.

mdavidsaver:
I think I've address @anjohnson comments.

modules/database/src/std/filters/filters.dbd.pod
anjohnson: Yeah, I noticed that a couple of days ago. Fix committed.

anjohnson:
Could you double-check the conditions in the `utag::filter()` method, in particular is the `first` check still appropriate? I thought you fixed QSRV to allow the initial monitor event to be dropped.

I have just updated your Release Notes text in the comment above. In the final Compatibility section could you clarify what the phrase "ensure that the variables referenced by output pointers are initialized" is supposed to mean, I can't really tell from that wording.

modules/database/src/ioc/db/dbLink.h
```diff
@@ -361,6 +361,29 @@ typedef struct lset {
...
+    /** @brief Extended version of getAlarm
+     *
+     * Equivalent of getAlarm() and also copy out alarm message string.
+     * The msgbuf argument may be NULL and/or msgbuflen==0, in which case
+     * the effect must be the same as a call to getAlarm().
+     *
+     * Implementations must write a trailing nil to msgbuf whenever
+     * @code msgbuf!=NULL && msgbuflen>0 @endcode .
+     *
+     * @since Added after 7.0.4.1
```
anjohnson: Reminder to fix `@since` comments...

modules/database/src/std/link/lnkCalc.c
```diff
@@ -53,6 +53,7 @@ typedef struct calc_link {
     } pstate;
     epicsEnum16 stat;
     epicsEnum16 sevr;
+    char amsg[MAX_STRING_SIZE];
```
anjohnson, line 56: Shouldn't this be `AMSG_SIZE`?

anjohnson (approved):
A couple more places to look at, but I think this will be good once they've been addressed.

mdavidsaver:
Merged as of 0edf986c31bb71ab7fcad3010cacadcf69390a40

PR #72, opened 2020-04-24
Add de-initHooks

mdavidsaver:
Extend the initHook idea to `exitDatabase()`.

### DIFF
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -710,20 +710,27 @@ int iocShutdown(void)
 {
     if (iocState == iocVoid) return 0;
 
+    initHookAnnounce(initHookAtShutdown);
+
     iterateRecords(doCloseLinks, NULL);
+    initHookAnnounce(initHookAfterCloseLinks);
 
     if (iocBuildMode == buildIsolated) {
         /* stop and "join" threads */
         scanStop();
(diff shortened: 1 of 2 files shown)


anjohnson:
Seems reasonable. My only request would be to expand slightly on the one-word "conditional" comment against the state definitions to say what the condition is. I'd like to avoid complaints or questions from external module authors when those hooks don't get fired at shutdown in a regular IOC; they shouldn't have to search the code to see what that condition is.

mdavidsaver:
Would "conditional.  testIocShutdownOk() only" be sufficient?

anjohnson:
Something like "triggered only by self-test code" would address the somewhat wider audience I'm thinking about; the word "conditional" isn't very enlightening here.

anjohnson:
Approve.

mdavidsaver:
> Approve.

Noted.  FYI. github has a way to record this [more prominently](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-proposed-changes-in-a-pull-request).

anjohnson (approved):
I was aware this is possible, but they don't make the approval process as easy as it could be – you have to know how to get the website to put up the button.

mdavidsaver:
Cherry picked as 5d5e552a7ec6ef69459c97b0081aa775372a6290

PR #73, opened 2020-05-01
Fix default/osdMessageQueue.cpp

anjohnson:
This branch contains my fixes for the message queue issue ([lp: #1868486](https://bugs.launchpad.net/epics-base/+bug/1868486)), but not the tests that I have also developed since they don't actually trigger any other bugs. Since I have already committed Heinz' fix for the issue that Mark Rivers was triggering I'm not as concerned whether this PR makes it into 3.15.8 or not, it would be nice to demonstrate the other issues that Michael thinks were present first, but I don't see that as essential.


mdavidsaver:
At a glance this looks like the right idea.  `freeEventNode()` should avoid returning a triggered epicsEvent to the free list, and its done safely under a lock.

(also to give Ben his due a9727fd5cbd6c27d4ecf6a61e6984537c3cbc5fe)

anjohnson:
Yay, merging.

PR #74, opened 2020-05-13
Applied all doxy-libcom changes to latest libcom headers

anjohnson:
Doxygen annotations developed at the Documentathon (ESS, 2019) and Codeathon (Diamond, 2020).


modules/libcom/src/osi/epicsEvent.h
```diff
@@ -4,42 +4,113 @@
...
+ * Defines the C++ and C API's for a simple binary semaphore. If multiple threads are
+ * waiting on the same event, only one of them will be woken when the event is signalled.
+ *
+ * The primary use of an event semaphore is for thread synchronization. An example of using an
+ * event semaphore is a consumer thread that processes requests from one or more producer threads.
+ * For example:
+ *
+ * When creating the consumer thread also create an epicsEvent.
+ @code
+   epicsEvent *pevent = new epicsEvent;
```
mdavidsaver, line 23: This is an example of bad c++ practice.  Examples should default to plain/stack construction unless there is a reason to do otherwise..  Doing so with a bare pointer is error prone (eg. example omits a delete).

modules/libcom/src/osi/epicsMessageQueue.h
```diff
@@ -26,22 +27,135 @@ typedef struct epicsMessageQueueOSD *epicsMessageQueueId;
...
+ *  first out basis. It is designed so that a single message queue can
+ *  be used with multiple writer and reader threads.
+ *
+ *  A C++ epicsMessageQueue cannot be assigned to, copy-constructed, or
+ *  constructed without giving the capacity and max-imumMessageSize
+ *  arguments.
+ *
+ *  The C++ compiler will object to some of the statements below:
+ *  \code{.cpp}
+ *  epicsMessageQueue mq0();   // Error: default constructor is private
```
mdavidsaver, line 40: It would be more useful and succinct to state that an epicsMessageQueue is not copyable or default constructable.

modules/libcom/src/osi/epicsMutex.h
```diff
@@ -3,52 +3,132 @@
...
+ *
+ * Mutual exclusion semaphores are for situations requiring exclusive access to
+ * resources. An epicsMutex may be claimed recursively, i.e. taken more than
+ * once by a thread, which must release it as many times as it was taken.
+ * Recursive usage is common for a set of routines that call each other while
+ * working on an exclusive resource.
+ *
+ * The typical C++ use of a mutual exclusion semaphore is:
+ @code
+     epicsMutex *plock = new epicsMutex;
```
mdavidsaver, line 22: Another example of bad practice.  No reason for heap allocation.

Further, modern examples of `epicsMutex` should use `epicsGuard<>`

modules/libcom/src/osi/epicsTime.h
```diff
@@ -19,139 +22,294 @@
...
+     * Returns an epicsTime containing the current time. For example:
+     * \code{.cpp}
+     *   epicsTime now = epicsTime::getCurrent();
+     * \endcode
+     */
     static epicsTime getCurrent ();
+    /** \brief Get current monotonic time
+     *
+     * Returns an epicsTime containing the current monotonic time, a
+     * high-resolution OS clock that counts at a steady rate, never
```
mdavidsaver, line 159: "steady rate" is not guaranteed on all targets.  (cf `man clock_gettime` and `CLOCK_MONOTONIC`).

modules/libcom/src/yajl/yajl_alloc.h
```diff
@@ -16,8 +16,12 @@
...
 /**
  * \file yajl_alloc.h
- * default memory allocation routines for yajl which use malloc/realloc and
- * free
+ * \brief Memory allocation macros for yajl
+ * \author Lloyd Hilaiel
+ *
+ * These macros are used inside YAJL instead of directly calling
+ * malloc(), realloc() or free(). They call the equivalent method
+ * in their \ref yajl_alloc_funcs parameter \a afs.
```
mdavidsaver, line 24: Should we be changing bundled headers in this way?  Seems like this work will either complicate and/or be lost when/if our copy of YAJL is refreshed.

modules/libcom/src/as/asTrapWrite.h
```diff
@@ -18,37 +25,65 @@
...
-    void *data;     /* Might be NULL if no data is available */
+    const char *userid; /**< @brief Userid of whoever orginated the request. */
+    const char *hostid; /**< @brief Hostid of whoever orginated the request. */
+    void *serverSpecific; /**< @brief A field for use by the server.
+        *
+        * Any listener that uses this field must know what type of
+        * server is forwarding the put requests. This pointer holds
+        * the value the server provides to asTrapWriteWithData(), which
+        * for RSRV is the dbChannel pointer for the target field.
+        */
```
mdavidsaver: End of line comments are typically just one line.

modules/libcom/src/calc/postfix.h
```diff
@@ -4,81 +4,357 @@
...
+ * @return Non-zero value in event of error
+ *
+ * It is the callers's responsibility to ensure that \c ppostfix points
+ * to sufficient storage to hold the postfix expression. The macro
+ * INFIX_TO_POSTFIX_SIZE(n) can be used to calculate an appropriate
+ * postfix buffer size from the length of the infix buffer.
+ *
+ * @note "n" must count the terminating nil byte too.
+ *
+ * -# The **infix expressions** that can be used are very similar
```
mdavidsaver, line 131: While good information, the remainder of this block doesn't directly pertain to `postfix()`.  An essay like this could easily be a separate [`@page calcsyntax CALC Syntax`](http://www.doxygen.nl/manual/commands.html#cmdpage) (with a stable URL) at the beginning or end of this file, which could then be referenced from with [`@ref calcsyntax`](http://www.doxygen.nl/manual/commands.html#cmdref) from here.

modules/libcom/src/cppStd/epicsAlgorithm.h
```diff
@@ -3,71 +3,105 @@
...
+ * @file epicsAlgorithm.h
+ * @author Jeff Hill & Andrew Johnson
+ *
+ * @brief Contains a few templates out of the C++ standard header algorithm
+ *
+ * @note The templates are provided here in a much smaller file. Standard algorithm
+ * contains many templates for sorting and searching through C++ template containers
+ * which are not used in EPICS. If all you need from there is std::min(),
+ * std::max() and/or std::swap() your code may compile faster if you include
+ * epicsAlgorithm.h and use epicsMin(), epicsMax() and epicsSwap() instead.
```
mdavidsaver, line 19: "faster" is a dubious claim.  IMO. If this header is to be documented, then it should also be deprecated.

modules/libcom/src/ellLib/ellLib.h
```diff
@@ -19,46 +32,185 @@
...
+ * @return The node's index, or -1 if it cannot be found on the list.
+ * @note The first node has index 1.
+ */
 epicsShareFunc int  ellFind (ELLLIST *pList, ELLNODE *pNode);
+
 typedef int (*pListCmp)(const ELLNODE* A, const ELLNODE* B);
-epicsShareFunc void ellSortStable(ELLLIST *pList, pListCmp);
+/**
+ * @author Michael Davidsaver
+ * @date 2016
```
mdavidsaver: I'm not enthused about in-line authorship in general now that we're >20 years into the VCS era.  Less so outside of file headers.  (and `@date` is as helpful as Jeff's phone number)  Anyone can `git blame` me going forward.

modules/libcom/src/misc/dbDefs.h
```diff
@@ -54,15 +68,18 @@
...
-/*Process Variable Name Size */
-/* PVNAME_STRINGSZ includes the nil terminator */
+/** @brief Size of a record name including the nil terminator */
 #define PVNAME_STRINGSZ 61
+/** @brief Size of a record name without the nil terminator */
 #define PVNAME_SZ (PVNAME_STRINGSZ - 1)
 
-/* Buffer size for the string representation of a DBF_*LINK field */
+/**
+ * @def PVLINK_STRINGSZ
```
mdavidsaver: `@def` stands out here.  Was this needed to quiet some warning?

modules/libcom/src/misc/epicsExport.h
```diff
@@ -1,43 +1,126 @@
...
 #include <shareLib.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 typedef void (*REGISTRAR)(void);
 
-#define EPICS_EXPORT_POBJ(typ,obj) pvar_ ## typ ## _ ## obj
-#define EPICS_EXPORT_PFUNC(obj) pvar_func_ ## obj
```
mdavidsaver, line 25: Are these two definitions considered public interface?

modules/libcom/src/misc/epicsUnitTest.h
```diff
@@ -5,8 +5,142 @@
...
+{
+    iocBuildIsolated() || iocRun();
+
+    ... test code ...
+
+    iocShutdown();
+    dbFreeBase(pdbbase);
+    registryFree();
+    pdbbase = NULL;
+    return testDone();
```
mdavidsaver, line 99: Example needs correcting.  And should be moved to `dbUnitTest.h`.  Current practice is:

```
void <dbdname>_registerRecordDeviceDriver(struct dbBase *);

MAIN(iocTest)
{
    testPlan(0);
    testdbPrepare();
    testdbReadDatabase("<dbdname>.dbd", 0, 0);
    <dbdname>_registerRecordDeviceDriver(pdbbase);
    testdbReadDatabase("some.db", 0, 0);
    testIocInitOk();
    /* IOC running */
    testIocShutdownOk();
    testdbCleanup();
    return  testDone();
```

mdavidsaver (commented):
Quite a few comments, but no blockers.  Although the example of epicsEvent and epicsMutex do need correction.

modules/libcom/src/osi/epicsEvent.h
anjohnson: Aaron Long took this example verbatim from the Application Developers Guide. I'm not disagreeing with you on the C++ best  practices aspect, but in this particular case we don't have any instances of creating an epicsEvent on the stack that I can find (not surprisingly). Maybe just remove this particular example completely?

modules/libcom/src/osi/epicsMessageQueue.h
anjohnson: It already does in the previous paragraph, although that wording isn't quite as succinct. Again, this is verbatim from the AppDevGuide. I don't have a problem with us changing any of the wording, but I don't intend to do all of that myself. I wasn't planning on publishing the Doxygen output yet anyway, so we don't have to make these edits before merging and generating the next release.

mdavidsaver: > we don't have to make these edits before merging and generating the next release.

Go ahead and merge (I said "no blockers").  I will then correct the three examples.

PR #75, opened 2020-05-20
Update and improve documentation of simulation mode

ralphlange:


mdavidsaver:
Looks fine.  I'll merge today.

I do wonder if there is a way to avoid the repetition?

ralphlange:
As long as the fields are in many record types, they have to appear in their DBDs.

I found only adding the link to the other page too short for a reference manual, so I added a telegram style paragraph to each of the record pages.

Note that there are subtle differences for input vs. output, conversion vs. none, SVAL buffering vs. none - I think there are four different versions. Also, the type of SVAL has to adapt to the type of VAL, so a DBD include would have to take parameters or use macros.

anjohnson (approved):
Looks good. There are a few places where the Common docs say "link can be a constant, a database link, or a channel access link" which doesn't encompass the new link types, but these are not the only places where those phrases appear so I'm okay with not bothering to fix them here.

Thanks!

PR #76, opened 2020-06-01
add dynamic_registerRecordDeviceDriver()

mdavidsaver:
Use epicsFindSymbol() to build a dynamic and idempotent equivalent
to the code generated by registerRecordDeviceDriver.pl

Also add runRegistrarOnce() to prevent registrars
from running multiple times, except from unittest code.

### DIFF
--- a/configure/CONFIG_DATABASE_MODULE
+++ b/configure/CONFIG_DATABASE_MODULE
@@ -24,3 +24,5 @@ MSI3_15                    = $(EPICS_DATABASE_HOST_BIN)/msi$(HOSTEXE)
 
 # Libraries needed to link a basic IOC
 EPICS_BASE_IOC_LIBS = dbRecStd dbCore ca Com
+
+HAS_registerAllRecordDeviceDrivers=YES
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -17,6 +17,19 @@ should also be read to understand what has changed since earlier releases.
 
 <!-- Insert new items immediately below here ... -->
 
+### Add registerAllRecordDeviceDrivers()
+
+Addition of registerAllRecordDeviceDrivers() as an iocsh function
+and in iocshRegisterCommon.h.  This function uses dynamic lookup with epicsFindSymbol()
+to perform the same function as a generated \*_registerRecordDeviceDriver() function.
+This allows dynamic loading/linking of support modules without code generation.
+
+This feature is not intended for use by IOCs constructed using the standard EPICS application
+build process and booted from a startup script in an iocBoot subdirectory, although it might
+work in some of those cases (the IOC's registerRecordDeviceDriver.cpp file is still required
+to link everything into the executable). It also won't work with some static build
+configurations or where the symbol table has been stripped from the executable.
+
 ## EPICS Release 7.0.4.1
 
 ### ARM Architecture Changes
--- a/modules/database/src/ioc/db/dbUnitTest.c
+++ b/modules/database/src/ioc/db/dbUnitTest.c
@@ -33,6 +33,7 @@
 #include "initHooks.h"
 #include "iocInit.h"
 #include "errSymTbl.h"
+#include "iocshRegisterCommon.h"
 
 static dbEventCtx testEvtCtx;
 static epicsMutexId testEvtLock;
@@ -98,6 +99,7 @@ void testdbCleanup(void)
     db_cleanup_events();
     initHookFree();
     registryFree();
+    clearRegistrarOnce();
     pdbbase = NULL;
     dbmfFreeChunks();
 }
--- a/modules/database/src/ioc/misc/Makefile
+++ b/modules/database/src/ioc/misc/Makefile
@@ -25,6 +25,6 @@ dbCore_SRCS += iocInit.c
 dbCore_SRCS += miscIocRegister.c
 dbCore_SRCS += dlload.c
 dbCore_SRCS += iocshRegisterCommon.c
-dbCore_SRCS += dynamic_registerRecordDeviceDriver.cpp
+dbCore_SRCS += registerAllRecordDeviceDrivers.cpp
 
 miscIocRegister_CFLAGS_iOS = -DSYSTEM_UNAVAILABLE
--- a/modules/database/src/ioc/misc/Makefile
+++ b/modules/database/src/ioc/misc/Makefile
@@ -25,5 +25,6 @@ dbCore_SRCS += iocInit.c
 dbCore_SRCS += miscIocRegister.c
 dbCore_SRCS += dlload.c
 dbCore_SRCS += iocshRegisterCommon.c
+dbCore_SRCS += dynamic_registerRecordDeviceDriver.cpp
 
 miscIocRegister_CFLAGS_iOS = -DSYSTEM_UNAVAILABLE
--- a/modules/database/src/ioc/misc/dynamic_registerRecordDeviceDriver.cpp
+++ b/modules/database/src/ioc/misc/dynamic_registerRecordDeviceDriver.cpp
@@ -1,0 +1,240 @@
+/*************************************************************************\
+* Copyright (c) 2020 Michael Davidsaver
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+#include <exception>
+#include <string>
+#include <set>
+#include <map>
+
+#include <string.h>
+
+#define EPICS_PRIVATE_API
+
+#include <iocsh.h>
+
+#include <epicsStdio.h>
+#include <epicsFindSymbol.h>
+#define epicsExportSharedSymbols
+#include <registryRecordType.h>
+#include <registryDeviceSupport.h>
+#include <registryDriverSupport.h>
+#include <iocshRegisterCommon.h>
+#include <registryCommon.h>
+#include <dbAccess.h>
+#include <dbBase.h>
+#include <dbStaticLib.h>
+
+namespace {
+
+struct compareLoc {
+    bool operator()(const recordTypeLocation& lhs, const recordTypeLocation& rhs) const
+    {
+        if(lhs.prset<rhs.prset)
+            return true;
+        else if(lhs.prset>rhs.prset)
+            return false;
+        return lhs.sizeOffset<rhs.sizeOffset;
+    }
+};
+
+// storage which will be referenced by pdbbase.  Must never be free'd
+std::set<std::string> names;
+std::set<recordTypeLocation, compareLoc> reclocs;
+
+template<typename T, typename Y>
+const T& intern(std::set<T,Y>& coll, const T& val)
+{
+    return *coll.insert(val).first;
+}
+
+std::set<void*> registrarsRun;
+
+// gcc circa 4.4 doesn't like iocshVarDef[2] as mapped_type
+struct varDef {
+    iocshVarDef def[2];
+};
+typedef std::map<std::string, varDef> vardefs_t;
+vardefs_t vardefs;
+
+template<typename T>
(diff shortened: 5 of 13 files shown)


mdavidsaver:
Now that epicsFindSymbol() is working on host targets, provide a way to forgo generated code in some situations (dynamic loading/linking).

@dirk-zimoch May be of interest.

dirk-zimoch:
What I do for dynamically loaded modules:
1. Let the EPICS build system generate <module>_registerRecordDeviceDriver (as for an application but make sure only the dbd files of the module are used, none of base or module dependencies like asyn).
2. Load the module library (resolve dependencies and check for version incompatibilities)
3. Load the dbd file with dbLoadDatabase
4. Pass `"<module>_registerRecordDeviceDriver"` to iocshCmd.

In principle, step 1 does not need to be done at build time but could be done instead of step 4 at load time (or as part of step 3). Is this what you mean?

mdavidsaver:
> Is this what you mean?

Yes, this is what `dynamic_registerRecordDeviceDriver()` does.

mdavidsaver:
I've added some usage in `analogMonitorTest.c` in order to have some test coverage.

Since I see this addition as low risk, I'm inclined to merge soon unless there are objections.

@anjohnson Any comments on naming?  I added `dynamic_registerRecordDeviceDriver()` to `iocshRegisterCommon.h` for lack of a better place.  I see it as primarily being call through iocsh, though I'll want to call it directly in some applications.

anjohnson:
Are the warnings below (from the first Appveyor dynamic build) expected from running analogMonitorTest? I think they imply that the `iocshRegisterVariable()` routine in iocsh.cpp is seeing a second copy of each of these variables with a different pointer, which would imply that bad things may be happening during linking on Windows dynamic.

```
perl -CSD analogMonitorTest.t -tap > analogMonitorTest.tap
Warning: iocshRegisterVariable redefining CASDEBUG.
Warning: iocshRegisterVariable redefining asCaDebug.
Warning: iocshRegisterVariable redefining atExitDebug.
Warning: iocshRegisterVariable redefining boHIGHlimit.
Warning: iocshRegisterVariable redefining boHIGHprecision.
Warning: iocshRegisterVariable redefining calcoutODLYlimit.
Warning: iocshRegisterVariable redefining calcoutODLYprecision.
Warning: iocshRegisterVariable redefining callbackParallelThreadsDefault.
Warning: iocshRegisterVariable redefining dbAccessDebugPUTF.
Warning: iocshRegisterVariable redefining dbBptNotMonotonic.
Warning: iocshRegisterVariable redefining dbConvertStrict.
Warning: iocshRegisterVariable redefining dbJLinkDebug.
Warning: iocshRegisterVariable redefining dbQuietMacroWarnings.
Warning: iocshRegisterVariable redefining dbRecordsAbcSorted.
Warning: iocshRegisterVariable redefining dbRecordsOnceOnly.
Warning: iocshRegisterVariable redefining dbTemplateMaxVars.
Warning: iocshRegisterVariable redefining dbThreadRealtimeLock.
Warning: iocshRegisterVariable redefining histogramSDELprecision.
Warning: iocshRegisterVariable redefining lnkDebug_debug.
Warning: iocshRegisterVariable redefining logClientDebug.
Warning: iocshRegisterVariable redefining seqDLYlimit.
Warning: iocshRegisterVariable redefining seqDLYprecision.
Starting iocInit
```

It looks a bit strange that the `analogMonitorTest` code is calling the `dynamic_registerRecordDeviceDriver()` routine twice in addition to `analogMonitorTest_registerRecordDeviceDriver()`. I wonder if recMiscTest.c might be a better place to put that particular check since it really has nothing to do with anything else going on in the analogMonitorTest code – it seems a bit more misc. There is no way to detect an error from the registration routine so the test (or an IOC booting) could actually fail?

The name `dynamic_registerRecordDeviceDriver()` seems fine except it would clash with any IOC out there called 'dynamic'. This doesn't bother me as long as you mention that in the Release Notes, but it might be a reason for changing the name to something that could never clash, like `registerAllRecordDeviceDrivers()` say (if that's what it actually does).

The new command registration is missing any `usage` text, and it would be nice to see the Release Notes entry for this change sooner rather than later. Are you planning to modify the makeBaseApp templates to use this instead after we've been running with it for a while?

mdavidsaver:
> Warning: iocshRegisterVariable redefining CASDEBUG

While not a bug in this context, this annoyance shows that my ebe7cf046e286d350cacf40afa0bcd6cc3b5ca46 was too strict.  The actual `iocshVarDef*` can be different depending on the caller code.  What is important is that the contents be the same (variable name, type, and address).

mdavidsaver:
> `registerAllRecordDeviceDrivers()` say (if that's what it actually does).

Ok, I'll rename it.

mdavidsaver:
> It looks a bit strange that the `analogMonitorTest` code is calling the `dynamic_registerRecordDeviceDriver()` routine twice in addition to `analogMonitorTest_registerRecordDeviceDriver()`.

I agree it is strange, but intentional.  Would this comment clarify?

> /* test registerAllRecordDeviceDrivers() to ensure that both methods of registration are idempotent */

anjohnson:
That comment clarifies the intent, although it isn't obviously testing what it says – AFAICT it can't actually fail, it just relies on the user seeing warning messages at runtime. My suggestion to move it to recMiscTest.c was to keep unrelated tests separate, and this check really has nothing to do with determining whether analog monitoring still works. It would be nice to have some tests that actually demonstrate failures as well (e.g. manually register a different address before calling `registerAllRecordDeviceDrivers()` say).

BTW Base-3.14 had a routine for use on VxWorks which could do registrations using symbol lookups, see [this file](https://github.com/epics-base/epics-base/blob/3.14/src/vxWorks/registerRecordDeviceDriver.c) for the code. I don't think it got much use though because we still needed the generated register_*.cpp file so the linker would know which object files to pull out of the libraries and into the IOC binary.

mdavidsaver:
Requested changes made.

mdavidsaver:
Unless further concerns raised in the next day or so, I'll consider this accepted.

ralphlange:
Code looks pretty good.

What I would like to see:
- a `HAS_*` macro (cf. our recent communication about programmatically checking for features).
- probably also one for make, since the difference is more likely to affect Makefiles than code, I guess.
- a paragraph in the release notes.

mdavidsaver:
>     * a `HAS_*` macro

I will add `HAS_registerAllRecordDeviceDrivers`.

I've started final [test](https://travis-ci.org/github/mdavidsaver/epics-base/builds/702489487) [builds](https://ci.appveyor.com/project/mdavidsaver/epics-base/builds/33764554) merging this branch and #79.

anjohnson:
Want to make sure you saw that even some succeeding Appveyor static builds are failing to find symbols [when running recMiscTest](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/33634316/job/o1ojmyt1vcu3v8c9#L10826):

```
make[5]: Entering directory 'C:/projects/epics-base/modules/database/test/std/rec/O.windows-x64-static'
perl -CSD recMiscTest.t -tap > recMiscTest.tap
Unable to find symbol 'pvar_rset_boRSET' : The specified procedure could not be found.
Unable to find symbol 'pvar_func_boRecordSizeOffset' : The specified procedure could not be found.
Unable to find support for record type 'bo' : The specified procedure could not be found.
Unable to find symbol 'pvar_jlif_lnkConstIf' : The specified procedure could not be found.
Unable to find link support 'lnkConstIf' : The specified procedure could not be found.
make[5]: Leaving directory 'C:/projects/epics-base/modules/database/test/std/rec/O.windows-x64-static'
```

Should the two new calls in recMiscTest.c have their return status value checked and trigger a test failure? I would want those kinds of failure in a real IOC startup script to trigger an iocsh "on error" handler in normal usage.

I also noticed that many of your `fprintf(stderr, ...)` calls seem to be missing a newline at the end of the string, and most of those symbols named in the errors above are not "procedures" so the second half of the "not found" message could use some adjustment (when used in a backtrace that message would be correct, but it isn't here).

mdavidsaver:
These "failures" are to be expected with static builds, especially on windows.  I wasn't so bother by the noise, but I could hide it with some `#ifdef`.  Since I'll be fixing the `fprintf`'s anyway, I'll do this.  This will let the test reasonably check the return value.

mdavidsaver:
> The specified procedure could not be found.

This is a Windows error message as returned via `epicsLoadError()`.  I would rather report it verbatim.

anjohnson:
> "failures" are to be expected with static builds

That needs documenting, so users understand when they can & can't expect to use `registerAllRecordDeviceDrivers()` in an IOC startup script. If it will never work with a static build (I don't know if you expect that to be the case or not) maybe the routine should explain that when built statically.

> This is a Windows error message

Ah, okay.

mdavidsaver:
Draft of release notes

```markdown
### Add registerAllRecordDeviceDrivers()

Addition of registerAllRecordDeviceDrivers() as an iocsh function
and in iocshRegisterCommon.h.  This function uses dynamic lookup with epicsFindSymbol()
to perform the same function as a generated \*_registerRecordDeviceDriver() function.
This allows dynamic loading/linking of support modules without code generation.
```

I don't really see `registerAllRecordDeviceDrivers()` appearing in start scripts a la. `makeBaseApp.pl` (though others may have different ideas).  The generated registration code is still needed to link an executable.  The closest I can imagine would be following some calls to `dlload,` or equivalent, which I think would be clearly different from the usual recipes.  eg. `softIoc` with (unmodified) autosave:

```
#!./bin/linux-x86_64/softIoc

dbLoadDatabase "../autosave/dbd/asSupport.dbd"
dlload "../autosave/lib/$(ARCH)/libautosave.so"
registerAllRecordDeviceDrivers()

dbLoadRecords("tick.db") # arbitrary records
```

mdavidsaver:
Fixed `printf()`s and made test conditional on dynamic linking.

anjohnson:
> I don't really see `registerAllRecordDeviceDrivers()` appearing in start scripts

I would like to see that made clear in the Release Notes entry, otherwise lots of people are likely to try it and ask on tech-talk why it doesn't work. It also isn't particularly useful for test code since tests run on static builds won't work (for some arch's at least). When/where do you see this being used? The "PSI approach to building IOCs" seems to be the only real user left AFAICT.

mdavidsaver:
> I would like to see that made clear in the Release Notes entry

Can you suggest some wording?  All I can think of is a sentence along the lines of:

```
registerAllRecordDeviceDrivers() requires that executable and/or libraries include export/symbol names,
and may not function correctly when linking statically (especially on Windows).
```

> It also isn't particularly useful for test code since tests run on static builds won't work

Which is why this test is now skipped when `STATIC_BUILD!=NO`.

> The "PSI approach to building IOCs" seems to be the only real user left AFAICT.

ESS is the other I know of with certainty, which at this point might be called a derivative.  One of the complications with these approaches has been the need to modify every module build to include generated registration code fragments.  `registerAllRecordDeviceDrivers()` avoids this, as can be seen with my previous example with softIoc and autosave.

> When/where do you see this being used?

So far, this patch has been shipping with my `epicscorelibs` python module, as part of unittests of libca bindings (@thomascobb 's project from the Codeathon).  This gives more assurance that the dynamic binding definitions match the libca ABI.

It makes getting a basic softIoc running quite a bit simpler (eg. for training classes).

```
pip install epicscorelibs
python -m epicscorelibs.ioc -d some.db
```

anjohnson:
> Can you suggest some wording?

"This feature is not intended for use by IOCs constructed using the standard EPICS application build process and booted from a startup script in an iocBoot subdirectory, although it might work in some of those cases (the IOC's registerRecordDeviceDriver.cpp file is still required to link everything into the executable). It also won't work with some static build configurations or where the symbol table has been stripped from the executable."

That wording is intended to be slightly scary, to dissuade people from trying very hard to use it.

mdavidsaver:
Applied as of df6981ceae12a2a0099f282540b0333cf9abbab5

PR #78, opened 2020-06-13
Restrict charactor set for record names.

mdavidsaver:
This change converts the warning in `dbRecordHead()` added by bdbada28a82ef88a7005653e979ae7bc2eb603be (in 2012) to an error, and also add `'\t'`.

It further adds a new warning for record names which begin with a number or `'{'`.

The bulk of the changeset ensure an error in `dbRecordHead()` doesn't escalate into a crash.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1065,15 +1065,12 @@ int dbRecordNameValidate(const char *name)
     for(; *pos; i++, pos++) {
         char c = *pos;
         if(i==0) {
-            /* first charactor restrictions */
-            if(c >= '0' && c <= '9') {
-                errlogPrintf("Warning: Record name '%s' should not begin with a number\n", name);
-
-            } else if(c=='-' || c=='+' || c=='[' || c=='{') {
+            /* first character restrictions */
+            if(c=='-' || c=='+' || c=='[' || c=='{') {
                 errlogPrintf("Warning: Record name '%s' should not begin with '%c'\n", name, c);
             }
         }
-        /* any charactor restrictions */
+        /* any character restrictions */
         if(c < ' ') {
             errlogPrintf("Warning: Record name '%s' should not contain non-printable 0x%02u\n",
                          name, (unsigned)c);
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1050,27 +1050,52 @@ static void dbBreakBody(void)
     }
     pgphentry->userPvt = pnewbrkTable;
 }
-
-static void dbRecordHead(char *recordType, char *name, int visible)
+
+static
+int dbRecordNameValidate(const char *name)
 {
-    char *badch;
-    DBENTRY *pdbentry;
-    long status;
+    size_t i=0u;
+    const char *pos = name;
 
     if (!*name) {
         yyerrorAbort("dbRecordHead: Record name can't be empty");
-        return;
+        return 1;
     }
-    badch = strpbrk(name, " \t\"'.$");
-    if (badch) {
-        epicsPrintf("Error: Bad character '%c' in record name \"%s\"\n",
-            *badch, name);
-        yyerrorAbort(NULL);
-        return;
-    } else if((*name >= '0' && *name <= '9') || *name=='{') {
-        epicsPrintf("Warning: Bad character '%c' begins record name \"%s\"\n",
-            *name, name);
+
+    for(; *pos; i++, pos++) {
+        char c = *pos;
+        if(i==0) {
+            /* first charactor restrictions */
+            if(c >= '0' && c <= '9') {
+                errlogPrintf("Warning: Record name '%s' should not begin with a number\n", name);
+
+            } else if(c=='-' || c=='+' || c=='[' || c=='{') {
+                errlogPrintf("Warning: Record name '%s' should not begin with '%c'\n", name, c);
+            }
+        }
+        /* any charactor restrictions */
+        if(c < ' ') {
+            errlogPrintf("Warning: Record name '%s' should not contain non-printable 0x%02u\n",
+                         name, (unsigned)c);
+
+        } else if(c==' ' || c=='\t' || c=='"' || c=='\'' || c=='.' || c=='$') {
+            epicsPrintf("Error: Bad character '%c' in record name \"%s\"\n",
+                c, name);
+            yyerrorAbort(NULL);
+            return 1;
+        }
     }
+
+    return 0;
+}
+
+static void dbRecordHead(char *recordType, char *name, int visible)
+{
+    DBENTRY *pdbentry;
+    long status;
+
+    if(dbRecordNameValidate(name))
+        return;
 
     pdbentry = dbAllocEntry(pdbbase);
     if (ellCount(&tempList))
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -136,12 +136,14 @@ static void allocTemp(void *pvoid)
 static void *popFirstTemp(void)
 {
     tempListNode        *ptempListNode;
-    void                *ptemp;
+    void                *ptemp = NULL;
 
     ptempListNode = (tempListNode *)ellFirst(&tempList);
-    ptemp = ptempListNode->item;
-    ellDelete(&tempList,(ELLNODE *)ptempListNode);
-    freeListFree(freeListPvt,ptempListNode);
+    if(ptempListNode) {
+        ptemp = ptempListNode->item;
+        ellDelete(&tempList,(ELLNODE *)ptempListNode);
+        freeListFree(freeListPvt,ptempListNode);
+    }
     return(ptemp);
 }
 
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1058,7 +1058,7 @@ int dbRecordNameValidate(const char *name)
     const char *pos = name;
 
     if (!*name) {
-        yyerrorAbort("dbRecordHead: Record name can't be empty");
+        yyerrorAbort("Error: Record/Alias name can't be empty");
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Additional warnings I am considering.

* ascii non-printable characters < 0x20 (space).  Bell, backspace, etc.
* First char in `[+-`.  Those which often appear in constants.

or going further

* First char not in `A-Za-z_` (like C)

I would not go so far as to exclude characters >= 0x80 (includes utf-8).  Nor would I add further restricts on characters after the first.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1045,9 +1061,14 @@ static void dbRecordHead(char *recordType, char *name, int visible)
...
+    badch = strpbrk(name, " \t\"'.$");
     if (badch) {
-        epicsPrintf("Bad character '%c' in record name \"%s\"\n",
+        epicsPrintf("Error: Bad character '%c' in record name \"%s\"\n",
+            *badch, name);
+        yyerrorAbort(NULL);
+        return;
+    } else if((*name >= '0' && *name <= '9') || *name=='{') {
+        epicsPrintf("Warning: Bad character '%c' begins record name \"%s\"\n",
             *badch, name);
```
simon-ess: Shouldn't this now be *name, not *badch?

mdavidsaver: Yes, it should.  Good catch.

mark0n:
I like @mdavidsaver's proposal for additional warnings. My feeling is that this will help us catch stupid mistakes and shouldn't cause too much noise.

dirk-zimoch:
Shouldn't we allow more chars instead of less? UTF-8 for example?
I agree with banning everything <= \u20, as well as `[{+-0-9` as the first char, but why can we not have record names in our native language?

ralphlange:
No part of EPICS Base is adequately handling multi-byte characters. Any of the C string handling functions will happily cut through a multi-byte sequence, rendering the result an illegal string. The assumption that Byte=Character is literally everywhere.
(Fun fact: `DBR_STRING` would effectively be variable-length, depending on which characters it contains.)

dirk-zimoch:
Yes, of course one has to search the code carefully to find potential problems. Luckily in UTF-8 0x00 is still the string terminator. Thus strlen, strcpy etc happily work on UTF-8. Only when splitting a string is necessary, don't do it in the middle of a multi-byte sequence. But that is easy enough to find: If (c&0xc0)==0x80 then c is a following byte of a multi-byte sequence. Cut the string earlier.

anjohnson:
It's not just where to cut strings although there are probably quite a lot of those, there are so many other potential sources of problems – e.g. printBuffer() called by `dbpr` in dbTest.c splits up long strings into chunks when displaying field values to provide nice tabulation and avoid going over the end of a line; in addition to finding the fight place to put the line breaks it would need to know how wide each character was, and how to handle right-to-left text (an IOC running on VxWorks or RTEMS through a serial port can't easily answer those questions). Should macros be allowed to contain strings of high-bit characters that don't make up a complete Unicode character? How is the code supposed to know which ones are and which ones aren't?

These aren't problems that are easy to solve, and any solution would significantly expand the code-base.

mdavidsaver:
I think we are straying from the topic at hand.  This PR is about potential restrictions on record names, not arbitrary field values (eg. DESC).

As I read it, @mark0n and @dirk-zimoch are in favor of restricting:

>     * ascii non-printable characters < 0x20 (space).  Bell, backspace, etc.
>     * First char in `[+-`.  Those which often appear in constants.

But not going so far as

>     * First char not in `A-Za-z_` (like C)

anjohnson:
For the record I am happy with the favored additional restrictions on PV names (you already have "first char not `{`" which is good). The APS has 240 records (out of over 650,000 total) that would fail the "First char not in `A-Za-z_`" rule, they all start with a number, either of an experimental sector or a building.

mdavidsaver:
Additional warnings added.  `dbRecordHead()` seems to be getting too long, so I've broken the name checking code out as `dbRecordNameValidate()`.

> APS has 240 records (out of over 650,000 total)

0.03% is a good place to be.  Do you have an estimate of `R` though?  Are they spreading?

anjohnson:
Spelling: s/charactor/character/g (at least twice in comments)

You said you were _not_ going as far as requiring a non-digit first character, but then you added code that prints a warning about PVs starting with a digit. I had previously checked the names for accelerator PVs and came up with that 240 figure, but Ned just pointed out that the [APS naming convention](https://www.aps.anl.gov/BCDA/EPICS-Process-Variable-Naming-Conventions) recommended that all beamline PV names start with their 2-digit sector number (or a lower-case letter), so there may be tens of thousands of PVs here which would be affected by that. Yes, that does mean those 240 accelerator were probably breaking our naming convention and should get fixed, but that's our problem.

I see no technical reason for the non-digit requirement, it might slow down parsing of those PV names in links very slightly, but that isn't a reason to disallow them.

ralphlange:
Field names are not allowed to be numbers, right?
Else `123.456` would be either a float or a `<record>.<field>` combination.

ralphlange:
Speaking of... would `0xFFFF` be parsed as a constant or as a PV?

anjohnson:
@ralphlange correct, field names must follow the C++ identifier rules so can't start with a digit (or be a C++ keyword). We already have checks for that in the DBD processing chain.

anjohnson:
Good question. Currently (with an EPICS R7.0.4.1 IOC):
```
INPA: CONSTANT 0xFFFF                   INPB: CA_LINK 0aFFFF NPP NMS            
INPC: CA_LINK 0bFFFF NPP NMS            INPD: CONSTANT 0XFFFF                   
INPE: CA_LINK 0oFFFF NPP NMS            INPF: CA_LINK 5u NPP NMS                
INPG: CA_LINK 0l NPP NMS                INPH: CONSTANT 0xdeadbeef               
```

mdavidsaver:
> You said you were _not_ going as far as requiring a non-digit first character

That's not quite what I've implemented, but this is a case of my thinking one thing and saying another.  My mistake.

> all beamline PV names start with their 2-digit sector number

Well, that's conclusive.  Applying the same logic as `{` wrt. NSLS2, we can't orphan BCDA.

mdavidsaver:
> Speaking of... would `0xFFFF` be parsed as a constant or as a PV?

Link parsing is done with `epicsParseDouble()` so I think a PV.

anjohnson:
As @mdavidsaver said we parse links using `epicsParseDouble()`:

```C
    /* Link is a constant if empty or it holds just a number */
    if (len == 0 || epicsParseDouble(pstr, &value, NULL) == 0) {
        pinfo->ltype = CONSTANT;
        return 0;
    }
```

He got the conclusion wrong though since C99 `strtod()` accepts hex numbers.

This is off-topic again, but an interesting corollary to that fact:

```
INPA: CONSTANT 0x1.2a4                  INPB: CONSTANT 0x1.a4                   
```

The above is on MacOS which should match Linux and RTEMS; Windows and VxWorks will behave differently though since this depends on how `epicsStrtod` is declared in osdStrtod.h, see below.

Both of the above "CONSTANT" strings are valid hex double literals too. `epicsParseDouble()` calls `epicsStrtod()` which may or may not be the OS-supplied `strtod()`. Our implementation is used when the OS-supplied version is broken, see the tail of the `epicsStdlibTest` output for test results on those OSs, and it checks for leading `0x` characters and passes the whole string to `strtol(..., 16)` or `strtoul(..., 16)` if it finds them. Hmm, there's a couple of bugs, it should now be using `strtoll()` and `strtoull()`, and even then that code still won't accept hex doubles that contain decimal parts.

I just looked at some recent epicsStdlibTest.tap files from Appveyor, and recent versions of `strtod()` do now pass our tests, so we could put some conditionals into the os/WIN32/osdStrtod.h header:

+ [vs2019 - pass](https://ci.appveyor.com/api/buildjobs/7m6ik8urd3lxs43j/artifacts/epicsStdlibTest.tap)
+ [vs2017 - pass](https://ci.appveyor.com/api/buildjobs/a2g63mqunp49lama/artifacts/epicsStdlibTest.tap)
+ [vs2015 - pass](https://ci.appveyor.com/api/buildjobs/m7j3h581dyyi8dqc/artifacts/epicsStdlibTest.tap)
+ [vs2013 - fail](https://ci.appveyor.com/api/buildjobs/02v8s9oac3sc3txc/artifacts/epicsStdlibTest.tap)
+ [vs2012 - fail](https://ci.appveyor.com/api/buildjobs/tjaisgt6c16v4bmo/artifacts/epicsStdlibTest.tap)
+ [vs2010 - fail](https://ci.appveyor.com/api/buildjobs/amnwqc1f0lnhbsth/artifacts/epicsStdlibTest.tap)
+ [mingw - pass](https://ci.appveyor.com/api/buildjobs/u8cjire3w0c6eear/artifacts/epicsStdlibTest.tap)

I'm not sure which CRT library the mingw version will be running, does it come with its own or use the vs2015 which I believe is the default for Appveyor if you don't ask for something different?

mdavidsaver:
> Spelling: s/charactor/character/g (at least twice in comments)

After some trial and error I have built and setup spell check for qtcreator (https://github.com/CJCombrink/SpellChecker-Plugin) which will hopefully help with my wonderful spelling.

mdavidsaver:
> 0x1.2a4

Definitely off-topic.  Good fodder for https://www.ioccc.org/ I suppose.

mdavidsaver:
On-topic.  I've removed the number prefix check.

anjohnson:
Thanks.

Suggestion: Now that you've added the nice `dbRecordNameValidate()` routine we should use it to check record alias names as well, in both `dbRecordAlias()` and `dbAlias()`.

Hmm, is dbLexRoutines.c even the best place for doing these checks? If you moved the validator into dbStaticLib.c it could be called from `dbCreateAlias()` and `dbCreateRecord()`. My bad for the original location, but if we ever replace the parser (and dbLexRoutines.c is definitely part of the parser) we'd still want the naming checks to be done.

mdavidsaver:
> we should use it to check record alias names as well

Good point.  Done.

> dbLexRoutines.c even the best place for doing these checks

In terms of convenience, and keeping the diffstat small, yes.  I'm going to leave this in place for now.  I've no objections to later re-factoring.

> if we ever replace the parser

I look forward to that happy day.  Though it would be preceded by some less than happy days to implement.

anjohnson (approved):
Looks good to me now.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1034,22 +1050,49 @@ static void dbBreakBody(void)
...
+            if(c=='-' || c=='+' || c=='[' || c=='{') {
+                errlogPrintf("Warning: Record/Alias name '%s' should not begin with '%c'\n", name, c);
+            }
+        }
+        /* any character restrictions */
+        if(c < ' ') {
+            errlogPrintf("Warning: Record/Alias name '%s' should not contain non-printable 0x%02u\n",
+                         name, (unsigned)c);
+
+        } else if(c==' ' || c=='\t' || c=='"' || c=='\'' || c=='.' || c=='$') {
```
dirk-zimoch, line 1078: There is some redundancy here. `c=='\t'` is already covered by `c < ' '` above (for any encoding? ASCII at least) and `c==' '` could have have been included abore as `c <= ' '`.
If we need to handle `\t` separately, then what about `\r`, `\l` ... ?

PR #79, opened 2020-06-20
libCom: test osdSockAddrReuse

mdavidsaver:
Ensure that `epicsSocketEnableAddressReuseDuringTimeWaitState()` and `epicsSocketEnableAddressUseForDatagramFanout()` have the desired effects.

### DIFF
--- a/.ci-local/travis-fixup.sh
+++ b/.ci-local/travis-fixup.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+set -e -u -x
+
+env|grep TRAVIS
+
+[ "$TRAVIS_OS_NAME" = "linux" ] || exit 0
+
+# Ensure there is an interface with a (correct) broadcast address
+# eg. 'trusty' VMs have interface broadcast address mis-configured
+# (why oh why do people insist on setting this explicitly?)
+
+sudo ip tuntap add dev tap42 mode tap
+
+sudo ip addr add 192.168.240.1/24 broadcast + dev tap42
+
+sudo ip link set dev tap42 up
+
+# note that this device will be UP but not RUNNING
+# so java will see it as not UP since java confuses UP and RUNNING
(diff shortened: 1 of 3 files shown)


mdavidsaver:
This change only adds to the `osiSockTest` unittest.  However, it is changed in a way which violates the principle of test isolation by attempting to send (properly formatted) CA search messages to all local interface broadcast addresses.  So there is some risk of external side-effects.  However, I think this risk is low, and offset by the benefit of hopefully detecting regressions in this OS dependent code.

The test is still not passing on travis-ci `os: trusty`, though I think this is due to VM network configuration.

mdavidsaver:
I've (re)discovered that the broadcast address is mis-configured on the trusty VM image.

```
eth0      Link encap:Ethernet  HWaddr 42:01:0a:14:00:66  
          inet addr:10.20.0.102  Bcast:10.20.0.102  Mask:255.255.255.255
          UP BROADCAST RUNNING MULTICAST  MTU:1460  Metric:1
          RX packets:13709 errors:0 dropped:0 overruns:0 frame:0
          TX packets:10715 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:237955857 (237.9 MB)  TX bytes:950413 (950.4 KB)
```

anjohnson:
I have no objections to the spamming behavior. I assume you've checked it with an IOC running and it doesn't trigger anything (other than the usual silent name-search). Will the test still pass if it gets a reply from an IOC that actually has a matching named PV?

mdavidsaver:
> I assume you've checked it with an IOC running

I've checked with a softIoc, and also snooped with wireshark to see that the search is properly formatted.  Though I should note that the search being sents protocol version 0 in the header, which newer servers will ignore out of hand.  The payload is just the PV name though, with proper alignment.

> reply from an IOC

The test is not looking at replies.  It will see other searches, which it will ignore.

mdavidsaver:
Unless further concerns raised in the next day or so, I'll consider this accepted.

ralphlange:
Fine with me.

PR #81, opened 2020-07-02
Add spdx license tags

anjohnson:
Note: This is a big PR but it mostly contains just 1 or 2-line changes to the header comments of most of the core source files. I haven't discussed it with the group before now since I didn't know when we'd be able to apply it.

About a month ago I received notification of a request spdx/license-list-XML#1012 to add the EPICS Open license to the [SPDX License List](https://spdx.org/licenses). That has now been accepted but it won't actually appear on that page until they publish the next version in about a month's time. However we can now use the EPICS license identifier in our source code, so I went ahead and created the changes to do that. I suspect the original request may have come from someone associated with ESS, but I'm not certain of that.

To add an SPDX license tag, just add this to/as a comment in the file:
```C
/* SPDX-License-Identifier: EPICS */
```
The [formatting requirements](https://spdx.dev/ids/) for tags are fairly loose, AFAICS they doesn't have to be a separate comment but they must be on one line and there should be white-space between the colon and the license identifier.

This PR adds the license tag to all the appropriate sources in the main repository on the 7.0 branch, unifies all our license header comments, and removes notices from a few files that should never have had them added in the first place. The easiest way to automate adding the license tag was to put it immediately after the license header. I had to add a license header to some of the more recent files but I did not add Copyright notices (except for some files that I wrote). I did not add the tag to any Makefiles or other build system files (I could do if someone thinks we should include them but they don't all have license headers), nor did I do any work on the external submodules at all.

### DIFF
--- a/configure/CONFIG.gnuCommon
+++ b/configure/CONFIG.gnuCommon
@@ -3,8 +3,7 @@
 #     National Laboratory.
 # Copyright (c) 2002 The Regents of the University of California, as
 #     Operator of Los Alamos National Laboratory.
-# EPICS BASE Versions 3.13.7
-# and higher are distributed subject to a Software License Agreement found
+# EPICS Base is distributed subject to a Software License Agreement found
 # in file LICENSE that is included with this distribution. 
 #*************************************************************************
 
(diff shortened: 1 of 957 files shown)


mdavidsaver (approved):
While I'm not sure I see the [value](https://spdx.dev/ids/#why), it's a low impact change.  So I've no objection.

PR #85, opened 2020-08-06
Check if input string to aToIPAddr is null

MaticPogacnik:
Some modules directly pass the address configuration from the startup script to aToIPAddr function.
This can cause segmentation fault if user did not specify the address.

### DIFF
--- a/modules/libcom/src/misc/aToIPAddr.c
+++ b/modules/libcom/src/misc/aToIPAddr.c
@@ -75,7 +75,7 @@ static int initIPAddr ( struct in_addr ipAddr, unsigned port,
  * "pAddrString" does not contain an address of the form
  * "n.n.n.n:p or host:p"
  */
-LIBCOM_API int epicsStdCall 
+LIBCOM_API int epicsStdCall
 aToIPAddr( const char *pAddrString, unsigned short defaultPort,
                 struct sockaddr_in *pIP )
 {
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Which modules are passing `NULL`?  Have you opened bug reports?  Changing `aToIPAddr()` to error in this case seems reasonable, but user code probably needs to be changed as well.  To me, passing NULL in this context seems indicative of a logic error.

anjohnson:
caPutLog, if you don't configure it correctly. @MaticPogacnik is fixing it there too, but the fact that you get a segfault inside that routine if you pass a NULL string by mistake could be made more friendly.

modules/libcom/src/misc/aToIPAddr.c
```diff
@@ -90,6 +90,13 @@ aToIPAddr( const char *pAddrString, unsigned short defaultPort,
     unsigned port;
     struct in_addr ina;
 
+    /*
+     * Check if input string is null
+     */
+    if ( pAddrString == NULL ) {
```
mdavidsaver, line 93: While it is in keeping with the existing style in this function, I think it can be assumed that a reader will understand `== NULL` to be a null check.  So the comment can be omitted.

MaticPogacnik:
Hi,

Yes, I found an issue while testing new features in caputlog module. 
I did already include the check at a module level, but I thought it would be nice to have a check like this here as it looks like that the main purpose of this function is to parse user configuration.

I amended the commit and removed the comment.

mdavidsaver:
Picked as 5b37663bcb99aee7f6d4c511a9c8e854997a33ec

PR #88, opened 2020-08-26
Update README.md

HiteshDhola:
Pointing out few err in the release version. As this is the first document anyone would Read. I also use this file to restore name of my "base" folder to older original version (like base-7.3.0.1) when new EPICS version is released. This way I can keep all of my compiled version isolated. So, I  thought I would report few things.

-Corrected released version. Removed old references.
-Document like "EPICS Application Developer's guide" is no longer available for Ver 7.x, I guess. You may remove/update it's reference.

### DIFF
--- a/documentation/README.md
+++ b/documentation/README.md
@@ -1,6 +1,6 @@
 # Installation Instructions {#install}
 
-## EPICS Base Release 7.0.3.1
+## EPICS Base Release 7.0.4.1
 
 -----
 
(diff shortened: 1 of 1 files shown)


ralphlange:
Leave the AppDev Guide in until there is a new document covering its scope for 7.x, please.
It's taking longer than we hoped for, but it's more work than we thought, and the Covid-19 situation hasn't made doing codeathons/documentathons easier...

PR #89, opened 2020-08-31
Extract compiler pre-defined macros as Make variables

mdavidsaver:
Related to #23.  Extract `__RTEMS_MAJOR__` and similar with the preprocessor to avoid needing to configure the RTEMS OS version/series manually.

### DIFF
--- a/configure/Makefile
+++ b/configure/Makefile
@@ -15,7 +15,8 @@ include $(TOP)/configure/CONFIG
 TOOLS = $(TOP)/src/tools
 
 CONFIGS += $(subst ../,,$(wildcard ../CONFIG*))
-CONFIGS += $(subst ../,,$(wildcard ../os/CONFIG*))
+CONFIGS += $(subst ../,,$(wildcard ../os/CONFIG.*))
+CONFIGS += $(subst ../,,$(wildcard ../os/CONFIG_SITE.*))
 
 CONFIGS += $(subst ../,,$(wildcard ../RELEASE*))
 CONFIGS += $(subst ../,,$(wildcard ../RULES*))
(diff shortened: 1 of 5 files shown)


ralphlange:
Nice!
Any particular reason why it should not go into 3.15?

configure/Makefile
```diff
@@ -31,3 +32,13 @@ CFG += CONFIG_DATABASE_VERSION
...
 include $(TOP)/configure/RULES
 
+ifdef T_A
+
+install: $(TOP)/configure/os/CONFIG_TOOLCHAIN.Common.$(T_A)
+
+$(TOP)/configure/os/CONFIG_TOOLCHAIN.Common.$(T_A): toolchain.c
+	$(PREPROCESS.cpp)
+
+CLEANS += ../os/CONFIG_TOOLCHAIN.Common.$(T_A)
```
mdavidsaver, line 42: It seems that `$(CLEANS)` isn't actually cleaned by `make distclean` and similar.  I'm thinking to switch to something like:

```make
ifndef T_A
realclean: clean_toolchain
clean: clean_toolchain

clean_toolchain:
	$(RM) $(wildcard os/CONFIG_TOOLCHAIN.Common.*)
endif
```

configure/toolchain.c
```diff
@@ -0,0 +1,35 @@
...
+
+#elif defined(_MSC_VER)
+MSVC_VER = _MSC_VER
+#endif
+
+#ifdef __rtems__
+#include <rtems/score/cpuopts.h>
+RTEMS_MAJOR = __RTEMS_MAJOR__
+RTEMS_MINOR = __RTEMS_MINOR__
+RTEMS_PATCH = __RTEMS_REVISION__
```
mdavidsaver, line 34: Since the `__RTEMS_*__` are not predefined macros, newer GCC (by default) expands these as eg.

```make
# 34 "../toolchain.c"
RTEMS_PATCH = 
# 34 "../toolchain.c" 3 4
             0
```

mdavidsaver:
This was applied as 077b41e6c10575f6f748ae1b5fb918c3c46e0747

PR #91, opened 2020-09-23
Fix asub documentation: alarm happens when return status is < 0

gabrielfedel:
Return status > 0 on aSub record doesn't trigger alarm.


anjohnson:
Thanks Gabriel, merging.

PR #92, opened 2020-11-03
epicsTime: fix overflow handling

mdavidsaver:
As I noted on [lp:1896295](https://bugs.launchpad.net/bugs/1896295), `epicsTimeDiffInSeconds()` and `class epicsTime` which underpins it, do not correctly handle 32-bit overflow on targets where `sizeof(long)==8` (eg. Linux).

This PR changes 'class epicsTime' to use `epicsUInt32` to match 'struct epicsTimeStamp', and tries to fix the various places in epicsTime.cpp where overflow is detected.  This passes existing tests, and a new test of `epicsTimeDiffInSeconds()` overflow.

I'm not certain that this is the best approach though.  I think a rewrite would be better.  Specifically to make `class epicsTime` a wrapper around `struct epicsTimeStamp`.

### DIFF
--- a/modules/libcom/src/osi/epicsTime.cpp
+++ b/modules/libcom/src/osi/epicsTime.cpp
@@ -101,7 +101,7 @@ epicsTimeLoadTimeInit::epicsTimeLoadTimeInit ()
         (double) POSIX_TIME_AT_EPICS_EPOCH / this->time_tSecPerTick;
 
     if (this->time_tSecPerTick == 1.0 &&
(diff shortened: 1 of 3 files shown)


mdavidsaver:
Recently, while doing some benchmarks of a driver with `perf`, I was surprised to see `epicsTimeDiffInSeconds()` appear as a hot spot. Another reason I'm inclined to rewrite epicsTime*, and simplify it assuming `time_t` is integer seconds (as it is on all known targets).

mdavidsaver:
Superseded by #185

PR #93, opened 2020-11-14
Deprecate epicsThreadExitMain()

mdavidsaver:
Also, replace `epicsThreadExitMain()` with a loop calling `epicsThreadSleep()` to make a non-interactive softIoc easier to interrupt or debug.

### DIFF
--- a/modules/database/src/std/softIoc/softMain.cpp
+++ b/modules/database/src/std/softIoc/softMain.cpp
@@ -249,7 +249,10 @@ int main(int argc, char *argv[])
 
         } else {
             if (loadedDb || ranScript) {
-                epicsThreadExitMain();
+                // non-interactive IOC.  spin forever
+                while(true) {
+                    epicsThreadSleep(1000.0);
+                }
 
             } else {
                 usage(argv[0], dbd_file);
--- a/modules/libcom/src/osi/epicsThread.h
+++ b/modules/libcom/src/osi/epicsThread.h
@@ -59,6 +59,7 @@
 #include <stddef.h>
 
 #include "libComAPI.h"
+#include "compilerDependencies.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -142,8 +143,12 @@ LIBCOM_API void epicsThreadRealtimeLock(void);
  * call this routine. This should be the last call in main, except the
  * final return. On most systems epicsThreadExitMain never returns.This
  * must only be called by the main thread.
+ *
+ * @deprecated Deprecated for lack of use.  Please report any usage.
+ * Recommended replacement is loop + epicsThreadSleep(),
+ * epicsEventMustWait(), or similar.
  **/
-LIBCOM_API void epicsStdCall epicsThreadExitMain(void);
+LIBCOM_API void epicsStdCall epicsThreadExitMain(void) EPICS_DEPRECATED;
 
 /** For use with epicsThreadCreateOpt() */
 typedef struct epicsThreadOpts {
--- a/modules/libcom/src/osi/os/RTEMS/osdThread.c
+++ b/modules/libcom/src/osi/os/RTEMS/osdThread.c
@@ -219,6 +219,8 @@ threadWrapper (rtems_task_argument arg)
  */
 void epicsThreadExitMain (void)
 {
+    cantProceed("epicsThreadExitMain() has been deprecated for lack of usage."
+                "  Please report if you see this message.");
 }
 
 static rtems_status_code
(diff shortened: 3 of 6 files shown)


modules/database/src/std/softIoc/softMain.cpp
```diff
@@ -247,8 +248,11 @@ int main(int argc, char *argv[])
             }
 
         } else {
-            if (loadedDb) {
-                epicsThreadExitMain();
```
anjohnson, line 252: I would much prefer that we change what epicsThreadExitMain() actually does if its current definition makes no sense than just stop using it here, as that would leave the code for the now-unused routine sitting in libCom expecting to be maintained. This is the only call to this routine in Base (there isn't even a test for it), and I'm pretty sure the routine was added for this specific use. Any out-of-tree users would suffer the same "doesn't make sense" problem.

modules/database/src/std/softIoc/softMain.cpp
```diff
@@ -228,7 +229,7 @@ int main(int argc, char *argv[])
                 std::cout<<"# End "<<argv[optind]<<"\n";
 
             epicsThreadSleep(0.2);
-            loadedDb = true;    /* Give it the benefit of the doubt... */
+            ranScript = true;    /* Assume the script has done any necessary initialization */
```
anjohnson: Since we don't really need to know which of the two happened we could combine the two flags into one and rename it to something a bit more meaningful like `workloaded`.

anjohnson (commented):
I'm not convinced this is the right fix.

modules/database/src/std/softIoc/softMain.cpp
mdavidsaver: The `loadedDb` flag is being separately tested shortly after this line to decided whether to call `iocInit()`.

modules/database/src/std/softIoc/softMain.cpp
mdavidsaver: If this is truly the only usage found in ~20 years (which I think it is) then wouldn't it make more sense to remove `epicsThreadExitMain()` from libCom?  Since there is no pressing reason to do so, I would first mark it deprecated for a release cycle or two.

mdavidsaver: fyi.  For this sort of "parking" in `main()`.  I see abstracting the signal handling as more useful than the blocking.  cf [pvxs::SigInt](https://mdavidsaver.github.io/pvxs/util.html#_CPPv4N4pvxs6SigIntE).

anjohnson: I may be assuming os-specific requirements that don't actually exist, but wouldn't it be better to have one routine (which is what `epicsThreadExitMain()` was supposed to be) that does any cleanup necessary for any application wants to suspend its `main()` thread until the process gets killed from outside?

It ought to be possible to Control-C out of a `softIoc -S` so fixing that would be welcome. I see that `camonitor` calls `ca_pend_event(0)` which we couldn't replace, but `pvget -m` waits on an epicsEvent that (presumably) never gets triggered; the nice thing about using a call to `epicsThreadExitMain()` instead would be that it explicitly documents the author's intention. A portable equivalent of `__attribute__ ((noreturn))` might be a good idea for it, which could also be applied to `cantProceeed()`.

Maybe we could just replace the os-specific versions of `epicsThreadExitMain()` with a generic one that has a looping call to `epicsThreadSuspendSelf()`, as used in `cantProceeed()`.

mdavidsaver: Actually, the unstated assumption may be mine.  I see "suspend until killed" as short-sighted.  It is true that CLIs often start with this, and many may never want more.  For these cases, the ~3 line loop+sleep seems utterly trivial.  Why bother with library assistance?

For more complicated cases where cleanup is needed.  imo. the only general way for something like `epicsThreadExitMain()` to handle application cleanup is to return to `main()` and let user code do its thing.  Given that this would break the contract which `epicsThreadExitMain()` has so far made, I don't see that it can reasonably by changed.

> `pvget -m` waits on an epicsEvent that (presumably) never gets triggered

It does, from a signal handler.

fyi. the PVXS version of this (`pvxmonitor`) will also `exit(0)` when all subscriptions "finish".

mdavidsaver:
I've gone ahead and committed the first (un-contentious) part as https://github.com/epics-base/epics-base/commit/b890d584bce92d8374ca76d09a68c49f9a8fad05.  This should avoid warning message which @dirk-zimoch sees.

anjohnson:
Approved at meeting on 5/5

mdavidsaver:
Applied as 1b8a1ebd072600389b9feef5d631ccf397c94767

PR #95, opened 2020-12-01
Break recursion during early osd* errors

mdavidsaver:
Errors during initialization of osdthread or osdmutex can trigger recursive calls to errlog or cantProceed, which escalate to a stack overflow.

Add (so far) GCC specific (probably also clang) detection and breaking of loops via a `__thread` local variable.  Note that suncc also understands `__thread`, and msvc has `__declspec(thread)`.

I'm not sure if this is the right approach, or the right behavior.  eg. maybe better for errlog to write directly to stderr?  maybe cantProceed() should become OSD?

### DIFF
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -41,6 +41,28 @@
 /*Declare storage for errVerbose */
 int errVerbose = 0;
 
+#if defined(__GNUC__)
+/* Recursive calls to errlogPrint() typically arise from osdThread or osdMutex
+ * initialization failures.  Keep them from escalating to a stack overflow
+ * by detecting recursive calls to errlog entry points.
+ */
+static __thread char errlogRecurse=0;
+
+#define BEGIN_RECURSE do { \
+    if(errlogRecurse) { \
+        return 0; \
+    } \
+    errlogRecurse = 1; \
+    } while(0)
+
+#define END_RECURSE do{ errlogRecurse = 0; } while(0)
+
+#else
+#define BEGIN_RECURSE do{}while(0)
+#define END_RECURSE do{}while(0)
+
+#endif
+
 static void errlogExitHandler(void *);
 static void errlogThread(void);
 
@@ -107,13 +129,15 @@ int errlogPrintf(const char *pFormat, ...)
 {
     va_list pvar;
     char *pbuffer;
-    int nchar;
+    int nchar = 0;
     int isOkToBlock;
+
+    BEGIN_RECURSE;
 
     if (epicsInterruptIsInterruptContext()) {
         epicsInterruptContextMessage
             ("errlogPrintf called from interrupt level\n");
-        return 0;
+        goto cleanup;
     }
 
     errlogInit(0);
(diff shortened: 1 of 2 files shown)


modules/libcom/src/error/errlog.c
```diff
@@ -41,6 +41,28 @@
 /*Declare storage for errVerbose */
 int errVerbose = 0;
 
+#if defined(__GNUC__)
+/* Recursive calls to errlogPrint() typically arise from osdThread or osdMutex
+ * initialization failures.  Keep them from escalating to a stack overflow
+ * by detecting recursive calls to errlog entry points.
+ */
+static __thread char errlogRecurse=0;
```
anjohnson, line 49: `__thread` variables _are_ available on VxWorks 6.9 (I just checked), but you'll get complaints from Dirk since they aren't in VxWorks 5.5 and this is a change to a core module which he still builds there.

modules/libcom/src/misc/cantProceed.c
```diff
@@ -54,6 +54,16 @@ LIBCOM_API void * mallocMustSucceed(size_t size, const char *msg)
 LIBCOM_API void cantProceed(const char *msg, ...)
 {
     va_list pvar;
+#if defined(__GNUC__)
+    static __thread char dead = 0;
+    if(dead) {
+        /* Probably here after a failure in OSD early initialization.
+         * So we don't know which, if any, of the OSI primitives are usable.
+         */
+        abort();
```
anjohnson, line 63: I assume you're proposing this PR after seeing actual loops. I am unable to trigger a problem by calling `cantProceed()` from the first line of `main()` so it has to be related to some earlier initialization in a static constructor.

My guess at the culprit is `epicsStackTrace()` which is a relatively recent addition to this code and arrived long after we fixed a bunch of OSD Thread Init problems early on, but I don't know if we ever wrote down the rules we had developed to prevent issues. If you can replicate the problem it would be good to check whether commenting out that call resolves it. There is a call path to `epicsMutexMustCreate()` in `epicsStackTrace()` which might explain the loop, although it's from inside a `once()` routine.

anjohnson (commented):
Other than a warning about `__thread` variables not being compatible with Dirk's VxWorks 5.5 builds I have no major objections to this. I would like to see whether `epicsStackTrace()` is the culprit though and determine whether that loop can be fixed properly.

modules/libcom/src/misc/cantProceed.c
mdavidsaver: I hit recursion while messing around with `PTHREAD_PRIO_INHERIT`.  Try calling `errlogPrintf()` from the early initialization code in osdThread or similar.

mdavidsaver:
So, I've been testing this change for the past few weeks in the fine tradition of "eat your own dog food".  Unfortunately, eating dog food sometimes makes a person sick.

It turns out what while `__thread` will compile and (apparently) link correctly, there is a subtle problem with RTEMS 4.x (at least).

`__thread` causes variables to be placed in a `.tbss` section, which the RTEMS 4.x linker scripts don't consider.  GNU ld handles such "orphaned" sections with silent stupidity.  With 4.9 the stack for the main thread ends up overlapping with `.bss`.  Hilarity ensues.

An excerpt from a linking map.

```
                0x0000000000346948                rpc_createerr
                0x0000000000346958                _null_auth
                0x0000000000346970                . = ALIGN (0x10)
 *fill*         0x0000000000346964        0xc 00
                0x0000000000346970                PROVIDE (__bss_end, .)

.tbss           0x00000000003226bc        0x2
 .tbss          0x00000000003226bc        0x1 /home/mdavidsaver/work/epics/base-git/lib/RTEMS-mvme3100/libCom.a(errlog.o)
 .tbss          0x00000000003226bd        0x1 /home/mdavidsaver/work/epics/base-git/lib/RTEMS-mvme3100/libCom.a(cantProceed.o
)
                0x00000000003226c0                . = ALIGN (0x10)
                0x00000000003226c0                _end = .
                0x00000000003226c0                __rtems_end = .
                0x00000000003226c0                PROVIDE (end, .)
```

Notice how the PC counter jumps backwards!  The early stack is computed based on `__rtems_end`, which should be the highest address in the image.

I take all of this as a sign that `__thread` isn't usable with RTEMS 4.x.

@hjunkes ftw.

modules/libcom/src/misc/cantProceed.c
anjohnson: > Try calling `errlogPrintf()` from the early initialization code in osdThread

I can see some problems, we have 4 calls to `checkStatusQuit()` instead of `checkStatusOnceQuit()` in some of the routines in posix/osdThread.c that get called before `epicsThreadInit()` has completed. This comment in the code explains:

```C
/* The following are for use by once, which is only invoked from epicsThreadInit*/
/* Until epicsThreadInit completes errlogInit will not work                     */
/* It must also be used by init_threadInfo otherwise errlogInit could get  */
/* called recursively                                                      */
```
Note that `errlogPrintf()` calls `errlogInit()`.

The initialization strategy in posix/osdThread.c was that API routines must first call `epicsThreadInit()` or otherwise ensure that they are safe to run earlier than that. Some do instead use `assert(epicsThreadOnceCalled)` if they would only be called after some other routine that calls `epicsThreadInit()`. Some newer APIs aren't doing either.

mdavidsaver:
On reflection, I don't think there is an easy solution to errors during early initialization of libCom.  Adding special cases to errlog is at best a partial solution.

PR #96, opened 2020-12-16
Force device entries for "Soft Channel" to index 0

anjohnson:
It used to be quite difficult to usurp the position of the Soft Channel device supports as the first loaded and hence default DTYP for records that don't set it. However now that dbdExpand.pl auto-creates a place-holder record type if a device names one that hasn't yet been seen it's no longer fatal to put another support DBD file in front of base.dbd when creating an IOC's DBD file, and that will cause devices in that support DBD file to become the default for records that don't set DTYP.

The changes here cause any device support named "Soft Channel" to be inserted into the zeroth position of the record device list whenever they are loaded, thus making them automatically the default.

Adds tests to dbStaticTest to ensure this works.

Fixes [lp: #1908305](https://bugs.launchpad.net/epics-base/+bug/1908305)


anjohnson:
RL: Can the dbdExpand take a flag to allow the auto-declaration to not be an error, which CONFIG_SITE could set?

PR #98, opened 2021-01-07
Use getifaddrs instead of SIOCGIFCONF.

MichaelRitzert:
The old code is preserved in osdNetIfConf.c, which is used by the default
platform directory.

Platforms that support getifaddrs use the new code from osdNetIfAddrs.c.

https://bugs.launchpad.net/epics-base/+bug/1910148


mdavidsaver (approved):
The logic in the new code looks correct to me.

mdavidsaver:
Applied as 410921b5e with a couple of changes to move files around and disable the DEBUG prints.

PR #99, opened 2021-01-12
Remove dbfl_type_rec

bfrk:
See https://code.launchpad.net/~bfrk/epics-base/+git/epics-base/+merge/381464 for details.

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -668,27 +668,21 @@ int db_post_extra_labor (dbEventCtx ctx)
     return DB_EVENT_OK;
 }
 
-/*
- *  DB_CREATE_EVENT_LOG()
- *
- *  NOTE: This assumes that the db scan lock is already applied
- *        (as it copies data from the record)
- */
-db_field_log* db_create_event_log (struct evSubscrip *pevent)
+static db_field_log* db_create_field_log (struct dbChannel *chan, int use_val)
 {
     db_field_log *pLog = (db_field_log *) freeListCalloc(dbevFieldLogFreeList);
 
     if (pLog) {
-        struct dbChannel *chan = pevent->chan;
         struct dbCommon  *prec = dbChannelRecord(chan);
-        pLog->ctx = dbfl_context_event;
-        if (pevent->useValque) {
+        pLog->stat = prec->stat;
+        pLog->sevr = prec->sevr;
+        pLog->time = prec->time;
+        pLog->field_type  = dbChannelFieldType(chan);
+        pLog->field_size  = dbChannelFieldSize(chan);
+        pLog->no_elements = dbChannelElements(chan);
+
+        if (use_val) {
             pLog->type = dbfl_type_val;
-            pLog->stat = prec->stat;
-            pLog->sevr = prec->sevr;
-            pLog->time = prec->time;
-            pLog->field_type  = dbChannelFieldType(chan);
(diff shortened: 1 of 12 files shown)


bfrk:
This is the appveyor failure:

```
testChannelAccess.tap ..... 
not ok 149 - void __thiscall ChannelAccessIFTest::test_channelArray_destroy(void): an array syncGet failed 
Failed 1/152 subtests 
	(less 12 skipped subtests: 139 okay)
```

It doesn't look like it has anyhing to do with the changes I made, more like something non-deterministic that may happen every once in a while.

anjohnson:
I just triggered that Appveyor job to rebuild, some of our tests do fail at random there and usually pass on a second try. I agree this can't have been caused by your last commit, that test passed previously and all you changed was comment text.

modules/database/src/std/filters/arr.c
```diff
@@ -87,85 +85,53 @@ static long wrapArrayIndices(long *start, const long increment, long *end,
...
-                    offset = (offset + start) % dbChannelElements(chan);
-                    dbExtractArrayFromRec(&chan->addr, pdst, nTarget, capacity,
-                        offset, my->incr);
-                    pfl->u.r.field = pdst;
-                }
-            }
-            dbScanUnlock(prec);
-            dbChannelField(chan) = pfieldsave;
-        }
+        /* TODO Treat scalars as arrays with 1 element */
```
anjohnson, line 108: TODO? Was this done?

anjohnson (commented):
Core 2021-Jan-27 Meeting:
We expect to be nervous about this change because of the complex area it's working in, so we will merge it _after_ the next release which should be before the Codeathon in March, but we don't see anything obviously wrong with it.

modules/database/src/std/filters/arr.c
bfrk: No, this was meant as a reminder for a possible future extension. I just saw the empty case here with the comment `/* Only filter arrays */` and had the idea to make array filters work in that case, too. I am no longer sure this would be such a great idea. Perhaps I should restore the old comment line.

bfrk: And yes, I understand your reservations wrt the complexity of the code. I'm fine with postponing it until after release if that lessens the risk.

bfrk: I do have changes ready that would simplify things, especially with dbGet. However, this involves a performance trade-off. I think I will make a PR nonetheless, to give you a preview.

mdavidsaver:
Since 7.0.5 is out, and the codeathon is looming, I'm going to merge tomorrow (Sunday) unless I hear objections.

modules/database/src/ioc/db/dbEvent.c
```diff
@@ -737,20 +754,6 @@ static void db_queue_event_log (evSubscrip *pevent, db_field_log *pLog)
...
-     * event on the queue and the current event are emtpy
-     * (i.e. of type dbfl_type_rec), simply ignore duplicate
-     * events (saving empty events serves no purpose)
-     */
-    if (pevent->npend > 0u &&
-        (*pevent->pLastLog)->type == dbfl_type_rec &&
-        pLog->type == dbfl_type_rec) {
-        db_delete_field_log(pLog);
-        UNLOCKEVQUE (ev_que);
-        return;
```
mdavidsaver, line 751: Removing this block allowed lots of duplicate array updates.

Should be restored with 4df48c91f4e7202899bd859312a2b68148bf0bad

PR #100, opened 2021-01-14
Scalar get optimization

bfrk:
See https://code.launchpad.net/~bfrk/epics-base/+git/epics-base/+merge/382204. ~~This branch is based on branch remove-dbfl_type_rec.~~

### DIFF
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -34,10 +34,12 @@
 #include "dbBase.h"
 #include "dbChannel.h"
 #include "dbCommon.h"
+#include "dbConvertFast.h"
 #include "dbEvent.h"
 #include "dbLock.h"
 #include "dbStaticLib.h"
 #include "link.h"
+#include "recGbl.h"
 #include "recSup.h"
 #include "special.h"
 #include "alarm.h"
@@ -630,14 +632,61 @@ long dbChannelOpen(dbChannel *chan)
 }
 
 /* Only use dbChannelGet() if the record is already locked. */
-long dbChannelGet(dbChannel *chan, short type, void *pbuffer,
-        long *options, long *nRequest, void *pfl)
+long dbChannelGet(dbChannel *chan, short dbrType, void *pbuffer,
+        long *options, long *nRequest, db_field_log *pfl)
 {
-    return dbGet(&chan->addr, type, pbuffer, options, nRequest, pfl);
+    dbAddr addr = chan->addr; /* structure copy */
+
+    if (pfl) {
+        addr.field_size = pfl->field_size;
+        addr.field_type = pfl->field_type;
+        addr.no_elements = pfl->no_elements;
+        addr.pfield = dbfl_pfield(pfl);
+    }
+
+    /*
+     * Try to optimize scalar requests by caching (just) the conversion
+     * routine. This is possible only if we have no options, the field and the
+     * request have exactly one element, we have no special processing to do,
+     * and the field type is not DBF_DEVICE or larger.
+     * Note that if the db_field_log has already copied the data, dbGet
+     * does not call get_array_info.
+     */
+    if (!(options && *options)
+        && addr.no_elements == 1
+        && (!nRequest || *nRequest == 1)
+        && (dbfl_has_copy(pfl) || addr.special != SPC_DBADDR)
+        && addr.special != SPC_ATTRIBUTE
+        && addr.field_type <= DBF_DEVICE)
+    {
+        if (chan->getCvt && chan->lastGetdbrType == dbrType) {
+            return chan->getCvt(addr.pfield, pbuffer, &addr);
+        } else {
+            unsigned short dbfType = addr.field_type;
+
+            if (INVALID_DB_REQ(dbrType)) {
+                char message[80];
+
+                sprintf(message, "dbChannelGet: Request type is %d\n", dbrType);
+                recGblDbaddrError(S_db_badDbrtype, &addr, message);
+                return S_db_badDbrtype;
+            }
+
+            chan->getCvt = dbFastGetConvertRoutine[dbfType][dbrType];
+            chan->lastGetdbrType = dbrType;
+            if (nRequest)
+                *nRequest = addr.no_elements;
+            return chan->getCvt(addr.pfield, pbuffer, &addr);
+        }
+    } else {
+        /* non-optimized case */
+        chan->getCvt = NULL;
+        return dbGet(&addr, dbrType, pbuffer, options, nRequest, pfl);
+    }
 }
 
 long dbChannelGetField(dbChannel *chan, short dbrType, void *pbuffer,
-        long *options, long *nRequest, void *pfl)
+        long *options, long *nRequest, db_field_log *pfl)
 {
     dbCommon *precord = chan->addr.precord;
     long status = 0;
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -665,10 +665,6 @@ long dbChannelGet(dbChannel *chan, short dbrType, void *pbuffer,
             unsigned short dbfType = addr.field_type;
 
             if (INVALID_DB_REQ(dbrType)) {
-                char message[80];
-
-                sprintf(message, "dbChannelGet: Request type is %d\n", dbrType);
-                recGblDbaddrError(S_db_badDbrtype, &addr, message);
                 return S_db_badDbrtype;
             }
 
--- a/modules/database/src/ioc/db/dbChannel.h
+++ b/modules/database/src/ioc/db/dbChannel.h
@@ -50,6 +50,8 @@ typedef struct evSubscrip {
 
 typedef struct chFilter chFilter;
 
+typedef long fastConvert(void *from, void *to, const dbAddr *paddr);
+
 /* A dbChannel points to a record field, and can have multiple filters */
 typedef struct dbChannel {
     const char *name;
@@ -60,6 +62,9 @@ typedef struct dbChannel {
     ELLLIST filters;          /* list of filters as created from JSON */
     ELLLIST pre_chain;        /* list of filters to be called pre-event-queue */
     ELLLIST post_chain;       /* list of filters to be called post-event-queue */
+    /* Support for optimizing scalar requests */
+    fastConvert *getCvt;      /* fast get conversion function */
+    short lastGetdbrType;     /* last dbChannelGet dbrType */
 } dbChannel;
 
 /* Prototype for the channel event function that is called in filter stacks
@@ -208,9 +213,9 @@ DBCORE_API extern unsigned short dbDBRnewToDBRold[];
 
 
 DBCORE_API long dbChannelGet(dbChannel *chan, short type,
-        void *pbuffer, long *options, long *nRequest, void *pfl);
+        void *pbuffer, long *options, long *nRequest, db_field_log *pfl);
 DBCORE_API long dbChannelGetField(dbChannel *chan, short type,
-        void *pbuffer, long *options, long *nRequest, void *pfl);
+        void *pbuffer, long *options, long *nRequest, db_field_log *pfl);
 DBCORE_API long dbChannelPut(dbChannel *chan, short type,
         const void *pbuffer, long nRequest);
 DBCORE_API long dbChannelPutField(dbChannel *chan, short type,
--- a/modules/database/src/ioc/db/dbDbLink.c
+++ b/modules/database/src/ioc/db/dbDbLink.c
@@ -58,7 +58,6 @@
 #include "dbBase.h"
 #include "dbBkpt.h"
 #include "dbCommonPvt.h"
-#include "dbConvertFast.h"
 #include "dbConvert.h"
 #include "db_field_log.h"
 #include "db_access_routines.h"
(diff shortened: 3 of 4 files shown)


bfrk:
All tests succeed when i run them on my machine (ubuntu-18, make -s runtests). I don't understand why they fail in the CI.

ralphlange:
Are you locally using the same commits of the git submodules? (E.g. `pva2pva`, where the failure happens?)

bfrk:
How do I make sure I have the same submodules? How can I see if they aren't up to date?

bfrk:
```
ben@home[5]:.../modules/database>git submodule status                                                                                                                         -3db08b5977efd7a3ac8eecfa068b5bfdfdf84dd0 ../../.ci
-7a2d264f2cb107bfd10adb23bc2b73d8323a79e4 ../normativeTypes
-e1c1a4bc1bad6933e57b199d58f74468401218b3 ../pvAccess
-d3b4976ea2b0d78075511f14d7f7bf9620dd4d14 ../pvData
-93a259cbde56668c1bbe495b15cc3ede8b42ce30 ../pvDatabase
-ad8b77e19f7e2ae50c48b5d871bdbe7b0ee23b61 ../pva2pva
-efb2631905cd61cc06041f5aac5be9017383a004 ../pvaClient
```

ralphlange:
In the CI logs, up in the "actions/checkout" part, under submodules, you will find what is used there.

bfrk:
Okay, I did this:
```
git submodule sync
git submodule update --init --force --depth=1
```
Rebuilt everything. Again all tests succeed for me.

ralphlange:
If the submodules are the same - any chance that base itself is more different than you think?

bfrk:
My time's up. I am going to take another look tomorrow.

bfrk:
I somehow managed to rebase my branch on a wrong commit which wasn't the HEAD of 7.0. No idea how that happened. Will fix and force push.

bfrk:
The one failure we get now looks interesting. My initial hunch is that this is not a bug I introduced; it looks more like a race condition that existed independently of the changes I made and that *may* have been uncovered due to the additional optimization happening now. It would be nice if someone familiar with the involved architectures could investigate this.

ralphlange:
Sorry - please ignore that one.

The simm tests are broken for a long time. I need to fix them and I know how.
Michael was relaxing the criteria when executing the tests on CI runners, which took some pressure off my back. (Too much, obviously.)

bfrk:
Okay, thanks! Ready for review then.

modules/database/src/ioc/db/dbChannel.c
```diff
@@ -630,14 +632,61 @@ long dbChannelOpen(dbChannel *chan)
...
+        if (chan->getCvt && chan->lastGetdbrType == dbrType) {
+            return chan->getCvt(addr.pfield, pbuffer, &addr);
+        } else {
+            unsigned short dbfType = addr.field_type;
+
+            if (INVALID_DB_REQ(dbrType)) {
+                char message[80];
+
+                sprintf(message, "dbChannelGet: Request type is %d\n", dbrType);
+                recGblDbaddrError(S_db_badDbrtype, &addr, message);
```
mdavidsaver: Printing a message will clash with my planned addition of `DBR_VFIELD` which depends on code like this silently returning `S_db_badDbrtype` when presented with an unknown dbrType.

modules/database/src/ioc/db/dbChannel.c
```diff
@@ -630,14 +632,61 @@ long dbChannelOpen(dbChannel *chan)
...
+                return S_db_badDbrtype;
+            }
+
+            chan->getCvt = dbFastGetConvertRoutine[dbfType][dbrType];
+            chan->lastGetdbrType = dbrType;
+            if (nRequest)
+                *nRequest = addr.no_elements;
```
mdavidsaver, line 674: I'm surprised to see that `nRequest` is only updated on the "slow" path.  I would expect this to be done by both or neither.

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -60,6 +62,9 @@ typedef struct dbChannel {
     ELLLIST filters;          /* list of filters as created from JSON */
     ELLLIST pre_chain;        /* list of filters to be called pre-event-queue */
     ELLLIST post_chain;       /* list of filters to be called post-event-queue */
+    /* Support for optimizing scalar requests */
+    fastConvert *getCvt;      /* fast get conversion function */
+    short lastGetdbrType;     /* last dbChannelGet dbrType */
```
mdavidsaver, line 67: I find myself wondering if the speedup from caching `dbFastGetConvertRoutine[dbfType][dbrType]` is worth the added code complexity.

modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -179,49 +175,19 @@ static long dbDbGetValue(struct link *plink, short dbrType, void *pbuffer,
...
-        status = dbChannelGet(chan, dbrType, pbuffer, NULL, pnRequest, pfl);
-
-        if (pfl)
-            db_delete_field_log(pfl);
-
-        if (status)
-            return status;
+    /* If filters are involved in a read, create field log and run filters */
+    if (ellCount(&chan->filters)) {
+        pfl = db_create_read_log(chan);
```
mdavidsaver, line 180: Maybe I've misunderstood the [discussion](https://code.launchpad.net/~dirk.zimoch/epics-base/+git/epics-base/+merge/378968), but it sounds like `db_create_read_log()` is what really needs to be optimized.  (which makes more sense to me wrt. global locking)  Maybe by adding a way to allow the first (and often only) `db_field_log` to be created on the stack?  eg. by adding a bit to control whether `db_delete_field_log()` calls `freeListFree()`.

mdavidsaver (commented):
As far as I can tell, this code change looks correct.  However, I find myself wondering if it is the best solution.  `dbGet()` is already one of the most complicated functions in Base, so I'm reluctant to let that complexity leak into `dbChannelGet()`.

modules/database/src/ioc/db/dbDbLink.c
mdavidsaver: Ok, the immediate issue with what I suggest would be filters like `dbnd` which store `db_field_log*`.

mdavidsaver:
I've made a quick attempt to characterize performance of this PR as well as #153 individually, and in combination.   I used my "favorite" test case for database processing performance, a counter record in an infinite scan loop involving a CA_LINK.  see [spin.db.txt](https://github.com/epics-base/epics-base/files/6259058/spin.db.txt)

I looked at measured scan rate before and after a single CA client connects.  I think this hits Ben's special case since the CA monitor is of a scalar DBR_TIME_DOUBLE.  Adding just one CA client has a very large negative effect on the count rate (~2.4MHz -> 0.18MHz).  Neither this PR, nor #153, seems to change this.  w/ #153 not surprising since no filter is involved.  Should this test have seen some improvement from this PR, or is a different scenario needed?

w/o This PR

spin

w/ This PR

spin2

Comparison of both runs

spin3

These tests were done on my laptop with power management active, and other tasks running.  I'm working from my [integration branch](https://github.com/mdavidsaver/epics-base/tree/integration), to which I apply Ben's change.

mdavidsaver:
I repeated my test looking at the effects of adding a deadband filter `{"dbnd":{"d":0}}` vs. no filter.  The zero deadband will not drop anything since the counter increments on each iteration, but it will mean a call to `db_create_read_log()` and filter processing.

This also seems to show no change.

spin4

I also tried a `{"dbnd":{"d":128}}`, which will drop 127 of every 128 events.  This has a large effect.  So in this (highly artificial) scenario, having to wake up the dbEvent thread seems to dominate.  Not so surprising to me as I started doing this test back when I looking at the effects of global locks in dbScanLock and general time.

So maybe the first question to address is whether there is whether there is a different test scenario (not a micro-benchmark) which benefits from this PR.

anjohnson:
I don't think this PR is supposed to make the IOC faster, it just moves where the optimization actually happens, in an attempt to simplify the code. The original first commit log message said "move optimization for scalar requests from link to channel". The optimization it's talking about is the one that calls the routines in dbFastLinkConv.c when the field holds a scalar. I measured the effect of this optimization a couple of years ago and found it to still be a worthwhile improvement over calling the more general routines in dbConvert.c for all copy/convert operations.

mdavidsaver:
> ... "move optimization for scalar requests from link to channel" ...

Which means it would be moved into the code path of my test, which does not involve a link.

> I measured the effect of this optimization a couple of years ago ...

You wouldn't happen to remember how?  I fully accept that I may not be testing the right case(s).

anjohnson:
Does the other path of your test end up calling `dbGet()` though? That's where the optimization originally came from, it's still present there, and that's where I actually ran my tests – I modified `dbGet()` to fail the check `if (offset == 0 && (!nRequest || no_elements == 1))` and recompiled.

mdavidsaver:
>  This has a large effect. So in this (highly artificial) scenario, having to wake up the dbEvent thread seems to dominate. Not so surprising to me as I started doing this test back when I looking at the effects of global locks in dbScanLock and general time.

Ha, surprise.  It is the scan lock which is being contended.  While I know I've looked at this code, I guess I had not appreciated that `dbChannel_get_count()` calls `dbScanLock()` unconditionally, even when it has a `db_field_log*` with all the necessary information.  Now this seems like a prime target for optimization...

mdavidsaver:
fyi.  I'm identifying contention with [futexs.stp](https://github.com/epics-base/epics-base/files/6261183/futexs.stp.txt), which I've modified from a [systemtap demo](https://sourceware.org/systemtap/SystemTap_Beginners_Guide/futexcontentionsect.html).

Skipping the scan lock for DBR_TIME_DOUBLE appears to work, but only gives a ~2x increase in the count rate.  aka. the -13.3x penalty becomes -12.6x.  Then contention for the dbEvent `writelock` dominates.  Oh well, it couldn't be that easy.

modules/database/src/ioc/db/dbChannel.c
bfrk: I merely copied that from `dbGet`. Easy to remove if it is not wanted (c34e7c67d4).

bfrk:
> `dbGet()` is already one of the most complicated functions in Base, so I'm reluctant to let that complexity leak into `dbChannelGet()`.

The complexity was moved from `dbDbGetValue` to `dbChannelGet`, not from
`dbGet`. I can't see how this leaks anything that hasn't leaked in exactly
the same way before.

modules/database/src/ioc/db/dbChannel.c
bfrk: What's the point if we already know that
```
        (!nRequest || *nRequest == 1)
```
?

modules/database/src/ioc/db/dbChannel.h
bfrk: I wondered, too. Dirk told us that he measured it and found it makes a great difference. I am too lazy right now to look up the message in the launchpad MR.

bfrk:
Regarding the optimisation, perhaps Dirk could chime in here? See https://code.launchpad.net/~dirk.zimoch/epics-base/+git/epics-base/+merge/378968/comments/1000415

mdavidsaver:
Note for this morning's discussion.

I was not expressing myself clearly.  My objection to moving the `dbFastGetConvertRoutine` lookup optimization from `dbDbGetValue()` to `dbChannelGet()` is that `dbGet()` already (maybe) calls `dbFastGetConvertRoutine`.  So the only added value I see in the `dbDbGetValue()` optimization is avoiding some conditionals, some call overhead, and the global indexing of the `dbFastGetConvertRoutine` table.  This doesn't seem to me likely to give much benefit on modern CPUs.

I am __not__ talking about removing `dbFastGetConvertRoutine` altogether.

Also, looking again I see that my test case above does __not__ actually cover dbGet() since I used `VAL+1` instead of a self-link with `INPA`.  oops....

As discussed, a better test might involve a series of input records with `INP` links forming a chain.  This would hit `dbGet()` and `dbDbGetValue()`.

bfrk:
I made some measurements. This is my test.db:
```
record(longin,"get") {
    field(INP,"get2")
    field(MDEL,"-1")
}
record(longin,"get2") {
    field(INP,"get CP")
    field(FLNK,"get")
}
```
This processes two longin records as fast as possible. The get.INP is a db link, to ensure that the "get" record really does a dbChannelGet (the other record has a CP input link and I wasn't sure such a link invokes dbChannelGet).
I measured the processing times by running `camonitor -t sI get` for a few minutes, redirected to a file, applied some filtering, and then counted the relative number of occurrences of the time differences. The two versions I tested were OPT, which is my scalar_optimization branch, and NO_OPT, which is the same branch but with the special optimization we are discussing here completely removed.
These are my results, as a histogram, with only percentages >= 1% shown:
```
timediff OPT  NO_OPT
0.000011 1%
0.000012 9%   8%
0.000013 9%   9%
0.000014 6%   5%
0.000015 12%  10%
0.000016 10%  11%
0.000017 5%   6%
0.000018 4%   4%
0.000019 4%   4%
0.000020 5%   5%
0.000021 6%   7%
0.000022 4%   4%
0.000023 2%   2%
0.000024 2%   2%
0.000025 2%   2%
0.000026 1%   1%
0.000027 1%   1%
0.000028 1%   1%
```
I ran this on my Linux laptop with no isolation, just like Michael did.
The difference between OPT and NO_OPT is barely measurable.

anjohnson:
Core Group 2022-10-26: In reviewing this again, we're wondering how much this optimization is worth pursuing. Benchmarking the existing (7.0.7) code against a version with the conversion cache removed from the dbLink would help answer that, measuring for these cases: No conversion, conversion (e.g. double to long or vice versa), and with/without a filter (to force it to allocate a db_field_log). Do this before rebasing again, in case the optimization proves insignificant.

PR #102, opened 2021-01-28
iocsh automatic paths

mdavidsaver:
Provide several new IOC shell variables computed from the locations of files (not `$PWD`).  Intended to help writing scripts with less dependence on the current directory (less sensitive to an added `cd ...`).

- `${EPICS_EXEC_DIR}` Directory containing the executable.  Exactly the string returned by `epicsGetExecDir()` (if not NULL)
- `${TOP}` Provide a default value of `${EPICS_EXEC_DIR}/../..`.  Any `$TOP` in the environment takes precedence).
- `${IOCSH_SCRIPT_DIR}` Directory containing the current IOC shell script

@anjohnson What should these names be?

Note that these are IOC shell variables, not environment variables.

 `${IOCSH_SCRIPT_DIR}` handles nested scripts.  It will expand as the directory containing the current fragment.  eg.

```
cat <<EOF > test.cmd
< subdir/included.cmd
EOF
install -d subdir
cat <<EOF > subdir/included.cmd
echo "IOCSH_SCRIPT_DIR=${IOCSH_SCRIPT_DIR}"
EOF
softIoc test.cmd
```

Will show a path which ends with `subdir/`.

The first patch in the stack is standalone, and adds TOP and EPICS_EXEC_DIR.

The rest add the mechanics (including an OSI path handling) needed for IOCSH_SCRIPT_DIR.  This includes `epicsPathAllocCWD()` which handles really long directory names, and a detour to use it.  To make `#!/.../softIoc` work I need to capture `argv[0]`, which necessitates a change to the IOC main template code.  I do this in the form of `iocshMain()` to allow the possibility of further features (like `softIoc`-ish arguments) without needing another change to the template. existing IOC applications will see IOCSH_SCRIPT_DIR default to `$PWD`. 

The last patch changes to template to include these new variables, and to mention `${EPICS_DB_INCLUDE_PATH}` which otherwise isn't documented (still).  I've no problem omitting this for the time being if it turns out to be contentious.

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.h
+++ b/modules/libcom/src/iocsh/iocsh.h
@@ -86,6 +86,20 @@ LIBCOM_API const iocshVarDef * epicsStdCall iocshFindVariable(
 /* This should only be called when iocsh is no longer needed*/
 LIBCOM_API void epicsStdCall iocshFree(void);
 
+/** Helper for implementing main() for an IOC.
+ *
+ * @code
+ * int main(int argc,char *argv[]) {
+ *     return iocshMain(argc, argv);
+ * }
+ * @endcode
+ *
+ * see the 'ioc' template of makeBaseApp.pl for an example with compatibility
+ * for older Base.
+ *
+ * @since UNRELEASED
+ */
+LIBCOM_API int iocshMain(int argc,char *argv[]);
 /** shorthand for \code iocshLoad(pathname, NULL) \endcode */
 LIBCOM_API int epicsStdCall iocsh(const char *pathname);
 /** shorthand for \code iocshRun(cmd, NULL) \endcode */
--- a/modules/libcom/src/iocsh/iocsh.h
+++ b/modules/libcom/src/iocsh/iocsh.h
@@ -86,6 +86,7 @@ LIBCOM_API const iocshVarDef * epicsStdCall iocshFindVariable(
 /* This should only be called when iocsh is no longer needed*/
 LIBCOM_API void epicsStdCall iocshFree(void);
 
+#define iocshMain iocshMain
 /** Helper for implementing main() for an IOC.
  *
  * @code
--- a/modules/libcom/src/osi/osiFileName.c
+++ b/modules/libcom/src/osi/osiFileName.c
@@ -128,16 +128,15 @@ char* epicsPathJoin(const char **fragments, size_t nfragments)
         if(epicsPathIsAbs(frag, flen)) {
             /* reset */
             firstfrag = i;
-            nfrags = 1;
+            nfrags = 0;
             totallen = 0;
-
-        } else {
-            nfrags++;
         }
 
         if(nfrags)
             totallen += 1; /* for separator */
+
         totallen += flen;
+        nfrags++;
     }
 
     ret = malloc(totallen + 1);
@@ -146,6 +145,7 @@ char* epicsPathJoin(const char **fragments, size_t nfragments)
 
         for(i=firstfrag, nfrags=0; i<nfragments; i++) {
             const char* frag = fragments[i];
+
             if(!frag|| frag[0]=='\0')
                 continue;
 
@@ -156,9 +156,11 @@ char* epicsPathJoin(const char **fragments, size_t nfragments)
 
             if(nfrags)
                 strcat(ret, OSI_PATH_SEPARATOR);
+
             strcat(ret, frag);
             nfrags++;
         }
+        assert(strlen(ret)==totallen);
     }
 
     free(cwd);
--- a/modules/libcom/src/osi/osiFileName.c
+++ b/modules/libcom/src/osi/osiFileName.c
@@ -1,0 +1,195 @@
+/*************************************************************************\
+* Copyright (c) 2021 Michael Davidsaver
+* SPDX-License-Identifier: EPICS
+* EPICS Base is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+#include <stdarg.h>
+#include <string.h>
+#include <errno.h>
+
+#include <epicsAssert.h>
+#include <osiUnistd.h>
+#include <epicsString.h>
+#include "osiFileName.h"
+
+#if defined(_WIN32) || defined(__CYGWIN__)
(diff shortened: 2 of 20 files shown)


mdavidsaver:
I think this is a fairly low risk change in terms of bugs since this is (mostly) only used on startup.

The risks in terms of compatibility are higher though.  The change to the IOC `main()` is optional, and existing IOC applications should continue to function as before.  The variable `EPICS_EXEC_DIR` and `IOCSH_SCRIPT_DIR` will override any existing definition, but I don't think these names have ever been used.  With `TOP` the new definition is omitted if `getenv("TOP")!=NULL`, so definitions from eg. an envPaths file are preserved.

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -86,6 +86,20 @@ LIBCOM_API const iocshVarDef * epicsStdCall iocshFindVariable(
...
+ * @code
+ * int main(int argc,char *argv[]) {
+ *     return iocshMain(argc, argv);
+ * }
+ * @endcode
+ *
+ * see the 'ioc' template of makeBaseApp.pl for an example with compatibility
+ * for older Base.
+ *
+ * @since UNRELEASED
```
mdavidsaver, line 101: @anjohnson Something I've started doing with [PVXS](https://mdavidsaver.github.io/pvxs/details.html#contributing) which I'd like to start doing with the 7.0 branch as well.  The idea is to later do a search+replace of `UNRELEASED` with the actual version number just prior to cutting a release.

mdavidsaver:
@anjohnson Is ok with the variable names.

That leaves the new API:

- `iocshMain()` used in new softMain.cpp skeleton
- `epicsPathDir()` in-place, re-entrant, equivalent to `dirname()`
- `epicsPathFile()` in-place, re-entrant, equivalent to `basename()`
- `epicsPathIsAbs()`
- `epicsPathJoin()`
- `epicsPathJoinCurDir` magic placeholder for current directory when calling `epicsPathJoin()`
- `epicsPathAllocCWD()` wraps `getcwd()` with handling of really long paths.

@FreddieAkeroyd There is some windows specific path handling in `epicsPathIsAbs()` which are in need of a second look.

modules/libcom/src/osi/osiFileName.c
```diff
@@ -0,0 +1,197 @@
...
+     * for windows... not so much.
+     * cf. https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#identify-the-path
+     *
+     * we treat as absolute:
+     *  - device and UNC paths.
+     *  - absolute on current drive.  eg. "\foo" -> "C:\foo" if %CD% starts with C:
+     *  - absolute w/ drive.  eg. "C:\foo"
+     *
+     * we treat as relative
+     *  - relative to some drive.  eg. "C:foo" -> "D:\bar\foo" if %CD% == "D:\bar"  (sneaky!)
```
FreddieAkeroyd, line 84: This one is horrible in that you need to consider the last current directory on the other drive. I think it may be C:foo -> c:\bar\foo if c:\bar was the last directory on c: and your current directory %CD% was e.g. e:\anything. However it is still  a relative path

modules/libcom/src/osi/osiFileName.c
```diff
@@ -0,0 +1,197 @@
...
+
+    /* search forward to ignore unused buffer */
+    pathlen = epicsStrnLen(path, pathlen);
+
+    /* search backwards for last separator */
+    pos = pathlen;
+    for(; pos>0; pos--) {
+        char c = path[pos-1u];
+        if(SEP(c)) {
+            /* found separator */
```
FreddieAkeroyd, line 40: Do you need to allow for `C:file.txt` on windows?

modules/libcom/src/osi/osiFileName.c
```diff
@@ -0,0 +1,197 @@
...
+            continue;
+
+        if(frag==magicCWD) {
+            if(!cwd)
+                cwd = epicsPathAllocCWD();
+            if(!cwd)
+                return NULL;
+            frag = cwd;
+        }
+
```
FreddieAkeroyd, line 125: If the process current directory was `c:\abc` and it's previous directory on d: was `d:\efg` then if it was passed ("d:", "<current_directory>") should it generate `c:\abc` or `d:\efg` ? I don't know how likely this is and whether we need to cover it, but we would need to use  `getdcwd` rather than `getcwd`

mdavidsaver: Assuming that the current directory (`getcwd()`) is always an absolute path, a join will ignore any prior fragments.  So `epicsPathJoin(["d:", epicsPathJoinCurDir, "foo"], 3)` will ignore `"d:"`.

mdavidsaver: Also, `epicsPathJoinCurDir` holds a magic address which must be used.  The fact that this address prints as `"<current_directory>"` is a convenience.

modules/libcom/src/osi/osiFileName.c
mdavidsaver: Yes, I think so...

modules/libcom/src/osi/osiFileName.c
mdavidsaver: So this case should be treated as absolute?

modules/libcom/src/osi/osiFileName.c
FreddieAkeroyd: Ah ok, we probably don't need to worry about windows having a drive specific current working directory then

mdavidsaver:
@FreddieAkeroyd Thanks for for taking a look at this.

Overall, I think the bulk of this change is on hold until something can be done about the fixed 256 character limit imposed by macLib.  While paths this long aren't common, it is known to happen.  Simply truncating them isn't so friendly.

I'll see about reworking the first part to at least add `TOP` and `EPICS_EXEC_DIR` as actual environment variables.

modules/libcom/src/osi/osiFileName.c
FreddieAkeroyd: It is still relative, you just have to be careful if you tried to reconstruct an absolute path from it yourself - it is relative to a directory that may not be your current working directory (i.e. `getdcwd` rather than `getcwd`)

PR #103, opened 2021-02-02
Add VALID_BUILDS support for "Command"

anjohnson:
This adds Makefile variables `PROD_CMD`, `LIBRARY_CMD` etc. for building command-line programs on targets that can support them but which are not build hosts (e.g. cross-building for other Linux arch's). It changes catools and other CA programs to build using these, instead of writing "build for all arch's except where the OS is VxWorks, RTEMS or iOS".


mdavidsaver:
Seems straightforward enough.

ralphlange:
Looks fine to me, too.

PR #104, opened 2021-02-04
Improve error message for DBF_MENU fields in .db files

mdavidsaver:
Arising from https://epics.anl.gov/core-talk/2021/msg00165.php

Adds some context eg `no such device support for 'ao' record type` or `using menu menuIvoa`when setting a DBF_MENU or DBF_DEVICE field from a .db file (not live).

Add `Did you mean "..."?` suggestion based using minimum [edit distance](https://en.wikipedia.org/wiki/Edit_distance) as a metric of "similarity".

```
$ cat badchoice.db 
record("ao", "foo") {
    field(DTYP, "Soft Chanel")
    field(PINI, "yes")
    field(IVOA, "Set output To IVOV")
    field(SCAN, "Pensive")
}
record("ai", "bar") {
    field(SCAN, "100")
}
$ softIoc -d badchoice.db
Can't set "foo.DTYP" to "Soft Chanel" no such device support for 'ao' record type : Illegal choice
    Did you mean "Soft Channel"?
Error at or before ")" in file "../../base-git/badchoice.db" line 2
Can't set "foo.PINI" to "yes" using menu menuPini : Illegal choice
    Did you mean "YES"?
ErrorCan't set "foo.IVOA" to "Set output To IVOV" using menu menuIvoa : Illegal choice
    Did you mean "Set output to IVOV"?
ErrorCan't set "foo.SCAN" to "Pensive" using menu menuScan : Illegal choice
    Did you mean "Passive"?
ErrorCan't set "bar.SCAN" to "100"  : Bad Field value
    Did you mean "10 second"?
ErrordbLoadRecords: failed to load '../../base-git/badchoice.db'
Error: Failed to load: ../../base-git/badchoice.db
```

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticRun.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticRun.c
@@ -28,6 +28,7 @@
 #include "dbCommonPvt.h"
 #include "dbStaticLib.h"
 #include "dbStaticPvt.h"
+#include "dbAccess.h"
 #include "devSup.h"
 #include "special.h"
 
@@ -479,8 +480,17 @@ long dbPutStringNum(DBENTRY *pdbentry, const char *pstring)
                 epicsEnum16 value;
                 long status = epicsParseUInt16(pstring, &value, 0, NULL);
 
-                if (status)
+                if (status) {
+                    status = S_db_badChoice;
+                    if(pflddes->field_type==DBF_MENU) {
+                        dbMenu  *pdbMenu = (dbMenu *)pflddes->ftPvt;
+                        dbMsgPrint(pdbentry, "using menu %s", pdbMenu->name);
+
+                    } else if(pflddes->field_type==DBF_DEVICE) {
+                        dbMsgPrint(pdbentry, "no such device support for '%s' record type", pdbentry->precordType->name);
+                    }
                     return status;
+                }
 
                 index = dbGetNMenuChoices(pdbentry);
                 if (value > index && index > 0 && value < USHRT_MAX)
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -23,9 +23,11 @@
 #include <errno.h>
 #include <ctype.h>
 
+#include "epicsAssert.h"
 #include "epicsStdio.h"
 #include "cantProceed.h"
 #include "epicsString.h"
+#include "epicsMath.h"
 
 /* Deprecated, use epicsStrnRawFromEscaped() instead */
 int dbTranslateEscape(char *dst, const char *src)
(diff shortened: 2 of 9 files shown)


mdavidsaver:
https://en.wikipedia.org/wiki/Levenshtein_distance should work better

mdavidsaver:
Switching the similarity algorithm significantly improves the suggestions.  I also renamed the new API call to `epicsStrSimilarCmp()` following the pattern of `epicsStrCaseCmp()`.

modules/database/src/ioc/dbStatic/dbStaticRun.c
```diff
@@ -479,8 +479,16 @@ long dbPutStringNum(DBENTRY *pdbentry, const char *pstring)
...
                 long status = epicsParseUInt16(pstring, &value, 0, NULL);
 
-                if (status)
+                if (status) {
+                    if(pflddes->field_type==DBF_MENU) {
+                        dbMenu  *pdbMenu = (dbMenu *)pflddes->ftPvt;
+                        dbMsgPrint(pdbentry, "using %s.dbd", pdbMenu->name);
+
+                    } else if(pflddes->field_type==DBF_DEVICE) {
+                        dbMsgPrint(pdbentry, "Not devsup for %s", pdbentry->precordType->name);
```
anjohnson: Many record types define their own menus which thus aren't found in a menuXxx.dbd file. Please change the first `dbMsgPrint()` text to `"using menu '%s'"` to avoid confusing users.
> Can't set "baz.OIF" to "incremental" using menu aoOIF : No digits to convert
>   Did you mean "Incremental"?

The second message text would read better if worded `"no such device support for '%s' record type"`.
> Can't set "bar.DTYP" to "Soft Chanel" no such device support for 'ai' record type : No digits to convert
>   Did you mean "Soft Channel"?

We should also change the "No digits to convert" which doesn't make much sense in the above context. Luckily this code is exactly the right place to do that, just change the value of `status` right here to something with a more appropriate message.

modules/libcom/src/misc/epicsString.c
```diff
@@ -358,3 +360,60 @@ unsigned int epicsMemHash(const char *str, size_t length, unsigned int seed)
...
+        dist1[0] = a+1;
+
+        for(b=0; b<lB; b++) {
+            size_t delcost = dist0[b+1] + 1,
+                   inscost = dist1[b] + 1,
+                   subcost = dist0[b],
+                   mincost = delcost;
+
+            if(A[a]!=B[b])
+                subcost++;
```
anjohnson, line 410: It's unfortunate that `YES` and `yes` give the same result of 0.0 as `YES` and `NO`. The "cost" of just changing the case of a character should be less than that of replacing it with a completely different character – I found [this site](https://asecuritysite.com/forensics/simstring) to be useful for confirming that, but it didn't give me any alternative algorithms that take case directly into account.

I'm not suggesting you should do or repeat the comparison after just upper- or lower-casing both strings, it should be possible to integrate case differences into the similarity calculation, although the result wouldn't be a true Levenshtein distance.

I originally wondered if we could just adjust `subcost` by different amounts here when there's only a case difference vs a completely different character, but I suspect it isn't quite that easy, `inscost` and `delcost` also get 1 added to higher up.

anjohnson (commented):
Some inline comments. Could use a release notes entry about the new routine `epicsStrSimilarCmp()` that mentions where it's being used.

anjohnson:
How about calling it `epicsStrSimilarity()` without the `Cmp` part since the return value doesn't really have the same meaning as the other `str*cmp()` routines?

modules/database/src/ioc/dbStatic/dbStaticRun.c
mdavidsaver: I'll reword the two messages.

wrt. "No digits to convert" I agree that it's confusing.  Are you thinking of testing for specific cases like `S_stdlib_noConversion`, or unconditionally overwriting with some other `S_*` code?  The only casualty I can imagine is someone trying to set at out of range number eg. SCAN=100.  Also, what would be a more appropriate message?

modules/libcom/src/misc/epicsString.c
mdavidsaver: hum.  I hadn't considered this.  Maybe multiply the current (integer) costs by 2x so the case in-sensitive comparison can be added in as (effectively) one half.  Since the cost is propagated without being tested, and the output is normalized, I think this could work.

I'll think about it.  If this turns out to be difficult, I'm inclined to leave it as a future improvement.

mdavidsaver:
> How about calling it `epicsStrSimilarity()`

Oh mighty Name Tzar, so it shall be!

modules/database/src/ioc/dbStatic/dbStaticRun.c
anjohnson: Setting SCAN=100 wouldn't fail in the `epicsParseUInt16()` routine anyway. I'm not too picky if the result makes sense for the error, just returning S_db_badChoice would probably work – "Illegal choice".

modules/libcom/src/misc/epicsString.c
anjohnson: Given that `lmax` which is used for normalization is initially set to the larger string length I concluded there might be other places where a factor might be needed. Also I wasn't sure if 2x was the right factor, probably best make that a macro and see what effect it has. Thanks for considering this.

modules/database/src/ioc/dbStatic/dbStaticRun.c
mdavidsaver: Updated.

modules/libcom/src/misc/epicsString.c
mdavidsaver: Giving half weight to case insensitive comparison seems to work without much effort.  I've updated the examples in https://github.com/epics-base/epics-base/pull/104#issue-567842371

mdavidsaver:
Draft release note

### Add epicsStrSimilarity()

Add epicsStrSimilarity() to epicsString.h which uses edit distance as an approximate comparison.
Enables a new "Did you mean ..." suggestion when a .db file provides an invalid value for a DBF_MENU or DBF_DEVICE field.

modules/libcom/src/misc/epicsString.c
```diff
@@ -358,3 +360,79 @@ unsigned int epicsMemHash(const char *str, size_t length, unsigned int seed)
     }
     return hash;
 }
+
+static
+char lower(char c)
```
anjohnson: We use `toupper()` in `epicsStrCaseCmp()` and `tolower()` in other libCom routines, not sure why you felt the need to define your own version.

anjohnson (approved):
Looks good, thanks.

modules/libcom/src/misc/epicsString.c
mdavidsaver: Because I forgot about the `to*` and only tried `man lower`...

anjohnson: Okay. We normally cast the argument to int to prevent a compiler warning on VxWorks.

mdavidsaver: Done.

mdavidsaver:
Applied as of 3f209efa3c4d25c7bf91af23288e021cfe2e44ea

PR #101, opened 2021-01-20
Merge RTEMS5 changes and rebase on a recent 7.0 base (was bcrtems5)

brendanchandler:
These changes start with a base of the current 7.0 branch and then merge mdavidsaver/rtems5 followed by heinz/7.0.  Hopefully that picks up all the rtems5 changes so far, and puts them on a recent base of the main 7.0 branch.

Unfortunately, I don't have everything working yet.  Compiling with RTEMS5 on pc686 works.  Compiling with RTEMS 4.10.2 on pc386 still has issues.

I'm going to try to fix the rest of the issues, but I thought I'd put this out here FYI in case anyone wants to help.

PR #90, opened 2020-09-05
Enable Epics for RTEMS5 (posix)

hjunkes:

Issue #29, opened 2019-07-03
libCom/calc optimization and alignment issues


anjohnson:
I recently noticed [this change](https://github.com/hir12111/epics-base/commit/121aa19025601391c44bdab74e5dc766ed28e5f0) in Diamond's RTEMS-4.11 work, which might explain some of the optimization issues we've been seeing with both calcPerform.c and postfix.c.

When the calls to `memcpy()` that copy double literals into/out of the byte-code stream are being optimized, maybe in some cases the optimizer doesn't realize that the byte-code end of the copy isn't guaranteed to be 8-byte aligned, although the other end is. A later commit removed this change to postfix.c with a comment that the `-mstrict-align` flag was sufficient to resolve their problem, so we definitely should add that to our RTEMS build flags — @hjunkes didn't you disable optimization for one or both of the libCom/calc files? We have also seen similar issues with the Microsoft optimizer though, which doesn't have `-mstrict-align`.

Diamond's solution in the above commit was to wrap the call to `memcpy()` which forced the optimizer to not make assumptions about the alignment. Is there a better, cross-platform way to do this?

mdavidsaver:
More targeted tests would be `-fno-builtin-memcpy` and/or explicitly casting the source to `char*`.

https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

@hjunkes @hir12111

EmilioPeJu:
I remember trying `-fno-builtin-memcpy` and not getting the expected result.
` -mstrict-align` was also preferred over it, because it doesn't have such performance impact.
I also attempted casting to (char *), but it didn't work as the factor that the compiler considered is the constant copying size, having a volatile variable for passing the size would work, because the compiler couldn't make any assumption about it, but that solution is a bit ugly.

hjunkes:
I disabled optimization just for postfix.c
...
#include "postfixPvt.h"
#include "shareLib.h"

#ifdef RTEMS_HAS_ALTIVEC
#pragma GCC push_options
#pragma GCC optimize ("O0")
#endif

/* declarations for postfix */

/* element types */
typedef enum {
        OPERAND,
...

mdavidsaver:
> I disabled optimization just for postfix.c

I agree that this is the appropriate scope for a workaround.

> I remember trying `-fno-builtin-memcpy` and not getting the expected result.

@hir12111 I wonder if this can be ignored, and if your "noinline" change work by itself?  Is `-mstrict-align` required as well?

[...]

PR #105, opened 2021-02-05
Rtems5 changes from hjunkes, anjohnson, and mdavidsaver (rebased instead of merged)

brendanchandler:
These are hjunkes, anjohnson, and mdavidsaver's changes for RTEMS5 plus some of my own to fix compile issues or test issues.

I had to create a new pull request after switching to rebasing the changes rather than merging them.  Github doesn't let you rewrite history of a pull request, and would have required adding a messy merge commit on top of my other pull request branch.


mdavidsaver:
Some previous discussion on #101.  Intended to supersede #90

mdavidsaver:
I see that the RTEMS 4.9/4.10 builds are failing in epicsMessageQueueTest.  It looks like the broken impl has been reintroduced.  cf. 1eeac6da2f3c1d5c626e4876f2bad22abd523317

```
  epicsMessageQueueTest.tap .... 
  not ok 67 - Sent 420 (should be 500)
  not ok 69 - Sent 420 (should be 500)
  not ok 71 - Sent 420 (should be 500)
```

mdavidsaver:
> ... undefined reference to `accept4'

The 5.1 build is breaking due to cf3173b6f424121f4239e8cdfcbc944749425c87 (my doing).  It seems that while the prototype for `accept4` is included (from newlib?), no implementation is provided by libbsd.

```
$ nm -g i386-rtems5/pc386/lib/libbsd.a |grep -w accept
00000000 T accept
         U accept
$ nm -g i386-rtems5/pc386/lib/libbsd.a |grep -w accept4
```

I've addressed this with 43bd5ee1c29806274a0ae8be852a944d9babdda1

modules/libcom/src/osi/os/RTEMS-posix/osdSock.h
```diff
@@ -0,0 +1,91 @@
...
+#include <net/if.h>
+#include <netdb.h>
+#include <unistd.h> /* close() and others */
+
+#ifdef RTEMS_LEGACY_STACK
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int select(int  n,  fd_set  *readfds,  fd_set  *writefds, fd_set *exceptfds, struct timeval *timeout);
```
mdavidsaver: Isn't this definition available from `<sys/select.h>`?

modules/libcom/src/osi/os/posix/osdFindSymbol.c
```diff
@@ -7,9 +7,29 @@
 \*************************************************************************/
 /* osi/os/posix/osdFindSymbol.c */
 
+#include "epicsFindSymbol.h"
+
+/* RTEMS posix but without libbsd */
+#ifdef RTEMS_LEGACY_STACK
```
mdavidsaver: Shouldn't this be `#ifdef __rtems__`?

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -25,7 +25,9 @@
 #include <unistd.h>
 
 #if defined(_POSIX_MEMLOCK) && _POSIX_MEMLOCK > 0
-#include <sys/mman.h>
+#ifndef RTEMS_LEGACY_STACK // seems to be part of libbsd?
```
mdavidsaver: Better to add `&& !defined(__rtems__)` to the condition above.  I don't think there is any reason to mlock() on RTEMS.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -426,6 +428,7 @@ LIBCOM_API
 void epicsThreadRealtimeLock(void)
 {
 #if defined(_POSIX_MEMLOCK) && _POSIX_MEMLOCK > 0
```
mdavidsaver: Maybe conditionally defined `HAVE_MLOCK` above to avoid an increasingly complex duplicate condition.

src/tools/makeTestfile.pl
```diff
@@ -24,75 +24,59 @@
...
+if ($TA =~ /^win32-x86/ && $HA !~ /^win/) {
+    # Use WINE to run win32-x86 executables on non-windows hosts.
+    # New Debian derivatives have wine32 and wine64, older ones have
+    # wine and wine64. We prefer wine32 if present.
+    my $wine32 = "/usr/bin/wine32";
+    $wine32 = "/usr/bin/wine" if ! -x $wine32;
+    $error = $exec = "$wine32 $exe";
 }
-
-# Ensure that Windows interactive error handling is disabled.
```
mdavidsaver: We need to keep the SEM mask handling.

configure/os/CONFIG.Common.RTEMS
```diff
@@ -95,15 +96,33 @@ MODEXT=.obj
...
 #--------------------------------------------------
-# Operating system flags
-OP_SYS_LDLIBS += -lrtemsCom -lc -lrtemscpu -lCom -lnfs -lm
-OP_SYS_LDFLAGS += $(CPU_CFLAGS) -u Init \
+# Operating system compile & link flags
+OP_SYS_CFLAGS += -D__LINUX_ERRNO_EXTENSIONS__
+
+OP_SYS_CFLAGS_NET_yes = -DRTEMS_LEGACY_STACK
+OP_SYS_CFLAGS_NET_no  = -DHAVE_SOCKADDR_SA_LEN=1
+OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_HAS_NETWORKING))
```
mdavidsaver, line 103: It would be good to find a better way to handle this condition.  In the absence of proper identification macros, I happen to know that the "classic" stack headers don't define `SOCK_CLOEXEC`.  So `osdSock.h` could test this:

```c
#if defined(__rtems__) && !defined(SOCK_CLOEXEC)
#  define RTEMS_LEGACY_STACK
#endif
```

I think that HAVE_SOCKADDR_SA_LEN is no longer used, and can be eliminated now.

configure/CONFIG_SITE
```diff
@@ -103,7 +103,6 @@
 # Which target architectures to cross-compile for.
 #  Definitions in configure/os/CONFIG_SITE.<host>.Common
 #  may override this setting.
-CROSS_COMPILER_TARGET_ARCHS=
```
mdavidsaver, line 106: This needs to stay.

configure/os/CONFIG.Common.RTEMS-beatnik
```diff
@@ -8,14 +8,18 @@ EXE = .elf
 RTEMS_BSP = beatnik
 RTEMS_TARGET_CPU = powerpc
 GNU_TARGET = powerpc-rtems
-ARCH_DEP_CFLAGS += -DMY_DO_BOOTP=NULL
+# optimization trouble in postfix.c
+ARCH_DEP_CFLAGS += -DRTEMS_HAS_ALTIVEC
```
mdavidsaver, line 12: Maybe just add `-mstrict-align` to CFLAGS and CXXFLAGS for this target?  see https://github.com/epics-base/epics-base/issues/29#issuecomment-508805177

configure/os/CONFIG.Common.RTEMS-qoriq_e500
```diff
@@ -0,0 +1,48 @@
...
+#will use bootp
+#ARCH_DEP_CFLAGS += -DMY_DO_BOOTP=NULL
+
+#ARCH_DEP_CFLAGS += -DRTEMS_NETWORK_CONFIG_MBUF_SPACE=2048
+#ARCH_DEP_CFLAGS += -DRTEMS_NETWORK_CONFIG_CLUSTER_SPACE=5120
+
+#netbsdlib
+#ARCH_DEP_CFLAGS += -I$(RTEMS_BASE)/powerpc-rtems5/qoriq_e500/lib/include
+
+#OP_SYS_LDLIBS += -lbspExt #does not use posix stuff ... want to ignore
```
mdavidsaver, line 22: Does this imply that the VME bus support isn't working for the mvme2500 target?

modules/database/test/std/link/lnkCalcTest.c
```diff
@@ -157,7 +157,7 @@ static void testCalc()
 
 MAIN(lnkCalcTest)
 {
-    testPlan(0);
+    testPlan(30);
```
mdavidsaver: oops.  This is an unrelated bug fix.

modules/libcom/RTEMS/Makefile
```diff
@@ -27,8 +31,12 @@ rtemsCom_SRCS += setBootConfigFromNVRAM.c
...
 rtemsCom_SRCS += epicsMemFs.c
+rtemsCom_SRCS += epicsNtp.c
 
-ifeq ($(RTEMS_BSP),pc386)
+ifeq ($(RTEMS_TARGET_CPU),i386)
+rtemsCom_SRCS += ne2kpci.c
+endif
+ifeq ($(RTEMS_BSP),pc686)
 rtemsCom_SRCS += ne2kpci.c
 endif
```
mdavidsaver: Can this condition include ne2kpci.c twice?

modules/libcom/RTEMS/posix/rtems_config.c
```diff
@@ -0,0 +1,164 @@
...
+#define CONFIGURE_INIT_TASK_STACK_SIZE (32 * 1024)
+#define CONFIGURE_INIT_TASK_INITIAL_MODES RTEMS_DEFAULT_MODES
+#define CONFIGURE_INIT_TASK_ATTRIBUTES RTEMS_FLOATING_POINT
+#endif
+*/
+
+//#define RTEMS_PCI_CONFIG_LIB
+//#define CONFIGURE_PCI_LIB PCI_LIB_AUTO
+
+#ifndef RTEMS_LEGACY_STACK
```
mdavidsaver: Should this be `__RTEMS_MAJOR__ > 4`?

modules/libcom/RTEMS/posix/rtems_config.c
```diff
@@ -0,0 +1,164 @@
...
+ */
+#if !defined(mpc604) && !defined(__mc68040__) && !defined(__mcf5200__) && \
+    !defined(mpc7455) && !defined(__arm__)  && !defined(__nios2__)
+    /* don't have RTC code */
+#define CONFIGURE_APPLICATION_NEEDS_RTC_DRIVER
+#endif
+
+#if defined(LIBBSP_I386_PC386_BSP_H)
+#define RTEMS_BSD_CONFIG_DOMAIN_PAGE_MBUFS_SIZE (64 * 1024 * 1024)
+#elif defined(LIBBSP_POWERPC_QORIQ_BSP_H)
```
mdavidsaver: Testing guard macros is brittle.  A better solution would be to add something like `rtems_config_CPPFLAGS += -DBSP_$(RTEMS_BSP)`.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1159 @@
...
+
+/* these settings are needed by the rtems startup
+ * may provide by dhcp/bootp
+ * or environments from the "BIOS" like u-boot, motboot etc.
+ * these are settings within FHI infrastructure (01/13/2021)
+ */
+char rtemsInit_NTP_server_ip[16] = "141.14.142.121";
+char bootp_server_name_init[128] = "1001.1001@141.14.128.9:/Volumes/Epics";
+char bootp_boot_file_name_init[128] = "/Volumes/Epics/myExample/bin/RTEMS-beatnik/myExample.boot";
+char bootp_cmdline_init[128] = "/Volumes/Epics/myExample/iocBoot/iocmyExample/st.cmd";
```
mdavidsaver: Need to remove this FHI site configuration.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1159 @@
...
+{
+        rtems_printer printer;
+
+        (void)arg;
+
+        rtems_print_printer_printf(&printer);
+        rtems_stack_checker_report_usage_with_plugin(&printer);
+
+        if (exit_code == 0) {
+                puts("*** END OF TEST ***");
```
mdavidsaver: Not really a functional issue, but someone is certain to ask if this appears during a production IOC reboot.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1159 @@
...
+static void
+default_network_dhcpcd(void)
+{
+    static const char default_cfg[] = "clientid FHI test client\n";
+    rtems_status_code sc;
+    int fd;
+    int rv;
+    ssize_t n;
+
+    fd = open("/etc/dhcpcd.conf", O_CREAT | O_WRONLY,
```
mdavidsaver: Should `dhcpd.conf` (and maybe `/etc/hosts`) be separate files captured at compile time?  It seems likely that someone will want site configuration in here.  I think the mechanics of `epicsMakeMemFs.pl` could be reused for this.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1159 @@
...
+/*
+ *  Telnet daemon configuration
+ */
+rtems_telnetd_config_table rtems_telnetd_config = {
+  .command = SHELL_ENTRY,
+  .arg = NULL,
+  .priority = 99, // if RTEMS_NETWORK and .priority == 0 bsd_network_prio should be used ...
+  .stack_size = 0,
+  .client_maximum = 5, // should be 1, on RTEMS and Epics it makes only sense for one connection a time
+  .login_check = NULL,
```
mdavidsaver, line 894: umm...  This doesn't seem like a good default.

modules/libcom/src/osi/os/RTEMS-posix/devLibVMEOSD.c
```diff
@@ -0,0 +1,6 @@
+#ifdef RTEMS_LEGACY_STACK
+#include "../RTEMS-kernel/devLibVMEOSD.c"
+#else
+#pragma message "\n VME-Support only with RTEMS Legacy stack\n"
```
mdavidsaver, line 4: This statement confuses me.  Maybe I don't understand the full scope of what is in `libbsd.a`.  I thought this was only network stuff, include NIC drivers?  This comment implies that some PCI bus infrastructure is there as well.

mdavidsaver (changes requested):
I've made another pass though most of the changes.  One confusion is about testing `RTEMS_LEGACY_STACK` vs. `__RTEMS_MAJOR__>4`.  There are more tests of `RTEMS_LEGACY_STACK` than I expect.

configure/CONFIG_SITE
brendanchandler: added it back

modules/libcom/src/osi/os/RTEMS-posix/osdSock.h
brendanchandler: I removed this unneeded declaration.

src/tools/makeTestfile.pl
brendanchandler: I switched makeTestfile.pl to the same version from my earlier merge request, which I gave more attention to while merging.  This issue and some others should be resolved, but it'll be worth a 2nd look to be sure.

modules/libcom/src/osi/os/posix/osdFindSymbol.c
brendanchandler: I'm not sure....  The comment indicates its for when RTEMS has posix support but is not using libbsd.  RTEMS_LEGACY_STACK seems to be the same as RTEMS_HAS_NETWORKING--I guess the assumption is when networking is enabled we're not using libbsd?  And I suppose without libbsd, we don't have these dlopen symbols defined?

@hjunkes Can you clarify?

modules/libcom/src/osi/os/posix/osdThread.c
brendanchandler: Switched to just ifndef __rtems__.  mlock() is defined with rtems posix, but throws an error if called.

21int mlock( const void *addr, size_t len )
22{
23  (void) addr;
24  (void) len;
25
26  rtems_set_errno_and_return_minus_one( ENOMEM );
27}

modules/libcom/src/osi/os/posix/osdThread.c
brendanchandler: agreed

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I removed it... presumably its not desired?

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I removed the FHI details, and almost replaced with empty strings, but decided to keep a generic example instead, just so its clear what format the strings are expected to be in.

  99char rtemsInit_NTP_server_ip[16] = "127.0.0.1";
 100char bootp_server_name_init[128] = "1001.1001@127.0.0.1:/epics";
 101char bootp_boot_file_name_init[128] = "/epics/myExample/bin/RTEMS-beatnik/myExample.boot";
 102char bootp_cmdline_init[128] = "/epics/myExample/iocBoot/iocmyExample/st.cmd";

configure/os/CONFIG.Common.RTEMS-qoriq_e500
anjohnson: I heard that from @hjunkes a few weeks ago. The APS will need it though, so unless Heinz gets to it first that's going to be another task for @brendanchandler after this port has been merged.

modules/libcom/RTEMS/epicsNtp.c
```diff
@@ -0,0 +1,86 @@
+/*************************************************************************\
+* (C) 2014 David Lettier.
+* http://www.lettier.com/
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
```
anjohnson: This file and the associated header appear to be heavily-modified versions of the C source from [this repo](https://github.com/lettier/ntpclient) which is provided under the BSD 3-clause license. Incorporating this code into EPICS is thus legal, but we can't change the license terms of someone else's software and the wording of our standard header might imply that we are trying to do that. Please remove lines 4 and 5 from both files, and add this instead:
```
* SPDX-License-Identifier: BSD-3-Clause
```

modules/libcom/RTEMS/epicsNtp.c
```diff
@@ -0,0 +1,86 @@
...
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netdb.h>
+#include <arpa/inet.h>
+#include <rtems.h>
+#include "epicsNtp.h"
+
+int epicsNtpGetTime(char *ntpIp, struct timespec *now)
```
anjohnson, line 21: I don't see where this routine is getting called, are these files actually needed at all?

modules/libcom/src/osi/os/RTEMS-posix/devLibVMEOSD.c
brendanchandler: I'm not 100% sure either, but I found this:
https://devel.rtems.org/browser/rtems-libbsd/libbsd.txt

"The libbsd makes FreeBSD subsystems like TCP/IP, USB, SD and some more usable for RTEMS."

modules/libcom/RTEMS/kernel/rtems_init.c
```diff
@@ -42,7 +42,7 @@
 #include "logClient.h"
 #include "osiUnistd.h"
 #include "iocsh.h"
-#include "osdTime.h"
+#include "../../src/osi/os/RTEMS-kernel/osdTime.h"
```
anjohnson: It shouldn't be necessary to have this horrendous relative path here (anything else you see like this is a bad code smell and should be examined very closely). The appropriate version of osdTime.h should have been installed into a standard include directory before this gets built, so I'm sure this change isn't actually needed.

modules/libcom/RTEMS/rtems_netconfig.c
```diff
@@ -15,8 +15,27 @@
  */
 #include <stdio.h>
 #include <bsp.h>
+
+#ifndef VERSION_INT
+#  define VERSION_INT(V,R,M,P) ( ((V)<<24) | ((R)<<16) | ((M)<<8) | (P))
```
anjohnson: `VERSION_INT()` is available in epicsVersion.h, please include that instead of redefining it here.

modules/libcom/RTEMS/epicsNtp.c
brendanchandler: It seems like its being called from modules/libcom/RTEMS/posix/rtems_init.c:
```
1027    if (epicsNtpGetTime(rtemsInit_NTP_server_ip, &now) < 0) {                                                                                                                        
1028      printf ("***** Can't get time from ntp ...\n");                                                                                                                                
1029    } else {                                                                                                                                                                         
1030      if (clock_settime(CLOCK_REALTIME, &now) < 0){                                                                                                                                  
1031        printf ("***** Can't set time: %s\n", rtems_status_text (sc));                                                                                                               
1032      } else {                                                                                                                                                                       
1033        strftime(timeBuff, sizeof timeBuff, "%D %T", gmtime(&now.tv_sec));                                                                                                           
1034        printf("time from ntp : %s.%09ld UTC\n", timeBuff, now.tv_nsec);   
```

anjohnson: Ah, and GitHub wasn't showing the contents of that new file so my browser search didn't find it – my bad!

modules/libcom/RTEMS/posix/rtems_init.c
anjohnson: Ack on the example, but I'd suggest not using the localhost IP which could be confusing. Maybe pick a non-routable address such as 192.168.5.1 or 10.0.5.1 say.

mdavidsaver: I think this highlights the complications of how we are configuring RTEMS.  eg. the collection of weak symbols in rtems_init.c seems organic (and brittle).  The switch from RTEMS 4 -> 5 is an opportune time to straighten this out.

As the comment above this block says, this is getting overwritten from DHCP.  It should, but isn't, earlier overwritten from local non-volatile storage (cf. the ugly untested mess in setBootConfigFromNVRAM.c).  (order of decreasing preference: DHCP, NVRAM, application defaults, static defaults).  "application defaults" being a wishlist item for me.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1159 @@
...
+        if (!strncmp(name, "interface", 9) && !strcmp(value,  ifnamebuf))
+			 strncpy(iName, value, 16);
+		if (!strncmp(name, "reason", 6) && !strncmp(value, "BOUND", 5)){
+            printf ("Interface %s bounded\n", iName);
+            bound = 1;
+		}
+		if (bound) {
+            // as there is no ntp-support in rtems-libbsd, we call our own client
+            if(!strncmp(name, "new_ntp_servers", 15))
+                strcpy(rtemsInit_NTP_server_ip,value);
```
mdavidsaver: Some use of `sizeof(rtemsInit_NTP_server_ip)` would be prudent.

modules/libcom/src/osi/os/RTEMS-posix/devLibVMEOSD.c
anjohnson: Should we ask the RTEMS developers whether/how they (plan to) support PCI and VME bus interfaces? For most applications that pick a board with an interface to it, VMEbus access is going to be an essential part of that application. This seems like a question for the RTEMS devel mailing list (which I don't follow) to get their take on what we should do.

Note that for targets that don't have a VME interface, the devLibVMEOSD.c file still needs to be compiled and define this dummy:
```C
/* No known VME interface here, provide a dummy */
devLibVME *pdevLibVME;
```
That makes it possible for IOC applications to be built for all targets without having to exclude the ones that don't have VME. Any attempt to access the VMEbus on such a target will fail at runtime though, unless the application also provides another implementation and sets that pointer (PCI to VME adaptors exist, and some EPICS sites use them via this mechanism).

hjunkes:
> I've made another pass though most of the changes. One confusion is about testing `RTEMS_LEGACY_STACK` vs. `__RTEMS_MAJOR__>4`. There are more tests of `RTEMS_LEGACY_STACK` than I expect.

It may be that some queries to RTEMS_LEGACY_STACK crept in, which are not necessary. It was because I had first started with __RTEMS_MAJOR__>4. But then the idea of the distinction between "kernel" and "posix" came into play. And when we had that, but RTEMS4 should also work with posix-support and then I had the problem with new stack (libbsd) or the old implementation. I will search everything again carefully.

modules/libcom/RTEMS/posix/rtems_init.c
hjunkes: I tried to get the whole thing to work with DHCP so we can debug reasonably in qemu. I left the stuff with "ConfigFromNVRAM" aside and wanted to look at it again when I have more time. With U-Boot everything looks different too.

modules/libcom/RTEMS/posix/rtems_init.c
hjunkes: Herewith I added the DHCP_option for "rtems_cmdline" (DHCP option 129).
"
nodhcp6
ipv4only
option ntp_servers
option  rtems_cmdline
option tftp_server_name
option bootfile_name
"
I'll be happy to take another look at epicsMakeMemFs.pl, though.

hjunkes: Just to keep you guys up to date.
I am currently working with qemu with xilinx_zynq_a9_qemu as target.
I have re-enabled the mechanism of epicsMakeMemFs.pl. So when the target boots and no DHCP information is received, there is a local directory into which one can "go" into:
...
***** Setting up file system *****
***** Using compiled in file data *****
-> /iocshTestBadArgIndirect.cmd - ok
-> /iocshTestBadArg.cmd - ok
-> /iocshTestSuccessIndirect.cmd - ok
-> /iocshTestSuccess.cmd - ok
***** Initializing NFS *****
check for time registered , C++ initialization ...
***** Preparing EPICS application *****
***** set directory to / *****
chdir("/")
***** Starting EPICS application *****
...
Parallel to the qemu - stuff I am setting up a vde-switch, dnsmask, ntp-server and nfs-server 
to be able to emulate a "normal" environment completely.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I changed most of these to use 0 or NULL, which lets RTEMS pick a default.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I changed a lot of this string handling to use sizeof() to check sizes, use snprintf() so that we can get reliable null termination (vs. strncpy), and use strtok_r rather than strtok.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I'll  take a look at this, but I propose handling it as a separate pull request.  Unless people think it needs to go in with this one?

mdavidsaver:
I see that test are now run for rtems5/pc686, and get quite far.  The failures all seem to be due to missing/incomplete network configuration.  Based on these two lines, I don't think there are any network interfaces configured.

```
  osiSockTest.tap .............. 
  not ok 19 - Found non-loopback interface
  not ok 20 - Successes 0
...
  testInetAddressUtils.tap .. 
  not ok 65 - Found 0 interfaces
...
```

However, the raw `osiSockTest.tap` file shows the loopback being configured, but not the `re0` interface.

```
...
 ---- dhcpDone Event Unknown state 1
-------------- IFCONFIG -----------------
re0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=8009b<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,VLAN_HWCSUM,LINKSTATE>
        ether 52:54:00:12:34:56
        inet6 fe80::5054:ff:fe12:3456%re0 prefixlen 64 scopeid 0x1
ifconfig: re0: no media types?
        inet 127.0.0.1 netmask 0xffffff00
        inet6 ::1 prefixlen 128
        groups: lo
...
```

This suggests to me that there are two problems.  First that DHCP isn't working.  Second that `osiSockDiscoverBroadcastAddresses` isn't working.

mdavidsaver:
FYI. In trying to understand the state of libbsd I've come across `rtemsbsd/include/bsp/nexus-devices.h` in that source which may be the place where BSP is mapped to NIC driver(s).  Based on this I think that the old/legacy network stack is still needed by all of our current `RTEMS-*` targets except: `pc*86`.  Two of the targets being added are also mentioned: `qoriq_e500`, `xilinx_zynq_a9_qemu`.

I notice that there are mentions of `RTEMS_BSD_DRIVER_TSEC` (present on the mvme3100), but this isn't being selected.

```sh
$ grep LIBBSP ./rtemsbsd/include/bsp/nexus-devices.h
#if defined(LIBBSP_ARM_REALVIEW_PBX_A9_BSP_H)
#elif defined(LIBBSP_ARM_BEAGLE_BSP_H)
#elif defined(LIBBSP_ARM_LPC32XX_BSP_H)
#elif defined(LIBBSP_M68K_GENMCF548X_BSP_H)
#elif defined(LIBBSP_ARM_XILINX_ZYNQ_BSP_H)
#elif defined(LIBBSP_ARM_ATSAM_BSP_H)
#elif defined(LIBBSP_ARM_ALTERA_CYCLONE_V_BSP_H)
#elif defined(LIBBSP_ARM_IMX_BSP_H)
#elif defined(LIBBSP_ARM_LPC24XX_BSP_H)
#elif defined(LIBBSP_I386_PC386_BSP_H)
#elif defined(LIBBSP_POWERPC_QORIQ_BSP_H)
#elif defined(LIBBSP_POWERPC_TQM8XX_BSP_H)
```

anjohnson:
I've been building this branch against RTEMS-5 for the various targets that APS is interested in.

The qoriq_e500 build was failing to find `mkimage` which is called by the target-specific `MUNCH_CMD` definition to create a bootable image. RTEMS-5 provides a `$(RTEMS_TOOLS)/bin/mkimage.py` script which works instead, requiring this patch:
```patch
index 41415c9ca..1a5fa1919 100644
--- a/configure/os/CONFIG.Common.RTEMS-qoriq_e500
+++ b/configure/os/CONFIG.Common.RTEMS-qoriq_e500
@@ -29,7 +29,7 @@ MUNCHNAME = $(PRODNAME:%$(EXE)=%$(MUNCH_SUFFIX))
 define MUNCH_CMD
        $(RTEMS_TOOLS)/bin/$(OBJCOPY_FOR_TARGET) -O binary $< $@
        gzip -9 -f $@
-       mkimage -A ppc -O linux -T kernel -a 0x4000 -e 0x4000 -n $* -d $@.gz $*.img 
+       $(RTEMS_TOOLS)/bin/mkimage.py -A ppc -O linux -T kernel -a 0x4000 -e 0x4000 -n $* -d $@.gz $*.img
 endef
 
 include $(CONFIG)/os/CONFIG.Common.RTEMS
```
That generates images for all the test harnesses in my `bin/RTEMS-qoriq_e500` directory, but I haven't worked out how to actually boot one on our MVME2500 board yet.

I also applied the workaround for the uC5282 that Michael suggested on the rtems-users list, although I haven't checked this change with the RTEMS-4 builds yet and that change may break other software that adds a different `-std=` flag, so I would prefer a different solution. At least this lets it me compile libCom.a though:
```patch
diff --git a/configure/os/CONFIG.Common.RTEMS-uC5282 b/configure/os/CONFIG.Common.RTEMS-uC5282
index 2cb215aca..4493b5d70 100644
--- a/configure/os/CONFIG.Common.RTEMS-uC5282
+++ b/configure/os/CONFIG.Common.RTEMS-uC5282
@@ -8,6 +8,7 @@
 RTEMS_BSP = uC5282
 RTEMS_TARGET_CPU = m68k
 ARCH_DEP_CFLAGS += -DMY_DO_BOOTP=NULL
+ARCH_DEP_CXXFLAGS += -std=c++98
 
 MUNCH_SUFFIX = .boot
 define MUNCH_CMD
```

This shows how far my builds are now getting:
```
tux% ls lib/RTEMS-*
lib/RTEMS-beatnik:
libCom.a

lib/RTEMS-mvme2100:
libCom.a

lib/RTEMS-mvme3100:
libCom.a

lib/RTEMS-pc686-qemu:
libca.a      libdbRecStd.a  libpvAccessCA.a   libpvData.a      librtemsCom.a
libCom.a     libnt.a	    libpvAccessIOC.a  libpvDatabase.a
libdbCore.a  libpvAccess.a  libpvaClient.a    libqsrv.a

lib/RTEMS-qoriq_e500:
libca.a      libdbRecStd.a  libpvAccessCA.a   libpvData.a      librtemsCom.a
libCom.a     libnt.a	    libpvAccessIOC.a  libpvDatabase.a
libdbCore.a  libpvAccess.a  libpvaClient.a    libqsrv.a

lib/RTEMS-uC5282:
libCom.a
```

We haven't talked about what targets we want this PR to be able to build before we merge it. I propose that the older RTEMS-4 targets should continue to build (so upgrading EPICS doesn't break existing builds) plus at least RTEMS-5/pc686. It looks like we may be able to get the qoriq_e500 to work, but I don't know where libbspExt fits into that; @hjunkes can devLib probe the PCI and/or VME bus on this board, and if so what code is doing the memProbe?

hjunkes:
To boot the image file on the mvme2500, you need the attached board description file 
[mvme2500.dtb.zip](https://github.com/epics-base/epics-base/files/5998321/mvme2500.dtb.zip).

I put it on the a CF-Card of the board

you have to set the environment
serverip (tftp-server)
ipaddr (own ip)

mmc rescan 0;fatload mmc 0 20000000 mvme2500.dtb;tftpboot 10000000 libComTestHarness.img;bootm 10000000 - 20000000;reset

MVME2500=> printenv
baudrate=115200
bdev=sda2
bootargs=root=/dev/ram rw console=ttyS0,9600n8 ramdisk_size=700000 cache-sram-size=0x10000
bootcmd=run rtems_mmc
bootdelay=-1
bootfile=uImage
bootsata=ext2load scsi 0:1 $loadaddr usb/MVME_uImage; ext2load scsi 0:1 0x2000000 usb/MVME_RamDisk; ext2load scsi 0:1 0xC00000 usb/MVME.dtb; bootm $loadad
dr 0x2000000 0xC00000;
bootsd=mmc rescan 0; fatload mmc 0 $loadaddr MVME_uImage; fatload mmc 0 0x2000000 MVME_RamDisk; fatload mmc 0 0xC00000 MVME.dtb; bootm $loadaddr 0x2000000
 0xC00000;
bootusb=usb start; fatload usb 0:1 $loadaddr MVME_uImage; fatload usb 0:1 0x2000000 MVME_RamDisk; fatload usb 0:1 0xC00000 MVME.dtb; bootm $loadaddr 0x200
0000 0xC00000;
cachemode=write
clearmem=on
dcache=on
enumerate=on
eth1addr=EC:9E:CD:19:3C:62
eth2addr=EC:9E:CD:19:3C:63
ethact=eTSEC1
ethaddr=EC:9E:CD:19:3C:61
ethport=all
ethprime=eTSEC1
fileaddr=10000000
filesize=0x2fdd
flashboot=run flashbootargs && sf probe 0 && sf read $loadaddr 120000 4E0000 && sf read c00000 A0000 20000 && bootm $loadaddr - c00000
flashbootargs=setenv bootargs root=/dev/$bdev ro init=$lnxinit ip=$ipaddr:$serverip:$gatewayip:$netmask:$hostname:$netdev:off console=ttyS0,$baudrate $oth
bootargs;
gatewayip=10.162.166.1
hostname=unknown
icache=on
ipaddr=141.14.128.89
l2cache=on
lnxinit=/sbin/init
loadaddr=1000000
loads_echo=1
model=MVME2500
netdev=eth0
netmask=255.255.240.0
nfsboot=run nfsbootargs && tftpboot $loadaddr MVME_uImage && tftpboot c00000 MVME.dtb && bootm $loadaddr - c00000
nfsbootargs=setenv bootargs root=/dev/nfs rw init=$lnxinit nfsroot=$serverip:$rootpath ip=$ipaddr:$serverip:$gatewayip:$netmask:$hostname:$netdev:off cons
ole=ttyS0,$baudrate $othbootargs;
partition=scsi 0:0
powerondiags=on
rootpath=/opt/nfsroot
rtems_mmc=mmc rescan 0;fatload mmc 0 20000000 mvme2500.dtb;fatload mmc 0 10000000 rtems.img;bootm 10000000 - 20000000;reset
rtems_myTests=mmc rescan 0;fatload mmc 0 20000000 mvme2500.dtb;tftpboot 10000000 $tfile.img;bootm 10000000 - 20000000;reset
rtems_test=mmc rescan 0;fatload mmc 0 20000000 mvme2500.dtb;tftpboot 10000000 rtems.img;bootm 10000000 - 20000000;reset
scsidevs=0
serial#=E170306
serverip=141.14.128.9
u1=tftp 0x1000000 spi-mvme2500.bin
u2=sf probe 0; sf erase 0 0x90000; sf write 0x1000000 0 0x90000
vxboot=tftpboot $vxbootfile && setenv bootargs $vxbootargs && bootvx
vxbootargs=motetsec(0,0)10.162.138.231:vxWorks h=10.162.138.231 e=10.162.166.25:ffffff00 b=10.162.166.1 u=vxworks pw=vxworks f=0x80
vxbootfile=vxWorks.st

Environment size: 2564/8188 bytes

hjunkes:
> We haven't talked about what targets we want this PR to be able to build before we merge it. I propose that the older RTEMS-4 targets should continue to build (so upgrading EPICS doesn't break existing builds) plus at least RTEMS-5/pc686. 

I'm relatively far along with testing xilinx_zynq_a9_qemu and would like to include that one in the targets list as well.

anjohnson:
@hjunkes Thanks. My board is an MVME2502 which doesn't have a CF card slot, but I think I see how to load everything over tftp and maybe even NFS.

anjohnson:
Progress on the MVME2502, but not complete success yet. I loaded the files into memory using the `tftb` command and started it:
```
MVME2502=> bootm 10000000 - 20000000; reset
WARNING: adjusting available memory to 30000000
## Booting kernel from Legacy Image at 10000000 ...
   Image Name:   libComTestHarness
   Created:      2021-02-17  18:15:01 UTC
   Image Type:   PowerPC Linux Kernel Image (gzip compressed)
   Data Size:    1880327 Bytes = 1.8 MiB
   Load Address: 00004000
   Entry Point:  00004000
   Verifying Checksum ... OK
## Flattened Device Tree blob at 20000000
   Booting using the fdt blob at 0x20000000
   Uncompressing Kernel Image ... OK
   Loading Device Tree to 00ff9000, end 00fff04a ... OK
�ʄ��}[k��V��◎�&Φ�<��SS/�J��!�SS/�J��C�cRb��c=��ĵ��cRb��c=��ĵ�cRb\c=��ĵ�Rb��c�Λ�·��Մ|Bakn���ƈ��J��ǌ�Fƈ����
                         b�K�����{
                                  �K:Χ�O˅�;BB!B+ǌBBSR!B^k�k�!�g��72�!r\+1�c���Ä�ǤÄ�Ǥ��/��c~g�Go��CN�C���GBw��bGBwbGbw��BZ�{�1B�BZ��1�BZ�B�bGw�m

...

                                          ¸j���c"rbSr�cCRcccs�5S�;S!Bsss��jscccs���3s��jscccCQ[XÏ�![XÏ�[XϏ��A@R��CA@R����ĕB�o�K��b����V��K�~O�����K��cw��=�!�b��C��[�����=osJCCK�       ��RK�H#����C�G�CFC��F��F��r����cƆBg_*uGc�D\CSB�R���w<�-ujb�ꩾ�k���CRb�@L�{+��R

U-Boot 2010.06 (Jun 18 2013 - 14:03:16)

...

Ser#: E10A722
I-cache enabled. (L1CSR1: 0x00000001)
D-cache enabled. (L1CSR0: 0x00000001) (write-through)
SCSI:  Net:   eTSEC1, eTSEC2, eTSEC3
Bootreg = a6 BootDev: SPI0   Switch: SPI0
MVME2502=> 
```

From the way that the random characters were displaying periodically I'm pretty sure it was running libCom tests, but RTEMS' idea of the console's baud rate evidently wasn't the same as U-Boot's, and I didn't have an easy way to change that on my end.

How/where do I tell RTEMS what baud rate to use?

mdavidsaver:
> c/src/lib/libbsp/powerpc/qoriq/configure.ac:RTEMS_BSPOPTS_SET([BSP_CONSOLE_BAUD],[*],[115200])

with rtems-source-builder cf. `--with-rtems-bspopts "BSP_CONSOLE_BAUD=blah"`.

FYI. with uboot `setenv baudrate 115200`.

hjunkes:
> FYI. with uboot `setenv baudrate 115200`.

To make it permanent
saveenv

anjohnson:
Thanks guys. I know this seems crazy but I prefer to keep all our VME Console ports running at 9600 baud since that's the way all our CPU boards have always come configured, even U-Boot on the MVME2500. Thus I need to fix it in RTEMS, not change the rate U-Boot uses.

I guess building it into our BSP is probably the right approach, but is there no way to override it in rtems_config.c or rtems_init.c? We have a file setBootConfigFromNVRAM.c so it would be nice if eventually a version of that could read and adopt the environment settings from U-Boot (even the network configuration parameters are available there) instead of compiling in our own defaults.

anjohnson:
```
    EPICS Test Harness Results
    ==========================

Failing Program           Tests  Faults
---------------------------------------
epicsAlgorithm               22       1
epicsCalcTest               630      88
epicsMathTest                35      12
epicsStdlibTest             199      14

Failed 4/37 test programs. 115/3040 subtests failed.
Programs=37, Tests=3040, 388 wallclock secs
```

Full output attached: [mvme2500-libcom.txt](https://github.com/epics-base/epics-base/files/6005278/mvme2500-libcom.txt)

mdavidsaver:
I believe RTEMS supports the "usual" `ioctl()` calls on "/dev/ttyS0".  I also see a `bsp_uboot_copy_board_info()` in `<u-boot.h>` as well as a `bsp_uboot_getenv()` in bsp.h for a couple of the PPC boards.

brendanchandler:
> This suggests to me that there are two problems. First that DHCP isn't working. Second that `osiSockDiscoverBroadcastAddresses` isn't working.

I noticed that we run the unit test's main() before DHCP has fully set up.  Using the rtems_shell in place of invoking main(), I can see ifconfig shows no ip for re0 interface for several seconds, and then starts showing it.  I think there might be some missing synchronization.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1174 @@
...
+        if (!strncmp(name, interface, strlen(interface)) &&
+            !strcmp(value,  ifnamebuf)) {
+            snprintf(iName, sizeof(iName), "%s", value);
+        }
+
+        if (!strncmp(name, reason, strlen(reason)) &&
+            !strncmp(value, bound_str, strlen(bound_str))) {
+            printf ("Interface %s bounded\n", iName);
+            bound = 1;
+        }
```
brendanchandler, line 732: @hjunkes, This code is searching for "reason=BOUND" in the dhcpcd hook environment, but I haven't seen that key/value pair in my runs.  If you search the following example output for "all out", there are no "all out ---> reason = BOUND" lines:  https://github.com/epics-base/epics-base/files/6005278/mvme2500-libcom.txt

Was there some documentation that indicated it should send reason=BOUND sometimes?

Rather than search for "reason=BOUND", do you think we should search for at least one interface with "if_up=true" instead, to know that one of the interfaces is up?

mdavidsaver: fyi. `IF_UP` does not imply completion.  While a DHCP client is searching an interface will by "up", but have no address assigned, and thus not be usable.

One place to start would be printing the complete `env` before the loop.

hjunkes: Exactly, that is the reason. This also works fine for me. The messages can also go away at some point. I'm also already in the process of what the presets of the argv's to change.
Unfortunately I failed grandiosely when I tried everything with RTEMS6 :-(

But a nice Makefile for creating RTEMS here:
https://gitlab.com/c-mauderer/rtems-simulation-environment

brendanchandler: Unless I mistake your meaning, the complete `env` is already being printed out as the messages starting with "all out".  The output below shows my qemu session and three sections of environment printed out for each of the three times rtems_dhcpcd_hook callback is called.

```
***** Primary Network interface : re0 *****
all out ---> interface = re0
all out ---> reason = PREINIT
all out ---> pid = 1
all out ---> ifmetric = 201
all out ---> ifwireless = 0
all out ---> ifflags = 4294936643
all out ---> ifmtu = 1500
all out ---> interface_order = re0
all out ---> if_up = false
all out ---> if_down = true

***** Primary Network interface : re0 *****
all out ---> interface = re0
all out ---> reason = CARRIER
all out ---> pid = 1
all out ---> ifmetric = 201
all out ---> ifwireless = 0
all out ---> ifflags = 4294936643
all out ---> ifmtu = 1500
all out ---> interface_order = re0
all out ---> if_up = false
all out ---> if_down = true


***** Primary Network interface : re0 *****
all out ---> interface = re0
all out ---> reason = IPV4LL
all out ---> pid = 1
all out ---> ifmetric = 201
all out ---> ifwireless = 0
all out ---> ifflags = 4294936643
all out ---> ifmtu = 1500
all out ---> interface_order = re0
all out ---> if_up = true
all out ---> if_down = false
all out ---> new_ip_address = 169.254.163.246
all out ---> new_subnet_cidr = 16
all out ---> new_network_number = 169.254.0.0
all out ---> new_subnet_mask = 255.255.0.0
all out ---> new_broadcast_address = 169.254.255.255

```

mdavidsaver: Ah, I wasn't reading closely enough this morning.

Looking at `make_env()` in `dhcpcd/script.c` from libbsd it looks like entries in this array may be relied upon to have a nil terminator.  So one way to simplify (or at least condense) this loop might be:

```c
for(; *env; env++) {
    const char *val;
#define IF(NAME) if(strncmp(*env, NAME "=", 1+strlen(NAME))==0 && (val=1+strlen(NAME)+*env))

    IF("interface") {
    }
    IF("reason") {
        printf("reason is '%s'\n", val);
    }
    ...
#undef IF
}
```

anjohnson:
```
all out ---> new_ip_address = 169.254.163.246
all out ---> new_subnet_cidr = 16
all out ---> new_network_number = 169.254.0.0
all out ---> new_subnet_mask = 255.255.0.0
all out ---> new_broadcast_address = 169.254.255.255
```
@brendanchandler That IP address is in the zeroconf range so it probably didn't come from dhcp at all. We will need to turn off DHCP and configure the network of our MVME2500 boards directly (which is why I'd like RTEMS to read its settings from U-Boot), but I have other stuff I have to work on now – let me know if you want access to the board.

mdavidsaver: > Was there some documentation that indicated it should send reason=BOUND sometimes?

fyi. The `env` array is exporting the same key/value pairs normally exported to dhcpcd hook scripts on Linux as environment variables.  cf. `man dhcpcd/dhcpcd-run-hooks.8.in` for a list of hooks.  Also `man dhcpcd/dhcpcd.conf.5.in`.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -0,0 +1,1174 @@
...
+    assert(n == (ssize_t) sizeof(default_cfg) - 1);
+
+    static const char fhi_cfg[] =
+        "nodhcp6\n" \
+        "ipv4only\n" \
+        "option ntp_servers\n" \
+        "option rtems_cmdline\n" \
+        "option tftp_server_name\n" \
+        "option bootfile_name\n" \
+        "define 129 string rtems_cmdline\n";
```
mdavidsaver: Probably should add `timeout 0` to disable DHCP timeout (imo. a must for a production IOC).  Could also add `noipv6`.

I see `waitip 4` which, when desirable, seems like it would remove the need for a custom hook in `rtems_init.c`.

Also, for testing, it looks like you could set something like:

```
interface eth0
static ip_address=192.168.0.10/24
static routers=192.168.0.1
static domain_name_servers=192.168.0.1
```

cf. `man dhcpcd/dhcpcd.conf.5.in` from the libbsd source.

mdavidsaver: Ah, right `ipv4only` and `noipv6` are practically equivalent.

modules/libcom/src/osi/os/posix/osdFindSymbol.c
brendanchandler: I couldn't find a reason we don't use dlopen() type symbols in RTEMS, so I just removed this #ifdef section specific to RTEMS.

dirk-zimoch (approved):

configure/os/CONFIG.Common.RTEMS
brendanchandler: I removed the unused HAVE_SOCKADDR_SA_LEN.

RTEMS_LEGACY_STACK is set based on RTEMS_HAS_NETWORKING, which is a macro that RTEMS provides.  It is set based on whether the user ran configure --with-networking enabled or disabled.  This seems like a pretty good way to identify whether RTEMS is using the legacy networking stack or not.. though perhaps we could just use the RTEMS_HAS_NETWORKING macro directly--I'm not sure the reason we created RTEMS_LEGACY_STACK, maybe @hjunkes or someone familiar with the history here knows the answer.

I suggest we might keep RTEMS_LEGACY_STACK set the way it is.

configure/os/CONFIG.Common.RTEMS-qoriq_e500
brendanchandler: I vote for addressing this in a separate pull request.

hjunkes:
I would like it if it did find its way into pr/105.
It only expands the target list and has no other influence on the PR?

The same applies to configure/os/CONFIG.Common.RTEMS-xilinx_zynq_a9_qemu

Heinz

configure/os/CONFIG.Common.RTEMS-qoriq_e500
brendanchandler: > I vote for addressing this in a separate pull request.

To clarify, I want to deal with VME bus support in a separate pull request, as Andrew suggested.  Not suggesting this change be put in a separate pull request.  Sorry for the confusion there.

configure/os/CONFIG.Common.RTEMS-beatnik
brendanchandler: Is the testing from #29 sufficient to include -mstrict-align?  I can add this flag, but don't have a means to verify.

modules/database/test/std/link/lnkCalcTest.c
brendanchandler: @mdavidsaver Should I take any action on this?  Leave it as 30?  Revert back to 0?

anjohnson: Is this is the only change in that commit? If so I will apply it directly to the 7.0 branch.

brendanchandler: > Is this is the only change in that commit? If so I will apply it directly to the 7.0 branch.

There were a few other changes, see 14140acd787e990f34bf1d4d827d6bb4903f282a

anjohnson: Yeah, 14140acd787e990f34bf1d4d827d6bb4903f282a also added building of the link-type tests on RTEMS to the modules/database/test/std/link/Makefile. I can cherry-pick this whole commit into 7.0 (in which case @brendanchandler can drop it when he next rebases), or just change the `testPlan()` call there. @mdavidsaver your call.

mdavidsaver: 14140ac looks reasonable.

anjohnson: @mdavidsaver I'm sure you're busy but your terse communication style doesn't let us know whether you think 14140acd787e990 should be cherry-picked into the 7.0 branch or left as a commit in @brendanchandler's rtems5 PR. The change you highlighted here was just part of that commit.

mdavidsaver: My apologies Andrew Johnson, I will endeavor to express myself in this particular instance with what I hope you will find to be adequate verbosity.  Having looked though the commit in question (14140ac), I have formed the opinion that it is reasonable to apply in its entirety, and without delay.  Building and running the link type unittests on RTEMS targets is something which should already be happening.  The fact that this is not the case may be seen as an omission on my part, which I am grateful to Brendan Chandler for noticing and correcting.

modules/libcom/RTEMS/Makefile
brendanchandler: > Can this condition include ne2kpci.c twice?

Yeah, looks like it was a merge issue.  I think only checking for CPU=i386 is sufficient, so I removed the BSP=pc686 check.

modules/libcom/RTEMS/posix/rtems_config.c
brendanchandler: > Should this be `__RTEMS_MAJOR__ > 4`?

I think so.  Changed to check rtems version instead.

modules/database/test/std/link/lnkCalcTest.c
anjohnson: Cherry-picked into 7.0 as bbb4d86f787b74579062aa77db9a98aee8e7ba80

brendanchandler:
I believe that all the review feedback's been addressed, except for things that won't be part of this pull request (vme bus support, etc)

configure/os/CONFIG.Common.RTEMS
```diff
@@ -95,15 +96,32 @@ MODEXT=.obj
...
+OP_SYS_CFLAGS += -D__LINUX_ERRNO_EXTENSIONS__
+
+OP_SYS_CFLAGS_NET_yes = -DRTEMS_LEGACY_STACK
+OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_HAS_NETWORKING))
+
+ifeq ($(RTEMS_HAS_POSIX_API),yes)
+POSIX_CPPFLAGS = -D_GNU_SOURCE -D_DEFAULT_SOURCE
+endif
+
+OP_SYS_LDLIBS_posix_NET_yes = -ltftpfs -lnfs -lz -telnetd
```
anjohnson: This needs to be `-ltelnetd`

anjohnson:
My RTEMS-5.1 builds were all originally built using the libbsd network stack, which was obviously a mistake since only the RTEMS-pc686-qemu and RTEMS-qoriq_e500 targets compile this branch to completion with that. Since it currently matters which RTEMS version(s) and/or network stack must be used it would be good to document that for each BSP in its `CONFIG.Common.RTEMS-<bsp>` file (add a "Configurations" section to the comments). The build could even error out if it finds a configuration that it knows is broken, just like I did for the pc386/686 BSPs.

After rebuilding the beatnik BSP with the old stack the EPICS build still fails because it needs the libbspExt library and header which I don't know how to install. I assume I'll have the same problem with the other mvme BSPs so I haven't tried them.

With the uC5282 BSP built against the old stack (after fixing `-ltelnetd`) I get a different failure, the rtems/netcmds-config.h header and these commands in `CONFIGURE_SHELL_USER_COMMANDS` are not available with the legacy stack:
```
    rtems_shell_HOSTNAME_Command
    rtems_shell_NETSTAT_Command
    rtems_shell_TCPDUMP_Command
    rtems_shell_PFCTL_Command
    rtems_shell_SYSCTL_Command
```

Those do seem to be the last problems with the uC5282 BSP though, with these fixed everything builds for that target. Since I've done all the changes I'm committing and pushing them now.

configure/os/CONFIG.Common.RTEMS
anjohnson: Fixed in my commit.

mdavidsaver:
> fails because it needs the libbspExt library and header which I don't know how to install.

You might have some luck picking off https://github.com/mdavidsaver/epics-base/commit/a33a193c40f1122f1c2f60c233edc1a5a64bbfe0

hjunkes:
You can find a libBspExt here: https://github.com/hjunkes/rtems-libbspext
This one works for beatnik and I fumbled my way to it. 
I would like to address (and if possible solve) this problem next week at the Codeathlon.

hjunkes:
I picked the cherries out of Michael commit ("RTEMS: Drop dependency on bspExt").
Could compile, looks good. Tests on VMEbus hardware on Tuesday during the codeathlon.

mdavidsaver:
So far I've only tested this change with a mvme3100.

mdavidsaver:
FYI. I have an updated set of RTEMS5 toolchains available, including additional BSPs (listed at https://github.com/epics-base/ci-scripts/pull/56#issuecomment-792071262).  This can be used by updating the `.ci` submodule, and changing the ubuntu version the RTEMS CI jobs use to `ubuntu-20.04`.  Additional jobs can be added which also set eg. `RTEMS_TARGET=mvme3100`.

brendanchandler:
> FYI. I have an updated set of RTEMS5 toolchains available, including additional BSPs (listed at [epics-base/ci-scripts#56 (comment)](https://github.com/epics-base/ci-scripts/pull/56#issuecomment-792071262)). This can be used by updating the `.ci` submodule, and changing the ubuntu version the RTEMS CI jobs use to `ubuntu-20.04`. Additional jobs can be added which also set eg. `RTEMS_TARGET=mvme3100`.

I tried these new targets over on my private branch and all targets build except mvme2100, mvme167, and uC5282.
https://github.com/brendanchandler/epics-base/runs/2067873106?check_suite_focus=true

Did these targets work in RTEMS 4.10?  I took a look at mvme2100 and there's a conflict with soft FPU flag between RTEMS and Epics....  epics is expecting a hard fpu, whereas RTEMS51 BSP is using a soft fpu.  I'm curious if this was working and tested under RTEMS 4.10 or if it could be an existing problem.

mdavidsaver:
> Did these targets work in RTEMS 4.10?

I can't speak for the m68k boards.  For the mvme2100, yes.  A recent [example](https://epics.anl.gov/tech-talk/2020/msg01714.php) plucked from tech-talk.

```
Welcome to rtems-4.10.2(PowerPC/Generic (classic FPU)/mvme2100) on MVME 2100
```

I wonder what "classic FPU" means.

hjunkes:
> I picked the cherries out of Michael commit ("RTEMS: Drop dependency on bspExt").
> Could compile, looks good. Tests on VMEbus hardware on Tuesday during the codeathlon.


[cherry_pick.zip](https://github.com/epics-base/epics-base/files/6118491/cherry_pick.zip)

anjohnson:
> I wonder what "classic FPU" means.

Different PowerPC CPUs have gone through several different FPUs over the years. The MVME2100, 2300, 2700, 5100, 5500 and 6100 boards support the original hardware floating point instructions; the MVME3100 and MVME4100 didn't, but at least one of them had something different (SPE?) which only handled 32-bit floats so VxWorks uses soft-float instead. The newer CPUs such as on the MVME2500 have yet another different FPU which doesn't follow the strict IEEE754 math rules in hardware but apparently you can add software to correct that using traps – this is almost certainly why our epicsMathTest checks using NaN and Inf are failing on that board.

mdavidsaver:
> You might have some luck picking off [mdavidsaver@a33a193](https://github.com/mdavidsaver/epics-base/commit/a33a193c40f1122f1c2f60c233edc1a5a64bbfe0)

I see now that this patch will only work with some of the powerpc BSPs.

hjunkes:
After Till adapted the network driver of the beatnik board for libbsd (nexus) I could build rtems5 for beatnik (MVME6100) with the new stack (libbsd). With this pr/105 I can build EPICS and use the VMEbus (SIS3316, MAXv, ...) without rtems-libbspext!  I have taken out the query for the LEGACY_STACK from libcom/src/osi/os/RTEMS-posix/devLibVMEOSD.c which I stupidly introduced. 
I still need to make changes to the dhcpcd. The option 129 (rtems_cmdline) is no longer recognized or requested. But this could be due to a new DHCP server at our site (infoblox). Here I still have to test.

brendanchandler:
> After Till adapted the network driver of the beatnik board for libbsd (nexus) I could build rtems5 for beatnik (MVME6100) with the new stack (libbsd). With this pr/105 I can build EPICS and use the VMEbus (SIS3316, MAXv, ...) without rtems-libbspext! I have taken out the query for the LEGACY_STACK from libcom/src/osi/os/RTEMS-posix/devLibVMEOSD.c which I stupidly introduced.

I am also playing around with VMEBus and removed the #if LEGACY_STACK check in RTEMS-posix/devLibVMEOSD.c, but I don't have any hardware locally to observe it working yet.  Its good to hear its working.

bfrk:
> I had to create a new pull request after switching to rebasing the changes rather than merging them. Github doesn't let you rewrite history of a pull request, and would have required adding a messy merge commit on top of my other pull request branch.

IME `git push --force <my-fork-on-github>` works fine.

brendanchandler:
> IME `git push --force <my-fork-on-github>` works fine.

I'll keep that in mind for next time.  Thanks for the tip.

anjohnson:
I solved the issue of building the uC5282 target, all m68k BSPs seem to build a 68k-specific `bspExtMemProbe()` routine (look for memProbe.c), so it was just a matter of excluding the new definition in that case to get it to build. I also discovered that the mvme167 BSP does not have the VMEbus interface routines that devLib relies on, so I assume that nobody has ever actually built EPICS for that board and will delete the configuration files for it (in a different branch).

I'm currently in the process of merging this into 7.0 and testing the result, I'd appreciate it if nobody commits/pushes to the 7.0 branch for the rest of today. I will push my merged result to this branch for GHA testing, then to the 7.0 branch if all goes well.

hjunkes:
I found one more problem (reccaster client was not running)
In modules/libcom/RTEMS/posix/rtems_config.c
#define  CONFIGURE_MAXIMUM_FILE_DESCRIPTORS 150 -> 64
It must not be higher than FD_SETSIZE which is defined by newlib to 64 in RTEMS.

anjohnson:
Okay, I'll make that change too with a comment – thanks!

I'm wondering if we should rename `RTEMS-kernel` to `RTEMS-score` since that's the name they now use for the older APIs (which would require various related changes in the build system). This is probably our last chance to do that, any comments on the idea?

anjohnson:
Only 2 target builds failed (I'm not planning to wait for Appveyor):
+ RTEMS-pc686-qemu timed out while running the tests, although they were still chugging along slowly. This was due to the slow startup under QEMU. All tests passed when I ran them on RHEL-7, and I don't think they were slow there.
+ Ubuntu 2016 gcc-4.8 failed a test.

I'm currently working on the Release Notes, then will commit and push everything to the 7.0 branch.

PR #107, opened 2021-02-05
Fix a bug which causes timers to expire early on modern Linux systems

mark0n:
Fixes #106. Tested on
- [X] Linux
- [ ] Darwin
- [ ] VxWorks
- [ ] RTEMS

Issue #106, opened 2021-02-05
Timers expire early


mark0n:
Generally, timers are supposed to sleep for _at least_ the amount of time they were set for. However, the current EPICS timer implementation behaves differently:
```
# Testing timer accuracy
ok  2 - timerCount == nTimers
ok  3 - 0.486553 < 5.000000, delay = 1.000000 s, error = -0.004866 s (0.5 %)
ok  4 - 0.440644 < 5.000000, delay = 1.100000 s, error = -0.004847 s (0.4 %)
ok  5 - 0.404795 < 5.000000, delay = 1.200000 s, error = -0.004858 s (0.4 %)
ok  6 - 0.373543 < 5.000000, delay = 1.300000 s, error = -0.004856 s (0.4 %)
ok  7 - 0.346924 < 5.000000, delay = 1.400000 s, error = -0.004857 s (0.3 %)
ok  8 - 0.326921 < 5.000000, delay = 1.500000 s, error = -0.004904 s (0.3 %)
ok  9 - 0.303757 < 5.000000, delay = 1.600000 s, error = -0.004860 s (0.3 %)
ok 10 - 0.285826 < 5.000000, delay = 1.700000 s, error = -0.004859 s (0.3 %)
ok 11 - 0.270031 < 5.000000, delay = 1.800000 s, error = -0.004861 s (0.3 %)
ok 12 - 0.255543 < 5.000000, delay = 1.900000 s, error = -0.004855 s (0.3 %)
ok 13 - 0.242964 < 5.000000, delay = 2.000000 s, error = -0.004859 s (0.2 %)
ok 14 - 0.231398 < 5.000000, delay = 2.100000 s, error = -0.004859 s (0.2 %)
ok 15 - 0.220428 < 5.000000, delay = 2.200000 s, error = -0.004849 s (0.2 %)
ok 16 - 0.211106 < 5.000000, delay = 2.300000 s, error = -0.004855 s (0.2 %)
ok 17 - 0.202014 < 5.000000, delay = 2.400000 s, error = -0.004848 s (0.2 %)
ok 18 - 0.194164 < 5.000000, delay = 2.500000 s, error = -0.004854 s (0.2 %)
ok 19 - 0.186316 < 5.000000, delay = 2.600000 s, error = -0.004844 s (0.2 %)
ok 20 - 0.181187 < 5.000000, delay = 2.700000 s, error = -0.004892 s (0.2 %)
[...]

PR #108, opened 2021-02-09
Add more timer tests

mark0n:
The existing tests are testing many things in each test. In contrast to that the tests I'm adding here are each testing one basic property (they can also act as a sort of documentation). Each test is designed to fail for one reason only - creating a fresh timer queue for each test helps with isolation.

Note that currently these tests are leveraging C++11 features. We could either

- merge these tests into a new 7.1 branch (which will be the first EPICS version which requires a C++11 compiler)
- refactor them to make them compatible with C++98 for now (and then refactor them for better readability later when we are working on 7.1)
- put `ifdef`s around the new code to include these tests only with compilers which support C++11

Which one do you prefer?

### DIFF
--- a/configure/os/CONFIG.win32-x86.win32-x86
+++ b/configure/os/CONFIG.win32-x86.win32-x86
@@ -185,7 +185,9 @@ RES=.res
 
 # MS Visual C++ doesn't recognize *.cc as a C++ source file,
 # so C++ compiles get the flag -TP
-COMPILER_CXXFLAGS = -TP
+# We want MSVC to report a reasonable C++ version in the __cplusplus macro so
+# we need to set -Zc:__cplusplus (this is at least required for MSVC 19.xx).
+COMPILER_CXXFLAGS = -TP -Zc:__cplusplus
 
 # Operating system flags
 OP_SYS_CFLAGS =
--- a/src/libCom/test/epicsTimerTest.cpp
+++ b/src/libCom/test/epicsTimerTest.cpp
@@ -77,13 +77,13 @@ void testTimerExpires() {
   testDiag("Timer expires");
   epicsTimerQueueActive &queue = epicsTimerQueueActive::allocate(false);
   {
-    handler h1(queue);
+    epicsEvent waitForExpire;
+    handler h1(queue, [&] { waitForExpire.trigger(); } );
     const double arbitrary_time { 0.3 };
     h1.start(arbitrary_time);
 
-    epicsThreadSleep(arbitrary_time + 0.1);
+    waitForExpire.wait();
 
-    testOk(h1.getExpireCount() == 1, "timer expired exactly once");
     verifyExpirationTime(h1.getDelay() - arbitrary_time, timeTolerance);
   } // destroy timer
   queue.release();
@@ -92,16 +92,20 @@ void testTimerExpires() {
 void testMultipleTimersExpire(std::vector<double> && sleepTime) {
   epicsTimerQueueActive &queue = epicsTimerQueueActive::allocate(false);
   {
+    std::atomic<int> count { 3 };
+    epicsEvent all_timers_expired;
     std::vector<handler> hv;
     hv.reserve(3);
     for (unsigned i = 0; i < 3; i++) {
-      hv.emplace_back(queue);
+      hv.emplace_back(queue, [&] {
+        if (!--count) { all_timers_expired.trigger(); }
+      });
     }
     for (unsigned i = 0; i < hv.size(); i++) {
       hv[i].start(sleepTime[i]);
     }
 
-    epicsThreadSleep(*std::max_element(sleepTime.cbegin(), sleepTime.cend()) + 0.1);
+    all_timers_expired.wait();
 
     for (unsigned i = 0; i < hv.size(); i++) {
       testOk(hv[i].getExpireCount() == 1, "timer expired exactly once");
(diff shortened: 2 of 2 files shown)


anjohnson:
The 3.15 branch doesn't provide `epicsTime::getMonotonic()`, and is never going to require C++11, so this PR seems to be targeting the wrong branch.

I don't have a problem with adding optional tests that require specific compiler features to run. The `epicsUnitTest.h` API even provides a placeholder for them so you don't have to adjust the number of tests in the plan, that's what `testSkip()` is for – described in the comments in that header file.

mark0n:
I replaced `getMonotonic()` by `getCurrent()`. With this change this PR can indeed go into the 3.15 branch. The new tests will now be skipped automatically when the compiler doesn't support C++11.

The failing tests on Ubuntu are due to #106, they should pass once #107 is merged.

mark0n:
The TAP file shows it is skipping my new tests on MacOS. Shouldn't clang 12 on MacOS support C++11?

mark0n:
This is ready for another round of review. Note that the timer tests sporadically failing on MacOS are the _old_ tests. The new tests are skipped on MacOS since the `__cplusplus` macro doesn't report compatibility with C++11 (can someone with a little more Mac experience please confirm that that's correct?).

configure/os/CONFIG.win32-x86.win32-x86
```diff
@@ -185,7 +185,9 @@ RES=.res
 
 # MS Visual C++ doesn't recognize *.cc as a C++ source file,
 # so C++ compiles get the flag -TP
-COMPILER_CXXFLAGS = -TP
+# We want MSVC to report a reasonable C++ version in the __cplusplus macro so
+# we need to set -Zc:__cplusplus (this is at least required for MSVC 19.xx).
+COMPILER_CXXFLAGS = -TP -Zc:__cplusplus
```
mdavidsaver, line 190: I hadn't come across the [`-Zc:*` family of options](https://docs.microsoft.com/en-us/cpp/build/reference/zc-conformance?view=msvc-160) before.  I had come across [`/std=*`](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-160), but that is >=2017.

[-Zc:__cplusplus](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-160) is documented as >=2017 (15.7).  Since the older appveyor builds succeed without additional warnings, I guess the older compilers are silently ignoring unknown `-Zc:*` operations?

And hey, with [`/Zc:preprocessor`](https://docs.microsoft.com/en-us/cpp/build/reference/zc-preprocessor?view=msvc-160) the MS preprocessor is finally come up to [C99 conformance](https://devblogs.microsoft.com/cppblog/msvc-preprocessor-progress-towards-conformance/).

src/libCom/test/epicsTimerTest.cpp
```diff
@@ -27,6 +35,150 @@
...
+  double getDelay() const { return expireTime - startTime; }
+private:
+  epicsTimer &timer;
+  int expireCount { 0 };
+  epicsTime startTime;
+  epicsTime expireTime;
+  std::function<void()> expireFct;
+};
+
+void verifyExpirationTime(double delta_t, double tolerance) {
```
mdavidsaver: `delta_t` is slightly confusion to me.   In most c++ code I've seen, `*_t` is a type.

src/libCom/test/epicsTimerTest.cpp
```diff
@@ -27,6 +35,150 @@
...
+  if (!testOk(std::abs(delta_t) < tolerance, "timer expired at the expected "
```
mdavidsaver: This tight unsigned test seems problematic.  Based on discussions elsewhere, I think we are in agreement that this delay should never by negative?  At the same time, on shared CI builder services, it could easily be longer.  Would a better pass condition be: `delta >= 0.0 && (testImpreciseTiming() || delta < tolerance)`?

src/libCom/test/epicsTimerTest.cpp
```diff
@@ -27,6 +35,150 @@
...
+
+void testTimerExpires() {
+  testDiag("Timer expires");
+  epicsTimerQueueActive &queue = epicsTimerQueueActive::allocate(false);
+  {
+    handler h1(queue);
+    const double arbitrary_time { 0.3 };
+    h1.start(arbitrary_time);
+
+    epicsThreadSleep(arbitrary_time + 0.1);
```
mdavidsaver: I'm not enthused by all of the sleeps, especially with hard-coded delays.  I would like to see all of these tests using `epicsEvent`.

mdavidsaver (changes requested):

src/libCom/test/epicsTimerTest.cpp
mdavidsaver: Also, to help with troubleshooting CI build failures, tests like this one should print the values being tested.

mdavidsaver: ... which I just noticed is being done on the next line (which was cropped out).

anjohnson: This PR is for 3.15, which doesn't have `testImpreciseTiming()`.

mdavidsaver: Shouldn't this addition be going into 7.0?

src/libCom/test/epicsTimerTest.cpp
mark0n: :+1: Fixed.

src/libCom/test/epicsTimerTest.cpp
mark0n: Done. The 8 failing tests on Linux are due to #106 and will go away when #107 gets merged. For now I didn't include the `testImpreciseTiming()` part to maintain compatibility with 3.15. I'm happy to add that once this change has been merged up to 7.0.

src/libCom/test/epicsTimerTest.cpp
mark0n: Seems like I forgot to commit my latest changes. Eliminating the first few `epicsThreadSleep`s was easy but I don't have any ideas how to get rid of the ones related to the cancel tests.

mark0n:
Superseded by #145.

PR #109, opened 2021-02-18
Print if some links still disconnected after 5 seconds

mdavidsaver:
Help catch typos resulting from unexpected implicit CA links by printing a short notice if any link are disconnected 5 seconds after `iocInit()`.

```
epics> Notice: 1 link(s) have not yet connected
```

Intended to be a hint that something may be off, which could trigger further investigation with `dbcar`.  Hopefully won't be too annoying in cases where remote links were deliberate.

"Borrows" access to the timer queue created in `callback.c` instead of creating a dedicated thread.

### DIFF
--- a/modules/database/src/ioc/db/callback.c
+++ b/modules/database/src/ioc/db/callback.c
@@ -45,6 +45,7 @@
 #include "dbStaticLib.h"
 #include "epicsExport.h"
 #include "link.h"
+#include "dbLinkPvt.h"
 #include "recSup.h"
 #include "dbUnitTest.h" /* for testSyncCallback() */
 
@@ -404,6 +405,24 @@ void callbackRequestProcessCallbackDelayed(epicsCallback *pcallback,
 {
     callbackSetProcess(pcallback, Priority, pRec);
     callbackRequestDelayed(pcallback, seconds);
+}
+
+static epicsTimerId linkCheck;
+
+static void checkLinks(void *unused)
+{
+    dbLinkConnCheck();
+    epicsTimerQueueDestroyTimer(timerQueue, linkCheck);
+}
+
+/* in callback.c for access to timerQueue */
+void dbLinkCheckStart(void)
+{
+    double delay = dbLinkConnectionCheckDelay;
+    if(delay < 0.0)
+        return;
+    linkCheck = epicsTimerQueueCreateTimer(timerQueue, &checkLinks, NULL);
+    epicsTimerStartDelay(linkCheck, 5.0);
 }
 
 /* Sync. process of testSyncCallback()
--- a/modules/database/src/ioc/db/dbLink.c
+++ b/modules/database/src/ioc/db/dbLink.c
@@ -25,6 +25,7 @@
 #include "ellLib.h"
 #include "epicsTime.h"
 #include "errlog.h"
+#include "epicsStdio.h"
 
 #include "caeventmask.h"
 
@@ -39,7 +40,7 @@
 #include "db_field_log.h"
 #include "dbFldTypes.h"
 #include "dbJLink.h"
-#include "dbLink.h"
+#include "dbLinkPvt.h"
 #include "dbLock.h"
 #include "dbScan.h"
 #include "dbStaticLib.h"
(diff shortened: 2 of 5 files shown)


anjohnson:
With more and more things making use of callback.c's `timerQueue` should we maybe move that (and the other things that use it which aren't callback code) out of callback.c and make it a bit more accessible (but still private to the IOC code)?

For many of our operational IOCs 5 seconds after iocInit is still in the middle of enabling hardware and starting sequence programs, so this message would get lost in the noise they generate. I'd suggest taking the delay from an exported variable so before iocInit the startup script could use `var` to disable it (by setting it to <0) or set to some longer delay appropriate to that IOC.

mdavidsaver:
> should we maybe move that

I thought about it, but wasn't sufficiently motivated this time.  I think I would still be adding the same number of LoC, just in different places.  And I'd feel obligated to rename the new global from `timerQueue` to something less likely to collide, which would add churn.

> taking the delay from an exported variable

`dbLinkInitCheckDelay`?

> lost in the noise

Digressing a bit.  I've been thinking about adding some color to iocsh.  Maybe something like what Linux `printk()` does, with macros for common prefixes: error, warning, info, etc.?  eg.

```c
#ifdef _WIN32
#  define EPICS_ERR "Error:"
#else
#  define EPICS_ERR "\033[31;1mError:\033[0m"
#endif
```

anjohnson:
On naming: This is actually doing a connection check, not an init check, so the variable should probably be `dbLinkConnectionCheckDelay` and the routine should be called `dbLinkConnectionCheck()` (or `dbLinkConnCheck()` if that seems too long although I'm fine with the long name).

Colors: As a name `EPICS_ERR` doesn't say what it's for. However I'm mostly concerned that it uses the target architecture as a proxy for choosing the terminal's escape sequences. I don't particularly want us relying on the terminfo database, but that would be the more correct solution. We should also be careful to print both the color set and clear sequences in the same string (which your example does); on VxWorks it's common for a printf() to get interrupted at one of the conversion boundaries and a message from another task be printed then (it even happens often between a `0x` and the following hex digits). I think this idea needs more discussion; you aren't the first person to propose it, although your solution looks safer than Keenan's was.

mdavidsaver:
> not an init check

I thought "Init" since this is a one time check at (or near) IOC startup.  Whatever, `dbLinkConnectionCheckDelay` it will be.



> I don't particularly want us relying on the terminfo database

Me neither.  This is why I was sticking to the simple/limited [color set](https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit) which is widely portable.

> uses the target architecture as a proxy for choosing the terminal's escape sequences

I was hoping to avoid a runtime condition, which would complicate usage to the point where I probably wouldn't bother.  This morning it occurs to me that sending escapes to errlogPrintf() may be annoying to anyone running a logger.  So I'm giving up on this idea for the time being.

ralphlange:
That's exactly the problem with color output (I think ASYN already does that):
The log files become completely unreadable.

Unless you do run-time checks if there is a terminal connected...

bfrk:
> This morning it occurs to me that sending escapes to errlogPrintf()
> may be annoying to anyone running a logger.  So I'm giving up on this
> idea for the time being.

I have played with the idea some time ago and gave up for the same reason.

mdavidsaver:
Actually... rather than injecting escapes if `istty()`, maybe the way to go is to strip them out if `!istty()`.  Making strings shorter doesn't require a re-alloc, and could be done by errlogThread.  I think the only remaining obstacle would the `isOkToBlock` logic (which directly `fprintf()`s from the calling thread).  Maybe these synchronous prints could emulated with something like `errlogFlush()`.  Getting rid of the `if(isOkToBlock)` conditions would be a nice code simplification.  Combined with eg. switching `errlogPrintf()` to just call `errlogVprintf()` `errlog.c` could give quite a bit shorter...

anjohnson:
> simple/limited color set which is widely portable.

It still uses ANSI escape sequences though. You may never have even seen an ADM-3A (to pick one pre-ANSI terminal type I last used 20+ years ago)...

If we do add color support I think it must be possible for sites to disable it at build-time (per-target arch) and maybe at run-time too. This PR isn't really the place to discuss it though.

mdavidsaver:
Updated with `dbLinkConnectionCheckDelay` control variable.

Draft Release note:

## One time notice of disconnected CA links

A new one-time check is performed after iocInit.
A message is printed if any volatile (eg. CA or PVA) links are disconnected.
This is intended to help catch typos resulting in inadvertent CA links.
By default  this check is performed 5 seconds after iocInit,
though this can be changed or disabled via the new `dbLinkConnectionCheckDelay` iocsh variable.

mdavidsaver:
I'm having second thoughts about this one. #111 seems like it could be the better solution.

anjohnson:
I agree, I think #111 is better.

mdavidsaver:
Abandoning...

PR #110, opened 2021-02-21
Embed GDB helper

mdavidsaver:
Something I've been trying out for the past few ~~months~~ years.  I'm finding it convenient, though so far only from the GDB CLI.  ~~Theoretically, qt-creater should be able to use this, but so far no luck.~~ QtC has its own helper mechanism.

This change adds a [GDB helper script](https://sourceware.org/gdb/current/onlinedocs/gdb/Python-API.html#Python-API) to pretty print record/pv names referenced by some common database types (`DBADDR`, `DBLINK`, `dbCommon`, and `*Record`).  Further the script is [embedded](https://sourceware.org/gdb/current/onlinedocs/gdb/dotdebug_005fgdb_005fscripts-section.html) in libdbCore.so for convenience (as is done by eg. libstdc++).

The result is that a stack frame like:

```gdb
#2  0x00007ffff7ed4751 in dbGet (paddr=0x7fffffffd800, dbrType=0, pbuffer=0x7fffffffd840, options=0x7fffffffd7d8,
    nRequest=0x7fffffffd7e0, pflin=0x0) at ../db/dbAccess.c:957
```

would instead include the PV name (`target.TSEL`) referenced by the `DBADDR*` argument.

```gdb
#2  0x00007ffff7ed4751 in dbGet (paddr=(dbAddr*)(0x7fffffffd800, "target.TSEL"), dbrType=0, pbuffer=0x7fffffffd840, 
    options=0x7fffffffd7d8, nRequest=0x7fffffffd7e0, pflin=0x0) at ../db/dbAccess.c:957
```

I also add three "convenience" functions `$rcast()` (up-cast `dbCommon`), `$record()` (lookup record), and `$elln()` (index `ELLLIST`).

For reasons of security, GDB will ignore embedded scripts by default.  Need to add a line to ~/.gdbinit

> add-auto-load-safe-path /path/to/base/checkout/

Then after libdbCore.so has been loaded.  Verify at GDB shell with:

```gdb
(gdb) info auto-load python-scripts
Loaded  Script                                                                 
...
Yes     epics-gdb-helper.py        
(gdb) info pretty-printer
global pretty-printers:
  EPICS
...                                            
```

### DIFF
--- a/documentation/gdb.md
+++ b/documentation/gdb.md
@@ -1,0 +1,111 @@
+# Debugging with GDB
+
+Starts a process in [GDB](https://sourceware.org/gdb/current/onlinedocs/gdb.html/):
+
+```sh
+gdb --args ../../bin/linux-x86_64-debug/myioc ./st.cmd
+```
+
+Attach to a running process by numeric PID:
+
+```sh
+gdb --pid=##
+```
+
+Shorthand if only one instance is running:
+
+```sh
+gdb --pid=`pgrep myioc`
+```
+
+## GDB shell Cheat sheet
+
+Print stack traces of all threads.  Optionally with pagination disabled:
+
+```gdb
+(gdb) set pagination off
+
+(gdb) thread apply all backtrace
+
+(gdb) set pagination on
+```
+
+Note: consider sending or posting this very long output as an attachment file.
+
+Break when a certain record begins processing:
+
+```gdb
+(gdb) break dbProcess if $_streq(precord.name, "my:record")
+```
+
+Break when a `PACT` field changes:
+
+```gdb
+(gdb) watch -l $record("my:record").pact
+```
+
+Note: Uses custom `$record()` helper.  See below.
+
+## Helpers
+
+If the bundled `epics-gdb-helper.py` is [enabled](#helper-enable),
+then a set of pretty printers are registered for some common database types.
+eg. `dbCommon*` will show the record name.
+`DBLINK*` will show the associated record name, and the link target.
+Also, several convenience are available.
+
+
+Use `$record()` to lookup a record by name:
+
+```gdb
+(gdb) print $record("my:record")
+$1 = (calcRecord*)(0x55555562f6f8, "my:record")
+```
+
+Note: `$record()` may only be called after record allocation during `iocInit()`.
+
+Use `$rcast()` to up-cast `dbCommon*` to the specific record type:
+
+```gdb
+(gdb) print precord
+$1 = (dbCommon*)(0x55555562f6f8, "my:record")
+(gdb) print $rcast(precord)
+$2 = (calcRecord*)(0x55555562f6f8, "my:record")
+```
+
+Use `$elln()` to index into `ELLLIST` linked lists.
+
+```gdb
+(gdb) print $elln(pdbbase.recordTypeList, 2)
+```
+
+### Enable (helper-enable)
+
+
+To automatically [enable](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Auto_002dloading-safe-path.html),
+add a line to `~/.gdbinit`.
+
+```
+add-auto-load-safe-path /path/to/base/checkout/
+```
+
+Or each time run:
+
+```
+(gdb) source /path/to/base/checkout/modules/database/src/ioc/epics-gdb-helper.py
+```
+
+
+Afterwards verify at GDB shell by running:
+
+```gdb
+(gdb) info auto-load python-scripts
+Loaded  Script
+...
+Yes     epics-gdb-helper.py
+
+(gdb) info pretty-printer
+global pretty-printers:
+  EPICS
+...
+```
--- a/documentation/gdb.md
+++ b/documentation/gdb.md
@@ -95,6 +95,9 @@ Or each time run:
 (gdb) source /path/to/base/checkout/modules/database/src/ioc/epics-gdb-helper.py
 ```
 
+Note: Entries in the `safe-path` list may be files (`.so` or `.py`), or directories.
+      When a directory is included, sub-directories are also trusted.
+      eg. `add-auto-load-safe-path /opt/epics` would also trust a `/opt/epics/lib`.
 
 Afterwards verify at GDB shell by running:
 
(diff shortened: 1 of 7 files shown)


mdavidsaver:
fyi. No changes where needed updating from GDB 8 -> 10.

The helper output makes it easier to avoid getting lost when working with eg. a stack trace of a record processing chain with 4x recursive calls to `dbProcess()`.  Definitely something I've gotten used to.

mdavidsaver:
Updated to improve handling of invalid pointers.

As a note, the helper script can be tested/used separately (without changing Base).  Save `epics-gdb-helper.py`, then start GDB and run:

```
(gdb) source /path/to/epics-gdb-helper.py
(gdb) info pretty-printer 
global pretty-printers:
  EPICS
...
```

`source ...` can also be added to `~/.gdbinit`.

mdavidsaver:
Needs update for GDB circa 16.

```
Python Exception <class 'AttributeError'>: module 'gdb' has no attribute 'printing'
```

mdavidsaver:
> Needs update for GDB circa 16.

Done.

fyi. a failure in a helper script is not fatal to GDB.

```
Starting program: /home/.../softIoc 
Python Exception <class 'AttributeError'>: module 'gdb' has no attribute 'printing'
...
[New Thread 0x7ffff777b6c0 (LWP 310879)]
[Thread 0x7ffff777b6c0 (LWP 310879) exited]

Thread 1 "softIoc" hit Breakpoint 1, main (argc=1, argv=0x7fffffffe1b8) at ../softIoc/softMain.cpp:133
```

mdavidsaver:
I have been using a version of this helper script for the past ~5 years.  Not daily, but often enough to take them for granted.  Do others think it is useful enough to include?

My latest iteration adds`documentation/gdb.md`.

If merged, then the only immediate effect would be a ~7kB increase in the size of libdbCore, and a message from GDB when it notices that a helper is present, but not permitted.

```
Starting program: /.../softIoc -d tick.db
warning: File "/.../libdbCore.so.3.25.1" auto-loading has been declined by your `auto-load safe-path' set to "$debugdir:$datadir/auto-load".
To enable execution of this file add
        add-auto-load-safe-path /home/dev/epics/base-git/lib/linux-x86_64-debug/libdbCore.so.3.25.1
line to your configuration file "/home/dev/.gdbinit".
```

ralphlange:
Looks good to me.
This will improve issue reports that include GDB backtrace information. Nice!

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -46,6 +46,27 @@
 #include "dbCommon.h"
 #include "dbJLink.h"
 
+#if defined(__GNUC__) && defined(__ELF__) && defined(__linux__)
+/* Loading of GDB python helper script.
+ *
+ * Need to include the following line in ~/.gdbinit
```
anjohnson: Could this be moved from the top of the source file to the bottom? The current location seems unnecessarily prominent.

documentation/gdb.md
```diff
@@ -0,0 +1,111 @@
...
+Use `$elln()` to index into `ELLLIST` linked lists.
+
+```gdb
+(gdb) print $elln(pdbbase.recordTypeList, 2)
+```
+
+### Enable (helper-enable)
+
+
+To automatically [enabled](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Auto_002dloading-safe-path.html),
```
anjohnson: s/enabled/enable/

documentation/gdb.md
```diff
@@ -0,0 +1,111 @@
...
+add a line to `~/.gdbinit`.
+
+```
+add-auto-load-safe-path /path/to/base/checkout/
+```
+
+Or each time run:
+
+```
+source /path/to/base/checkout/modules/database/src/ioc/epics-gdb-helper.py
```
anjohnson: Should this line start with a `(gdb) ` prompt, and have `gdb` on the block-start line? These also apply to the code-block above.

anjohnson (changes requested):
I'm fine with this in principle, looks useful.

In the event that some architecture/compiler has a problem with the embedded helper, I'd like to see an entry in the `CONFIG_SITE` file which can remove it, say add
```makefile
EMBED_GDB_HELPER = YES
```
so a more specific `CONFIG_SITE` file can set it to `NO`.

Questions:
1. Are there alternatives such as an environment variable which could be used instead of editing `~/.gdbinit`? That doesn't scale well for sites with multiple Base installations.
2. What is the earliest version of GDB that supports this? It probably should be documented (in gdb.md) given that some sites still run RHEL-6 and I doubt that has it.

A few inline comments below too.

mdavidsaver:
> I'd like to see an entry in the `CONFIG_SITE` file which can remove it, say add
> 
> ```makefile
> EMBED_GDB_HELPER = YES
> ```

This seems like excessive paranoia to me.  If it really does fail, then there is just this one block to remove.  If we are going to add __yet another untested build condition__, then I think we would need a better name.  `EPICS_BUILD_BINARY_EMIT_ELF_SPECIAL_SECTION_FOR_GDB_HELPER = YES`.  Which is less than 60 characters!

> 1. Are there alternatives such as an environment variable which could be used instead of editing `~/.gdbinit`

[Not as I read it](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Auto_002dloading-safe-path.html#Auto_002dloading-safe-path).

> 2\. What is the earliest version of GDB that supports this?

I'm not sure, and at the moment I am not inclined to test.  I think it would be difficult, but possible, to unit test this helper script.

What I am falling back on is that the special section will be ignored by really old GDB.  Newer GDB will only load it if permitted, and even then a failure will only result in easily disabled noise during a debugging session.

modules/database/src/ioc/dbStatic/dbStaticLib.c
mdavidsaver: Fair enough :)

mdavidsaver: This section is now consigned to the bottom of this ~4k line file!

anjohnson:
Core: Approved, after minor doc addition about the safe-path.

PR #111, opened 2021-02-21
Add DB link modifier

mdavidsaver:
Follow on, or maybe alternative to, #109 explores adding a new link modifier `DB` to mark a link which should not become a CA link.

```
record(ai, "src") {
    field(INP, "oops DB")
}
```

Results in an error during `iocInit()`.  Further, attempting to Put "oops DB" at runtime will error as well.

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -1104,9 +1104,9 @@ static long dbPutFieldLink(DBADDR *paddr,
     if (link_info.ltype == PV_LINK &&
         (link_info.modifiers & (pvlOptCA | pvlOptCP | pvlOptCPP)) == 0) {
         chan = dbChannelCreate(link_info.target);
-        if (chan && dbChannelOpen(chan) != 0) {
-            errlogPrintf("ERROR: dbPutFieldLink %s.%s=%s: dbChannelOpen() failed\n",
-                precord->name, pfldDes->name, link_info.target);
+        if (chan && (status = dbChannelOpen(chan)) != 0) {
+            errlogPrintf(ERL_ERROR ": dbPutFieldLink %s.%s=%s: dbChannelOpen() failed w/ 0x%lx\n",
+                precord->name, pfldDes->name, link_info.target, status);
             goto cleanup;
         }
     }
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -24,6 +24,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#define EPICS_PRIVATE_API
+
 #include "alarm.h"
 #include "cantProceed.h"
 #include "cvtFast.h"
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -1111,6 +1111,13 @@ static long dbPutFieldLink(DBADDR *paddr,
                 precord->name, pfldDes->name, link_info.target, status);
             goto cleanup;
         }
+        if (!chan && (link_info.modifiers & pvlOptDB)) {
+            errlogPrintf("%s.%s " ERL_ERROR ": Unable to create DB link to \"%s\"\n",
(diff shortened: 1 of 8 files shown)


bfrk:
I haven't looked at the code but I like the idea. I think it should be an error though, not just a warning.

anjohnson:
I like this idea, but have minor concerns about the implementation. A link that has both `CA` and `DB` flags is obviously broken, but the `DB` flag silently overrides the `CA` one. I hadn't realized before that we don't actually report such conflicts in the other flags either. I'm not objecting to the priority order used, but it would be nice if the conflicts got reported to the user at load/parse time (this isn't new though, 3.15.9 and 3.14.12.8 behave just the same). We don't check the flags against what is legal for the direction of the link either (IN/OUT/FWD), but since subroutine records have been known to use their INPx links for output doing that could break existing databases.

Merging this PR should be a prerequisite for adding a `TPP` flag which would tell DB output links to use the target field's `pp()` attribute to decide whether to process the target record or not.

My answers to the questions posed above:
1. Yes, I think this is worth doing.
2. We should put adding support for this on the VDCT to-do list, but it isn't a blocker.
3. Your warning message in `dbInitLink()` looks reasonable, except use `dbLinkFieldName()` to show the field name of the link with the error. I would expect to see a similar message in `dbAddLink()`, those routines usually have very similar functionality.
4. Was the `dbPutFieldLink()` question really about `dbAddLink()`? If not please clarify.

mdavidsaver:
> A link that has both `CA` and `DB` flags is obviously broken ...
> ... we don't actually report such conflicts in the other flags either ...
> ... it would be nice if the conflicts got reported to the user at load/parse time ...

I agree about warning during `iocInit()`.  However, not afterwards as remote client triggered "warning" `errlogPrintf()` s are an invitation for log spam since a client won't know that anything is amiss.  As this aspect isn't a regression, and isn't as simple as adding a print, I'm not planning to address it here.

>     4. Was the `dbPutFieldLink()` question really about `dbAddLink()`?

Yes

mdavidsaver:
Updated.

1. Signal an error if the `DB` modifier can't be satisfied.  This errors `iocInit()` early (like a .dbd validation or `asInit()` failure).  `dbPutFieldLink()` errors early to reject w/o modification at runtime.
2. In dbLink.h make `dbInitLink()`/`dbAddLink()` private.  Not strictly required as only the return type changes, but I don't think these details should ever have appeared in a public header.

mdavidsaver:
> I agree about warning during `iocInit()` ...

Then again, maybe inconsistent modifiers should be an error.

anjohnson:
Group 12/15/21 approved subject to adding documentation. A load-time test would be nice but not essential.

mdavidsaver:
> ... A link that has both `CA` and `DB` flags is obviously broken ...

Except maybe it shouldn't be...  It occurs to me that this "DB" flag could be applied to local CA links with little change.  The logic moves up a few lines in `dbInitLink()` and `dbPutFieldLink()`, and errors if the flag is set and `dbChannelTest(pvname)` returns non-zero.

Obviously "DB" wouldn't be the best name anymore.  Maybe "LOC"?

This would also be somewhat less efficient as both `dbChannelTest()` and then `dbChannelCreate()` would be called each time.  Probably not a significant difference though.

anjohnson:
It occurred to me in the shower this morning that we might want to think a little more big-picture about this.

Should we have a flag `NCA` to match the other flags which have the similar `N` prefix to negate them? The semantics of that should be obvious to anyone who understands the `CA` flag, the negation means this link must not be resolved using CA.

The work you have done to add the `DB` flag isn't wasted though, it's still reasonable and I suggest adding a negated version of that, `NDB` which would mean this must not become a DB link.

Why? Well that leaves the possibility of adding an API for other "simple" link types in the future that don't require JSON syntax. A link such as `nowhere NDB NCA` would never resolve today, but might be able to some day.

ralphlange:
Well, `NDB NCA` would resolve over pvAccess, wouldn't it?
Happy Holidays!

mdavidsaver:
Dropping in favor of #222, which is imo. different enough to be a new PR.

PR #113, opened 2021-02-22
errlog rewrite w/ color removal

mdavidsaver:
Following up the digression on https://github.com/epics-base/epics-base/pull/109#issuecomment-781816959.  There are two parts to this PI.  A re-write of errlog around a double buffer, and teaching errlog to strip out ANSI terminal CSI escape sequences before passing along to log handlers.

The "WIP: add some color" commit is intended as a demo, and should not be merged.  Starting softIoc on Linux should show a blue `iocInit` when `$TERM` is set, and not when it is cleared.  In either case log clients should see plain text without escapes.

The buffering of log entries is conceptually similar, with the additional of a second buffer to allow errlogThread to print without holding a mutex.  This might be a benefit by itself wrt. PI and latency.  I was able to rewrite eg. `errlogPrintf()` to simply delegate to `errlogVprintf()` to avoid duplication, although this savings in LoC is offset by adding `errlogStripANSI()` (which seems in need of improvement).

The only change I needed to make to the errlog unittest was to allow for an additional message in one case.  I think this is due to the added capacity of double buffering.

I suspect that the logic in `isATTY()` which is testing `$TERM` may need to be excluded on RTEMS/vxWorks, but I haven't checked.  As it stands, escapes will be striped unless `$TERM` is set to a non-empty string.

The main omission I can see is that `printf()`, `epicsPrintf()`, and friends aren't covered by this PR.  I couldn't see a straightforward and portable way to do this.

I also haven't investigated the the support for ANSI escapes in Windows 10.

### DIFF
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -12,6 +12,11 @@
  *      Date:            07JAN1998
  */
 
+#ifdef _WIN32
+#  define VC_EXTRALEAN
+#  include <windows.h>
+#endif
+
 #include <stdlib.h>
 #include <stddef.h>
 #include <stdio.h>
@@ -205,11 +210,11 @@ void errlogSequence(void)
         epicsEventMustTrigger(pvt.waitForSeq);
 }
 
+#if !defined(_WIN32)
 static
 int isATTY(FILE* fp)
 {
     int ret = 0;
-#if !defined(_WIN32)
     const char* term = getenv("TERM");
     int fd = fileno(fp);
 
@@ -222,9 +227,33 @@ int isATTY(FILE* fp)
     /* only attempt to use ANSI escapes if some terminal type is specified */
     if(ret && (!term || !term[0]))
         ret = 0;
-#endif
     return ret;
 }
+
+#else /* _WIN32 */
+static
+int isATTY(FILE* fp)
+{
+    HANDLE hand = NULL;
+    DWORD mode = 0;
+    if(fp==stdout)
+        hand = GetStdHandle(STD_OUTPUT_HANDLE);
+    else if(fp==stderr)
+        hand = GetStdHandle(STD_ERROR_HANDLE);
+#ifdef ENABLE_VIRTUAL_TERMINAL_PROCESSING
+    if(hand && GetConsoleMode(hand, &mode)) {
+        (void)SetConsoleMode(hand, mode|ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+        mode = 0u;
+        if(GetConsoleMode(hand, &mode) && (mode&ENABLE_VIRTUAL_TERMINAL_PROCESSING))
+            return 1;
+    }
+#else
+    (void)hand;
+    (void)mode;
+#endif
+    return 0;
+}
+#endif
 
 /* in-place removal of ANSI terminal escape sequences.
  * exported for use by unit-test only
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -353,6 +353,7 @@ int eltc(int yesno)
     epicsMutexMustLock(pvt.msgQueueLock);
     pvt.toConsole = yesno;
     epicsMutexUnlock(pvt.msgQueueLock);
+    errlogFlush();
     return 0;
 }
 
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -33,6 +33,7 @@
 #include "errlog.h"
 #include "epicsStdio.h"
 #include "epicsExit.h"
+#include "osiUnistd.h"
 
 
 #define MIN_BUFFER_SIZE 1280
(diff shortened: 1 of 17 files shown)


mdavidsaver:
image

mdavidsaver:
Weird.  Removing the `errlogFlush()` from `eltc()` causes `dbCaLinkTest()` to hang on RTEMS.  Clearly there is a synchronization problem with this test which is being hidden by the flush.  I can't recall ever seeing sporadic failures on Linux or Windows though.  I've re-added the flush until this can be investigated further.

`eltc()` no longer needs to flush since the `toConsole` flag is passed through the queue with each message.

FreddieAkeroyd:
Windows 10 does support ANSI escapes in some terminals, @dirk-zimoch has added coloured output to Stream Device and some recent changes to make this on/off configurable https://github.com/paulscherrerinstitute/StreamDevice/pull/64 will point you at the lines needed to check for and enable it on Windows 10

dirk-zimoch:
How to enable ANSI colors in Windows: https://github.com/paulscherrerinstitute/StreamDevice/blob/ea8873becd656c33aa230731d3890d57cd28ace1/src/StreamError.cc#L52

mdavidsaver:
Updated.  I've made an untested attempt at enabling windows 10 terminal escapes.  I've also done a search/replace to colorize "error" and "warning" in many internal error messages.  I picked red and magenta since these have good contrast on both dark and light backgrounds.  (recent gcc does the same)

This update introduces a bunch of `ANSI_*` macros to `errlog.h` for specific colors.  I'm not certain if this is worthwhile, so these could easily be omitted.  The internal usage is `ERL_ERROR` and `ERL_WARNING` for these two specific usage contexts, and so could accommodate other formatting (__bold__, _italic_, etc.).

```c
// equivalent
errlogPrintf(ERL_ERROR ": something is amiss\n");
errlogPrintf(ANSI_RED("ERROR") ": something is amiss\n");
errlogPrintf(ANSI_ESC_RED "ERROR" ANSI_ESC_RESET ": something is amiss\n");
errlogPrintf("\033[31;1mERROR\033[0m: something is amiss\n");
```

dark
light

mdavidsaver:
As discussed during the last Core meeting, I'm planning to merge this change now that `R7.0.6.1` has been tagged.

This also changes the `ring buffer full` message to explicitly say "error".

https://github.com/epics-base/epics-base/blob/78b41f5bbf63381afbbb63ed69fe6a4d1481d839/modules/database/src/ioc/db/callback.c#L88-L90

FreddieAkeroyd:
I ran on my windows desktop, seemed to work as planned (coloured output on screen, not coloured to file/procserv)

mdavidsaver:
Merged as of b9899213d4e61de8cc28e0006bf5196e605f676e

modules/libcom/test/epicsErrlogTest.c
```diff
@@ -370,7 +403,8 @@ MAIN(epicsErrlogTest)
 
     testDiag("Logged %u messages", pvt.count);
     epicsEventMustWait(pvt.done);
-    testEqInt(pvt.count, N+1);
+    /* Expect N+1 messages +- 1 depending on impl */
+    testOk(pvt.count >= N && pvt.count<=N+2, "Logged %u messages, expected %zu", pvt.count, N+1);
```
dirk-zimoch, line 407: Another %z. I suggest to use unsigned long for counters.

modules/libcom/src/error/errlog.c
```diff
@@ -524,172 +597,97 @@ int errlogInit2(int bufsize, int maxMsgSize)
...
+
+            while(pos < print->pos) {
+                listenerNode *plistenerNode;
+                char* base = print->base + pos;
+                size_t mlen = epicsStrnLen(base+1u, pvt.bufSize - pos);
+                int stripped = 0;
+
+                if((base[0]&ERL_STATE_MASK) != ERL_STATE_READY || mlen>=pvt.bufSize - pos) {
+                    fprintf(stderr, "Logic Error: errlog buffer corruption. %02x, %zu\n",
+                            (unsigned)base[0], mlen);
```
dirk-zimoch, line 652: VxWorks does not support %z, not even in version 6.9.
I suggest to use unsigned int for mlen, pos, bufsize, maxMsgSize. On all supported architectures, unsigned int is large enough for MAX_MESSAGE_SIZE and BUFFER_SIZE.

modules/libcom/src/error/errlog.c
```diff
@@ -524,172 +597,97 @@ int errlogInit2(int bufsize, int maxMsgSize)
...
-        }
+            memset(print->base, 0, pvt.bufSize);
+            print->pos = 0u;
 
-        if (pvtData.atExit)
-            break;
-        if (epicsEventTryWait(pvtData.flush) != epicsEventWaitOK)
-            continue;
+            if(nLost && console)
+                fprintf(console, "errlog: lost %zu messages\n", nLost);
```
dirk-zimoch, line 683: %z again. Is suggest to use unsigned long for nLost.

PR #114, opened 2021-02-25
Post array events against the field pointer, not the array pointer

### LINKED ISSUE #97: Segfault when reading from compress record into aai record
Commit 542353aedbf521e822f06d04d737f94dffc80088 seems to have introduced a regression and segfault when reading from `compress` records into an `aai` record. Test db:
```
record(compress, "compress")
{
  field(ALG,  "Circular Buffer")
  field(INP,  "jaai PP")
  field(NSAM, "5")
  field(PINI, "YES")
}

record(aai, "caai")
{
  field(INP,  "42")
  field(NELM, "5")
  field(FTVL, "DOUBLE")
}

record(aai, "jaai")
{
  field(INP,  {const:[1,2,3,4,5]})
  field(NELM, "5")
  field(FTVL, "DOUBLE")
}

record(aai, "aai")
{
  field(INP,  "compress")
  field(NELM, "5")
  field(FTVL, "DOUBLE")
}

record(ao, "kill")
{
  field(FLNK, "aai")
}
```
`dbpf kill 1` results in a segfault.

Changing `if (dbLinkIsConstant(plink))` to `if (plink->type == CONSTANT || plink->type == JSON_LINK)` and moving `dbInitLink(plink, DBF_INLINK);` inside the `if` seems to fix the issue and correctly initializes the array:
```
diff --git a/modules/database/src/std/dev/devAaiSoft.c b/modules/database/src/std/dev/devAaiSoft.c
index 55975ba..99b46a7 100644
--- a/modules/database/src/std/dev/devAaiSoft.c
+++ b/modules/database/src/std/dev/devAaiSoft.c
@@ -47,13 +47,13 @@ static long init_record(dbCommon *pcommon)
     aaiRecord *prec = (aaiRecord *)pcommon;
     DBLINK *plink = &prec->inp;
 
-    /* This is pass 0, link hasn't been initialized yet */
-    dbInitLink(plink, DBF_INLINK);
-
-    if (dbLinkIsConstant(plink)) {
+    if (plink->type == CONSTANT || plink->type == JSON_LINK) {
         long nRequest = prec->nelm;
         long status;
 
+        /* This is pass 0, link hasn't been initialized yet */
+        dbInitLink(plink, DBF_INLINK);
+
         /* Allocate a buffer, record support hasn't done that yet */
         if (!prec->bptr) {
             prec->bptr = callocMustSucceed(nRequest, dbValueSize(prec->ftvl),
```
I don't know if it causes other problems though...

krisztianloki:
This fixes github issue #97:
Reading into an aai record from a compress or histogram or subArray record
could cause a segfault if the aai record was initialized before the
linked one.

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -18,6 +18,21 @@ should also be read to understand what has changed since earlier releases.
 <!-- Insert new items immediately below here ... -->
 
 
+### Fix aai's Device Support Initialization
+
+Krisztian Loki [reported](https://github.com/epics-base/epics-base/issues/97)
+segfaults occurring when a Soft Channel aai record INP field was a DB link to
+an array field of a compress record. This was caused by the aai record's
+pass-0 device support initialization clashing with the semantics of the new
+link support API.
+
+The aai record
+[has been modified](https://github.com/epics-base/epics-base/pull/114) to
+allow the Soft Channel device support to request a pass-1 initialization
+callback. See the Device Support section of the Array Analogue Input Record
+Reference pages in this release for the API changes, which are fully backwards
+compatible for existing aai device support.
+
 
 ### Priority inversion safe posix mutexes
 
--- a/modules/database/src/std/dev/devAaiSoft.c
+++ b/modules/database/src/std/dev/devAaiSoft.c
@@ -47,18 +47,14 @@ static long init_record(dbCommon *pcommon)
     aaiRecord *prec = (aaiRecord *)pcommon;
     DBLINK *plink = &prec->inp;
 
-    /* This is pass 0, link hasn't been initialized yet */
-    dbInitLink(plink, DBF_INLINK);
+    /* Ask record to call us in pass 1 instead */
+    if (prec->pact != 2) {
+        return 2;
+    }
 
     if (dbLinkIsConstant(plink)) {
         long nRequest = prec->nelm;
         long status;
-
-        /* Allocate a buffer, record support hasn't done that yet */
-        if (!prec->bptr) {
-            prec->bptr = callocMustSucceed(nRequest, dbValueSize(prec->ftvl),
-                "devAaiSoft: buffer calloc failed");
-        }
 
         status = dbLoadLinkArray(plink, prec->ftvl, prec->bptr, &nRequest);
         if (!status) {
--- a/modules/database/src/std/dev/devAaiSoft.c
+++ b/modules/database/src/std/dev/devAaiSoft.c
@@ -48,8 +48,8 @@ static long init_record(dbCommon *pcommon)
     DBLINK *plink = &prec->inp;
 
     /* Ask record to call us in pass 1 instead */
-    if (prec->pact != 2) {
-        return 2;
+    if (prec->pact != AAI_DEVINIT_PASS1) {
+        return AAI_DEVINIT_PASS1;
     }
 
     if (dbLinkIsConstant(plink)) {
--- a/modules/database/src/std/rec/aaiRecord.c
+++ b/modules/database/src/std/rec/aaiRecord.c
@@ -112,16 +112,18 @@ static long init_record(struct dbCommon *pcommon, int pass)
             prec->ftvl = DBF_UCHAR;
         prec->nord = (prec->nelm == 1);
 
-        /* we must call pdset->init_record in pass 0
-           because it may set prec->bptr which must
-           not change after links are established before pass 1
-        */
-
+        /* call pdset->init_record() in pass 0 so it can do its own
+         * memory allocation and set prec->bptr, which must be set by
+         * the end of pass 0.
+         */
         if (pdset->common.init_record) {
             long status = pdset->common.init_record(pcommon);
 
-            /* init_record may set the bptr to point to the data */
-            if (status)
+            if (status == 2) {
+                /* requesting pass 1 callback, remember to do that */
+                prec->pact = 2;
+            }
+            else if (status)
                 return status;
         }
         if (!prec->bptr) {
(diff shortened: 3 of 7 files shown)


mdavidsaver:
This all looks good.  (I see that I missed subArray yesterday)  I do notice that that there is a second `db_post_events()` in histogram which needs changing.

```
$ git grep db_post_events.*bptr
modules/database/src/std/rec/compressRecord.c:    db_post_events(prec, prec->bptr, monitor_mask);
modules/database/src/std/rec/histogramRecord.c:        db_post_events(prec, prec->bptr, DBE_VALUE | DBE_LOG);
modules/database/src/std/rec/histogramRecord.c:        db_post_events(prec, prec->bptr, monitor_mask);
modules/database/src/std/rec/subArrayRecord.c:    db_post_events(prec, prec->bptr, monitor_mask);
```

anjohnson:
Since the bug that this PR fixes could still exist in external array record types too, should we add a check and warning to the `dbEntryToAddr()` routine in in dbAccess.c if a record's `cvt_dbaddr()` method changes `paddr->pfield`? That's the only place this method is called and it can easily detect the cases we're concerned about. I don't think this change would break backwards-compatibility with older Base releases:
```patch
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -621,7 +621,13 @@ long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
 
         /* Let record type modify paddr */
         if (prset && prset->cvt_dbaddr) {
-            return prset->cvt_dbaddr(paddr);
+            long status = prset->cvt_dbaddr(paddr);
+            if (paddr->pfield != pdbentry->pfield) {
+                errlogPrintf("Warning: %s::cvt_dbaddr('%s') set paddr->pfield\n"
+                    "  See https://github.com/epics-base/epics-base/pull/114\n",
+                    pdbentry->precordType->name, paddr->precord->name);
+            }
+            return status;
         }
     }
     return 0;
```
The message points to this PR which should make it obvious what changes are needed in the record code and already points back to the problem that this fixes. Should I limit how many times it appears? I may adjust the PR's title and/or description slightly to make it more obvious though.

@krisztianloki I'm not asking you to do any more with this PR, I'm in the process of merging it (by hand) and this idea struck me while I was doing that.

mdavidsaver:
> ... should we add a check and warning to the `dbEntryToAddr()` routine ...

I wondered about this as well.  I was thinking about a error check of `status || pfield!=NULL` after `cvt_dbaddr()` to detect this specific condition.  A stricter warning seems likely to produce a lot of noise in an IOC with eg. an `sscanRecord`.  I think it would be reasonable if it could be limited to say, once per field per record type (maybe by hiding a flag in `dbFldDes`?).

anjohnson:
The warning is only shown when something (either a DB link or a network client) actually connects to that field, but I like the idea of a flag in dbFldDes to make it only display once, so I've added that (ignoring the fact that there are no locks protecting that structure so _in theory_ this could race if RSRV and QSRV searches collide on it).

Merging.

mdavidsaver:
> Merging.

Maybe not yet?  This might be worth a courtesy warning to the synapps gang, since they will likely be the first to see the warnings, and receive the complaints.

anjohnson:
The potential for a segfault when an aai.INP points to any such field dates back several releases so I want to put the warning detection code in, but I will enable it only if `getenv("EPICS_PFIELD_WARNING")` exists. Initially I thought the synapps modules looked like a lot of work, but only 12 of the 25 record types actually have a `cvt_dbaddr()` method. We missed 4 of the built-in record types too – bptr isn't the only field used to hold array pointers...

Checking and fixing all the synapps record types that have that method would be an excellent codeathon task:
- allenBradley/allenBradleyApp/1771DCMSrc/abDcmRecord.c
- allenBradley/allenBradleyApp/1771IFESrc/ab1771IFERecord.c
- allenBradley/allenBradleyApp/1771IXSrc/ab1771IXRecord.c
- allenBradley/allenBradleyApp/1771NSeriesSrc/ab1771NRecord.c
- asyn/asyn/asynRecord/asynRecord.c
- calc/calcApp/src/aCalcoutRecord.c
- calc/calcApp/src/sCalcoutRecord.c
- camac/camacApp/src/camacRecord.c
- mca/mcaApp/mcaSrc/mcaRecord.c
- optics/opticsApp/src/tableRecord.c
- sscan/sscanApp/src/sscanRecord.c
- vme/vmeApp/src/vmeRecord.c

anjohnson:
Change of plan. There's a major problem with requiring the pfield change to fields that hold long strings (lsi, lso, printf). The way that these work is that they present as a DBF_STRING but with a field size matching the buffer length, and the `$` field modifier knows that it can convert them into an array of that length. The `cvt_dbaddr()` method thus _has_ to modify the `paddr->pfield` pointer to point to the buffer in case the the field is accessed as a `DBF_STRING`, in which case there is no subsequent call to `get_array_info()`.

The changes in this PR are still fine so I will probably merge them along with some slightly more complicated equivalent changes to the aSub record type, although I need sleep first.

We should try to get the synapps record types similarly fixed at the codeathon, but the check and warning aspects just aren't possible without doing more work, so later if at all.

anjohnson:
I'm not happy with asking all of the external record types to be changed just because the aai record has some strange semantics. I have a solution that doesn't require that, although the changes in this PR are still sensible and I will include them.

The Soft Channel support's `init_record()` routine can now advise the record (by returning the status value 2) that it needs to be called again in pass 1. The record will make temporary use of the PACT field to remember that it needs to reinitialize the device support. The updated Soft Channel device support can handle its constant initialization duties in pass 1 without demanding special behavior from the other record types. Existing device support will continue to work as before since it has always returned 0, -1, or an error status value – 2 would have been treated as an error.

I have added a Release Note, and updated the record reference documentation in the .dbd.pod file to describe the new initialization semantics. I am pushing my changes to this branch now, and plan to merge them tomorrow if there are no major objections.

mdavidsaver:
I'm not thrilled with another magic `return 2;`, but agree that this is preferable to trying to force a change to the external record types.  As far as I can see, this wouldn't break existing device support using HW link types (I've written some).

> The record will make temporary use of the PACT field to remember that it needs to reinitialize the device support

No objection.  fyi. This kind of extra/private state could instead be added as a bit field in `struct dbCommonPvt`.

anjohnson:
> a bit field in `struct dbCommonPvt`

Since in this case it's specific to the aai record type, it would make more sense to add another field to the record type if I couldn't use PACT for it.

I have replaced the magic `2` with a macro `AAI_DEVINIT_PASS1` defined in the aaiRecord.h file.

Thanks.

PR #120, opened 2021-03-08
Substitute strncpy by memcpy to avoid gcc 8 warnings

gabrielfedel:
This MR is related to this issue: https://bugs.launchpad.net/epics-base/+bug/1905159
These changes remove the strncpy warnings on gcc 8.
I choose to apply the PR on the 3.15 because these changes could be propagated to the 7.0

It looks safer to use memcpy instead of disable this kind of warnings

### DIFF
--- a/src/ca/legacy/pcas/generic/casStrmClient.cc
+++ b/src/ca/legacy/pcas/generic/casStrmClient.cc
@@ -1487,7 +1487,7 @@ caStatus casStrmClient::hostNameAction ( epicsGuard < casClientMutex > & guard )
         }
         return S_cas_internal;
     }
-    strncpy ( pMalloc, pName, size - 1 );
+    memcpy ( pMalloc, pName, (size - 1)*sizeof(char));
     pMalloc[ size - 1 ]='\0';
 
     if ( this->pHostName ) {
@@ -1531,7 +1531,7 @@ caStatus casStrmClient::clientNameAction (
         }
         return S_cas_internal;
     }
-    strncpy ( pMalloc, pName, size - 1 );
+    memcpy ( pMalloc, pName, (size - 1)*sizeof(char));
     pMalloc[size-1]='\0';
 
     if ( this->pUserName ) {
--- a/src/ioc/dbStatic/dbStaticLib.c
+++ b/src/ioc/dbStatic/dbStaticLib.c
@@ -742,7 +742,7 @@ static long dbAddOnePath (DBBASE *pdbbase, const char *path, unsigned length)
 
     pdbPathNode = (dbPathNode *)dbCalloc(1, sizeof(dbPathNode));
     pdbPathNode->directory = (char *)dbCalloc(length+1, sizeof(char));
-    strncpy(pdbPathNode->directory, path, length);
+    memcpy(pdbPathNode->directory, path, length*sizeof(char));
     pdbPathNode->directory[length] = '\0';
     ellAdd(ppathList, &pdbPathNode->node);
     return 0;
--- a/src/libCom/error/errSymLib.c
+++ b/src/libCom/error/errSymLib.c
@@ -162,7 +162,7 @@ static void errRawCopy ( long statusToDecode, char *pBuf, unsigned bufLength )
                 status = sprintf ( pBuf, "%d", errnum );
             }
             else {
-                strncpy ( pBuf,"<err copy fail>", bufLength );
+                memcpy(pBuf, "<err copy fail>", bufLength*sizeof(char) );
                 pBuf[bufLength-1] = '\0';
                 status = 0;
             }
@@ -181,8 +181,8 @@ static void errRawCopy ( long statusToDecode, char *pBuf, unsigned bufLength )
                     "(%d,%d)", modnum, errnum );
             }
             else {
-                strncpy ( pBuf, 
-                    "<err copy fail>", bufLength);
+                memcpy ( pBuf, 
+                    "<err copy fail>", bufLength*sizeof(char));
                 pBuf[bufLength-1] = '\0';
                 status = 0;
             }
--- a/src/libCom/osi/epicsTime.cpp
+++ b/src/libCom/osi/epicsTime.cpp
@@ -658,7 +658,7 @@ size_t epicsTime::strftime (
                     if ( tmpLen >= bufLenLeft ) {
                         tmpLen = bufLenLeft - 1;
                     }
-                    strncpy ( pBufCur, pOVF, tmpLen );
+                    memcpy(pBufCur, pOVF, tmpLen*sizeof(char));
                     pBufCur[tmpLen] = '\0';
                     pBufCur += tmpLen;
                     bufLenLeft -= tmpLen;
@@ -670,7 +670,7 @@ size_t epicsTime::strftime (
                 if ( tmpLen >= bufLenLeft ) {
                     tmpLen = bufLenLeft - 1;
                 }
-                strncpy ( pBufCur, pDoesntFit, tmpLen );
+                memcpy(pBufCur, pDoesntFit, tmpLen*sizeof(char));
                 pBufCur[tmpLen] = '\0';
                 pBufCur += tmpLen;
                 bufLenLeft -= tmpLen;
--- a/src/libCom/osi/osiSock.c
+++ b/src/libCom/osi/osiSock.c
@@ -74,7 +74,7 @@ unsigned epicsShareAPI sockAddrToA (
             return len;
         }
         else {
-		    strncpy ( pBuf, "<Ukn Addr Type>", bufSize-1 );
+		    memcpy ( pBuf, "<Ukn Addr Type>", (bufSize-1)*sizeof(char));
 		    pBuf[bufSize-1] = '\0';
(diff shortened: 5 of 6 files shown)


gabrielfedel:
If this approach is ok I will include a new PR for base 7.0 for files not present on 3.15

tboegi (commented):
Hm,
The commit message says "it looks safer".
What exactly does this mean ?
Actually, it is unsafer (to use memcpy() here.
Lets say your destination has 500.000 bytes.
The source has 22 bytes, including the '\0'
After the 22 bytes the memory page is empty, there is no memory.
What happens ?
memcpy() will try to copy 500.000 bytes, as we asked it to do.
After some bytes, it will try to access memory, which is not there ->
All OS which have a memory protection will dump.

(And, as a side note: Why do we multiply with sizeof(char) ?
 It is not in the commit message, and it is not needed either)

gabrielfedel:
@tboegi  I see, you are right. 
I've took a look and it looks safer, but actually there is a better solution to "remove" the warnings without change the strncpy, just locally. So I will follow this approach.

I've added the sizeof(char) because I thought was the safer way to define the number of bytes to be copied.

Anyway I will close this PR, and try another approach for remove the warnings.

ralphlange:
What I have seen in a few places (libraries mostly) is that they created their own, safe, thread-safe implementation of `strncpy()`.
Often based on one of the publicly available solutions.

tboegi:
strncpy() is pretty much standard, isn't it ?
What we -could- use is a strlcpy() (or make our own epicsStrlcpy(), if that is better ?
BSD has it, and Linux
https://linux.die.net/man/3/strlcpy
But not MacOs and the rest of the gang

ralphlange:
Windows has banned the use of `strncpy()` for some years, creating lots of colorful warnings in compilers and static code analysis.

FreddieAkeroyd:
What about `strcpy_s` ? https://en.cppreference.com/w/c/string/byte/strcpy

gabrielfedel:
I've read about the problem with strncpy here: https://the-flat-trantor-society.blogspot.com/2012/03/no-strncpy-is-not-safer-strcpy.html

Anyway, the first reason of the PR was because of this kind of message from gcc 8:

```
../../../src/libCom/error/errSymLib.c:165:17: warning: ‘strncpy’ output truncated copying between 1 and 5 bytes from a string of length 15 [-Wstringop-truncation]                                                                                             
                 strncpy ( pBuf,"<err copy fail>", bufLength );                                                                                                                                                                                                
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                 

```

But there is a solution to "avoid" this warnings, that look like this:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-truncation"
    strncpy ( pBuf,"<err copy fail>", bufLength );                
#pragma GCC diagnostic pop
```

This was my second idea to remove that warnings. 
Any comments are welcome :)

src/ca/legacy/pcas/generic/casStrmClient.cc
```diff
@@ -1487,7 +1487,7 @@ caStatus casStrmClient::hostNameAction ( epicsGuard < casClientMutex > & guard )
         }
         return S_cas_internal;
     }
-    strncpy ( pMalloc, pName, size - 1 );
+    memcpy ( pMalloc, pName, (size - 1)*sizeof(char));
     pMalloc[ size - 1 ]='\0';
```
mdavidsaver, line 1491: What warning is coming from this instance?  This looks like as "safe" a string duplication as I can imagine.

```c++
size = strlen(pName)+1u;
pMalloc = new char [size];
if(!pMalloc) { ... }
strncpy ( pMalloc, pName, size - 1 );
pMalloc[ size - 1 ]='\0';
```

The bug I see here is expecting c++ `new[]` to return `NULL` when it will actually throw an exception.

mdavidsaver: imo. plain `strcpy()` here would be appropriate since local code has just allocated the destination buffer.

FreddieAkeroyd:
strncpy  of `"<err copy fail>"` will only get executed if bufLen <=5 so is the solution to shorten the error message as it will never fit anyway?

src/ca/legacy/pcas/generic/casStrmClient.cc
gabrielfedel: This warning

```
../../../../../../src/ca/legacy/pcas/generic/casStrmClient.cc:1490:13: warning: ‘char* strncpy(char*, const char*, size_t)’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]                  
     strncpy ( pMalloc, pName, size - 1 );                                                                                                                                                                                                                     
     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                      

```

I can remove this kind of warning using this:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-truncation"
    strncpy ( pMalloc, pName, size - 1 );
#pragma GCC diagnostic pop
```

As this is not really an issue

mdavidsaver: I don't like disabling whole classes of warnings now and "forever".  Better to replace with plain `strcpy()` since there is no danger of truncation or overflow in this instance.

FreddieAkeroyd: Would a non gcc compiler generate an "unknown pragma" warning?

gabrielfedel:
This might solve for this case specifically but there are other similar messages in places where this solution cannot be applied (I think), for example:

```
../../../src/libCom/osi/epicsTime.cpp: In function ‘size_t epicsTime::strftime(char*, size_t, const char*) const’:                                                                                                                                             
../../../src/libCom/osi/epicsTime.cpp:661:29: warning: ‘char* strncpy(char*, const char*, size_t)’ output truncated copying between 1 and 3 bytes from a string of length 3 [-Wstringop-truncation]                                                            
                     strncpy ( pBufCur, pOVF, tmpLen );                                                                                                                                                                                                        
                     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                         

```


I think this kind of warning looks more an information than an issue, so we can include the pragma code to ignore locally if the code is doing what is intended to do.

src/ca/legacy/pcas/generic/casStrmClient.cc
gabrielfedel: @mdavidsaver , but this would be applied only for one command specific, not for all the code.  But if you still think is better to replace by strcpy I can do that.

gabrielfedel: @FreddieAkeroyd , not sure, I will check.

FreddieAkeroyd:
Does changing `size_t tmpLen = sizeof ( pOVF ) - 1;` to `size_t tmpLen = sizeof ( pOVF ) ;` on line 657 help at all?

src/ca/legacy/pcas/generic/casStrmClient.cc
mdavidsaver: Also, I think that GCC is specifically complaining that both of the following lines _could_ write to `pMalloc[size-1]`.

```c
strncpy ( pMalloc, pName, size - 1 );
pMalloc[ size - 1 ]='\0';
```

So the warning would also go away by changing the `strncpy()` to `size-2`, although that would also require handing the `size==1` case specially.

mdavidsaver:
> ... will only get executed if bufLen <=5 ...

This statement doesn't seem quite correct, at least with [glibc](https://github.com/bminor/glibc/blob/60167dc24b09b25c6ca3353f9ce38c41340a792f/string/strncpy.c#L26-L33).  `strncpy()` will happily copy partial strings.

src/ca/legacy/pcas/generic/casStrmClient.cc
gabrielfedel: You are right, it is possible to solve this warning this way.

mdavidsaver: > ... would be applied only for one command specific, not for all the code.

If you can find a one-size-fits-all solution for `char*` operations in C, then you will be famous!  I think these warnings need to be considered case by case.

FreddieAkeroyd:
The compiler is looking at all the if tests and realises that the branch requesting a copy of `"<err copy fail>"` will only get executed when `bufLen <= 5` so strncpy() will work ok, but it will always have to truncate the string in this case, which is what it is warning about.

src/ca/legacy/pcas/generic/casStrmClient.cc
gabrielfedel: Yes, you are right. I will start this again, looking case by case.

mdavidsaver:
fyi. `sizeof(char)` is, and will always be, `1`.  So multiplying sizes by `sizeof(char)` is redundant.

PR #122, opened 2021-03-08
Adds Doxygen annotations to epicsType.h

DominicOram:
As per the codeathon project detailed in https://github.com/epics-base/epics-base/wiki/Header-annotations-project.

I modified the pre-existing comments and added a few more where it was obvious to do so.

### DIFF
--- a/modules/libcom/src/misc/epicsTypes.h
+++ b/modules/libcom/src/misc/epicsTypes.h
@@ -56,6 +56,11 @@ typedef double          epicsFloat64;
 typedef epicsInt32      epicsStatus;
  /** @} */
 
+#define MAX_STRING_SIZE 40
+
+/**
+ * \brief !! Dont use this - it may vanish in the future !!
+ */
 typedef struct {
     unsigned    length;
     char        *pString;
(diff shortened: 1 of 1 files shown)


modules/libcom/src/misc/epicsTypes.h
```diff
@@ -49,15 +54,15 @@ typedef epicsUInt16     epicsEnum16;
...
 
 typedef struct {
     unsigned    length;
     char        *pString;
 } epicsString;
 
-/*
- * !! Dont use this - it may vanish in the future !!
+/**
+ * \brief !! Dont use this - it may vanish in the future !!
```
mdavidsaver, line 70: @anjohnson Is this warning still appropriate?  Use of `MAX_STRING_SIZE` remains common.

anjohnson: I would like to see a "Don't use this" comment on the `epicsOldString` and `epicsString` typedefs, but not on the `MAX_STRING_SIZE` macro which is far too widely used to be deprecated.

mdavidsaver: Agreed.

DominicOram: I'm happy to change this tomorrow

mdavidsaver (approved):
Looks ok to me.

mdavidsaver:
Applied as 1fcbdad5e991af5b8c7a4f6996c9a301e849bfc1.  Thanks.

PR #123, opened 2021-03-08
First approach for the macro PWD inside the iocShell

marciodo:
This is the first approach. The function that creates the macro was called 3 times with the IOC example from makeBaseApp.pl:

1. Right in the beginning of the boot.
2. After the call to envPaths.
3. After IOC Init.

So, the macro is created during the boot, but it doesn't update when the user uses the cd IOC shell function. I thought about calling the function on libcomRegister.c, but my function addPwdMacro() doesn't seem to fit the scope of the IOC Shell API. I could duplicate the code on libcomRegister.c, too.

Any suggestion is appreciated.

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -19,6 +19,7 @@
 #include <stdio.h>
 #include <ctype.h>
 #include <errno.h>
+#include <unistd.h>
 
 #include "epicsMath.h"
 #include "errlog.h"
@@ -545,6 +546,16 @@ int iocshSetError(int err)
     return err;
 }
 
+void addPwdMacro(iocshContext *context) {
+    if (context != NULL) {
+        char buf[256];
+        char *pwd = getcwd ( buf, sizeof(buf) - 1 );
+        if ( pwd ) {
+            macPutValue(context->handle, "PWD", pwd);
+        }
+    }
+}
+
 /*
  * The body of the command interpreter
  */
@@ -579,7 +590,6 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
     int ret = 0;
 
     iocshInit();
-
     /*
      * See if command interpreter is interactive
      */
@@ -652,6 +662,8 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
             return -1;
         }
 
+        addPwdMacro(context);
+
         epicsThreadPrivateSet(iocshContextId, (void *) context);
     }
     MAC_HANDLE *handle = context->handle;
@@ -661,6 +673,7 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
 
     macPushScope(handle);
     macInstallMacros(handle, defines);
+    //addPwdMacro();
 
     wasOkToBlock = epicsThreadIsOkToBlock();
     epicsThreadSetOkToBlock(1);
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -546,7 +546,7 @@ int iocshSetError(int err)
     return err;
 }
 
-void addPwdMacro(iocshContext *context) {
+static void addPwdMacro(iocshContext *context) {
     if (context != NULL) {
         char buf[1024];
         char *pwd = getcwd(buf, sizeof(buf) - 1);
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -19,7 +19,7 @@
 #include <stdio.h>
 #include <ctype.h>
 #include <errno.h>
-#include <unistd.h>
+#include "osiUnistd.h"
 
 #include "epicsMath.h"
 #include "errlog.h"
@@ -548,10 +548,14 @@ int iocshSetError(int err)
 
 void addPwdMacro(iocshContext *context) {
     if (context != NULL) {
-        char buf[256];
-        char *pwd = getcwd ( buf, sizeof(buf) - 1 );
-        if ( pwd ) {
+        char buf[1024];
+        char *pwd = getcwd(buf, sizeof(buf) - 1);
+        if (pwd) {
             macPutValue(context->handle, "PWD", pwd);
+        } else {
+            if (errno == ERANGE) {
+                perror("Current path is too long");
+            }
         }
     }
 }
@@ -590,6 +594,7 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
     int ret = 0;
 
     iocshInit();
+
     /*
      * See if command interpreter is interactive
      */
@@ -673,7 +678,6 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
 
     macPushScope(handle);
     macInstallMacros(handle, defines);
-    //addPwdMacro();
 
     wasOkToBlock = epicsThreadIsOkToBlock();
     epicsThreadSetOkToBlock(1);
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -19,7 +19,6 @@
 #include <stdio.h>
 #include <ctype.h>
 #include <errno.h>
-#include "osiUnistd.h"
 
 #include "epicsMath.h"
 #include "errlog.h"
@@ -546,20 +545,6 @@ int iocshSetError(int err)
     return err;
 }
 
-static void addPwdMacro(iocshContext *context) {
-    if (context != NULL) {
-        char buf[1024];
-        char *pwd = getcwd(buf, sizeof(buf) - 1);
-        if (pwd) {
-            macPutValue(context->handle, "PWD", pwd);
-        } else {
-            if (errno == ERANGE) {
-                perror("Current path is too long");
-            }
-        }
-    }
-}
-
 /*
  * The body of the command interpreter
  */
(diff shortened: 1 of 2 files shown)


modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -19,6 +19,7 @@
 #include <stdio.h>
 #include <ctype.h>
 #include <errno.h>
+#include <unistd.h>
```
anjohnson: Try osiUnistd.h instead of this, it has replacements for most routines when building on Windows. I think this is why your build failed on Windows.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -545,6 +546,16 @@ int iocshSetError(int err)
     return err;
 }
 
+void addPwdMacro(iocshContext *context) {
+    if (context != NULL) {
+        char buf[256];
```
anjohnson: 256 may be too small, but sys/param.h isn't a portable header; can you make this 1024 instead? That should be enough for most sites. It should probably inform the user if it gets an ERANGE error though.

anjohnson (commented):
A couple of suggestions.

modules/libcom/src/iocsh/iocsh.cpp
mdavidsaver: @anjohnson FYI #102 would add `epicsPathAllocCWD()` to avoid this sort of situation.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -545,6 +546,20 @@ int iocshSetError(int err)
...
 
+void addPwdMacro(iocshContext *context) {
+    if (context != NULL) {
+        char buf[1024];
+        char *pwd = getcwd(buf, sizeof(buf) - 1);
+        if (pwd) {
+            macPutValue(context->handle, "PWD", pwd);
+        } else {
+            if (errno == ERANGE) {
+                perror("Current path is too long");
```
mdavidsaver: With `perror()` there is no need to test `errno` explicitly.  A simple `perror("getcwd()");` would do.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -545,6 +546,20 @@ int iocshSetError(int err)
     return err;
 }
 
+void addPwdMacro(iocshContext *context) {
+    if (context != NULL) {
```
mdavidsaver: fyi. `context` can never be null where this function is (or should) be called.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -545,6 +546,20 @@ int iocshSetError(int err)
     return err;
 }
 
+void addPwdMacro(iocshContext *context) {
```
mdavidsaver: Internal functions like this should be marked `static`.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -652,6 +667,8 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
             return -1;
         }
 
+        addPwdMacro(context);
+
         epicsThreadPrivateSet(iocshContextId, (void *) context);
     }
```
mdavidsaver: This `if(!context)` condition above is part of handling recursive calls.  eg. when an iocsh script sources another iocsh script.  `!context` is only true for the outermost call.  Since the inner scripts may call `cd` I think the call to `addPwdMacro()` should be moved after the condition.

mdavidsaver (changes requested):

modules/libcom/src/iocsh/iocsh.cpp
marciodo: I can't assume that this function will ever be called from this single point in code in the future. Don't you think it is better to prevent now?

modules/libcom/src/iocsh/iocsh.cpp
marciodo: I tested it here but didn't like the result. Take a look:
- When the path is smaller than 1024, perror will print "<path name>: Success" every time addPwdMacro() is called. On my test IOC that happens 3 times.
- For big paths, it will print "Numerical result out of range" in the same 3 moments.

Don't you think a more descriptive text is better?

mdavidsaver: >     * On my test IOC that happens 3 times.

Maybe an example will help.  I'm thinking of the follow, where `getcwd()` is only called when `getcwd()` indicates an error.

```c++;
        if (char *pwd = getcwd(buf, sizeof(buf) - 1)) {
            macPutValue(context->handle, "PWD", pwd);
        } else {
            perror("getcwd()");
        }
```

> Don't you think a more descriptive text is better?

```c
errno = ERANGE;
perror("Current path is too long");
```

For me prints `Current path is too long: Numerical result out of range`, which seems redundant.  `man 3 getcwd` also lists eg. `EINVAL` as a possible error condition.  This prints `Current path is too long: Invalid argument`, which seems contradictory.

The argument of `perror()` is meant to describe the operation which has failed.  `perror()` then adds information about why this operation has failed.

modules/libcom/src/iocsh/iocsh.cpp
mdavidsaver: No code is fully safe from future regressions.  Personally, I don't think I would have bothered.  Or at most `assert(context);`. Having a function `addPwdMacro()` which can silently not add the PWD macro is also an avenue for future surprise.

marciodo: Ok, I'll study better how this context structure works. With your two comments about it, I think I'm missing an important point.

modules/libcom/src/iocsh/iocsh.cpp
marciodo: Yes, looks like perror is not adequate at all to show these messages.

modules/libcom/src/iocsh/iocsh.cpp
mdavidsaver: We can talk though this if you like.

marciodo: Thank you. I usually need a moment alone with the code to figure things out. But I'll reach to you if I see it is taking too long.

modules/libcom/src/iocsh/iocsh.cpp
marciodo: @anjohnson and @mdavidsaver, I'm going through macLib and I've found this in macLib.h:

`/** \brief Maximum size of a macro name or value
 */
#define MAC_SIZE 256`

Although it seems that you can register a macro with any amount of characters, this constant is used for the macro expansion.

mdavidsaver: Ah, yes.  I had forgotten about that limit.  And it isn't so straightforward to eliminate, or even extend as it is used to size some local array variables (on the stack).

I guess use of macLib needs to be reconsidered...

marciodo: Are you thinking about hard-code ${PWD} or $(PWD) as a "special command" inside iocshBody instead of using macLib?

mdavidsaver: I'm thinking about having `cd` (`chdirCallFunc()` in libComRegister.c) call `epicsSetEnv()`, or alternately `putenv()`.  Not as good of a solution as it wouldn't capture direct calls to `chdir()`.

mdavidsaver: ... but is necessary since iocsh should be reentrant and changes to process environment aren't necessarily.

mdavidsaver: #137 would also come into play.

The macLib limit also effects my #102.

mdavidsaver: `${PWD}` isn't the only path I'd like to expose via macLib, so I'm not inclined to make a special case of it.

marciodo: I've just pushed the code with the epicsEnvSet idea. It works regarding that the user issues a cd on the shell. Can you think about a boot-up trigger to make sure PWD will always exist?

mdavidsaver:
> ... Can you think about a boot-up trigger to make sure PWD will always exist?

My first thought would be in libComRegister.c where the "cd" command is being registered.

marciodo:
@mdavidsaver and @anjohnson, I've just pushed a new solution with the result from the recent discussion.

modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -396,13 +398,39 @@ static void installLastResortEventProviderCallFunc(const iocshArgBuf *args)
 
 static iocshVarDef asCheckClientIPDef[] = { { "asCheckClientIP", iocshArgInt, 0 }, { NULL, iocshArgInt, NULL } };
 
+// Register the PWD environment variable when the cd IOC shell function is 
+// registered. This variable contains the current directory path.
+static void registerPwd() {
+    // On Linux, the IOC already boots with the PWD env variable initialized.
+    // We check if it exists thinking about other OSs.
+    char* currentPwd = getenv("PWD");
+    if (currentPwd == NULL) {
```
mdavidsaver: I think this test should be removed.  `$PWD` should always be in sync with `getcwd()`, so there is no harm in enforcing this.  As written this code path will almost never run on Linux, which is a recipe for rare bugs.

mdavidsaver (changes requested):
The approach looks ok.  Though I think #140 needs to be merged first, which hopefully won't take too long.

modules/libcom/src/iocsh/libComRegister.c
marciodo: Ok, I removed the test.

mdavidsaver:
Applied as 614961da170ae7094cf55c27ed3183723c07fc90 with some changes.  Thanks!

PR #124, opened 2021-03-09
[WIP] Remove strncpy warnings that are exclusively from branch 7.0

gabrielfedel:
These changes removes warnings from strncpy on gcc 8. 
These warnings appears in files presents only on branch 7.0

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -834,7 +834,8 @@ static long getLinkValue(DBADDR *paddr, short dbrType,
     {
         const char *rtnString = dbGetString(&dbEntry);
 
-        strncpy(pbuf, rtnString, maxlen-1);
+        if (maxlen > 1)
+            strncpy(pbuf, rtnString, maxlen-1);
         pbuf[maxlen-1] = 0;
         if(dbrType!=DBR_STRING)
             nReq = strlen(pbuf)+1;


modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -736,14 +736,17 @@ static long dbAddOnePath (DBBASE *pdbbase, const char *path, unsigned length)
...
 
     pdbPathNode = (dbPathNode *)dbCalloc(1, sizeof(dbPathNode));
     pdbPathNode->directory = (char *)dbCalloc(length+1, sizeof(char));
-    strncpy(pdbPathNode->directory, path, length);
-    pdbPathNode->directory[length] = '\0';
+    sz = sizeof(pdbPathNode->directory);
+    if (sz > 0)
+        strncpy(pdbPathNode->directory, path, sz);
+    pdbPathNode->directory[sz] = '\0';
     ellAdd(ppathList, &pdbPathNode->node);
```
FreddieAkeroyd: I think you need `length` rather than `sz`  in these lines - I believe `sz` will always be `sizeof(char*)`

gabrielfedel: You are right about the use of sizeof. I thought it will get the size of allocated memory for pdbPathNode->directory ...
Anyway, if I use length gcc continues with the warning. I will try another solution.

bfrk: Indeed.
What is the purpose of the test? Does strncpy do something bad when the size parameter is 0?
Also, since we used calloc, there is no need to set `pdbPathNode->directory[length] = '\0';`, so this line could be removed.
What is the exact warning you get?

bfrk (commented):
We have
```
    if(!nReq)
        return 0;
```
a few lines up so the test looks redundant, but I guess the compiler is too stupid to figure that out.
However, if you want to guard against that case, you should include the next line, too!

bfrk:
General remark: since this is tricky (easy to make off-by-one errors with fatal consequences), it might be helpful to add a short comment. Something like
```
/* test is redundant because of (!nReq) guard above; fixes compiler warning with <insert compiler & version> */
if (...) ...
```

bfrk:
I was wrong. Sorry. The `nReq` is of type long, so could be negative. That means the test is not redundant. I would try to change the guard `(!nReq)` to `(nReq <= 0)` and see if that makes the warning go away.

tboegi:
@gabrielfedel @bfrk 

if  (nReq <= 0) may better written as
if (nReq < 1) to give space to at least the '\0'

bfrk:
> if (nReq < 1) to give space to at least the '\0'

We are talking about the change in dbAccess.c, right? The guard at line 811 affects all array requests. It is correct to return 0 when the caller requests an array with 0 elements. The caller may mistakenly request a negative number of elements, which should be an error:
```
    if(!nReq)
        return 0;
    if(nReq<0)
        return some_errorcode;
```
The strncpy and setting the terminator afterwards are now guarded against maxlen <= 0. If the caller requested 1 element, what they will get is a properly terminated string of length 0, precisely what they requested. There should be no warning. If there is, because the compiler doesn't see that, add the redundant condition at that point together with a comment.

bfrk:
BTW, "give space" is not describing the situation correctly. The caller has allocated the buffer. We must assume it has space for at least *nRequest elements. And we already subtract 1 from the request size for the length of the string to account for the terminating null byte.

modules/database/src/ioc/dbStatic/dbStaticLib.c
gabrielfedel: Are you sure about not be necessary the last '\0' ?

The warning I'm having, on the original code, is:

```
In function ‘dbAddOnePath’,                                                                                                                                         
    inlined from ‘dbAddPath.part.7’ at ../dbStatic/dbStaticLib.c:730:16,                                                                                            
    inlined from ‘dbAddPath’ at ../dbStatic/dbStaticLib.c:663:6:                                                                                                    
../dbStatic/dbStaticLib.c:745:5: warning: ‘strncpy’ output truncated before terminating nul copying 1 byte from a string of the same length [-Wstringop-truncation] 
     strncpy(pdbPathNode->directory, path, length);                                                                                                                 
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~      
```

gabrielfedel:
@bfrk , I didn't understand your suggestion for the dbAccess.c . Do you think I can keep the redundant verification I've added but including a comment to explain it?

modules/database/src/ioc/dbStatic/dbStaticLib.c
gabrielfedel: I've already removed change from dbStaticLib.c, until I discover a better way to avoid this warning

gabrielfedel:
@bfrk actually, if I include a verification for nreq < 0, the strncpy warning changes to : 

`../db/dbAccess.c:842:9: warning: ‘strncpy’ destination unchanged after copying no bytes [-Wstringop-truncation]     `                                                                                                                                           
                                                                                         `
So my understand is that  it is not redundant the test, but it is to avoid the warning where no bytes are copied.

bfrk:
It is logically redundant according to C semantics if you have previously made sure that maxlen>0. The compiler warning (the new one you get after making sure that maxlen > 0) is stupid. There are lots of procedures that have no effect when called with certain values for their parameters. Forcing users to add arbitrary redundant branching in their code is bad for quality. Sigh.

So yes, add the test but also add a comment that says the check is redundant and only there to make stupid warning go away.

FreddieAkeroyd:
Is there an (undocumented)  `-Wno-stupid` compiler option we can use?

bfrk:
I think the goal is to annoy us with these warnings until we give up and don't use strncpy anymore ;-)

tboegi:
The original code looks a little bit weird:

 strncpy(pdbPathNode->directory, path, length);
 pdbPathNode->directory[length] = '\0';
 
 A (may be) slightly better way would be
 strncpy(pdbPathNode->directory, path, length-1);
 pdbPathNode->directory[length] = '\0';
 
The thing is, that gcc tries to be helpful (and sometimes fails).
Different versions of gcc have evolved differently.
It may be even, but that is speculation, the strncpy() is "obsolete" under Linux.
The new strlcpy does what we need.
We could try to define a epicsStrlcpy() on our own, and use that one.
Inside epicsStrlcpy() we can try to make compilers happy, and there is even one
place to fight.
Opinions ?

bfrk:
+1 strlcpy is the right thing

gabrielfedel:
> It is logically redundant according to C semantics if you have previously made sure that maxlen>0. The compiler warning (the new one you get after making sure that maxlen > 0) is stupid. There are lots of procedures that have no effect when called with certain values for their parameters. Forcing users to add arbitrary redundant branching in their code is bad for quality. Sigh.
> 
> So yes, add the test but also add a comment that says the check is redundant and only there to make stupid warning go away.

Sorry, I think I was not completely clear. The test I included was  like this (same as on the PR):

```
          if (maxlen > 1)                        
              strncpy(pbuf, rtnString, maxlen-1);
```
And this is needed for the case where maxlen == 1, otherwise gcc will emits a warning because of the copy of 0 bytes.
So I think it is not a redundant.

mdavidsaver:
I'm somewhat reluctant to go with defining our out `str*()` functions.  This would certainly quiet these static analysis warnings by circumventing whatever hooks GCC is using.  This seems undesirable as the long term trend is for the SNR of these warnings to improve.

gabrielfedel:
@tboegi 
I think that changes the code for this

```
 strncpy(pdbPathNode->directory, path, length-1);
pdbPathNode->directory[length] = '\0';
```

will not do the same thing...as it will copy length-1 bytes , instead of length bytes....Or am I missing something?

mdavidsaver:
> The compiler warning (the new one you get after making sure that maxlen > 0) is stupid.

Quoting from https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82944

> The -Wstringop-truncation warning added in GCC 8.0 via r254630 for bug 81117 is specifically intended to highlight likely unintended uses of the strncpy function that truncate the terminating NUL charcter from the source string.

tboegi:
@mdavidsaver : are we talking about the same thing here ?
One change could be to replace all the 
 strncpy(pSrc, pDst, length-1);
 pDst[length] = '\0';
 hand-made-crafted code with one line:
 strlcpy(pSrc, pDst, length);
 
This would make the written code shorter - and easier to read.
Is this a good idea ?
(The problem is, that many, but not all platforms/libraries-delivered-with-this-platform
have strlcpy()
Linux has it not ? MacOs has strlcpy(), how about vxWorks, Windows, RTEMS ?

mdavidsaver:
> ... We could try to define a epicsStrlcpy() on our own, and use that one. ...

I should have quoted.  I'm objecting to this approach specifically.

gabrielfedel:
> > The compiler warning (the new one you get after making sure that maxlen > 0) is stupid.
> 
> Quoting from https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82944
> 
> > The -Wstringop-truncation warning added in GCC 8.0 via r254630 for bug 81117 is specifically intended to highlight likely unintended uses of the strncpy function that truncate the terminating NUL charcter from the source string.

Interesting... I could solve the warning on dbStaticLib.c with this change:

```
-    strncpy(pdbPathNode->directory, path, length);   
+    strncpy(pdbPathNode->directory, path, length+1); 
```

It looks safe for me, considering pdbPathNode->directory is allocated with length + 1... 

Any issue with this approach?

mdavidsaver:
hum..  I should have checked this earlier.  Actually, GCC 8.3.0 (w/ debian 10) isn't complaining about this one for me.  Let me do a full rebuild and I'll attach a list of warnings.

mdavidsaver:
GCC 8.3.0 on amd64 (w/ Debian 10) actually gives me a surprisingly short [list of warnings](https://github.com/epics-base/epics-base/files/6109937/log880.txt).  The flags used are eg.

> /usr/bin/gcc  -D_GNU_SOURCE -D_DEFAULT_SOURCE         -DUSE_TYPED_RSET -DUSE_TYPED_DSET   -D_X86_64_  -DUNIX  -Dlinux -Wdate-time -D_FORTIFY_SOURCE=2   -DBUILDING_dbCore_API   -O3 -g  -pipe -Wall -Werror-implicit-function-declaration     -Werror=implicit-function-declaration -fvisibility=hidden  -mtune=generic      -m64 -fstack-protector-strong -Wformat -Werror=format-security -fPIC -I. -I../O.Common -I. -I. -I.. -I../as -I../bpt -I../db -I../dbStatic -I../dbtemplate -I../misc -I../registry -I../rsrv -I../../../../../include/compiler/gcc -I../../../../../include/os/Linux -I../../../../../include         -c ../dbStatic/dbStaticLib.c

I see `-Wall`, but am I missing some other flags?

mdavidsaver:
> These changes removes warnings from strncpy on gcc 8.

@gabrielfedel Which specific 8.x are you testing with?  It seems that 8.1 fixed a couple of issues with the then new `-Wstringop-truncation`.  cf. [84468](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84468) and [84228](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84228).

mdavidsaver:
> Linux has it not ? MacOs has strlcpy(), how about vxWorks, Windows, RTEMS ?

Right, I remember reading about [this](https://lwn.net/Articles/507319/) some time ago.  It seems that the [situation](https://sourceware.org/glibc/wiki/FAQ#Why_no_strlcpy_.2F_strlcat.3F) wrt. glibc hasn't changed significantly since 2012.  So `strl*()` seem destined to remain *BSD extensions.

mdavidsaver:
I've added a CI build with gcc 9 (currently 9.3.0).  https://github.com/epics-base/epics-base/runs/2069013700?check_suite_focus=true

Edit: [Selected warnings](https://github.com/epics-base/epics-base/files/6110672/log930.txt)

bfrk:
> > It is logically redundant according to C semantics if you have previously made sure that maxlen>0. The compiler warning (the new one you get after making sure that maxlen > 0) is stupid. There are lots of procedures that have no effect when called with certain values for their parameters. Forcing users to add arbitrary redundant branching in their code is bad for quality. Sigh.
> > So yes, add the test but also add a comment that says the check is redundant and only there to make stupid warning go away.
> 
> Sorry, I think I was not completely clear. The test I included was like this (same as on the PR):
> 
> ```
>           if (maxlen > 1)                        
>               strncpy(pbuf, rtnString, maxlen-1);
> ```
> 
> And this is needed for the case where maxlen == 1, otherwise gcc will emits a warning because of the copy of 0 bytes.
> So I think it is not a redundant.

It depends what you care about. I care about the code being correct. You care about it being warning free.

> The -Wstringop-truncation warning added in GCC 8.0 via r254630 for bug 81117 is specifically intended to highlight likely unintended uses of the strncpy function that truncate the terminating NUL charcter from the source string.

If the compiler assumes that the source string is null-terminated, then of course this warning will flag all uses of strncpy where the size parameter is zero. This *is* stupid, sorry. It is (or should be) well known that strncpy does *not* guarantee termination. The man page is very clear on that. Correct code will *always* overwrite the buffer with a null byte somewhere after calling strncpy. Incorrect code that does not do so will fail because the input string is longer than the destination. In general this cannot be detected statically. These warnings, issued in the extremely few cases where it can be statically detected, don't take these precautions, or their absence, into account. They are not helpful at all because they only serve to give a false sense of security ("my code is warning free, I am great!") and entice inexperienced programmers to add contortions like unnecessary branching into their code without understanding what they do.

I really no longer understand the GNU folks. Instead of solving the problem at the root, by deprecating the dangerous strncpy and replacing it with strlcpy, they invent these ridiculous warnings. And the completely ridiculous rationale is that strlcpy is not "standard". So every year we get more compiler warnings for "standard" library functions that are unsafe by design. I guess this is what nowadays is called "innovation".

gabrielfedel:
@mdavidsaver I'm running gcc 8.3.0 also (on Debian 10), but it looks the output is different from yours (weird):
https://gist.github.com/gabrielfedel/bf4fcf8c5d160ba94902cafb3f3fb09d

Anyway, I've compiled with gcc 10.2.0:
https://gist.github.com/gabrielfedel/4e06dc9851300076c2930819cfc18031

And I think the result is the same from gcc 9.3.0 you listed.
So I think I will focus on that ones. I might close this PR, as the warnings it try to solve doesn't appear on the newest gcc

gabrielfedel:
@mdavidsaver I'm running gcc 8.3.0 also (on Debian 10), but it looks the output is different from yours (weird):
https://gist.github.com/gabrielfedel/bf4fcf8c5d160ba94902cafb3f3fb09d

I've compiled with gcc 10.2.0 also and got similar output (different from yours with gcc 9)
https://gist.github.com/gabrielfedel/4e06dc9851300076c2930819cfc18031

Am I missing some step on the compilation? 
To get the list of warnings I'm just running:

make  2> warning_outputs

PR #125, opened 2021-03-09
#80 - Use OS strtod if MSVC 1900 or above

### LINKED ISSUE #80: Windows target: vs2015 and later have working strtod()
The `epicsParseDouble()` routine calls `epicsStrtod()` which we introduced because the `strtod()` implementation on some OSs doesn't meet standards. This is implemented in libCom in an osi/os/ header osdStrtod.h, and our own implementation in misc/epicsStdlib.c which is defined if the header requests it. The `epicsStdlibTest` program checks the OS version of `strtod()` if we are currently using our own, so it's easy to tell when we can switch back.

The WIN32 version of osdStrtod.h has selected our own implementation for a while, but recent versions have been fixed:

+ [vs2019 - pass](https://ci.appveyor.com/api/buildjobs/7m6ik8urd3lxs43j/artifacts/epicsStdlibTest.tap)
+ [vs2017 - pass](https://ci.appveyor.com/api/buildjobs/a2g63mqunp49lama/artifacts/epicsStdlibTest.tap)
+ [vs2015 - pass](https://ci.appveyor.com/api/buildjobs/m7j3h581dyyi8dqc/artifacts/epicsStdlibTest.tap)
+ [vs2013 - fail](https://ci.appveyor.com/api/buildjobs/02v8s9oac3sc3txc/artifacts/epicsStdlibTest.tap)
+ [vs2012 - fail](https://ci.appveyor.com/api/buildjobs/tjaisgt6c16v4bmo/artifacts/epicsStdlibTest.tap)
+ [vs2010 - fail](https://ci.appveyor.com/api/buildjobs/amnwqc1f0lnhbsth/artifacts/epicsStdlibTest.tap)
+ [mingw - pass](https://ci.appveyor.com/api/buildjobs/u8cjire3w0c6eear/artifacts/epicsStdlibTest.tap)

This issue is a reminder for someone to fix osi/os/WIN32/osdStrtod.h so we use the OS version when compiling with vs2015 or later. This fix should probably be done on the 3.15 branch.

rerpha:
Closes #80 

Hopefully i've done this correctly! 

Should now use the OS strtod implementation if using MINGw or MSVC 1900 (2015) or above and still use our own impl if not.

### DIFF
--- a/src/libCom/osi/os/WIN32/osdStrtod.h
+++ b/src/libCom/osi/os/WIN32/osdStrtod.h
@@ -13,10 +13,6 @@
 extern "C" {
 #endif
 
-/*
- * epicsStrtod() for systems with broken strtod() routine
- */
-epicsShareFunc double epicsStrtod(const char *str, char **endp); 
 
 /*
  * Microsoft apparently added strto[u]ll() in VS2013
@@ -26,6 +22,15 @@ epicsShareFunc double epicsStrtod(const char *str, char **endp);
 #if !defined(_MINGW) && (_MSC_VER < 1800)
 #  define strtoll _strtoi64
 #  define strtoull _strtoui64
+#endif
+
+#if (_MSC_VER < 1800) && defined(_MINGW)
+/*
+ * epicsStrtod() for systems with broken strtod() routine
+ */
+epicsShareFunc double epicsStrtod(const char *str, char **endp);
+#else
+#  define epicsStrtod strtod
 #endif
 
 #ifdef __cplusplus
--- a/src/libCom/osi/os/WIN32/osdStrtod.h
+++ b/src/libCom/osi/os/WIN32/osdStrtod.h
@@ -24,7 +24,11 @@ extern "C" {
 #  define strtoull _strtoui64
 #endif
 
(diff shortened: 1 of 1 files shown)


mdavidsaver:
@anjohnson @FreddieAkeroyd Since the appveyor builds passed, I guess this is good to go?

anjohnson (approved):
Looks good to me. Should we do merges during the week or later? The only reason to delay might be the Appveyor builds, but since this is 3.15 it should be fine.

mdavidsaver:
> The only reason to delay might be the Appveyor builds

We're going to give appveyor a good workout this week!  We should probably cancel jobs for some of the doxygen only PRs.

FreddieAkeroyd:
The code looks fine, but if `epicsStrtod` is defined as `strtod` then all the strtod tests look to be skipped? Would it be safer to always run the tests in case a subsequent compiler change caused the native strtod to fail?

anjohnson:
The `strtod()` tests should just be duplicates of the main set of tests that call `epicsStrtod()`, so if MS breaks `strtod()` those tests would fail. I think the setup does what we need.

PR #126, opened 2021-03-09
Extend timestamp channel filter

exzombie:
Based on the discussions summarized in the Launchpad issue on [making the `TIME` field directly accessible](https://bugs.launchpad.net/epics-base/+bug/1182091), here is the implementation of the extended channel filter.

Usage:
- `pvname.{"ts":{}}` will get you the same behaviour as before: a new timestamp will be generated to replace existing one.
- `pvname.{"ts":{"num": "<type>"}}` will give you a numeric timestamp. `<type>` can be:
  - `"sec"` for the number of seconds past epoch (`DBF_ULONG`)
  - `"nsec"` for the nanoseconds (`DBF_ULONG`)
  - `"dbl"` for `seconds.fraction` format (`DBF_DOUBLE`)
  - `"ts"` for a "timestamp-like" format, a two elements array of seconds and nanoseconds, respectively (`DBF_ULONG`)
- `pvname.{"ts":{"num": "<type>", "epoch": "unix"}}` will change the epoch
- `pvname.{"ts":{"str": "<type>"}}` will return a `DBF_STRING`. Type can be `"epics"` for a `caget`-like timestamp, or `iso`.

There is one open issue with `pvget` that I describe in a [comment](https://github.com/epics-base/epics-base/pull/126#issuecomment-795972206), but this was later confirmed to [affect other filters as well](https://bugs.launchpad.net/epics-base/+bug/1918909).

Connecting to the `TIME` field will be part of a separate merge request.

### DIFF
--- a/modules/database/src/std/filters/ts.c
+++ b/modules/database/src/std/filters/ts.c
@@ -1,4 +1,5 @@
 /*************************************************************************\
+* Copyright (c) 2021 Cosylab d.d
 * Copyright (c) 2010 Brookhaven National Laboratory.
 * Copyright (c) 2010 Helmholtz-Zentrum Berlin
 *     fuer Materialien und Energie GmbH.
@@ -9,6 +10,7 @@
 
 /*
  *  Author: Ralph Lange <Ralph.Lange@bessy.de>
+ *  Author: Jure Varlec <jure.varlec@cosylab.com>
  */
 
 #include <stdio.h>
@@ -16,20 +18,122 @@
 #include <string.h>
 
 #include "chfPlugin.h"
-#include "db_field_log.h"
 #include "dbExtractArray.h"
 #include "dbLock.h"
+#include "db_field_log.h"
 #include "epicsExport.h"
+#include "freeList.h"
+#include "cantProceed.h"
 
-/*
- * The size of the data is different for each channel, and can even
- * change at runtime, so a freeList doesn't make much sense here.
- */
+/* Allocation size for freelists */
+#define ALLOC_NUM_ELEMENTS 32
+
+
+/* Filter settings */
+
+enum tsMode {
+    tsModeInvalid = 0,
+    tsModeGenerate = 1,
+    tsModeDouble = 2,
+    tsModeSec = 3,
+    tsModeNsec = 4,
+    tsModeArray = 5,
+    tsModeString = 6,
+};
+
+static const chfPluginEnumType ts_numeric_enum[] = {
+    {"dbl", 2}, {"sec", 3}, {"nsec", 4}, {"ts", 5}};
+
+enum tsEpoch {
+    tsEpochEpics = 0,
+    tsEpochUnix = 1,
+};
+
+static const chfPluginEnumType ts_epoch_enum[] = {{"epics", 0}, {"unix", 1}};
+
+enum tsString {
+    tsStringInvalid = 0,
+    tsStringEpics = 1,
+    tsStringIso = 2,
+};
+
+static const chfPluginEnumType ts_string_enum[] = {{"epics", 1}, {"iso", 2}};
+
+typedef struct tsPrivate {
+    enum tsMode mode;
+    enum tsEpoch epoch;
+    enum tsString str;
+} tsPrivate;
+
+static const chfPluginArgDef ts_args[] = {
+    chfEnum(tsPrivate, mode, "num", 0, 0, ts_numeric_enum),
+    chfEnum(tsPrivate, epoch, "epoch", 0, 0, ts_epoch_enum),
+    chfEnum(tsPrivate, str, "str", 0, 0, ts_string_enum),
+    chfPluginArgEnd
+};
+
+static int parse_finished(void *pvt) {
+    tsPrivate *settings = (tsPrivate *)pvt;
+    if (settings->str != tsStringInvalid) {
+        settings->mode = tsModeString;
+#if defined _MSC_VER && _MSC_VER <= 1700
+        // VS 2012 crashes in ISO mode, doesn't support timezones
+        if (settings->str == tsStringIso) {
+            return -1;
+        }
+#endif
+    } else if (settings->mode == tsModeInvalid) {
+        settings->mode = tsModeGenerate;
+    }
+    return 0;
+}
+
+
+/* Allocation of filter settings */
+static void *private_free_list;
+
+static void * allocPvt() {
+    return freeListCalloc(private_free_list);
+}
+
+static void freePvt(void *pvt) {
+    freeListFree(private_free_list, pvt);
+}
+
+
+/* Allocation of two-element arrays for second+nanosecond pairs */
+static void *ts_array_free_list;
+
+static void *allocTsArray() {
+    return freeListCalloc(ts_array_free_list);
+}
+
+static void freeTsArray(db_field_log *pfl) {
+    freeListFree(ts_array_free_list, pfl->u.r.field);
+}
+
+/* Allocation of strings */
+static void *string_free_list;
+
+static void *allocString() {
+    return freeListCalloc(string_free_list);
+}
+
+static void freeString(db_field_log *pfl) {
+    freeListFree(string_free_list, pfl->u.r.field);
+}
+
+
+/* The dtor for waveform data for the case when we have to copy it. */
 static void freeArray(db_field_log *pfl) {
+    /*
+     * The size of the data is different for each channel, and can even
+     * change at runtime, so a freeList doesn't make much sense here.
+     */
     free(pfl->u.r.field);
 }
 
-static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
+static db_field_log* generate(void* pvt, dbChannel *chan, db_field_log *pfl) {
     epicsTimeStamp now;
     epicsTimeGetCurrent(&now);
 
@@ -44,7 +148,7 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
             dbScanLock(dbChannelRecord(chan));
             dbChannelGetArrayInfo(chan, &pSource, &nSource, &offset);
             dbExtractArray(pSource, pTarget, pfl->field_size,
-                nSource, pfl->no_elements, offset, 1);
+                           nSource, pfl->no_elements, offset, 1);
             pfl->u.r.field = pTarget;
             pfl->dtor = freeArray;
             pfl->u.r.pvt = pvt;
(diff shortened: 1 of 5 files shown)


mdavidsaver:
>     * CA protocol and channel filters do not support unsigned integers, so I could only use doubles to avoid data loss.

fyi. PVA protocol does support unsigned integers, including 64-bit.  So a `DBF_UINT64` can be transported exactly.

mdavidsaver:
If would be good to have some test coverage of this filter.  eg. setting up a database with one record, force that .TIME to a known value, and then do some `testdbGetFieldEqual()` with the various options.

exzombie:
> > ```
> > * CA protocol and channel filters do not support unsigned integers, so I could only use doubles to avoid data loss.
> > ```
> 
> fyi. PVA protocol does support unsigned integers, including 64-bit. So a `DBF_UINT64` can be transported exactly.

Good to know! I'd love to see that supported. But how do the values get to the PVA server?

- In `db_field_log.h`, `union native_value` supports only a limited number of types. Can it be easily extended? The comment above the definition could be interpreted that way, but I have my doubts.
- How will the CA server handle an unsigned value? Will it convert it to a signed type, or a double?

mdavidsaver:
>     * `union native_value` supports only a limited number of types. Can it be easily extended?

I think it should be, though this might turn out to be a separate, and larger, project.

>     * How will the CA server handle an unsigned value?

DBF_UINT64 is implicitly converted to DBF_DOUBLE before CA transport.  fyi. You could experiment with eg `int64outRecord` if you are curious.

mdavidsaver:
> I think it should be, though this might turn out to be a separate, and larger, project.

... or maybe not.  It seems that at least now that #99 is merged, this union is already having epicsUInt64 values `memcpy()`'d to it.


Also, maybe worth pointing out that some of the existing tests are now failing:

```
   tsTest.tap .... 
  not ok  4 - register_pre registers one filter w/o argument
  not ok  7 - ts has one filter w/o argument in pre chain
```

exzombie:
> DBF_UINT64 is implicitly converted to DBF_DOUBLE before CA transport. fyi. You could experiment with eg `int64outRecord` if you are curious.

I was more concerned about UINT32; if it was converted to INT32, it would be troublesome. You don't want to lose a bit in seconds-since-epoch. But I just checked with mbbi which has fields of type ULONG, and they are converted to DOUBLE, so all is good.

> ... or maybe not. It seems that at least now that #99 is merged, this union is already having epicsUInt64 values memcpy()'d to it.

Yes, I see that that is the method of getting the data in there. And I also see that getting the data out goes through `void*` and the conversion tables based on the field type. Which means (to me at least) that `union native_value` only serves as convenience for writing filters. If you agree, I would extend it with all supported types, i.e. like the already existing `union epics_any`, but with optional string member.

Speaking of which, what is the purpose of `DB_EVENT_LOG_STRINGS`? As I see it, one cannot use strings this way at all because they're not there in a default build.

> Also, maybe worth pointing out that some of the existing tests are now failing:

Not much surprise there, tests are still TODO. I was a *bit* surprised because I aimed for backwards compatibility, but I see now that the failing tests check that the number of installed filters is one, which is not the case anymore.

exzombie:
I pushed a new version which now has a pretty complete suite of tests, can return integers and has two predefined string formats. Apart from documentation (which I will provide tomorrow), there are two outstanding issues.

-----------

The first problem is that this filter somehow doesn't work well with PVA. It works with `pvmonitor`, but not `pvget`. The latter sees the correct type of the variable, but the value comes from the `VAL` field instead of what the filter provides. For example:
```
record(calc, "test:counter") {
    field(SCAN, "1 second")
    field(CALC, "VAL+1")
    field(MDEL, -1)
}
```
Loading this with `softIocPVA` gives the following results:
```
$ caget test:counter
test:counter                   24
$ caget -f9 test:counter.'{"ts": {"num": "float"}}'
test:counter.{"ts": {"num": "float"}} 984252569.202834606
$ pvget test:counter.'{"ts": {"num": "float"}}'
test:counter.{"ts": {"num": "float"}} 2021-03-10 20:29:53.203  113
$ caget -f9 test:counter.NAME'{"ts": {"num": "float"}}'
test:counter.NAME{"ts": {"num": "float"}} 984252615.202881813
$ pvget test:counter.NAME'{"ts": {"num": "float"}}'
test:counter.NAME{"ts": {"num": "float"}} Error Disconnect
```
On the last command, the IOC console prints `an exception caught while in receiveThread at ../../src/remote/codec.cpp:1133: dbGet for meta fails`. I don't know what to make of this.

----------------------------

The other issue is with installing the channel filter to handle the `TIME` field. I have implemented this, but I haven't pushed yet because it causes some tests to fail. Specifically, `dbPutLinkTest` has two failures related to the `TSE` field. I have not investigated this yet; I'm told that `TSE` has special handling of `TIME` and if anyone can tell me that we can't hijack it, I'd like to know before I spend more time on this.

anjohnson:
> I'm told that `TSE` has special handling of `TIME` and if anyone can tell me that we can't hijack it, I'd like to know before I spend more time on this.

As per our video chat, the special TSEL link handling doesn't actually try to connect to the TIME field (which would fail anyway), it actually replaces the field name with VAL (or maybe it just removes it) when you point TSEL to a .TIME field before it attempts to make the connection. I don't think you need to worry about that, although there may be things in the tests that you have to adjust to avoid them failing.

dirk-zimoch:
> $ pvget test:counter.NAME'{"ts": {"num": "float"}}'
test:counter.NAME{"ts": {"num": "float"}} Error Disconnect

It seems the PVA server does not handle filters correctly. I filed a bug report: https://bugs.launchpad.net/epics-base/+bug/1918909

mdavidsaver:
>  seems the PVA server does not handle filters correctly.

Should now be fixed.  see https://github.com/epics-base/pva2pva/issues/41

Also, with epics-base/epics-base@ca3ef9c61e57a7413ad9626f17ce9205cdabdc5f I changed `testdbGet*()` to use `dbChannel*` instead of `DBADDR` internally, so they can now be used to test filters.  eg.

```c
testdbGetFieldEqual("blah.{}", DBF_LONG 42);
```

or

```c
    dbCommon *prec =  testdbRecordPtr("TEST");
    static epicsInt32 arr[] = {9, 8, 7, 6, 5, 4, 3, 2, 1};
    testdbGetArrFieldEqual("TEST", DBF_LONG, 10, 9, arr);
    testdbGetArrFieldEqual("TEST.{arr:{s:5}}", DBF_LONG, 10, 4, arr+5);
```

modules/database/src/ioc/db/dbChannel.c
```diff
@@ -462,15 +493,36 @@ dbChannel * dbChannelCreate(const char *name)
...
 
-    chan = freeListCalloc(dbChannelFreeList);
-    if (!chan)
-        goto finish;
     cname = malloc(strlen(name) + 1);
     if (!cname)
         goto finish;
 
-    strcpy(cname, name);
+    if (strcmp(dbEntry.pflddes->name, "TIME")) {
```
mdavidsaver: Did you explore re-writing the PV name string as it is copied?  eg. to match on strings ending with ".TIME" and append some `{"ts":...` before further parsing?  It seems like this would avoid much special case handling of "ts".

exzombie:
> > seems the PVA server does not handle filters correctly.
> 
> Should now be fixed. see [epics-base/pva2pva#41](https://github.com/epics-base/pva2pva/issues/41)

I can't confirm this, I'm probably missing something.
- I rebased on the latest `7.0` branch
- I checked out the latest `master` of the pva2pva submodule, `git log` confirms that the fix is there
- `make clean`, rebuild, try with `pvget`
- ... and there is no difference. Am I missing a step there?

modules/database/src/ioc/db/dbChannel.c
exzombie: Yes. I chose the current implementation for two reasons:
- I see no reason not to allow subscribing to `test:counter.TIME{"dec": {"n": 3}}`. Simply appending `{"ts": ...}` would break that.
- I like calling `pvNameLookup()` first because it ensures that `name` has the form I assume. I do not know what assumptions I can make about `name` by the time it reaches `dbChannelCreate()`.

exzombie:
My last attempt to fix vs2012 didn't work, either. Somehow, the tests stop early, and the only reason I can think of is that `epicsTimeToStrftime()` still crashes. I can't reproduce the environment to debug properly, and appveyor takes way too long for me to iterate blindly, so I don't see what else I can do.

Another problem is that vs2010 static-debug fails at `testChannelAccess` while vs2010 dynamic-debug works. Go figure.  Perhaps there's some other reason the tests sometimes get terminated early?

mdavidsaver:
> Somehow, the tests stop early

Looking at the log (`tsTest.tap` artifact):

```
1..111
...
ok 106 - ts has one filter with argument in post chain
ok 107 - ts has no filter in pre chain
```

Do your local runs show all 111 tests?


> `testChannelAccess`

This test is known to be flaky, and I don't think it is using any of the code changed by this PR.

anjohnson:
```diff
$ diff -u tsTest.tap-vs201[23]
--- tsTest.tap-vs2012	2021-03-25 14:54:09.000000000 -0500
+++ tsTest.tap-vs2013	2021-03-25 14:56:24.000000000 -0500
@@ -120,3 +120,7 @@
 ok 105 - dbChannel with plugin ts opened
 ok 106 - ts has one filter with argument in post chain
 ok 107 - ts has no filter in pre chain
+ok 108 - ts filter does not drop or replace field_log
+ok 109 - ts filter gives correct field type
+ok 110 - ts filter does not touch non-value fields of field_log
+ok 111 - ts filter gives correct field value
```

The failure is when running the tests on the channel `x.VAL{ts:{"str": "iso"}}`. Looking at the code I see this:

```C
case tsStringIso:
        fmt = "%Y-%m-%dT%H:%M:%S.%06f%z";
```
I'm going to guess that the `%z` is what's causing the problem, and that it's actually failing in the filter. Windows didn't used to properly support the timezone conversion specification in their `strftime()` implementation, so I'd suggest looking at their documentation for what might have been different between vs2012 and vs2013. I'm not sure how to fix this though.

FreddieAkeroyd:
From https://bugs.python.org/issue20010 it looks like %z used to return the full timezone name (like %Z) but on later compilers correctly returns the offset, returning the full timezone name could probably make the result longer than 40 chars.

modules/database/src/ioc/db/dbChannel.c
mdavidsaver: I am thinking the same thing wrt. other filters.  My main motivation though is code simplification.  To minimize the special case handling of the `ts` filter.

Since the channel name string is always copied, it seems reasonable to do so a little earlier (before calling `pvNameLookup()`), where some re-writing would be less invasive.  Something like the following:

```c
...
/* implied default filter */
#define TIMEfilt ".TIME{\"ts\":..."
size_t namelen = strlen(name);
if(namelen>=5 && strcmp(name+namelen-5u, ".TIME")==0) { /* aka. if name ends with ".TIME" */
    cname = malloc(namelen-5u+sizeof(TIMEfilt));
    memcpy(cname, name, namelen-5u);
    memcpy(cname+namelen-5u, TIMEfilt, sizeof(TIMEfilt));
} else {
    cname = malloc(namelen+namelen+1u);
    stcpy(cname, name);
}
pname = cname;

status = pvNameLookup(&dbEntry, &pname); /* ... handle as before ... */
...
```

exzombie: Sorry for taking this long to respond, I've been quite busy.

I followed your approach to see what the end result looks like, and I'm not entirely happy. Personally, I hate working with strings in C because it takes a *lot* of care to get even trivial things right, and reading such code is not my cup of tea. That's why I still prefer the previous approach where `pvNameLookup()` was reused to parse the string, the name was merely shortened and the string code was straightforwad. 

That's just my opinion FYI, of course; being the maintainer, it's up to you to decide which approach you find more maintainable. The change is in commit 7ba4b22.

I think applying additional filters on top of TIME should now be added to the tests. I didn't test that before because the filter specification in the PV name wasn't touched, but I feel the new implementation requires it. I'll wait for your decision on which implementation you prefer before spending time on additional tests.

exzombie:
@FreddieAkeroyd thanks for finding the root cause! It seems to me that this needs to be worked around in `epicsTime::strftime()`. I'm still puzzled as to why the test crashes: `strftime()` is supposed to return 0 if the string doesn't fit, and both `epicsTime::strftime()` and the ts filter truncate the string in that case.

Anyway, it seems that python has a solution. Only `time.strftime()` passes the string directly to the system-provided `strftime()` while `datetime.strftime()` handles `%z` (and some other specifiers) itself. Perhaps this could be ported to `epicsTime.strftime()`. I'll open an issue when I get around to it.

modules/database/src/ioc/db/dbChannel.c
mdavidsaver: In your latest change, it looks like you are trying to be more general than I was thinking by "merging" two JSON filter expressions. If I'm misunderstanding this code, please give an example of what a rewritten PV name should look like before and after.

> ... I see no reason not to allow subscribing to `test:counter.TIME{"dec": {"n": 3}}` ...

I am strongly inclined towards strictness in parsing.  It is much easier to later relax restrictions than it is to later impose them.

With this in mind, I don't see this generality as necessary at present.  It seems reasonable to me to assume that someone explicitly giving a filter for .TIME knows what this entails.  Further, an explicit filter spec may well include "TS" filter, in which case automatically prepending a second instance may be surprising or unhelpful.

> ... Personally, I hate working with strings in C ...

I absolutely agree, this is why for me short and simple are important criteria.

exzombie: > In your latest change, it looks like you are trying to be more general than I was thinking by "merging" two JSON filter expressions. If I'm misunderstanding this code, please give an example of what a rewritten PV name should look like before and after.

No, you are quite correct, merging them is the point. In the earlier implementation, this happened naturally, and I saw no reason to disallow it then. Now, the complexity of string manipulation is reason enough.

>With this in mind, I don't see this generality as necessary at present. It seems reasonable to me to assume that someone explicitly giving a filter for .TIME knows what this entails.

"knows what this entails" might be a bit too much to expect from a casual user. After this change, the TIME field will, from the outside, look like any other field, and it will not be obvious that it is not. Only a careful reader of the RRM will notice that it is supposed to be NOACCESS and wonder why it can still be accessed, and a note in the Release Notes is easily overlooked or forgotten. It is not obvious to the user that channel filters are involved and that there may be incompatibilities. I think there is merit in avoiding surprises.

Let me summarize the three approaches that are now on the table:
- the original approach that reuses `pvNameLookup()`
  - in the common case where the field is *not* TIME, the only additional step is a 4 byte comparison of `dbEntry.pflddes->name` with `"TIME"`
  - in the case where the field *is* TIME, there is an additional call to `pvNameLookup()` and minimal string manipulation which removes the `"TIME"` part of the string
  - the most complex part is the `installTimestampFilter()` function which is still straightforward
  - any filters specified after `"TIME"` are automatically preserved
- ad-hoc parsing of the PV name
  - even in the common case where the field is *not* TIME, there is a substring search
  - when the field *is* TIME, there is a change in memory allocation and a fair bit of string manipulation
  - ... which is the cost of preserving any filters specified after `"TIME"`
- simpler ad-hoc parsing of the PV name
  - even in the common case where the field is *not* TIME, there is a substring search
  - when the field *is* TIME, there is a change in memory allocation; the string is truncated at `"TIME"` and replaced with the "ts" filter specification
  - the cost of simplicity is disallowing additional filters

It still seems to me that reusing `pvNameLookup()` is the better approach.
- It's not much more complex than the simpler variant of ad-hoc parsing
- it preserves any JSON attached (I expect the code would work even if you extended the JSON beyond just channel filters) 
- in fact, it preserves *any* text after the field, so it doesn't make syntactically wrong PV names magically work like simple truncation would
- it is more performant because it doesn't do the additional substring search in the common case (which, admittedly, probably doesn't matter).

Just my two cents. I don't really mind going with the third approach if you still prefer it, I'm not that heavily invested in the first one :)

anjohnson:
Core group 7/28:
Make the VS2010 test failures be ToDo.
AI on ANJ to review filter code.
AI on MAD to review name handling again & comment.

exzombie:
I'm picking this up again, and while going through my notes, I found that we have initially discussed being able to access the timestamp also as `uint64`. @FreddieAkeroyd , @till-s  , you had a usecase for this, if I recall correctly. I don't remember why it's not implemented, perhaps there was a technical limitation, perhaps not. It would only be available via PVAccess. I'd like to know: is having `uint64` datatype genuinely useful, or is `uint32[2]` (returned by `{"ts": {"num": "ts"}}`) sufficient?

till-s:
IMHO it would simply be nice if
`caget xyz.TIME`
would yield something unsurprising. Other than that I don't have a specific use case. I just need this rarely enough (usually from some scripting environment like matlab or python) for me to have forgotten that taking a quick peek at a timestamp is not straightforward and requires jumping through some hoops.

exzombie:
- The code has been rebased.
- To intercept the TIME field, the "TIME" string is chopped off and replaced with the "ts" filter
- The ISO string format is disabled on VS 2012 or older to avoid crashes caused by the timezone issue.

modules/database/src/std/filters/ts.c
```diff
@@ -9,27 +10,130 @@
...
+static const chfPluginArgDef ts_args[] = {
+    chfEnum(tsPrivate, mode, "num", 0, 0, ts_numeric_enum),
+    chfEnum(tsPrivate, epoch, "epoch", 0, 0, ts_epoch_enum),
+    chfEnum(tsPrivate, str, "str", 0, 0, ts_string_enum),
+    chfPluginArgEnd
+};
+
+static int parse_finished(void *pvt) {
+    tsPrivate *settings = (tsPrivate *)pvt;
+    if (settings->str != tsStringInvalid) {
```
mdavidsaver, line 79: Is this the condition which justifies the `cantProceed()`s below?

modules/database/src/std/filters/ts.c
```diff
@@ -56,34 +160,207 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
...
-static void channelRegisterPre(dbChannel *chan, void *pvt,
-                               chPostEventFunc **cb_out, void **arg_out, db_field_log *probe)
-{
+static db_field_log *replace_fl_value(tsPrivate const *pvt,
+                                      db_field_log *pfl,
+                                      void (*func)(tsPrivate const *,
+                                                   db_field_log *)) {
+    /* Get rid of the old value */
+    if (pfl->type == dbfl_type_ref && pfl->u.r.dtor) {
+        pfl->u.r.dtor(pfl);
```
mdavidsaver: It may only be paranoia, but can you add `pfl->u.r.dtor = NULL;` after running the old dtor.

modules/database/src/std/filters/ts.c
```diff
@@ -56,34 +160,207 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
...
+        return;
+    }
+
     *cb_out = filter;
+    *arg_out = pvt;
+
+    /* Get rid of the value of the probe because we will be changing the
+       datatype */
+    if (probe->type == dbfl_type_ref && probe->u.r.dtor) {
+        probe->u.r.dtor(probe);
```
mdavidsaver: And here.

modules/database/src/std/filters/ts.c
```diff
@@ -56,34 +160,207 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
...
+    pfl->u.v.field.dbf_double = arr[0] + arr[1] * 1e-9;
+}
+
+static void ts_array(tsPrivate const *settings, db_field_log *pfl) {
+    pfl->field_type = DBF_ULONG;
+    pfl->field_size = sizeof(epicsUInt32);
+    pfl->no_elements = 2;
+    pfl->type = dbfl_type_ref;
+    pfl->u.r.pvt = NULL;
+    pfl->u.r.field = allocTsArray();
```
mdavidsaver, line 228: `freeListCalloc()` can return NULL, which needs to be handled in some way.  (eg. return 0 elements?)

modules/database/src/std/filters/ts.c
```diff
@@ -56,34 +160,207 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
...
+static void ts_string(tsPrivate const *settings, db_field_log *pfl) {
+    char const *fmt;
+    char *field;
+    size_t n;
+
+    pfl->field_type = DBF_STRING;
+    pfl->field_size = MAX_STRING_SIZE;
+    pfl->type = dbfl_type_ref;
+    pfl->u.r.pvt = NULL;
+    pfl->u.r.field = allocString();
```
mdavidsaver: Also here.

mdavidsaver (commented):
> AI on MAD to review name handling again & comment.

Without the changes to `dbChannel.c`, my objections in that area are mote.

Otherwise, only a couple of minor comments.

modules/database/src/std/filters/ts.c
exzombie: I'm not sure what you mean with "justifies". But let me explain. The possible states after parsing are:
- no arguments were given, mode is `tsModeInvalid`, and we change that to `tsModeGenerate`;
- "num" was given, mode is one of the numeric modes, nothing needs to be done here;
- "str" was given, so we set mode accordingly.

So in this sense, `cantProceed()` calls further down are justified because those situations shouldn't be possible, and I wanted the `switch` statements to be exhaustive.

modules/database/src/std/filters/ts.c
exzombie: Good point, I'm too used to OS where allocations can't fail ...

modules/database/src/std/filters/ts.c
exzombie: I chose to return 0 elements here as well. I was a bit afraid that the client could be confused by this, but it seems to work well with camonitor. I thought that making a reference to a statically allocated null string would be safer, but it behaves strangely. Perhaps downstream code does something I don't expect when the `dtor` is NULL. If returning 0 elements concerns anyone, I will investigate further.

modules/database/src/std/filters/ts.c
anjohnson: Core group 8 June: Failing with `cantProceed()` is not appropriate in code that can be triggered remotely by any client, that could be a DoS which we'd rather avoid, please handle this some other way. It is definitely appropriate to log a message in the event of a logic error though.

exzombie: I see your point. Even if there was a straightforward way to prove the code was unreachable, it may not remain that way when someone changes it. Fixed in [e23c4bc](https://github.com/epics-base/epics-base/pull/126/commits/e23c4bc91332ebdb10f52ab68081c25c4c6cdd96).

exzombie:
I have rebased this PR to the latest 7.0 branch, including the changes to `db_field_log`. Is there something more that needs to be done to push this forward?

anjohnson:
Hi Jure, we will review this PR again in our meeting next week, thanks.

anjohnson:
Core Group 2022-10-26: Looks good, thanks.

modules/database/src/std/filters/ts.c
```diff
@@ -56,34 +162,248 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
...
+        field[0] = 0;
+    }
+
+    return 0;
+}
+
+static db_field_log *filter(void *pvt, dbChannel *chan, db_field_log *pfl) {
+    (void)chan;
+
+    tsPrivate *settings = (tsPrivate *)pvt;
```
mdavidsaver: This fails with older compilers which insist that C code have local definitions before statements.

Looking more closely, the appveyor builds didn't run for some reason (why this wasn't found sooner).  @exzombie could you rebase to pick up any appveyor config changes?

From a vs2012 build:

```
ts.c(284) : error C2275: 'tsPrivate' : illegal use of this type as an expression
        ts.c(68) : see declaration of 'tsPrivate'
ts.c(284) : error C2065: 'settings' : undeclared identifier
ts.c(286) : error C2065: 'settings' : undeclared identifier
ts.c(286) : error C2223: left of '->mode' must point to struct/union
ts.c(318) : error C2275: 'tsPrivate' : illegal use of this type as an expression
        ts.c(68) : see declaration of 'tsPrivate'
ts.c(318) : error C2065: 'settings' : undeclared identifier
ts.c(319) : error C2065: 'settings' : undeclared identifier
ts.c(319) : error C2223: left of '->mode' must point to struct/union
ts.c(329) : error C2275: 'tsPrivate' : illegal use of this type as an expression
        ts.c(68) : see declaration of 'tsPrivate'
ts.c(329) : error C2065: 'settings' : undeclared identifier
ts.c(331) : error C2065: 'settings' : undeclared identifier
ts.c(331) : error C2223: left of '->mode' must point to struct/union
ts.c(348) : error C2065: 'settings' : undeclared identifier
ts.c(348) : error C2223: left of '->mode' must point to struct/union
ts.c(379) : error C2275: 'tsPrivate' : illegal use of this type as an expression
        ts.c(68) : see declaration of 'tsPrivate'
ts.c(379) : error C2065: 'settings' : undeclared identifier
ts.c(381) : error C2065: 'settings' : undeclared identifier
ts.c(381) : error C2223: left of '->mode' must point to struct/union
ts.c(381) : error C2223: left of '->epoch' must point to struct/union
ts.c(381) : error C2223: left of '->str' must point to struct/union
ts.c(385) : warning C4113: 'void *(__cdecl *)()' differs in parameter lists from 'void *(__cdecl *)(void)'
```

mdavidsaver: > ... the appveyor builds didn't run for some reason ...

Looking at the appveyor event log, I understand what is happening.  That log shows:

> Commit "fd25f63d" of branch "7.0" skipped as commit files matched "skip_commits.files"

The build is being skipped because the last commit only updates the release notes.  I recall being confused by this before.  GHA seems to be smart enough to consider all commits being pushed, but appveyor isn't and only considers the last.

@exzombie The easiest workaround would be to re-shuffle your patches so that the release notes change is the first rather than last commit (or squash it with one of the others).

exzombie: Ugh, I'll try to remember that, thanks. I prefer to treat release notes separately because they *always* have conflicts. Anyway, when I noticed that Appveyor jobs aren't running, I had them run overnight [against my fork](https://ci.appveyor.com/project/exzombie/epics-base) instead, and they were all successful.

anjohnson: Thanks. We just discussed that Appveyor problem in today's Core group meeting and will be removing the documentation directory from Appveyor's `skip_commits.files` setting to avoid this in the future.

anjohnson:
Core Group: Approved, whoever merges needs to wait for Appveyor to complete before pushing to epics-base project.

mdavidsaver:
Applied as of 9655b78e1108d2f805a134c396c71fc7d9d69791 @exzombie Thank you for persevering!

PR #127, opened 2021-03-09
errlog.h: added doxygen comments.

mp49:


modules/libcom/src/error/errlog.h
```diff
@@ -22,8 +36,17 @@
...
+ * errlogListener function type. 
+ * 
+ * This is used when adding or removing log listeners in ::errlogAddListener 
+ * and ::errlogRemoveListeners.
+ */
 typedef void (*errlogListener)(void *pPrivate, const char *message);
 
+/**
+ * errlog severity enums
+ */
```
mdavidsaver: A minor nit.  Please avoid two extra lines when there is only one line of content.  eg.

```c
/** errlog severity enums */
```

modules/libcom/src/error/errlog.h
```diff
@@ -45,44 +68,191 @@ LIBCOM_API extern int errVerbose;
...
+ * \param listener Function pointer of type ::errlogListener
+ * \param pPrivate
+ */
 LIBCOM_API void errlogAddListener(errlogListener listener, void *pPrivate);
+
+/**
+ * This function will remove a listener callback.
+ *
+ * \param listener Function pointer of type ::errlogListener
+ * \param pPrivate
```
mdavidsaver: While it is common practice for C callbacks, it might be worth saying that `pPrivate` will be passed as the first argument of `listener()`.

modules/libcom/src/error/errlog.h
mp49: I've compressed the single line comments in the latest commit

modules/libcom/src/error/errlog.h
mp49: Thanks, I wasn't sure how to describe that. I've added that description.

mdavidsaver (approved):
Looks fine to me.

mdavidsaver (approved):
Looks ok to me.

mdavidsaver:
Applied as 1fcbdad5e991af5b8c7a4f6996c9a301e849bfc1.  Thanks.

PR #129, opened 2021-03-09
WIP: Fix msvc warnings

FreddieAkeroyd:
Fix various warning on Visualo Studio 2019 - Work In Progress

### DIFF
--- a/configure/os/CONFIG.win32-x86.win32-x86
+++ b/configure/os/CONFIG.win32-x86.win32-x86
@@ -32,12 +32,16 @@ HDEPENDS_METHOD = MKMF
 
 #
 # -W<d> display warnings at level d
+#    -Wall is -W4 plus some additional warnings that are
+#          off by default (https://docs.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default?view=msvc-160)
 #    -W4 is for maximum (lint type) warnings
 #    -W3 is for production quality warnings
 #    -W2 displays significant warnings
 #    -W1 is the default and shows severe warnings only
-# -w<d><n> Set warning C<n> to be shown at level <d>
-WARN_CFLAGS_YES = -W3
+# -we<n> Make warning C<n> an error
+# -wd<n> Disable warning C<n>
+# -we4013 makes it an error to implicitly define a function
+WARN_CFLAGS_YES = -W3 -we4013
 WARN_CFLAGS_NO  = -W1
 
 #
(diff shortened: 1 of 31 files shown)


FreddieAkeroyd:
So there are some
```
unsafe use of type 'bool' in operation     in epicsCalcTest.c
unary minus operator applied to unsigned type, result still unsigned    in epicsStdlibTest.c
```
which I think is part of the test. There is
```
 warning C4996: 'GetVersionExA': was declared deprecated
```
which we can leave for another day. 

There are some issues with SOCKET that could be resolved with lots of casts to integer, but cleaner would be  changing the return type of some functions from int to SOCKET. I wasn't sure if that broke API compatibility somewhere (SOCKET on win64 is 64bit, but only the lower 32 bits are used hence casting to an integer is safe).
```
warning C4297: 'casAsyncPVAttachIO::~casAsyncPVAttachIO': function assumed not to throw an exception but does
warning C4297: 'casAsyncPVExistIO::~casAsyncPVExistIO': function assumed not to throw an exception but does
warning C4297: 'casAsyncReadIO::~casAsyncReadIO': function assumed not to throw an exception but does
warning C4297: 'casAsyncWriteIO::~casAsyncWriteIO': function assumed not to throw an exception but does
```
these would either need a `noexcept(false)` protected for older c++, or the mechanism changing.

mdavidsaver:
I don't think there would be much value to fixing up warnings in the PCAS code in 3.15 since it has been removed in 7.0.

> I wasn't sure if that broke API compatibility somewhere ...

Wouldn't this be a bug fix?  Any breakage would be in cases where the user code is incorrectly truncating `SOCKET`.

mdavidsaver:
>  warning C4996: 'GetVersionExA': was declared deprecated

Both osdThread and osdTime are calling `InitializeCriticalSection()` unconditionally, so the NT check could simply be removed.

https://github.com/epics-base/epics-base/blob/5593103c11157aada4e40c5f00c6b033387fb57e/modules/libcom/src/osi/os/WIN32/osdMutex.c#L73-L74

mdavidsaver:
Oh never mind, I got that test backwards.  `CreateMutex` is the "newer".

FreddieAkeroyd:
The SOCKET issue I was thinking of was binary DLL compatibility, but maybe we don't promise this? The function would have the same name but return 64bits rather than 32 bits. If user code compiled against an old library activated the new DLL then I didn't know if this unexpectedly larger return type would cause an issue (either as it would pass through a different return register, or cause an overwrite). But maybe everything would pass through the same 64bit register/64bit aligned stack and it would be fine. But if we don't guarantee a new DLL working with old code then there is no problem.

FreddieAkeroyd:
@mdavidsaver you are right that the check can be removed - `thisIsNT` will be true for everything we are likely to find these days and we could just remove the `CreateMutex` based code.

mdavidsaver:
> binary DLL compatibility, but maybe we don't promise this?

We've never promised that any two Base releases would be ABI compatibility on any target.

FreddieAkeroyd:
We could also drop the `_WIN32_WINNT` define too in `osdThread.c` - it always sets this to `0x0400` and then has a bit of code that only gets included if this is `0x0501`

PR #130, opened 2021-03-10
Fix type comparison on msi.cpp

gabrielfedel:
This change fix the comparison of different signedness (int and long unsigned int).
The pull request is on 3.15 so it can be propagated to 7.0 also.

### DIFF
--- a/src/ioc/dbtemplate/msi.cpp
+++ b/src/ioc/dbtemplate/msi.cpp
@@ -300,11 +300,11 @@ static void makeSubstitutions(inputData * const inputPvt,
             char *pstart;
             char *pend;
             int  cmdind=-1;
-            long unsigned int  i;
+            size_t  i;
 
             for (i = 0; i < NELEMENTS(cmdNames); i++) {
                 if (strstr(command, cmdNames[i])) {
-                    cmdind = i;
+                    cmdind = (int)i;
                 }
             }
             if (cmdind < 0) goto endcmd;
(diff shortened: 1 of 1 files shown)


src/ioc/dbtemplate/msi.cpp
```diff
@@ -300,7 +300,7 @@ static void makeSubstitutions(inputData * const inputPvt,
             char *pstart;
             char *pend;
             int  cmdind=-1;
-            int  i;
+            long unsigned int  i;
```
FreddieAkeroyd: Don't you now get a warning on line 307 with `cmdind = i;` instead? It might be you need to leave it as integer and put an integer cast on NELEMENTS.

gabrielfedel: Actually no.

Do you think is better to use the cast ?

FreddieAkeroyd: I get a warning on windows if I make the change
```
C:\devel\epics-base\src\ioc\dbtemplate\msi.cpp(307): warning C4365: '=': conversion from 'unsigned long' to 'int', signe
d/unsigned mismatch
```
given `cmdind` needs to be signed for `-1` there will need to be a cast somewhere. Maybe using `size_t i` and then putting the cast on `cmdind = (int)i` is better than on `NELEMENTS`

gabrielfedel: Looks good. I've done that.

mdavidsaver (approved):
Looks fine to me.

PR #131, opened 2021-03-10
Use waitable timers on WIN32

FreddieAkeroyd:
Use waitable timers, potentially a high resolution version, to try and avoid early wakeups.

Probably missing some additional timer cleanup code at the moment, also need to check on older windows versions.

See #106

### DIFF
--- a/src/libCom/osi/os/WIN32/osdEvent.c
+++ b/src/libCom/osi/os/WIN32/osdEvent.c
@@ -26,6 +26,8 @@
 #include "shareLib.h"
 #include "epicsEvent.h"
 
+#include "osdThreadPvt.h"
+
 typedef struct epicsEventOSD {
     HANDLE handle;
 } epicsEventOSD;
@@ -83,8 +85,6 @@ epicsShareFunc epicsEventStatus epicsEventWait ( epicsEventId pSem )
         return epicsEventError;
     }
 }
-
-extern HANDLE osdThreadGetTimer(void); /* from osdThread.c */
 
 /*
  * epicsEventWaitWithTimeout ()
--- a/src/libCom/osi/os/WIN32/osdEvent.c
+++ b/src/libCom/osi/os/WIN32/osdEvent.c
@@ -84,36 +84,49 @@ epicsShareFunc epicsEventStatus epicsEventWait ( epicsEventId pSem )
     }
 }
 
+extern HANDLE osdThreadGetTimer();
+
 /*
  * epicsEventWaitWithTimeout ()
  */
 epicsShareFunc epicsEventStatus epicsEventWaitWithTimeout (
     epicsEventId pSem, double timeOut )
 { 
-    static const unsigned mSecPerSec = 1000;
+    static const unsigned nSec100PerSec = 10000000;
+    HANDLE handles[2];
     DWORD status;
-    DWORD tmo;
+    LARGE_INTEGER tmo;
+    HANDLE timer;
 
     if ( timeOut <= 0.0 ) {
-        tmo = 0u;
-    }
-    else if ( timeOut >= INFINITE / mSecPerSec ) {
-        tmo = INFINITE - 1;
+        tmo.QuadPart = 0u;
     }
     else {
-        tmo = ( DWORD ) ( ( timeOut * mSecPerSec ) + 0.5 );
-        if ( tmo == 0 ) {
-            tmo = 1;
-        }
+        tmo.QuadPart = -((LONGLONG)(timeOut * nSec100PerSec + 0.5));  // +0.99999999 ?
     }
-    status = WaitForSingleObject ( pSem->handle, tmo );
+
+    if (tmo.QuadPart < 0) {
+        timer = osdThreadGetTimer();
+        if (!SetWaitableTimer(timer, &tmo, 0, NULL, NULL, 0))
+        {
+            printf("event error %d\n", GetLastError());
+            return epicsEventError;
+        }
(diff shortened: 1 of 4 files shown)


src/libCom/osi/os/WIN32/osdEvent.c
```diff
@@ -84,33 +84,45 @@ epicsShareFunc epicsEventStatus epicsEventWait ( epicsEventId pSem )
     }
 }
 
+extern HANDLE osdThreadGetTimer(void); /* from osdThread.c */
```
mdavidsaver: In similar cases (recent posix PI mutex changes) definitions like this have gone into a local private header under `os/*/`.

src/libCom/osi/os/WIN32/osdThread.c
```diff
@@ -526,6 +526,14 @@ static win32ThreadParam * epicsThreadParmCreate ( const char *pName )
...
         strcpy ( pParmWIN32->pName, pName );
         pParmWIN32->isSuspended = 0;
+#ifdef CREATE_WAITABLE_TIMER_HIGH_RESOLUTION
+        pParmWIN32->timer = CreateWaitableTimerEx(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
+        if (pParmWIN32->timer == NULL) {
+            pParmWIN32->timer = CreateWaitableTimer(NULL, 0, NULL);
+        }
+#else
+        pParmWIN32->timer = CreateWaitableTimer(NULL, 0, NULL);
+#endif
```
mdavidsaver: Need some check if `CreateWaitableTimer()` can fail.  Also, since `pParmWIN32` is allocated with `calloc()`, this could be slightly shorten to:

```c
#ifdef CREATE_WAITABLE_TIMER_HIGH_RESOLUTION
        pParmWIN32->timer = CreateWaitableTimerEx(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
#endif
        if (pParmWIN32->timer == NULL) {
            pParmWIN32->timer = CreateWaitableTimer(NULL, 0, NULL);
        }
        if (pParmWIN32->timer == NULL) {
            cantProceed("CreateWaitableTimer()");
        }
```

mdavidsaver:
Well, this looks more straightforward than I was expecting!

src/libCom/osi/os/WIN32/osdEvent.c
FreddieAkeroyd: I did consider putting it in `osdThread.h` but this is included in `epicsThread.h` and so the data seemed to be getting leaked further. Is that OK i.e. people should not rely on functions in osd? it doesn't need to be `epicsShareExtern` so that would block DLL access. Or should i create a new private header file that is only used by `osdEvent.c` and `osdThread.c` ?

mdavidsaver: I think the best way forward is to create eg. `os/WIN32/osdThreadPvt.h`.  Creating the file is all that is necessary.  Private headers don't need any Makefile change.

Of course, historically Base has been quite sloppy about putting "private" definitions in public (installed) headers.  Over time of course some of these definitions get used, complicating later maintenance.  We've been trying to improve this situation.  I admit that we're still sometimes slightly sloppy and wrap such definitions in an `#ifdef EPICS_PRIVATE_API`.

src/libCom/osi/os/WIN32/osdThread.c
FreddieAkeroyd: Thanks for the shortening. Should it return `NULL` on error (plus free memory) rather than calling `cantProceed` for consistency, or instead be calling `callocMustSucceed` rather than `calloc` earlier?

mdavidsaver: ya ya ya.  I wasn't paying attention to the context...

FreddieAkeroyd:
The waitable timer API is Windows XP or better. The `CreateWaitableTimerEx` function is Windows vista or better, but it will not get compiled into code unless high resolution timers are present (windows 10 1803 or higher). At the moment i don't know what would happen if  CREATE_WAITABLE_TIMER_HIGH_RESOLUTION was used on W10 prior to 1803. I may be able to use Windows server 2016 to do a test.

anjohnson:
Is this PR now ready to merge? I would like to see a note in `documentation/ReleaseNotes.md` about the change to the timers so anybody who reads them can see what you did without having to delve into the commits or the actual code changes. Doesn't need to be very detailed, but a including link to the document you found might be good.

FreddieAkeroyd:
Hi Andrew, I am pretty much busy all day today, but will do the release notes bit at end of day, probably in time for Americas wrap up

FreddieAkeroyd:
The only remaining things are a bit of testing the behaviour of mixing binaries across windows versions. If the program was compiled on W10 1803 or more recent (so compiled in support for high resolution timers) but was then run on an older version of windows 10 without such support. If passing the high resolution flag to the createTimer function is either ignored or returns an error, then everything will be handled. If it were to instead return a "broken" timer then we would have problems (either lots of error messages when we set a delay on the timer, or some other issue). 1803 is an old Window 10 version, but Windows Server 2016 (Long-Term Servicing Channel) is based on version 1607 and in standard support until November 2022. So I would like to do this test, did you want to wait until I have done this or merge as is and I'll open a bug report if I find a problem.

anjohnson:
Merging as agreed in today's Core Group meeting.

PR #132, opened 2021-03-10
Use epicsStrtod, remove some warnings

FreddieAkeroyd:
Use `epicsStrtod`

### DIFF
--- a/src/libCom/yajl/yajl_parser.c
+++ b/src/libCom/yajl/yajl_parser.c
@@ -45,6 +45,8 @@
 #include "yajl_encode.h"
 #include "yajl_bytestack.h"
 
+#include <osdStrtod.h>
+
 unsigned char *
 yajl_render_error_string(yajl_handle hand, const unsigned char * jsonText,
                          unsigned int jsonTextLen, int verbose)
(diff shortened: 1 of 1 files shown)


src/libCom/yajl/yajl_parser.c
```diff
@@ -45,6 +45,8 @@
 #include "yajl_encode.h"
 #include "yajl_bytestack.h"
 
+#include <osdStrtod.h>
+
```
anjohnson, line 49: I'd prefer that you replace the `#include <stdlib.h>` with `#include <epicsStdlib.h>` than pull in this header, which was really meant to be an implementation detail.

anjohnson (approved):
Needs care when merging up to 7.0 to avoid regressions since both `memneeded` and `YA_MALLOC()` have already been modified to use `size_t` there.

src/libCom/yajl/yajl_parser.c
FreddieAkeroyd: Thanks - sorry, I should have looked more closely how it was used elsewhere in the code

FreddieAkeroyd:
@anjohnson Should I bring the YA_MALLOC (and any other 7.0 differences) back onto the 3.15 branch?

anjohnson:
@FreddieAkeroyd No, that was part of a major reimport of the yajl code from upstream. This looks good now, thanks!

PR #133, opened 2021-03-10
Make C implicit function declarations an error on win32

FreddieAkeroyd:
Make C implicit function declarations an error on win32


FreddieAkeroyd:
CI build failing due to 
```
../../../src/ioc/dbtemplate/dbLoadTemplate_lex.l(223): error C4013: 'read' undefined; assuming extern returning int
```
which is now an error - this was fixed as part of  #129

mdavidsaver:
Maybe this PR should be combined with #129?

FreddieAkeroyd:
I've merged it into #129 - when I created this PR I had forgotten it would currently fail...

PR #139, opened 2021-03-11
epicsStdlib.h: add doxygen comments.

mp49:


modules/libcom/src/misc/epicsStdlib.h
```diff
@@ -25,57 +39,171 @@
 extern "C" {
 #endif
 
-#define S_stdlib_noConversion (M_stdlib | 1) /* No digits to convert */
-#define S_stdlib_extraneous   (M_stdlib | 2) /* Extraneous characters */
-#define S_stdlib_underflow    (M_stdlib | 3) /* Too small to represent */
-#define S_stdlib_overflow     (M_stdlib | 4) /* Too large to represent */
-#define S_stdlib_badBase      (M_stdlib | 5) /* Number base not supported */
```
mdavidsaver, line 32: These `S_*` macros with end of line comments are "magic".  They are parsed out by `modules/libcom/src/error/makeStatTbl.pl` which generates the lookup table behind eg. `errSymLookup()`.

They either need to stay as they were, or `makeStatTbl.pl` needs to be taught about `/**< ... */` style comments.

mdavidsaver (changes requested):

modules/libcom/src/misc/epicsStdlib.h
mdavidsaver: fyi.

https://github.com/epics-base/epics-base/blob/ca3ef9c61e57a7413ad9626f17ce9205cdabdc5f/modules/libcom/src/error/Makefile#L23-L30

mp49: Would it be ok to duplicate the comment (so it's in the doxygen part and the end of line part)? If not, I'll just revert that section.

anjohnson: @mp49 Yes, add your `/** ... */` lines for Doxygen just before the original unmodified `#define` line.

mp49: I've made that change, and also rebased against the latest 7.0 branch.

mdavidsaver (approved):
Looks good.

mdavidsaver:
Applied as 5daf4fc932feda6468ca0141b354cc8b6c0e217c.  Thanks.

Issue #137, opened 2021-03-10
Avoid intentional memory leak in `epicsEnvSet()`


mdavidsaver:
`epicsEnvSet()` has different implementations, some of which use `setenv()`, and some `putenv()`.  Those using `putenv()` currently leak on each call since the buffer passed to `putenv()` may be referenced by the libc for some indeterminate time.  Currently this is the case for: default, WIN32, and vxWorks.  The others use `setenv()`

The default impl (used by eg. Linux) could probably be switched to use `setenv()` at this point.  Windows has [`_putenv_s`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-160) which seems to work like `setenv()`.  @anjohnson says that with at least recent vxWorks (>=6.8), `putenv()` is making a copy of the provided buffer, which we could then free.

```
$ git ls-files|grep osdEnv
modules/libcom/src/osi/os/Darwin/osdEnv.c
modules/libcom/src/osi/os/RTEMS/osdEnv.c
modules/libcom/src/osi/os/WIN32/osdEnv.c
modules/libcom/src/osi/os/default/osdEnv.c
modules/libcom/src/osi/os/iOS/osdEnv.c
modules/libcom/src/osi/os/solaris/osdEnv.c
modules/libcom/src/osi/os/vxWorks/osdEnv.c
```

dirk-zimoch:
> @anjohnson says that with at least recent vxWorks (>=6.8), putenv() is making a copy of the provided buffer, which we could then free.

The same was already true for vxWorks 5. No need to keep a copy of the buffer in vxWorks.

dirk-zimoch:
From the VxWorks 5.5 manual:

>    STATUS putenv  ( char * pEnvString  )
>
>    This routine sets an environment variable to a value by altering an existing variable or creating a new one. The parameter points to a string of the form "variableName=value". Unlike the UNIX implementation, the string passed as a parameter is copied to a private buffer.
> RETURNS
>    OK, or ERROR if space cannot be malloc'd.

PR #140, opened 2021-03-11
epicsEnvSet remove intential memory leak

mdavidsaver:
Address #137


mdavidsaver:
This also get us down to three, actually distinct, implementations.

```
$ git ls-files|grep osdEnv
modules/libcom/src/osi/os/WIN32/osdEnv.c
modules/libcom/src/osi/os/default/osdEnv.c
modules/libcom/src/osi/os/vxWorks/osdEnv.c
```

modules/libcom/src/osi/os/WIN32/osdEnv.c
```diff
@@ -34,25 +34,8 @@
...
-                __FILE__,
-                __LINE__,
-                "Failed to set environment parameter \"%s\" to \"%s\": %s\n",
-                name,
-                value,
-                strerror (errno));
-        free (cp);
-    }
+    _putenv_s(name, value);
 }
```
FreddieAkeroyd, line 46: Do we need to check the return code,  or that value does not exceed `_MAX_ENV` ?

mdavidsaver: Well, probably yes.  Although none of the other impls check for `ENOMEM` from `putenv()` or `setenv()`.

How should an error be communicated?  Change to a non-void return type?  (I hesitate to printf)

FreddieAkeroyd: Changing the return type feels like a lot of work might potentially need to be done to trace the error through, though maybe that is the most correct.  Do you feel using `errPrintf` creates a dependency between the osi and other layers? It is probably unlikely for `_putenv_s` to fail, but if it did could we end up with a hard to trace error? That was probably more my concern, though errors caused by corrupt/missing environment variables are probably clearer than others to identify, so maybe it is not such an issue.

modules/libcom/src/osi/os/vxWorks/osdEnv.c
```diff
@@ -52,8 +52,9 @@ LIBCOM_API void epicsStdCall epicsEnvSet (const char *name, const char *value)
         errPrintf(-1L, __FILE__, __LINE__,
             "Failed to set environment parameter \"%s\" to \"%s\": %s\n",
             name, value, strerror (errno));
-        free (cp);
     }
+    /* from at least vxWorks 5.5 putenv() is making a copy, so we can free */
+    free (cp);
```
anjohnson: It would be nice to make `cp` a stack variable here, thus completely avoiding the need to call `malloc()` and `free()`. Since we're using gcc the following should be legal (but we still have to abide by the 'allocations before statements' rule so it must appear at the start of a block):

```C
    iocshEnvClear(name);
    {
        char cp[strlen (name) + strlen (value) + 2];

        strcpy (cp, name);
        strcat (cp, "=");
        strcat (cp, value);
        if (putenv(cp) < 0) {
```
The "Leaks memory" comment higher up could also be deleted since we won't any more.

modules/libcom/src/osi/os/WIN32/osdEnv.c
anjohnson: The code Freddie is deleting was already calling `mallocMustSucceed()` which would suspend and dump a stack trace if it failed, and `errPrintf()` if the `putenv()` failed. Sticking with the latter wouldn't make matters any worse, so I say stick with that.

anjohnson (commented):
A couple of suggestions.

mdavidsaver:
I think I've addressed the comments.  Errors are now reported with `errlogPrintf()`, and calls to `cantProceed()` are removed.

On vxWorks, I've also made an (untested) attempt to use gcc on-stack dynamic arrays to avoid allocation for short strings.  The security conscious C programmer in me forces a fallback to `malloc()` for longer strings.

mdavidsaver:
@anjohnson Do you plan to build/run test the vxWorks change prior to merge?  Otherwise, I will deal with any build fail which Jenkins turns up.

modules/libcom/src/osi/os/default/osdEnv.c
```diff
@@ -20,40 +20,27 @@
...
-#include "cantProceed.h"
 #include "envDefs.h"
 #include "osiUnistd.h"
 #include "epicsFindSymbol.h"
 #include "iocsh.h"
 
+#ifdef __rtems__
+#  define RTEMS_VERSION_INT VERSION_INT(__RTEMS_MAJOR__, __RTEMS_MINOR__, 0, 0)
+#endif
+
```
anjohnson, line 34: The whole point of our having the `osi/os/$(OS_CLASS)/` source structure was to avoid putting OS-specific conditionals into the source files since that makes them harder to read. The differences here are pretty minor so I understand the desire to not duplicate most of this code in an `osi/os/RTEMS/osdEnv.c` file, but they could be encapsulated even better like this:

```C
#ifdef __rtems__
#  define RTEMS_VERSION_INT VERSION_INT(__RTEMS_MAJOR__, __RTEMS_MINOR__, 0, 0)
#endif

#if defined(__rtems__) && RTEMS_VERSION_INT<VERSION_INT(4,10,0,0)
   /* newlib w/ RTEMS <=4.9 returns void */
#  define unsetenv_M(name) ({unsetenv(name); 0})
#else
#  define unsetenv_M(name) unsetenv(name)
#endif
```
Then remove the RTEMS-4.9 conditional below and call `unsetenv_M()` instead of `unsetenv()`.

Note that there is also a comment about `putenv()` immediately above the `epicsEnvUnset()` routine below which needs to be removed.

modules/libcom/src/osi/os/vxWorks/osdEnv.c
```diff
@@ -35,25 +34,32 @@
  */
 LIBCOM_API void epicsStdCall epicsEnvSet (const char *name, const char *value)
 {
-    char *cp;
+    size_t alen = (!name || !value) ? 2u : strlen(name) + strlen(value) + 2u; /* <NAME> '=' <VALUE> '\0' */
+    const int onstack = alen <= 64u; /* use on-stack dynamic array for small strings */
```
anjohnson: I would make the transition point much larger, say 512 bytes, which we would rarely if ever hit. There is enough stack for this kind of thing.

modules/libcom/test/epicsEnvTest.c
```diff
@@ -72,8 +70,43 @@ MAIN(epicsEnvTest)
     value = getenv(PARENT);
     testOk(value && (strcmp(value, PARENT) == 0),
             "PARENT environment variable not modified");
+}
+
+static void testChangeEnv(void)
+{
+    const char *foo = "foo", *bar = "bar", *name = "testChangeEnv";
```
anjohnson, line 77: Why does the code below sometimes use `name` and sometimes a literal `"testChangeEnv"`? It's a little confusing to read, and up to the compiler whether all the literals would return the same pointer.

anjohnson (commented):
I could build it on VxWorks but I don't have the ability to run any tests at the moment, my VME systems at work are turned off (except for the MVME2500 for Brendan to test RTEMS on). I suggest we wait until you do the merge to check the VxWorks build.

I added some additional comments here with some more code improvements though, sorry!

modules/libcom/test/epicsEnvTest.c
mdavidsaver: A (perhaps futile) attempt to detect if `epicsSetEnv()` stores, and later clobbers, the `const char*` passed to it.

modules/libcom/src/osi/os/vxWorks/osdEnv.c
mdavidsaver: Ok, your call.

modules/libcom/src/osi/os/default/osdEnv.c
mdavidsaver: Right, I hardly ever use the `({...})` construct, but I guess it should be compatible with old enough gcc and clang.  Would it work, and does it need to work, with suncc?

Digression...

I was thinking about putting the (conditional) definition of `RTEMS_VERSION_INT` in `epicsVersion.h`.  Maybe also `GCC_VERSION_INT`.  We use these somewhat widely, and the definitions should always be the same.

anjohnson: The `({ ... })` construct is only needed when building for RTEMS <= 4.9, the other branch of my suggestion doesn't need/use it. Are you asking whether RTEMS 4.9 was built with suncc? 😉 

Exporting `RTEMS_VERSION_INT` might have a deleterious effect on external modules which also define it and would not expect it to be already defined. I'm not objecting to the idea, but it might have some repercussions downstream. There is only one place each where we currently define and use that and `VXWORKS_VERSION_INT` although I haven't checked the RTEMS-5 branch.

mdavidsaver: > ... Are you asking whether RTEMS 4.9 was built with suncc? wink

Brain fart...

> Exporting `RTEMS_VERSION_INT` might have a deleterious effect ...

True, although at worst this would force adding another `#ifdef` (which I would include in `epicsVersion.h` as well).  I see both the benefits and costs as minor.  So I may add these definitions the next time it would be convenient, or not.

mdavidsaver:
I've addressed the round 2 comments.

anjohnson (approved):
Thanks!

mdavidsaver:
Merged as of 3eeebb74cd83757756acb2024560e602ee8f9118

PR #141, opened 2021-03-11
Improve message sanity checks in camessage.c

### LINKED ISSUE #128: Improve the CAS error message generated when an IOC gets port-scanned
I'm copying here the issue published in the codeathon ideas list: [improve-the-cas-error-message-generated-when-an-ioc-gets-port-scanned](https://github.com/epics-base/epics-base/wiki/code-a-thon-project-ideas#improve-the-cas-error-message-generated-when-an-ioc-gets-port-scanned)

>Read [this tech-talk thread](https://epics.anl.gov/tech-talk/2021/msg00064.php).
>
>The code that needs changing is in [camessage.c::camessage()](https://github.com/epics-base/epics-base/blob/7050bded319cafcb92b14ff82a95716aa44dffcd/modules/database/src/ioc/rsrv/camessage.c#L2453). In the case described in tech-talk above, checking the condition msg.m_cmmd < NELEMENTS(tcpJumpTable) would have indicated that the TCP socket was not opened by a real CA client. There may be better ways to distinguish the first message received over TCP from a pre-3.12.0-beta1 client, but the IOC should attempt to detect non-CA clients and report them as such.

jesusvasquez333:
This PR improves the sanity checks perform to incoming messages. This improvements will resolve #128 .

The following improvements are implemented:
- ~One issue is that the [test to detect deprecated client versions](https://github.com/epics-base/epics-base/blob/f30e9533c49b50cbdddc361b327f97e5e3d9612c/modules/database/src/ioc/rsrv/camessage.c#L2454) fails when we haven't receive a client version yet (and therefore `client->minor_version_number` is `0`). So, I added an extra condition to that test [to check is  client->minor_version_number!=0](https://github.com/epics-base/epics-base/commit/1bcf55e29e6cdbda5688c8fbd12ad5e381dce07b).~ (This situation happens with clients < `CA_V49`, but the proposed change was not enough to make this test backward compatible, so I removed it from the PR). 
- I also added 2 extra sanity checks to incoming messages, before they are processed, for the 2 fields of the headers that are fixed:
  - A [check to detect message with invalid command ID](https://github.com/epics-base/epics-base/commit/2ef15ed8299a68403d0739d9fbb92c3f910d1892) right after the message is received. This will caught messages like the one described in #128, and
  - A [check to detect invalid payload sizes, considering both standard and extended forms](https://github.com/epics-base/epics-base/commit/9e1a8bf67e22ac344dd7577c358c48c1092d766f).

I couldn't figure out a way to detect deprecated clients on the first message. When the first message arrives, we still do not know the client version; clients starting at `CA_V49` will send a `CA_PROTO_VERSION` message which does includes the version number. However, the deprecated clients we are trying to detect ( < `CA_V44`) do not implement the `CA_PROTO_VERSION` messages, but instead use the old `CA_PROTO_NOOP` message which doesn't include the version number. ~So, I just [added a comment in the code describing this](https://github.com/epics-base/epics-base/commit/fdb498abda5251dde6cf9a3bed49baa7859b5e68).~ (this topic is left to be revisit in the future).

I also added a few more fixes:
- According to the documentation, the extended message form header should have both fields `payload` set to `0xffff` and `data count` set to `0`. The code however, was only checking the payload field. So, I [added the check for the data count filed as well](https://github.com/epics-base/epics-base/commit/5db754131a31502c728b3b9f930d2edb863c8084).
- The `CA_PROTO_EVENT_CANCEL` message contains the `data type` filed. Although the type is not actually important for this command, I [added a check for that field](https://github.com/epics-base/epics-base/commit/29f9d4db197487f9d81a52e5b5f3235532990172) as a sanity check of the received message.
- I [replaced one data type check](https://github.com/epics-base/epics-base/commit/7eafd07d213fad791f1a0a84b7cd566ee97ad581), which was using directly `LAST_BUFFER_TYPE` , for the macro `INVALID_DB_REQ()` instead.
- Some other minor formatting issues, and added some more comments.


jesusvasquez333:
With this changes, when I reproduce the issue as described in #128 (i.e. sending a HTTP GET request to the IOC), I see this error instead in the IOC shell:
```
epics> CAS: request from 172.16.125.3:50770 => CAS: Invalid command rejected
CAS: Request from 172.16.125.3:50770 => cmmd=18245 cid=0x54502f31 type=12064 count=18516 postsize=21536
CAS: Request from 172.16.125.3:50770 =>   available=0x2e310d0a 	N=0 paddr=(nil)
CAS: forcing disconnect from 172.16.125.3:50770
```

Note:  After some discussions, this log message is now controlled using `CASDEBUG`. If the variable is `0`, the message won't be printed.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -2434,24 +2438,77 @@ int camessage ( struct client *client )
...
             pBody = ( void * ) ( mp + 1 );
         }
 
-        /* ignore deprecated clients, but let newer clients identify themselves. */
-        if (msg.m_cmmd!=CA_PROTO_VERSION && !CA_VSUPPORTED(client->minor_version_number)) {
+        /*
+         * Ignore deprecated clients, but let newer clients identify themselves.
+         *
+         * If "client->minor_version_number" is 0, it means that we haven't received the minor
+         * version number from the client yet, so we don't check the version.
```
mdavidsaver: `CA_PROTO_VERSION` will be the first message which either peer sends.  So seeing any other message first is a violation.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -2434,24 +2438,77 @@ int camessage ( struct client *client )
         msg.m_cid       = ntohl ( mp->m_cid );
         msg.m_available = ntohl ( mp->m_available );
 
-        if ( CA_V49(client->minor_version_number) && msg.m_postsize == 0xffff  ) {
+        /* disconnect clients that send an invalid command */
+        if (msg.m_cmmd > CA_PROTO_LAST_CMMD) {
```
mdavidsaver: I had deliberately avoided adding this test as a way to preserve the possibility of future extensions of the protocol.

It this point in the lifetime of CA, it's probably reasonable to expect that there won't be too many more.  So maybe this test could be adjusted to `CA_PROTO_LAST_CMMD + 16`?

modules/database/src/ioc/rsrv/camessage.c
jesusvasquez333: I think that is true only for >= `CA_V49` clients? Previous version of the protocol didn't contain the `CA_PROTO_VERSION` message, but instead the `CA_PROTO_NOOP` which doesn't include the minor version number in the header. Is that correct?

And `CA_VSUPPORTED()` will compare if the version is >= `CA_V44` instead.

modules/database/src/ioc/rsrv/camessage.c
jesusvasquez333: But this particular version of the server only implements `CA_PROTO_LAST_CMMD` commands. So, even if a future client, using a newer version of the protocol with more commands, if it sends a request for a command not implemented in this version (i.e. `m_cmmd > CA_PROTO_LAST_CMMD`), that request should still be rejected as the server does not implement it. Of course, command lower than `CA_PROTO_LAST_CMMD` will still be processed.

In order to use those new commands, you will need a server using the new protocol version as well. And that server will have a higher `CA_PROTO_LAST_CMMD`.

What do you think?

anjohnson: We generally don't want to _disconnect_ clients or servers just because they support a newer protocol version than we do; as long as they exchange version numbers the newer end is supposed to adapt its behavior to the older end, so both can still talk using the older of the two protocol versions. That's why the RSRV code is littered with version checks. It has always been possible for different version clients and servers to talk because they are designed for exactly that.

jesusvasquez333: Oh, yes, that is a good point. Then, what about doing the same that was done for deprecated clients:
https://github.com/epics-base/epics-base/blob/ca3ef9c61e57a7413ad9626f17ce9205cdabdc5f/modules/database/src/ioc/rsrv/camessage.c#L2456-L2467

But with a different error message (`CAS: Invalid command rejected`)?

mdavidsaver: > .. request should still be rejected as the server does not implement it.

Yes, but silently.  Otherwise introducing a new message will result in copious log spam.

This would be especially acute with UDP broadcasts.  How can I reasonably introduce a new UDP search message (eg. maybe search w/ IPv6 address) without flooding the logs of all existing IOCs?

While I know that is not the case now.  As a general rule, I think that validation errors with UDP packets should be silent by default.

modules/database/src/ioc/rsrv/camessage.c
mdavidsaver: Well... does this mean that with 95fd255dd6b888b90624547165062b8a97278fab I actually broke support for clients before v4.9 (aka before 3.14)?

jesusvasquez333: Good question. I want to say no, but I'm not sure. 

If a  <`CA_V49`  client tries to send a command via TCP, it will be caught by the `CA_VSUPPORTED(client->minor_version_number)` statement as `client->minor_version_number = 0`. Versions previous to `CA_V49` (up to [R3.14.0-alpha2](https://github.com/epics-base/epics-base/releases/tag/R3.14.0-alpha2), which uses protocol minor [version 8](https://github.com/epics-base/epics-base/blob/R3.14.0-alpha2/src/ca/caProto.h#L39)) implements the [CA_PROTO_NOOP](https://github.com/epics-base/epics-base/blob/R3.14.0-alpha2/src/ca/caProto.h#L101) message instead; that message does not include the minor version number, but `0` instead, in the [header of the CA_PROTO_NOOP message](https://github.com/epics-base/epics-base/blob/R3.14.0-alpha2/src/ca/tcpiiu.cpp#L1007). [In fact, [the documentation says](https://github.com/epics-docs/epics-docs/blob/master/specs/ca_protocol.rst#ca_proto_version), that the version is not available for < `CA_V411`, but I think it should say `CA_V49` instead, see [epics-docs#2](https://github.com/epics-docs/epics-docs/issues/2)].

As the minor version number can be extracted from other types of messages (like for example the [CA_PROTO_CLAIM_CIU](https://github.com/epics-base/epics-base/blob/R3.14.0-alpha2/src/ca/tcpiiu.cpp#L1574), which correspond to the newer `CA_PROTO_CREATE_CHANNEL`, and the [CA_PROTO_SEARCH](https://github.com/epics-base/epics-base/blob/R3.14.0-alpha2/src/ca/nciu.cpp#L195) ), I added the extra condition `client->minor_version_number != 0` to allow those commands to be processed. 

This will however, let pass all commands until a version number is extracted; not  sure if that is better or worse.

modules/database/src/ioc/rsrv/camessage.c
jesusvasquez333: That makes sense to me. The invalid commands can then be rejected, without disconnecting the client and without generating a log error; only sending the error back to the client? 

It will look something like this:
```c
/*
 * Reject invalid commands.
 * Send the error to the client, but keep the connection open to avoid a
 * re-connect loops.
 * Do not generate a log message for this type of error to avoid flooding
 * the logs.
 */
if (msg.m_cmmd > CA_PROTO_LAST_CMMD) {
    SEND_LOCK(client);
    send_err(&msg, ECA_NOSUPPORT, client, "CAS: Invalid command rejected");
    SEND_UNLOCK(client);
    client->recvBytesToDrain = msgsize - bytes_left;
    client->recv.stk = client->recv.cnt;
    status = RSRV_OK;
    break;
}
```

mdavidsaver: I think we need to be clear about distinguishing TCP from UDP since `camessage()` handles both.  What you show is reasonable for a TCP connection.  It isn't reasonable for UDP where responding with an error to an unknown or even invalid UDP packet would invite the sort of "amplification" DDoS scenarios seen with other UDP based protocols like DNS or NTP.  Especially if responding to a broadcast.

So I think the new validations you add are totally reasonable.  Sending a remote error to a TCP peer is also reasonable (eg. `client->proto==IPPROTO_TCP`).

Maybe these console log messages should be disabled by default?  eg. conditioned on a new `rsrvDebug` global.

anjohnson: > conditioned on a new `rsrvDebug` global

Shouldn't that use the already-existing `CASDEBUG` variable? Do you really want a new one?

mdavidsaver: Ah, yes.  I had forgotten about that, and looked at the wrong end of the alphabet.

modules/database/src/ioc/rsrv/camessage.c
mdavidsaver: So... I was able to get a capture of a 3.13 era CA client searching,  The UDP packet has only CA_PROTO_SEARCH messages, so I think it would be ignored.  A fix, if we choose to make one, would be to expand this condition:

```c
if (msg.m_cmmd!=CA_PROTO_VERSION && !CA_VSUPPORTED(client->minor_version_number)) {
```

with exceptions for the other messages which include the peer version.

modules/database/src/ioc/rsrv/camessage.c
jesusvasquez333: I think I added all the requested changes this section. Invalid commands are now rejected without disconnecting the client. The error is sent only to TCP clients, and the log message is disabled by default but can be enabled (only for TCP clients as well) using the `CASDEBUG` variable.

modules/database/src/ioc/rsrv/camessage.c
jesusvasquez333: Yes. I think the original condition will ignore a `CA_PROTO_SEARCH` message, as `client->minor_version_number` will be `0`.  

I think your proposal of filtering by message type is a better solution respect to my original `client->minor_version_number != 0` condition, which will let pass all type of messages.

From the documentation, the messages that contain the minor version number, beside from `CA_PROTO_VERSION` of course, are: `CA_PROTO_SEARCH` and `CA_PROTO_CREATE_CHANNEL`. So, we could expand the condition to be:
```c
if (msg.m_cmmd!=CA_PROTO_VERSION &&
    msg.m_cmmd!=CA_PROTO_SEARCH &&
    msg.m_cmmd!=CA_PROTO_CREATE_CHANNEL &&
    !CA_VSUPPORTED(client->minor_version_number)) {
```

By the way, looking at the code, it seems like the version number is not being extracted from the `CA_PROTO_SEARCH` messages. It is extracted from a `CA_PROTO_CREATE_CHANNEL` message [here](https://github.com/epics-base/epics-base/blob/54e9d3f5d1256edf8e6eca06191f51ba0aaf4cf2/modules/database/src/ioc/rsrv/camessage.c#L1223), but the `CA_PROTO_SEARCH` messages only do a version check [here for TCP](https://github.com/epics-base/epics-base/blob/54e9d3f5d1256edf8e6eca06191f51ba0aaf4cf2/modules/database/src/ioc/rsrv/camessage.c#L2265-L2268) and [here for UDP](https://github.com/epics-base/epics-base/blob/54e9d3f5d1256edf8e6eca06191f51ba0aaf4cf2/modules/database/src/ioc/rsrv/camessage.c#L2178-L2181). So I think we need to add:
```c
client->minor_version_number = mp->m_count;
```
to extract version number from those messages as well.

mdavidsaver: I'm not sure this would be enough.  In the >=3.14 CA traffic I've seen, the `CA_PROTO_CLIENT_NAME` and `CA_PROTO_HOST_NAME` message are sent before `CA_PROTO_CREATE_CHANNEL`.  The 3.13 capture I was able to get didn't include the TCP phase, so I don't know if these messages were present.

This situation was briefly discussed at the core meeting on the 24th.  The outcome was that, absent more developer interest, to recommend that sites still having 3.13 IOCs, and wishing to mix in base 7.0, to put the old IOCs behind a cagateway built against base 3.14 or 3.15.

So I don't think this PR needs to be so concerned with 3.13 compatibility as long as it isn't making the situation more complicated in case someone wants to revisit this topic in future.

jesusvasquez333: Ok, sounds good to me. So, I will then revert back this condition to its original state:
```c
if (msg.m_cmmd!=CA_PROTO_VERSION && !CA_VSUPPORTED(client->minor_version_number)) {
```
as that will work as expect for client >= `CA_V49` (i.e. base >= `R3.14.0-beta1`).

BTW, maybe you want to considered bumping the [CA_MINIMUM_SUPPORTED_VERSION](https://github.com/epics-base/epics-base/blob/7.0/modules/ca/src/client/caProto.h#L31)  to `9`? so that [CA_VSUPPORTED](https://github.com/epics-base/epics-base/blob/7.0/modules/ca/src/client/caProto.h#L32) will compare against `CA_V49`. As it seems like previous version are not actually supported.

jesusvasquez333: I reverted back this check to its original state.

anjohnson:
Core Group 7/28: AI on MAD.

marciodo:
I've created the function validate_camessage to check for valid commands (current last command number + 16 for making space for newer clients with more commands) and a few combinations of expected value for a field based on the command type.

When trying an HTTP GET, the IOC prints the message below and close the connection:
```
CAS: request from 134.79.228.91:52850 => CAS: Invalid channel access message rejected
CAS: Request from 134.79.228.91:52850 => cmmd=18245 cid=0x54502f31 type=12064 count=18516 postsize=21536
CAS: Request from 134.79.228.91:52850 =>   available=0x2e310d0a         N=0 paddr=(nil)
CAS: forcing disconnect from 134.79.228.91:52850
``` 

I've tested with caput, caget, and camonitor, using scalars and waveform with elements below the standard protocol size and above it, to have tests covering both standard and extended sizes. Versions of caput, caget, and camonitor used: 3.14.12, 3.15.5, and the ones from this pull request.

Let me know if you'd like to print the message only if CASDEBUG is set.

I've merged the branch of this pull request with the most recent 7.0 to make sure the previous changes in 7.0 would work with the changes made by this PR and check for conflicts.

anjohnson:
Ping @mdavidsaver please.

mdavidsaver:
This change _looks_ ok to me.

However, on reflection I'm not comfortable taking sole responsibility for review and testing of this PR.  Changes to message handling code carry a risk of breaking wire compatibility.  After having been responsible for one such accidental break, I hesitate to risk another.

wrt. #128 my now less than long term answer for confusing RSRV log messages is to disable RSRV in favor of PVA (with its own confusing messages ;) ).

anjohnson:
Group to @anjohnson Rework to reduce scope creep, remove message validation checks, risk to break compat.

PR #143, opened 2021-03-12
Stdio doxygen annotations

KathrynBaker:


modules/libcom/src/osi/epicsStdio.h
```diff
@@ -52,23 +83,72 @@ extern "C" {
 #  define putchar epicsStdoutPutchar
 #endif
 
+/**
+ * \brief `epicsSnprintf` is meant to have the same semantics as the C99 function `snprintf`
+ *
+ * \details
```
mdavidsaver, line 89: Can the duplication of `\details` be avoided?  Maybe by moving to the `\file` block and linking from individual functions?  (cf `\ref` and `\section` an similar)

anjohnson:
Merged as 1cacd058cd3bd1d9b6cabc6219c48217913eb1e8

PR #145, opened 2021-03-12
Timer improvements

### LINKED ISSUE #106: Timers expire early

mark0n:
- Use high-resolution timers on WIN32 (waitable timers)
- Additional tests (in case the compiler is C++11 compatible)
- Fix a bug which often caused timers to expire half a quantum early on Linux and WIN32 (even if high-precision timers were used)

This PR fixes #106. It supersedes #107 and #108.

### DIFF
--- a/configure/os/CONFIG.win32-x86.win32-x86
+++ b/configure/os/CONFIG.win32-x86.win32-x86
@@ -188,7 +188,9 @@ RES=.res
 
 # MS Visual C++ doesn't recognize *.cc as a C++ source file,
 # so C++ compiles get the flag -TP
-COMPILER_CXXFLAGS = -TP
+# We want MSVC to report a reasonable C++ version in the __cplusplus macro so
+# we need to set -Zc:__cplusplus (this is at least required for MSVC 19.xx).
+COMPILER_CXXFLAGS = -TP -Zc:__cplusplus
 
 # Operating system flags
 OP_SYS_CFLAGS =
--- a/modules/libcom/src/osi/Makefile
+++ b/modules/libcom/src/osi/Makefile
@@ -39,6 +39,7 @@ INC += epicsGeneralTime.h
 INC += osdTime.h
 INC += generalTimeSup.h
 INC += osiClockTime.h
+INC += osdTimer.h
 INC += epicsSignal.h
 INC += osiProcess.h
 INC += osiUnistd.h
--- a/modules/libcom/src/osi/os/Linux/osdTimer.h
+++ b/modules/libcom/src/osi/os/Linux/osdTimer.h
@@ -1,0 +1,14 @@
+/*************************************************************************\
+* Copyright (c) 2021 Facility for Rare Isotope Beams
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+#ifndef osdTimerh
+#define osdTimerh
+
+/* Linux supports high-precision timers (in contrast to quantized timers which
+ * only support sleeping for a multiple of the quantum). */
+#define HAS_HIGH_PREC_TIMERS
+
+#endif /* osdTimerh */
--- a/modules/libcom/src/osi/os/WIN32/osdEvent.c
+++ b/modules/libcom/src/osi/os/WIN32/osdEvent.c
@@ -26,6 +26,8 @@
 #include "libComAPI.h"
 #include "epicsEvent.h"
 
+#include "osdThreadPvt.h"
+
 typedef struct epicsEventOSD {
     HANDLE handle;
 } epicsEventOSD;
@@ -83,8 +85,6 @@ LIBCOM_API epicsEventStatus epicsEventWait ( epicsEventId pSem )
         return epicsEventError;
     }
 }
-
-extern HANDLE osdThreadGetTimer(void); /* from osdThread.c */
 
 /*
  * epicsEventWaitWithTimeout ()
--- a/modules/libcom/src/osi/os/WIN32/osdEvent.c
+++ b/modules/libcom/src/osi/os/WIN32/osdEvent.c
@@ -84,7 +84,7 @@ LIBCOM_API epicsEventStatus epicsEventWait ( epicsEventId pSem )
     }
 }
 
-extern HANDLE osdThreadGetTimer();
+extern HANDLE osdThreadGetTimer(void); /* from osdThread.c */
 
 /*
  * epicsEventWaitWithTimeout ()
@@ -92,7 +92,7 @@ extern HANDLE osdThreadGetTimer();
 LIBCOM_API epicsEventStatus epicsEventWaitWithTimeout (
     epicsEventId pSem, double timeOut )
 {
-    static const unsigned nSec100PerSec = 10000000;
+    static const unsigned nSec100PerSec = 10000000u;
     HANDLE handles[2];
     DWORD status;
     LARGE_INTEGER tmo;
@@ -102,14 +102,12 @@ LIBCOM_API epicsEventStatus epicsEventWaitWithTimeout (
         tmo.QuadPart = 0u;
     }
     else {
-        tmo.QuadPart = -((LONGLONG)(timeOut * nSec100PerSec + 0.5));  // +0.99999999 ?
+        tmo.QuadPart = -((LONGLONG)(timeOut * nSec100PerSec + 0.5));
     }
 
     if (tmo.QuadPart < 0) {
         timer = osdThreadGetTimer();
-        if (!SetWaitableTimer(timer, &tmo, 0, NULL, NULL, 0))
-        {
-            printf("event error %d\n", GetLastError());
+        if (!SetWaitableTimer(timer, &tmo, 0, NULL, NULL, 0)) {
             return epicsEventError;
         }
         handles[0] = pSem->handle;
@@ -123,10 +121,11 @@ LIBCOM_API epicsEventStatus epicsEventWaitWithTimeout (
         return epicsEventOK;
     }
     else if ( status == WAIT_OBJECT_0 + 1 || status == WAIT_TIMEOUT ) {
+        /* WaitForMultipleObjects will trigger WAIT_OBJECT_0 + 1,
+           WaitForSingleObject will trigger WAIT_TIMEOUT */
         return epicsEventWaitTimeout;
     }
     else {
-        printf("event error %d\n", GetLastError());
         return epicsEventError;
     }
 }
--- a/modules/libcom/src/osi/os/WIN32/osdEvent.c
+++ b/modules/libcom/src/osi/os/WIN32/osdEvent.c
@@ -84,36 +84,49 @@ LIBCOM_API epicsEventStatus epicsEventWait ( epicsEventId pSem )
     }
 }
 
+extern HANDLE osdThreadGetTimer();
+
 /*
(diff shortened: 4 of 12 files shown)


FreddieAkeroyd:
I've added a few diagnostics to  https://github.com/FreddieAkeroyd/epics-base/tree/timers  and also updated my home machine to 20H2 so I can compare to the work machine for pre W10 2004 behaviour.

One thing I have noticed (on my computer at least) is that `epicsThreadSleepQuantum` always returned 15.625ms whatever my timer resolution was set to by `timeBeginPeriod` or otherwise. It uses `GetSystemTimeAdjustment` which is the rate at which the system will adjust time discrepancies, which always seems to be at the lower resolution end of the supported range. So I think the logic of subtracting half a "tick" from the requested time was working out worse than would be expected as it was using the wrong "tick" duration as a basic. 

The current tests pass most of the time, I need to look a bit deeper. Calling `timeBeginPeriod` does not affect things now we are using high resolution timers, I verified that on W10 20H2 where I could see `Sleep` and `WaitForSingleObject` averaging 16ms to do a 1ms wait (with no timeBeginPeriod call) whereas `epicsThreadSleep` and `epicsEventWaitWithTimeout` were still averaging ~1.7 ms

mdavidsaver:
I notice that one of my test in pva2pva is failing for windows/mingw builds.  I've only just fixed the mingw CI jobs, which had been incorrectly configured, so I'm not 100% certain that that this branch is responsible.  I can't recall seeing this failure before.  The nature of the error suggests to me a possible spurious wakeup of an epicsEvent.

```
Bailout called.  Further testing stopped:  Data event w/o data
FAILED--Further testing stopped: Data event w/o data
Tests failed in:
    modules/pva2pva/testApp/O.win32-x86-mingw
        testpdb.tap
```

FreddieAkeroyd:
I'll check the mingw distribution, it is possible their win api headers do not include the define that is used to test for the presence of high precision timers and so they are running the standard ones. If this was the problem though, then i  would have thought that some of the appveyor builds would have shown this too - the image used to run the VS2010 build is windows server 2012 so certainly doesn't have high precision timers.

FreddieAkeroyd:
So,
- mingw isn't using high resolution timers
- when I run under mingw `testpdb` passes, but I instead get a failure in `testpvalink`
```
not ok  3 - dbGetField("src:i1.INP", 0) -> "{"pva":"target:i"}" == "{pva:"target:i"}"
not ok 12 - dbGetField("src:o2.OUT", 0) -> "{"pva":"target:i2"}" == "{pva:"target:i2"}"
testpvalink.t ...... Dubious, test returned 1 (wstat 256, 0x100)
Failed 2/15 subtests
```

FreddieAkeroyd:
The `testpdb` test fails in appveyor automated build, but if I remote desktop into the node after it has failed and run the test it passes...

mdavidsaver:
@FreddieAkeroyd  I know there are lingering race conditions with testpdb and testpvalink.  If you don't specifically see `Data event w/o data` then it's probably not worth your time investigating.

mdavidsaver:
Also, it would be good to see if the TODO can be removed from epicsEventTest (eg. by removing the `_WIN32` case.  (cf. cd32a7cb1e8c3bca78683c54ea42ede04cf698a6).
https://github.com/epics-base/epics-base/blob/7eb7988e55e2bc572030984b03f12d864eb6017e/modules/libcom/test/epicsEventTest.cpp#L175-L177

FreddieAkeroyd:
I've re-enabled the delay calculation on my machine and it seems fine, I'll see what appveyor makes of it

mdavidsaver:
> TODO can be removed from epicsEventTest

I realized that I could do this myself (I haven't used this aspect of github permissions before).  Some of my builds with this change fail with small negative delays eg.

```
epicsEventTest.tap ........... 
not ok 18 - epicsEventWaitWithTimeout(0.031250)  delay error -0.000011 sec
```

https://ci.appveyor.com/project/mdavidsaver/epics-base/build/job/b4m1lqpnxnawd286#L17817

FreddieAkeroyd:
I've been getting that too on my appveyor builds. First issue was that MinGW defaults to Windows XP if you don't explicitly specify `_WIN32_WINNT` and so will by default not use the high resolution timers (i will need to start a separate discussion about specifying `_WIN32_WINNT` etc). Then only the VS2019 appveyor build image runs a version of windows that supports such timers, and mingw doesn't use that image.  Also the clock tick on the appveyor is 16ms - there seems to be nothing around to have called e.g. `timeBeginPeriod` to drop it. So I'm looking at a combination of small things ... it will need to be a bit cleverer about when it is using high resolution timers, and maybe we have a rounding error somewhere. I had a feeling the non high resolution timers could wake up a bit early, but i'll need to confirm that is definitily the case.

mdavidsaver:
I'm glad you're on top of this :)  It might help to reduce confusion (mine at least) to add some prints to show what compile/runtime support is present.  Maybe something like the `epicsMutexOsdShowAll()` I added (5593103c11157aada4e40c5f00c6b033387fb57e) to show if PI mutexes are actually being created.

mdavidsaver:
fyi. I'm testing a change to switch the appveyor mingw builds to the 2019 image.  To me at least, it seems reasonable to ask mingw users to update more often.

wrt. `_WIN32_WINNT`, there are a couple of places where this is set to `0x400` (aka. `_WIN32_WINNT_NT4`).  In PVXS I'm [setting this](https://github.com/mdavidsaver/pvxs/blob/32debd3e40921dc85eb14689b599cffac1e994d1/configure/CONFIG_SITE#L44-L45) to `_WIN32_WINNT_VISTA` (`0x0600`) in order to use [R/W locks](https://docs.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks).

FreddieAkeroyd:
Ideally we should check MinGW on three of the build image (2015, 2017, 2019) as they all represent currently supported versions of windows. Windows 8.1 / Server 2012 (vs1015 image) is supported until January 2023 and server 2016 (vs 2017 image) until 2027

FreddieAkeroyd:
Appveyor looks to mix mingw-w64 versions on different images, i'm currently trying the MSYS2 distribution that is available on all images https://github.com/FreddieAkeroyd/ci-scripts/commits/mingw_vs_2019

I am also now setting my `_WIN32_WINNT` at 0x0600 in MinGW as this provides `CreateWaitableTimerEx` - I then check for high res timer support separately at run time. I've not quite sorted out what to do with tests, if high res timers are available then all runs fine. If they are not, then either waitable timers or the previous mechanism are both prone to early wake up.

mdavidsaver:
> Appveyor looks to mix mingw-w64 versions on different images ...

Yup...  Except for the added hours to appveyor builds, I wouldn't mind testing newest mingw builds on older windows.  I mostly don't want to spend time on older mingw.

ralphlange:
+1
Especially as there is MSYS2, I have no issue asking users to update their MinGW installation. That's a really easy thing to do, especially compared to asking for an update of the Windows installation.

AppVeyor seems to have been using whatever they felt was appropriate at the time they generated the image, and were careful not to change their images more than necessary after they were public, which is understandable.

mdavidsaver:
> ... I've not quite sorted out what to do with tests ...

Since there isn't much we can do about older windows, I think these tests need to be marked "todo".  Preferably this would be condition on the presence of build+runtime support.  This could mean adding another (exported) function alongside `osdThreadGetTimer()` in `osdThreadPvt.h` (which should be under `WIN32/`), and conditionally including this from test code.

anjohnson:
Core Group 2022-10-26: Since we've lost @mark0n (and the timer code is a bit scary), we're not inclined to continue working on these changes ourselves. If someone else wants to take it up (or to rewrite timers completely) we will be happy to consider that, but we've agreed to close this PR for now.

PR #146, opened 2021-03-20
Timeouts for running unit test programs

anjohnson:
The generated .t file now kills the test program if it hasn't
completed within a defined interval, 5 minutes by default.
Separate implementations for Windows and Unix hosts.


anjohnson:
Well that's interesting, Appveyor is passing the tests run by the new scripts while the Windows targets on GHA won't let the script spawn any of the test programs, they are getting "Access is denied" errors. Hopefully I can fix that in the Perl code, I have some pointers to possible solutions.

anjohnson:
```
epicsTypesTest.t: Can't spawn Process './epicsTypesTest.exe': Access is denied
```
[The Win32::Job::spawn() method](https://metacpan.org/pod/Win32::Job#Using-Win32::Job) documentation says:

> One reason this will fail is if the calling process is itself part of a job, and the job's security context does not allow child processes to be created in a different job context than the parent.

Looking at the implementation, the `spawn()` method [returns `undef`](https://metacpan.org/source/JDB/Win32-Job-0.05/Job.xs#L506) when Windows' `CreateProcess()` routine returns 0. The [MS Documentation](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa) isn't very clear what might cause that, but the Perl module does set `CREATE_BREAKAWAY_FROM_JOB` in the `dwCreationFlags` argument and that flag's documentation says about it:

> If the calling process is associated with a job, the job must set the `JOB_OBJECT_LIMIT_BREAKAWAY_OK` limit.

It appears that GHA is not setting that flag [as described here](https://docs.microsoft.com/en-us/windows/win32/procthread/job-objects#managing-a-process-tree-that-uses-job-objects), which might be quite reasonable from their perspective, but it probably means that I can't use the Win32::Job module.

So let't try again using the (less well-documented) Win32::Process API instead, which also has the wait-for-timeout and kill-child features we need.

anjohnson:
This seems to be working now (all GHA checks complete with one test failing; first Appveyor job passed). I still have to work on fixing netget.plt to use the same API when running under Windows and to write some Release Notes, so I haven't finished but it should be sufficient for use in the @mdavidsaver integration branch.

documentation/RELEASE_NOTES.md
```diff
@@ -18,6 +18,33 @@ should also be read to understand what has changed since earlier releases.
...
+default 5 minutes, 300 seconds) without exiting it will be killed after that
+period has elapsed. The limit can be extended (or shortened) for individual
+tests as necessary by adding a line to the Makefile which runs them, setting
+the environment variable `EPICS_UNITTEST_TIMEOUT` to a value in seconds while
+building the test (`.t`) script, like this:
+
+```
+  TESTSCRIPTS_HOST += hourLongTest.t
+  hourLongTest.t: export EPICS_UNITTEST_TIMEOUT=3600
+```
```
mdavidsaver, line 126: I don't think I like encouraging hard coded overrides like this.

The author of a test can in principle determine a minimum execution time.  However, timeout is about a maximum execution time, which may vary widely depending on the runtime environment.  So I would expect that `$EPICS_UNITTEST_TIMEOUT` will always be set from CI configuration.  Practically speaking, based on the longest observed test runtime.

I suppose timeout configuration could be split, with test authors able to provide a minimum time, with maybe a multiplier provided from CI environment.  However, I'm not sure this complexity would be worthwhile.  Current best practice encourages short unittests anyway.

anjohnson: Having some kind of CI-specific multiplier wouldn't really work when the test runs for a specific elapsed time, e.g. epicsMessageQueueTest, and I agree that trying to do some kind of combination adds unnecessary complexity.

A test author has to be able to override the default timeout for an individual test. The way this branch is currently coded the environment variable is checked by `makeTestfile.pl` when it is generating the `.t` script to set the default timeout written into the script, but the script itself _also_ checks for it when it's called by `make runtests` or `make tapfiles` and if provided will use that value instead of the compiled-in timeout even if that's shorter. Maybe it should display a warning if it sees that, or always use the longer of the two values?

Remember that an author's Makefile can always override any value provided by the CI environment anyway. However if they use the documented approach that you're commenting on of only setting the value when building that specific target their setting will only override the CI when the `.t` script is built, not when it's run, so the CI's value would still be used at runtime. The author can always override the CI if they want to though, you can't prevent a Makefile from doing this, which will take effect for all targets including `make tapfiles`:

```makefile
  export EPICS_UNITTEST_TIMEOUT=3600
```

anjohnson:
From [here](https://github.com/epics-base/epics-base/pull/146/checks?check_run_id=2176357310#step:8:392):
```
epicsMessageQueueTest.t: Timed out 'qemu-system-i386 -m 64 -no-reboot -serial stdio -display none -net nic,model=ne2k_pci -net user,restrict=yes -kernel epicsMessageQueueTest' after 300 seconds
```
My 5 minute default evidently isn't long enough for RTEMS-4.9 runs. We could add a test-specific override here, increase my default (for all tests), or set `EPICS_UNITTEST_TIMEOUT` in the CI builder for RTEMS runs. Preferences?

mdavidsaver:
> set `EPICS_UNITTEST_TIMEOUT` in the CI builder for RTEMS runs.

This is my preference.  Also, my $0.02 just set this long (eg. 20 minutes).  Hopefully RTEMS test hangs will remain infrequent enough for this not to be annoying.

documentation/RELEASE_NOTES.md
```diff
@@ -18,6 +18,33 @@ should also be read to understand what has changed since earlier releases.
...
+The time limit only applies to tests that use `makeTestfile.pl` to generate a
+test (`.t`) script that runs the actual test program. Tests which are provided
+as a `.plt` script should implement their own timeout if they can hang while
+running in a Continuous Integration system (the Base "netget" test now does
+this in a way that works on Windows as well as Posix hosts).
+
+The same environment variable can be set while running the tests to override
+the time limit given in the Makefile, but it cannot be configured to provide a
+separate limit for each program when using `make runtests`. This feature may
+be useful for debugging purposes.
```
mdavidsaver: Is this statement about overriding correct?  It looks to me like it would not be possible for CI environment to override a hard coded timeout.

```make
all: target

target: export FOO=baz
target:
        echo "FOO|$$FOO|${FOO}"
        env| grep FOO
```

```sh
$ FOO=bar make
echo "FOO|$FOO|baz"
FOO|baz|baz
env| grep FOO
FOO=baz
```

anjohnson: Your Makefile is too simple because the `target` target isn't creating a script that later gets run from a different make rule. EPICS has one set of rules for generating the `.t` scripts and another for running them and creating the `.tap` files. My notes recommend using the target-specific syntax so the variable will only be set when building the `.t` scripts, not when running them:
```makefile
all: build run

build: export FOO=baz

build:
	echo "build: FOO|$$FOO|${FOO}"
run:
	echo "run: FOO|$$FOO|${FOO}"
```

```
$ FOO=bar make
echo "build: FOO|$FOO|baz"
build: FOO|baz|baz
echo "run: FOO|$FOO|bar"
run: FOO|bar|bar
```

mdavidsaver: > Your Makefile is too simple ...

Ah, if only this could be true more often...

Still, I accept your point.

documentation/RELEASE_NOTES.md
mdavidsaver: I see that I've been getting confused by the wording.  As I read it, what you're trying to say is:

Unittests based on `.t` Makefile rules (eg. `TESTSCRIPTS_HOST += $(TESTS:%=%.t)`) will now have a default timeout of 5 minutes.  A test process will be killed if it runs longer.  This __default__ timeout can be overridden for individual tests like: ...

anjohnson:
> > set `EPICS_UNITTEST_TIMEOUT` in the CI builder for RTEMS runs.

> This is my preference.

I would prefer to not have to explain to users who run our tests for themselves on RTEMS why that test sometimes fails by timing out, and that they have to set the environment variable to prevent it. Thus I think our default timeouts need to be long enough that none of our tests normally fail on any of our supported targets (assuming a relatively recent machine is running QEMU).

I will bump up my default from 5 minutes to 15 or maybe the 20 minutes you suggested, which should resolve this problem. I will also clarify the wording of my Release Notes about where the timeout applies.

Thanks.

documentation/RELEASE_NOTES.md
anjohnson: I rewrote the notes, hopefully this is easier to understand:

### Timeouts for Unit Test Programs

The unit test programs that are run by the `make runtests` or `make tapfiles`
commands get executed by a `.t` wrapper script which is normally generated by
the EPICS `makeTestfile.pl` program. Those generated wrapper scripts now
impose a time-limit on the test program they execute, and will kill it if it
runs for longer than 1000 seconds (16 minutes 40) without exiting. That
time-limit can be changed for any such test by modifying the Makefile which
creates and runs the `.t` wrapper script.

Setting the environment variable `EPICS_UNITTEST_TIMEOUT` to the desired
number of seconds while the Makefile is generating the test script changes the
timeout in that script. For example:

```
  TESTSCRIPTS_HOST += hourLongTest.t
  hourLongTest.t: export EPICS_UNITTEST_TIMEOUT=3600
```

When selecting such a timeout remember that different Continuous Integration
systems such as GitHub Actions and Appveyor run on processors with different
speeds, so allow enough head-room for slower systems to complete the test.

Test programs written directly in Perl as a `.plt` script should implement a
similar timeout for themselves. The "netget" test in Base does this in a way
that works on Windows as well as Unix-like hosts.

anjohnson:
So the last Appveyor failure of this PR was apparently the `netget.t` script not exiting, despite my adding a new process watchdog around the `caget` and `pvget` commands that it runs. That might mean my new Perl `system_timeout()` facility isn't working in this particular circumstance, or that the part that's hanging is related to the communication between the test script and the iocsh that it controls – that part could get confused by a message being printed at the wrong moment, breaking the "end of command" detection logic in the `EPICS::IOC` module. The `system_timeout()` facility works fine when I run it on the APS Jenkins build agent, but it's possible one of the processes can somehow become unkillable.

I'm almost ready to throw in the towel for the `netget.t` test program on Windows. There's one more rewrite I could try, dropping the use of the `EPICS::IOC` module and running the IOC as a killable process without attempting to control it through iocsh commands at all. Is it worth trying to do that?

PR #147, opened 2021-03-23
epicsUnitTest timeout

mdavidsaver:
An external timeout like #146 seems to me a better solution.  So this change is redundant, and so may be unnecessary.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -22,6 +22,7 @@ env:
     SETUP_PATH: .ci-local:.ci
     BASE: SELF
     EPICS_TEST_IMPRECISE_TIMING: YES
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-scripts-build.yml
```diff
@@ -22,6 +22,7 @@ env:
     SETUP_PATH: .ci-local:.ci
     BASE: SELF
     EPICS_TEST_IMPRECISE_TIMING: YES
+    EPICS_TEST_TIMEOUT: 1200 # 20 min (RTEMS is slowest)
```
anjohnson, line 25: RTEMS may take 20 minutes to run all the tests, but this implies there's a specific test that needs 20 minutes, which can't be true.

mdavidsaver: What you say is true, but well... who cares? :)  20 minutes is still well short of the 5 hour job timeout.

PR #148, opened 2021-03-26
Windows: use -Z7 instead of -Zi to generate debug symbols

ralphlange:
_Based on a suggestion by Freddie Akeroyd:_

"C7 compatible" or "old-style" debug information is kept local
in the translation unit (.obj file) and does not create issues
with parallel builds.

### DIFF
--- a/configure/os/CONFIG.win32-x86.win32-x86
+++ b/configure/os/CONFIG.win32-x86.win32-x86
@@ -49,9 +49,9 @@ OPT_CFLAGS_YES_NO = -Ox -Oy-
 OPT_CFLAGS_YES = $(OPT_CFLAGS_YES_$(OPT_WHOLE_PROGRAM))
 
 #
-# -Zi generate program database for debugging information
+# -Z7 generate C7 compatible debugging information (inside .obj)
 # -RTCsu enable run-time error checks
-OPT_CFLAGS_NO = -Zi -RTCsu
+OPT_CFLAGS_NO = -Z7 -RTCsu
 
 # specify object file name and location
 OBJ_CFLAG = -Fo
@@ -116,9 +116,9 @@ OPT_CXXFLAGS_YES_NO = -Ox -Oy-
 OPT_CXXFLAGS_YES = $(OPT_CXXFLAGS_YES_$(OPT_WHOLE_PROGRAM))
 
 #
-# -Zi generate program database for debugging information
+# -Z7 generate C7 compatible debugging information (inside .obj)
 # -RTCsu enable run-time error checks
-OPT_CXXFLAGS_NO = -RTCsu -Zi
+OPT_CXXFLAGS_NO = -RTCsu -Z7
 
 # specify object file name and location
 OBJ_CXXFLAG = -Fo
--- a/configure/os/CONFIG.win32-x86.win32-x86
+++ b/configure/os/CONFIG.win32-x86.win32-x86
@@ -140,20 +140,6 @@ STATIC_LDLIBS_NO=
 STATIC_LDFLAGS=
 RANLIB=
 
-#
-# option needed for parallel builds with Visual Studio 2013 onward
-# VS2012 and above have VisualStudioVersion, so just need to exclude 2012 (11.0)
-# -FS Force Synchronous PDB Writes
-#
-ifneq ($(VisualStudioVersion),)
-ifneq ($(VisualStudioVersion),11.0)
-  OPT_CXXFLAGS_NO += -FS
-  OPT_CFLAGS_NO += -FS
-endif
-endif
-
-
-#
 # add -profile here to run the ms profiler
 # -LTCG whole program optimization
 # -incremental:no full linking


mdavidsaver:
For reference [MS documentation](https://docs.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-160) as of vs2019 does not refer to `/Z7` as "old" or "deprecated".

anjohnson:
I think this is related to 1454f42a2737fd476531a1476 which we applied in Feb 2018 and has a conditional. Should we keep the `-Zi` flag for newer VS versions and just use `-Z7` when `VisualStudioVersion` is not set?

mdavidsaver:
From my (limited) experience, I think having debug information in object files (and I guess libraries and executables?) would be preferable for our use cases where simplified debugging is more important than implementation hiding.  eg. My [ci-core-dumper](https://github.com/mdavidsaver/ci-core-dumper/blob/9a98d282dc3d074be17fe3132109b88852726b3a/ci_core_dumper/windows.py#L92) plays games to build a .pdb file search path, which I know doesn't cover all cases.

anjohnson:
These settings are only applied to `-debug` builds (i.e. when `HOST_OPT=NO`) so I guess there's no big down-side for non-developers who wouldn't be building that version anyway. Do we still want/need the `-FS` flag that was introduced by the commit I mentioned above (and excluded from vs2012 builds by 4aee25e8e2f911c08c8ebaf4) then?

ralphlange:
That's a good point.
The whole trickery with (or without) `-FS` is not needed anymore.

FreddieAkeroyd:
As Michael mentions, the docs used to talk about `/Z7` as "leacy CodeView format" but I was never quite sure what you lost over pdb files. You can't do "edit and continue" but I never did anyway.  

pdb files are a pain with static builds - the name of the pdb file is put into the object file and then into the library, and if you install the library and link later you may never find the pdb when making the executable. My discussions with Ralph were related to missing debug symbols on one of the google action builds for ci-scripts during linking of the sequencer, and I thought it was mixing debug and release, but it was just linking statically to base and not finding the pdb. I believe you can still create a pdb file from the final executable static link if you wish even when all the object files/static libraries are compiled with `-Z7` or you can put everything into a large exe

As an aside, you could also avoid the `-FS` trickery for parallel pdb by adding `-Fd$*.pdb` as the CFLAG - this creates a pdb file per object file, but is a bit messy!

At our facility we use `-Zi` for both debug and release DLL builds. We install the pdb files into a central network symbol store using `SymStore` and set the  `_NT_SYMBOL_PATH` search path environment variable on computers so they use this store at runtime for lookups. So we get symbols if anything goes wrong, but still have a smaller install on each machine. If you put the pdb files into a symbol store it doesn't matter about the embedded path within the DLL pointing to somewhere that doesn't exist anymore - it locates the correct file from the symbol store index via a fingerprint/checksum id.

So as a default for EPICS base, `-Z7` is probably a good compromise for debug builds to make testing and smaller deployments easier. What I should do is write a page about using `-Zi` and symbol stores for the EPICS wiki in case somebody else wants to also have optimised DLL builds with symbol but save on the disk space of deploying duplicate copies of symbols everywhere.

FreddieAkeroyd:
Hmmm ... when I said above "I believe you can still create a pdb file from the final executable static link..." with `-Z7` it seems that you have to do that or you get no symbols at all. As per https://docs.microsoft.com/en-us/cpp/build/reference/debug-generate-debug-info?view=msvc-160 "It is not possible to create an .exe or .dll that contains debug information. Debug information is always placed in a .obj or .pdb file."   So if we compile with `-Z7` and link `-debug` we get a pdb, if you don't specify `-debug` then it looks like there are no symbols in the exe. So we can't get away from pdb files entirely, but using `-Z7` during compile would allow us to remove `-FS` and other parallel build issues.

mdavidsaver:
Ah, I seemed too good to be true...

FreddieAkeroyd:
We could look at using `-PDBALTPATH` https://docs.microsoft.com/en-us/cpp/build/reference/pdbaltpath-use-alternate-pdb-path?view=msvc-160 and installing pdb files during the build? I've not used that myself as I found `SymStore` worked quite well - it will recursively index a directory tree and build an index for you. It might be that could work on the CI build with core dumper - just build an index after the build and point at it before running tests.  Normally symstore copies the symbol file into the store as well as building an index pointing to the new location, but for using it as a temporary lookup during build/test telling it to store pointers to the original locations in the index could work better

mdavidsaver:
Interesting, maybe in `configure/os/CONFIG.win32-x86.win32-x86` adding `-PDBALTPATH "$(FINAL_DIR)/%_PDB%"` to `LINK.shrlib` and `LINK.cpp`?  `FINAL_DIR` would need to become conditionally `INSTALL_BIN` on windows.  Now it's always `INSTALL_SHRLIB`.

There is also the problems of installing the .pdb files alongside the executables/dlls.  I'm not sure if this would need additional rules, or maybe it could be appended to `MT_EXE_COMMAND` and `MT_DLL_COMMAND`?

ralphlange:
Reverted the two `-FS` related commits that Andrew mentioned.

anjohnson:
Merging as agreed in today's Core Group meeting.

PR #149, opened 2021-03-31
Address modifiers

bfrk:
For discussion see https://code.launchpad.net/~bfrk/epics-base/+git/epics-base/+merge/381628
The branch has been rebased from the above MR.


bfrk:
It turned out I had rebased this branch from an out-dated commit. It also turned out that this PR does in fact _not_ depend on the scalar-get-optimization branch, so I have removed this dependency.

bfrk:
I don't understand the build failures.

ralphlange:
Those?

```
  make[5]: *** [../../../../../configure/RULES_BUILD:259: arrayRangeModifier.obj] Error 2
  make[5]: Entering directory 'D:/a/epics-base/epics-base/modules/database/src/ioc/O.windows-x64'
  cl          -DUSE_TYPED_RSET -DUSE_TYPED_DSET     -nologo -FC -D__STDC__=0 -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DBUILDING_dbCore_API -DBUILDING_dbCore_API   -Ox -GL -Oy-   -W3        -MD -DEPICS_BUILD_DLL -DEPICS_CALL_DLL   -I. -I../O.Common -I. -I. -I.. -I../as -I../bpt -I../db -I../dbStatic -I../dbtemplate -I../misc -I../registry -I../rsrv -I../../../../../include/compiler/msvc -I../../../../../include/os/WIN32 -I../../../../../include         -c ../db/arrayRangeModifier.c
  arrayRangeModifier.c
  D:\a\epics-base\epics-base\modules\database\src\ioc\db\arrayRangeModifier.c(23): warning C4273: 'wrapArrayIndices': inconsistent dll linkage
  make[4]: *** [../../../../configure/RULES_ARCHS:58: install.windows-x64] Error 2
  D:\a\epics-base\epics-base\modules\database\src\ioc\db\arrayRangeModifier.h(23): note: see previous definition of 'wrapArrayIndices'
  make[3]: *** [../../../configure/RULES_DIRS:85: ioc.install] Error 2
  D:\a\epics-base\epics-base\modules\database\src\ioc\db\arrayRangeModifier.c(58): error C2036: 'const void *': unknown size
  D:\a\epics-base\epics-base\modules\database\src\ioc\db\arrayRangeModifier.c(59): error C2036: 'void *': unknown size
  make[2]: *** [../../configure/RULES_DIRS:85: src.install] Error 2
  make[1]: *** [../configure/RULES_DIRS:85: database.install] Error 2
```

bfrk:
Ah, thanks, I saw the bottom part of the failure details and concluded wrongly that it was concerned with the tests. So, yes, those errors. Probably something wrong about the `epicsShareFunc`.

bfrk:
I am always confused about `epicsShareFunc` and where and where not to add it. Should I follow the example of e.g. dbEvent.h where it says:
```
#ifdef epicsExportSharedSymbols
#   undef epicsExportSharedSymbols
#   define INCLdbEventhExporting
#endif

#include "epicsThread.h"

#ifdef INCLdbEventhExporting
#   define epicsExportSharedSymbols
#endif

#include "shareLib.h"
```
?

bfrk:
Okay, looks better. I guess it is ready for review then.

mdavidsaver:
> ... Should I follow the example of ...

This is no longer necessary for `epicsThread.h` as the libCom headers are decorated with `LIBCOM_API` instead of `epicsShare*`.

modules/database/test/std/rec/addrModifierTest.c
```diff
@@ -0,0 +1,247 @@
...
+    testDiag("expecting success from %s", pv->name);
+    sprintf(fieldname, "%s.PROC", pv->name);
+    testdbPutFieldOk(fieldname, DBF_LONG, 1);
+    sprintf(fieldname, "%s.SEVR", pv->name);
+    testdbGetFieldEqual(fieldname, DBF_LONG, NO_ALARM);
+    sprintf(fieldname, "%s.STAT", pv->name);
+    testdbGetFieldEqual(fieldname, DBF_LONG, NO_ALARM);
+}
+
+#if 0
```
mdavidsaver: Is this intended as a TODO?

modules/database/test/std/rec/addrModifierTest.c
```diff
@@ -0,0 +1,247 @@
...
+        buf[i] = initial[i];
+    for (i=0, j=start; j<=end; i++, j+=incr)
+        buf[j] = values[i];
+    testdbGetFieldEqual("tgt.NORD", DBF_LONG, 8);
+    testdbGetFieldEqual("tgt.SEVR", DBF_LONG, NO_ALARM);
+    testdbGetFieldEqual("tgt.STAT", DBF_LONG, NO_ALARM);
+    testdbGetArrFieldEqual("tgt.VAL", DBF_DOUBLE, 10, 8, buf);
+}
+
+#if 0
```
mdavidsaver: ... also

modules/database/src/ioc/db/Makefile
```diff
@@ -11,10 +11,12 @@
 
 SRC_DIRS += $(IOCDIR)/db
 
+INC += arrayRangeModifier.h
```
mdavidsaver: Does `arrayRangeModifier.h` need to be public API?

modules/database/test/std/rec/addrModifierTest.c
bfrk: I don't think so. It's been a while since I wrote that so not 100% sure. My current best guess is left-over from copy-pasting from other tests. Will delete.

modules/database/src/ioc/db/Makefile
bfrk: `wrapArrayIndices` is used in the array filter code, otherwise: no. I could give this function its own C and header file. Or perhaps you have a better idea?

modules/database/test/std/rec/addrModifierTest.c
bfrk: > > ... Should I follow the example of ...
> 
> This is no longer necessary for `epicsThread.h` as the libCom headers are decorated with `LIBCOM_API` instead of `epicsShare*`.

I meant the general procedure, not the particular header file. In abf9e68a878 I tried to follow this example. This made some of the windows build problems go away, but I readily admit I have no idea what I am doing here Any advice is welcome. Is there a general guide-line I can study?

modules/database/src/ioc/db/Makefile
mdavidsaver: If this header doesn't need to be public, then simply remove the `INC += `.

mdavidsaver:
The ugly undef/def pattern is no longer needed around any of the libCom headers.  At some point we'll convert dbCore and dbRecStd to use the same `*_API` convention, and then these include ordering hacks will be a thing of the past.

wrt. removing `epicsShareFunc` from the functions in `arrayRangeModifier.h` w/o breaking anything.  This confirms that the header should not be installed.  Functions in public headers must be decorated with `epicsShareFunc` or some `*_API` macro.

mdavidsaver:
fyi. I do my development builds with:

```
$ cat configure/CONFIG_SITE.local
USR_CFLAGS_Linux += -fvisibility=hidden
USR_CXXFLAGS_Linux += -fvisibility=hidden -fvisibility-inlines-hidden
```

Which can catch some of the situations which lead to DLL linking errors on Windows, and give an idea where `epicsShareFunc` or `*_API` decorations are needed.

bfrk:
Am 01.04.21 um 18:57 schrieb mdavidsaver:
> The ugly undef/def pattern is no longer needed around any of the libCom headers.  At some point we'll convert dbCore and dbRecStd to use the same `*_API` convention, and then these include ordering hacks will be a thing of the past.

Good, thanks.

> wrt. removing `epicsShareFunc` from the functions in `arrayRangeModifier.h` w/o breaking anything.  This confirms that the header should not be installed.  Functions in public headers must be decorated with `epicsShareFunc` or some `*_API` macro.

Any suggestion what to do about `wrapArrayIndices`?

mdavidsaver:
... I had forgotten that `dbCoreAPI.h` and `DBCORE_API` already exist, and can be used in code under `modules/database/src/ioc/`.  Coupled with `-fvisibility=hidden`, this should catch most if not all situations which would lead to dllimport/export errors with MSVC.

mdavidsaver:
> Any suggestion what to do about `wrapArrayIndices`?

Include `dbCoreAPI.h`, and decorate with `DBCORE_API` to export it from dbCore.  This will allow it to be used from dbRecStd.  You may need to add `USR_CPPFLAGS=-I$(TOP)/modules/database/src/ioc/db` in `modules/database/src/std/Makefile` to reach the now private `arrayRangeModifier.h`.

mdavidsaver:
> At some point we'll convert dbCore ...

Actually, there is no reason to wait.  cf. #150

bfrk:
Am 01.04.21 um 19:55 schrieb mdavidsaver:
>> Any suggestion what to do about `wrapArrayIndices`?
> 
> Include `dbCoreAPI.h`, and decorate with `DBCORE_API` to export it from dbCore.  This will allow it to be used from dbRecStd.  You may need to add `USR_CPPFLAGS=-I$(TOP)/modules/database/src/ioc/db` in `modules/database/src/std/Makefile` to reach the now private `arrayRangeModifier.h`.

Okay, I am now using DBCORE_API. I opted to give wrapArrayIndices its
own header file and install only that. Declaring functions in a header
that cannot be called feels wrong.

mdavidsaver:
So you intend `wrapArrayIndices()` to become a public API?

bfrk:
Am 01.04.21 um 20:35 schrieb mdavidsaver:
> So you intend `wrapArrayIndices()` to become a public API?

I was operating under the assumption that this is the only way this
function can be shared between the filters and dbCore. Was that wrong?

mdavidsaver:
Private headers should not be installed, but rather should be referenced though `USR_CPPFLAGS+=-I...`.

eg.
https://github.com/epics-base/epics-base/blob/c829ca39d873d59947d851a7b22d74d14748e336/modules/database/test/ioc/db/Makefile#L14-L15

bfrk:
> Private headers should not be installed, but rather should be referenced though `USR_CPPFLAGS=+-I...`.

Fine with me, I can do that. But I still need to include dbCoreAPI.h and
declare the function DBCORE_API, else I get problems on Windows, right?

mdavidsaver:
Yes.

bfrk:
Okay, done. If you ask me, declaring something as DBCORE_API without it
actually being part of the (public) API feels like lying to me. Perhaps
epicsShareFunc wasn't that bad a name after all.

bfrk:
> USR_CFLAGS_Linux += -fvisibility=hidden
> USR_CXXFLAGS_Linux += -fvisibility=hidden -fvisibility-inlines-hidden

BTW, many thanks for that hint, very useful. Is there any reason not to add that to the Linux/gcc config by default?

anjohnson:
> Is there any reason not to add that to the Linux/gcc config by default?

Getting the decorations right only matters for applications that need to be supported on Windows. Putting those flags in Base might cause unnecessary problems when building local EPICS applications, and I'd rather we didn't expose our users to the kinds of errors they can cause unnecessarily.

As a developer if you want these and other flags to always apply to your personal builds you can put them in your home/configure directory, I have this file in mine:

```makefile
$ cat ~/configure/CONFIG_USER.Common.linux-x86_64
# target-specific config for linux-x64 targets

# This is needed on RHEL
COMMANDLINE_LIBRARY = READLINE_NCURSES

OP_SYS_CFLAGS += -Wpointer-arith

ifdef BASE_3_15
  OP_SYS_CFLAGS += -fvisibility=hidden
  OP_SYS_CXXFLAGS += -fvisibility=hidden -fvisibility-inlines-hidden
endif

USE_POSIX_THREAD_PRIORITY_SCHEDULING = YES
```

See also `ENABLE_ASAN=YES` if you have a new enough gcc.

mdavidsaver:
> Base might cause unnecessary problems when building local EPICS applications

I can erase the "might" from this.  The site local code I've seen mostly doesn't bother with `epicsShare*`.

> ... I'd rather we didn't expose our users to the kinds of errors they can cause unnecessarily.

DItto.  It's hard enough to motivate sites to upgrade as it is.

bfrk:
> ... I'd rather we didn't expose our users to the kinds of errors they can cause unnecessarily.

Understood. Thanks for explaining.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -1332,25 +1345,33 @@ long dbPut(DBADDR *paddr, short dbrType,
...
     if (paddr->pfldDes->special == SPC_DBADDR &&
         prset && prset->get_array_info) {
-        long dummy;
 
-        status = prset->get_array_info(paddr, &dummy, &offset);
+        status = prset->get_array_info(paddr, &available_no_elements, &offset);
         /* paddr->pfield may be modified */
         if (status) goto done;
-        if (no_elements < nRequest)
-            nRequest = no_elements;
```
mdavidsaver, line 1360: I guess this bounds limit is now expected to be enforced elsewhere?

modules/database/src/ioc/db/dbAddrModifier.h
```diff
@@ -0,0 +1,48 @@
...
+
+/** @brief Structure of an address modifier.
+ *
+ * This will normally be allocated by user code. An address modifier
+ * implementation should supply a function to create the private data 'pvt'
+ * and initialize 'handle' with a function to handle the modifier.
+ */
+typedef struct {
+    void *pvt;                      /** @brief Private modifier data */
+    dbHandleAddrModifier *handle;   /** @brief Function to handle the modifier */
```
mdavidsaver, line 45: It looks like this function pointer will always be `&handleArrayRangeModifier` at present.  Are alternatives possible?  Maybe a premature abstraction?

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -54,6 +55,8 @@ typedef struct chFilter chFilter;
 typedef struct dbChannel {
     const char *name;
     dbAddr addr;              /* address structure for record/field */
+    dbAddrModifier addrModifier;
+                              /* optional: which indices are targeted */
```
mdavidsaver, line 59: Could `dbHandleAddrModifier` be embedding directly in `dbChannel`?  Seems like this could avoid the need for a new public `dbPutModifier()` and similar.

mdavidsaver (changes requested):
Looking again, I'm wondering about the design of `dbAddrModifier` and `dbHandleAddrModifier`, which are embedded in `dbChannel`, but also handled separately.  The inclusion of a function pointer in `dbAddrModifier` suggests an extension point.  However, limiting to a single function pointer gives no obvious way to handle chaining (as with Get/Monitor filters).  Adding this, perhaps insufficient, extension point now seems to me likely to complicate any later project to add filter-like chaining for Put.  This indirection also seems to be motivating much of the new public API, which might otherwise be avoided.

modules/database/test/std/rec/addrModifierTest.c
```diff
@@ -0,0 +1,223 @@
...
+    sprintf(fieldname, "%s.PROC", pv->name);
+    testdbPutFieldOk(fieldname, DBF_LONG, 1);
+    sprintf(fieldname, "%s.SEVR", pv->name);
+    testdbGetFieldEqual(fieldname, DBF_LONG, NO_ALARM);
+    sprintf(fieldname, "%s.STAT", pv->name);
+    testdbGetFieldEqual(fieldname, DBF_LONG, NO_ALARM);
+}
+
+static double initial[] = {0,1,2,3,4,5,6,7,8,9};
+static double buf[10];
```
mdavidsaver: Looks like this variable could become a local of `expectRange()`.

mdavidsaver (commented):
The added test coverage looks adequate.

modules/database/test/std/rec/addrModifierTest.c
bfrk: Keeping the declaration at the top level, right after that of `initial`, was deliberate because they have to have the same number of elements.

modules/database/src/ioc/db/dbAccess.c
bfrk: Yes, it is done now immediately before the call to `dbPutConvertRoutine` (for the other cases it is not needed).

modules/database/src/ioc/db/dbAddrModifier.h
bfrk: I tend to agree, but see my other reply regarding the extension point.

modules/database/src/ioc/db/dbChannel.h
bfrk: Again, see my general comment about the extension point.

bfrk:
I agree with your reservations about the extension point. I added that in response to a suggestion by Andrew, see the discussion right at the start of https://code.launchpad.net/~bfrk/epics-base/+git/epics-base/+merge/381628, in particular my reply to Andrew here: https://code.launchpad.net/~bfrk/epics-base/+git/epics-base/+merge/381628/comments/1002334

modules/database/test/std/rec/addrModifierTest.c
mdavidsaver: Have a look at `NELEMENTS(X)`, which is our name for the `sizeof(X)/sizeof((X)[0])` trick.

```c
#include <dbDefs.h>
...
    double buf[NELEMENTS(initial)];
```

mdavidsaver:
We'll have to find out what @anjohnson thinks of this idea now that you have made it concrete.

modules/database/test/std/rec/addrModifierTest.c
bfrk: Nice. Fixed in 72c5cf6a8e3

bfrk:
I am trying to rebase to resolve the conflicts. I first checked out 7.0 and did a git pull. Then I did `git rebase 7.0 address-modifiers`. As expected I am getting conflicts. However, when the rebase stops, prompting me to resolve conflicts, I also see this completely unrelated change:
```
diff --git a/.ci b/.ci
index 246428427..3db08b597 160000
--- a/.ci
+++ b/.ci
@@ -1 +1 @@
-Subproject commit 2464284271581a66cb92b704bd9b7c4039282b99
+Subproject commit 3db08b5977efd7a3ac8eecfa068b5bfdfdf84dd0
```
What should I do with that?

ralphlange:
That's an update (or not update) of `ci-scripts`.
(See https://github.com/epics-base/ci-scripts/commits/master)
I would not update `ci-scripts` as part of this PR unless it is required.

bfrk:
Ah, got it. I need to do `git pull --recurse-submodules` when updating my 7.0 branch. Now the diff is gone.

bfrk:
> We'll have to find out what @anjohnson thinks of this idea now that you have made it concrete.

Yes, that would be helpful.

anjohnson:
Core Group 2022-10-26: We like the idea of adding this functionality, but we are concerned about committing to API changes which are not extensible to support chaining, as that would probably complicate doing that significantly if/when we do come up with such a design.

mdavidsaver:
It occurs to me that since `struct dbChannel` is always allocated in Base, we could add hidden the additional members in super struct.  As is currently done with `struct dbCommonPvt`.

I think this would avoid public API changes, so the only user visible change would be the `.[]` syntax for channel names.  Which seems reasonable since we already support this as a special case for get/monitor.

PR #150, opened 2021-04-01
Use DBCORE_API

mdavidsaver:
Great big mechanical change from `epicsShare*` to `DBCORE_API`.  Some manual fixups after:

```sh
git ls-files modules/database/src/ioc|egrep '\.[hc]p*$' | xargs sed -i \
     -e 's|epicsShareFunc|DBCORE_API|g' \
     -e 's|epicsShareClass|DBCORE_API|g' \
     -e 's|epicsShareExtern|DBCORE_API extern|g' \
     -e 's|epicsShareDef\s*||g' \
     -e 's|shareLib\.h|dbCoreAPI.h|g' \
     -e 's|epicsShareAPI|epicsStdCall|g' \
     -e '/#define\s*epicsExportSharedSymbols/d'
```

### DIFF
--- a/modules/database/src/ioc/as/asCa.c
+++ b/modules/database/src/ioc/as/asCa.c
@@ -33,7 +33,6 @@
 #include "caerr.h"
 #include "caeventmask.h"
 
-#define epicsExportSharedSymbols
 #include "asCa.h"
 #include "asDbLib.h"
 #include "callback.h"
--- a/modules/database/src/ioc/as/asCa.h
+++ b/modules/database/src/ioc/as/asCa.h
@@ -14,17 +14,17 @@
 
 #include <stdio.h>
 
-#include "shareLib.h"
+#include "dbCoreAPI.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-epicsShareFunc void asCaStart(void);
-epicsShareFunc void asCaStop(void);
-epicsShareFunc int ascar(int level);
-epicsShareFunc int ascarFP(FILE *fp, int level);
-epicsShareFunc void ascaStats(int *pchans, int *pdiscon);
+DBCORE_API void asCaStart(void);
+DBCORE_API void asCaStop(void);
+DBCORE_API int ascar(int level);
+DBCORE_API int ascarFP(FILE *fp, int level);
+DBCORE_API void ascaStats(int *pchans, int *pdiscon);
 
 #ifdef __cplusplus
 }
--- a/modules/database/src/ioc/as/asDbLib.c
+++ b/modules/database/src/ioc/as/asDbLib.c
@@ -25,7 +25,6 @@
 
 #include "caeventmask.h"
 
-#define epicsExportSharedSymbols
 #include "asCa.h"
 #include "asDbLib.h"
 #include "callback.h"
--- a/modules/database/src/ioc/as/asDbLib.h
+++ b/modules/database/src/ioc/as/asDbLib.h
@@ -15,7 +15,7 @@
 #include <stdio.h>
 
 #include "callback.h"
-#include "shareLib.h"
+#include "dbCoreAPI.h"
 
 typedef struct {
     epicsCallback   callback;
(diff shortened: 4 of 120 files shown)


anjohnson (approved):
No objections, I was surprised several weeks back that we hadn't already done this.

mdavidsaver:
I also don't see this as a high risk change, so if the appveyor builds succeed (some time next week...) I'll go ahead and merge.

bfrk:
> Great big mechanical change from `epicsShare*` to `DBCORE_API`.
Sorry to disrupt the happy consensus here, but I don't think this is a
good idea. DBCORE_API *suggests* that this is a semantic annotation i.e.
that the item marked DBCORE_API is considered part of the external API.
But this is not the case and therefore misleading. If you want to share
a function between e.g. dbCore and dbRecStd, like I did in
https://github.com/epics-base/epics-base/pull/149, that is clearly an
internal implementation detail, you are nevertheless forced to mark it
DBCORE_API in the header file. IMO the old epicsShareFunc better
reflects its actual meaning and purpose.

mdavidsaver:
@bfrk It sounds like you are objecting to the name(s), and not the mechanism?  Should I take it that you have no objection to the idea of giving each library its own import/export macro?  This avoids most of the problems of `#include` ordering which make `epicsShare*` so tedious.  I hope you aren't suggesting that we go back to shifting `#include`s around trying to sort out strange linking errors?

As for the specific names (`LIBCOM_API`, `DBCORE_API`, `LIBCA_API`).  These are essentially implementation details, and could be changed with few external complications.  However, I hope you'll understand that I'm reluctant to do so.  This would mean a lot of churn for what I see as little benefit.

The specific point you make about meaning and private headers only applies to Base developers.  Users see only public headers, and in these will see functions so marked as the usable public API.  This seems unlikely to me to cause additional confusion.

anjohnson:
Even though it may be only an internal implementation detail that requires a routine to be exported, it's still an internal API since it's a call between two shared libraries/DLLs, thus I don't have a problem with the names `DBCORE_API` `LIBCA_API` `LIBCOM_API` and `DBRECSTD_API` for decorators. I agree that for internal APIs we should either put declarations in private header files that don't appear in the regular include path (although I don't particularly like the approach of adding the include flags to `USR_CPPFLAGS` and am wondering if we should install these headers into an `include/private` directory say that doesn't normally get searched), or require source files that need them to define a macro such as `EPICS_PRIVATE_xxx_API` to be able to see those declarations which is the other approach we have used.

You could always create a more explicit decorator like this in your header file if you really want to make the distinction more obvious:
```C
#include "dbCoreApi.h"
#define DBCORE_PRIVATE_API DBCORE_API
```

mdavidsaver:
> ... adding the include flags to `USR_CPPFLAGS` ...

I kind of like that this approach acts as a mild disincentive to overusing this pattern.

> ... install these headers into an `include/private` ...

To what end?  I don't like the idea of exposing Base internal details to users only to avoid further complicating already complicated Makefiles.

anjohnson:
> To what end?

Unbundling PCAS from Base-3.16 revealed that the CA Gateway [needed access to](https://github.com/epics-extensions/ca-gateway/blob/e54abf568aedf4c93bd994536b154c7253b82a78/src/Makefile#L68) the private header files `casCtx.h` and `caHdrLargeArray.h` from it, which [had never before been installed](https://github.com/epics-modules/pcas/commit/9caf4cf56f338dac32ecb856b9e3e9311a3a7dae). If we'd had something like an `include/private` install directory we could have avoided having to make those public. The CA Gateway would also have failed to build against a 3.15 or earlier Base that had `INSTALL_LOCATION` set since its Makefile assumed the PCAS sources would always be present.

However I'm not really suggesting we should actually add that. The (undiagnosed) OCD part of my brain just doesn't like to see a potentially compiler-specific flag `-I` used in generic Makefile variables inside the source tree (likewise it isn't very happy about `-D` flags, but it does accept that all our compilers implement both of them). It also balks on seeing more than two `../` components in a path, it thinks that's starting to embed too much knowledge about the overall structure of the source tree into the code (separation of concerns).

bfrk:
> @bfrk It sounds like you are objecting to the name(s), and not the mechanism?

Yes.

>  Should I take it that you have no objection to the idea of giving each library its own import/export macro?  This avoids most of the problems of `#include` ordering which make `epicsShare*` so tedious.  I hope you aren't suggesting that we go back to shifting `#include`s around trying to sort out strange linking errors?

I understand that there are valid reasons for separate macros for each
shared library. (I am still hazy about the details, but I guess that's
because reasoning about the CPP is inherently difficult.)

> As for the specific names (`LIBCOM_API`, `DBCORE_API`, `LIBCA_API`).  These are essentially implementation details, and could be changed with few external complications.  However, I hope you'll understand that I'm reluctant to do so.  This would mean a lot of churn for what I see as little benefit.

I would find it easier to buy this argument if you guys didn't do this
sort of thing on a regular basis for (IMHO) far more questionable
benefit (adapting copyright headers, yeah). Anecdotal evidence: when I
looked at the branches I proposed for merging a few months after I wrote
the code, I tried to get a feel for what changed in base in the
meantime. The sheer amount of files modified in mostly trivial ways made
that completely impossible.

> The specific point you make about meaning and private headers only applies to Base developers.  Users see only public headers, and in these will see functions so marked as the usable public API.  This seems unlikely to me to cause additional confusion.

External modules (e.g. the sequencer) also have to decorate exported
items if they are expected to work on Windows. It would be nice if EPICS
base continues to provide some mechanism that can be used by these modules.

ralphlange:
> I understand that there are valid reasons for separate macros for each shared library. (I am still hazy about the details, but I guess that's because reasoning about the CPP is inherently difficult.)

Each shared library that uses the mechanism has to treat its *own* symbols (that need to get exported) different from other libraries' symbols (that need to get imported).

ralphlange:
Most of the header file discussion dissolves if we leave the black and white thinking of public vs. private and recognize that there are different APIs for different consumers.
A simple three-section (or -tier) concept of user / system / private APIs fits our use case much better.
User APIs and system APIs have to be installed, private API headers (in the narrow meaning of "only used inside this module") don't.

I would put the system headers in a sub directory (makes it easier for packaging), but not add it to the compiler command line and have the other system modules instead
```C++
#include <system/casCtx.h>
```
which makes the "system" property of the header very obvious, at the point where it is being used.

bfrk:
Am 03.04.21 um 09:59 schrieb Ralph Lange:
> Most of the header file discussion dissolves if we leave the black and white thinking of public vs. private and recognize that there are different APIs for different consumers.
> A simple three-section (or -tier) concept of user / system / private APIs fits our use case much better.

I understand "user" and "private", but I find it hard to develop an
intuition for "system". Can you give a generic guide line for when to
mark an API as system? I understand from your example and the preceeding
discussion that the ca-gateway would be a consumer for the system tier
API but I find it difficult to generalise that example.

ralphlange:
> I understand "user" and "private", but I find it hard to develop an intuition for "system". Can you give a generic guide line for when to mark an API as system? I understand from your example and the preceeding discussion that the ca-gateway would be a consumer for the system tier API but I find it difficult to generalise that example.

Hm.
APIs that are neither user nor private?
APIs to be used by other support modules?
APIs that expose some level of internal implementation and likely need consumers to be changed when the implementation does?

EPICS is a software framework.
Software frameworks always have different user groups with different levels of interaction and knowledge. Looking at e.g. ITER, I can easily distinguish - with very little to no overlap:
Operator / IOC instantiator (DB creator) / IOC application developer / Support module developer / Base developer.
In the realm of headers/APIs I would map the last three groups to "user" / "system" / "private".

bfrk:
Am 03.04.21 um 10:46 schrieb Ralph Lange:
>> I understand "user" and "private", but I find it hard to develop an intuition for "system". Can you give a generic guide line for when to mark an API as system? I understand from your example and the preceeding discussion that the ca-gateway would be a consumer for the system tier API but I find it difficult to generalise that example.
> 
> Hm.
> APIs that are neither user nor private?
> APIs to be used by other support modules?
> APIs that expose some level of internal implementation and likely need consumers to be changed when the implementation does?

I am not convinced. You cannot provide a clear definition what "system"
level means. I think what you are trying to do here is misguided. An API
is either public or private, there is no such thing as "in between". An
API that is provided for just a single (legacy) consumer should not be
blessed by categorizing it as "system". Instead, it should be marked as
public, but /deprecated/.

ralphlange:
Objection!
Software frameworks, especially when using the [pImpl](https://en.cppreference.com/w/cpp/language/pimpl) technique, can have clearly distinguished APIs / headers for modules that use the framework and modules that implement some of the framework's interfaces.
The "user" APIs often don't require recompilation when the implementation changes, the "system" APIs always do.

mdavidsaver:
I'm going to go ahead and merge.  Discussions about macro naming, and varying degrees of "public", can of course continue.

> ... regular basis for (IMHO) ...

Many of the recent changes are of lines which hadn't been modified since the 90's.  I see these as down payments on ~3 decades of [technical debt](https://en.wikipedia.org/wiki/Technical_debt).

> It would be nice if EPICS base continues to provide some mechanism that can be used by these modules.

`shareLib.h` isn't going away.  It continues to be used in Base for simple cases (eg. `epicsExport`ing variables for device support).

PR #151, opened 2021-04-03
Fix (another) race in `db_close_events()`

mdavidsaver:
Attempt to fix race condition in `db_close_events()`.  Begins be reverting 37a76b433a9e7d5a8d26a13fd21ad62f20a0c1c1, then makes a simpler addition to join the worker on targets where this is possible.

https://bugs.launchpad.net/epics-base/+bug/1922442

Since this is not the first time I've fixed, and then un-fixed, this issue.  Please have a good close look...

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -61,6 +61,7 @@
 #include "recGbl.h"
 #include "recSup.h"
 #include "special.h"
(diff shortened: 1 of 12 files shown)


mdavidsaver:
Looks like I pushed too soon after merging #150 and this huge delta is included...

mdavidsaver:
Now #152

PR #152, opened 2021-04-03
Fix (another) race in db_close_events()

mdavidsaver:
Attempt to fix race condition in db_close_events(). Begins by reverting 37a76b4, then makes a simpler addition to join the worker on targets where this is possible.

https://bugs.launchpad.net/epics-base/+bug/1922442

Since this is not the first time I've fixed, and then un-fixed, this issue. Please have a good close look...

... duplicate of #151


mdavidsaver:
I've been running with this fix for the last few months, and haven't seen any ill effects.  I'm inclined to merge it this week.

mdavidsaver:
Applied as ca2ea14082bdadb3ea8f7b7ad967fd42b41e6a41

PR #153, opened 2021-04-05
Allow runtime bypass of freeListLib

mdavidsaver:
Allow the free list behavior to be bypassed/disabled at runtime by environment or iocsh variable.

Something I did to facilitate testing/exploring the effects of freeListLib.  Maybe useful to others.  I'm not sure if this should be merged though.

### DIFF
--- a/modules/libcom/src/freeList/freeList.h
+++ b/modules/libcom/src/freeList/freeList.h
@@ -29,6 +29,8 @@
 extern "C" {
 #endif
 
+LIBCOM_API extern int freeListBypass;
+
 LIBCOM_API void epicsStdCall freeListInitPvt(void **ppvt, int size, int malloc);
 LIBCOM_API void * epicsStdCall freeListCalloc(void *pvt);
 LIBCOM_API void * epicsStdCall freeListMalloc(void *pvt);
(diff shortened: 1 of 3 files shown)


mdavidsaver:
`EPICS_FREELIST_DEBUG` -> `EPICS_FREELIST_BYPASS`.

mdavidsaver:
Renamed environment variable.

Draft release note...

### Allow runtime bypass of free list allocator

The environment variable `$EPICS_FREELIST_BYPASS` may be set to `YES` to cause the `freeListLib` functions to always call directly to `malloc()`/`free()`.  May be useful when troubleshooting some kinds of memory allocation bugs which would otherwise be "hidden".  eg. use-after-free data races.  This may also improve the results of dynamic analysis tools which are not aware of this internal free list.

mdavidsaver:
I've encountered another situation where `freeListLib` was masking allocation bugs.  valgrind gave more accurate results with `freeListLib` bypassed, which makes me wonder whether its annotations need to be updated.

anjohnson:
Core Group 3/1/2023: Approved.

mdavidsaver:
Applied as eea361bf5e3971124c2cc82bd595cebff722b0fd

PR #154, opened 2021-04-05
Add DBR_VFIELD

mdavidsaver:
While not acceptable in it present form, I would like to get feedback before another iteration as I would like to see a solution merged.

The problem I'm trying to solve is how to usefully access complex types as database fields.  My immediate application is to the PVA data containers `shared_vector` and `PVStructure`, although there are others (eg. areaDetector's NDArray).

I'd like to be able to move these types through `dbGet()` and `dbPut()`, and by extension the derivative interfaces like links.

I want to do this in a way which would avoid allocating a new `DBF_*` code for every possible data type to avoid close coupling of currently external types with the database code.  My solution to this is to add one new `DBR_*` code, and use it to pass extra arguments around most of the existing code of `dbGet()` and `dbPut()`.  The logic to handle this is split between the caller, and record support.

So a caller would look like:

```c++
DBADDR *paddress = ...;
shared_vector<const void> aval;
struct VField {
    const VFieldType *vtype;
    shared_vector<const void> *aval;
} varg = {&vfSharedVector, &aval};
dbGet(paddress, DBR_VFIELD, NULL, &varg, NULL);
```

So in effect, `DBR_VFIELD` indicates that additional arguments are passed through the existing 'pbuffer' argument.  At present this ugliness is left exposed, though could eventually be hidden in a wrapper.

The other side of `dbGet()` is one of two new record support functions

```c++
long get_vfield(struct dbAddr *paddr, struct VField *p)
{
    stdstringinRecord *prec = (stdstringinRecord*)paddr->precord;

    if(p->vtype==&vfStdString) {
        VString *pstr = (VString*)p;
        if(dbGetFieldIndex(paddr)==stdstringinRecordVAL) {
            pstr->value = prec->val;
            return 0;
        } else if(dbGetFieldIndex(paddr)==stdstringinRecordOVAL) {
            pstr->value = prec->oval;
            return 0;
        }
    }
    return S_db_badChoice;
}
```

This PR supersedes https://github.com/mdavidsaver/epics-base/pull/1 now that the pre-requisite #71 is merged.

### DIFF
--- a/configure/CONFIG_DATABASE_MODULE
+++ b/configure/CONFIG_DATABASE_MODULE
@@ -26,3 +26,4 @@ MSI3_15                    = $(EPICS_DATABASE_HOST_BIN)/msi$(HOSTEXE)
 EPICS_BASE_IOC_LIBS = dbRecStd dbCore ca Com
 
 HAS_registerAllRecordDeviceDrivers=YES
+HAS_DBR_VFIELD=YES
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -634,6 +634,8 @@ long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
     paddr->field_size  = pflddes->size;
     paddr->special     = pflddes->special;
     paddr->dbr_field_type = mapDBFToDBR[dbfType];
+    paddr->ro = 0;
+    paddr->vfields = NULL;
 
     if (paddr->special == SPC_DBADDR) {
         const rset *prset = dbGetRset(paddr);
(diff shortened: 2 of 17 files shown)


mdavidsaver:
Current status:

* @anjohnson Suggests rename V -> Q https://github.com/mdavidsaver/epics-base/pull/1#discussion_r468964263
* Usage in QSRV https://github.com/mdavidsaver/pva2pva/tree/vfield
* Question about arguments for `rset::get_vfield`.  Include `db_field_log*`?
* I can't find a way to represent a PVA style delta with existing dbEvent queue.  So delta is limited to mapping from `DBE_*`.  (eg. can't indicate partial table update)

mdavidsaver:
@bfrk I would be interested (although slightly frightened ;) ) to having your opinion on this PR since you have been working in this area.

mdavidsaver:
Also, `stdstringinRecord` is intended primarily as demonstration/test.  It could be removed, or "demoted" to test code.

bfrk:
From a first cursory look it looks like a clever idea to me. It would help to have a bit of high-level documentation to get a feeling for how powerful this mechanism is. It would also help to have another somewhat more artificial example (a structure with say three fields of different scalar types) to compare to the stdstringinrecord example. As I understand it, with this we can transport arbitrary structures "atomically" i.e. with a single call to `dbGet` or `dbPut`. One thing I don't quite understand is this: I would have thought that at some point a DBR_VFIELD request has to iterate over the available fields, but I can't find the place where this happens.

PR #156, opened 2021-04-08
`dbChannel_get_count()`unlocked

mdavidsaver:
A change I tried while playing around with https://github.com/epics-base/epics-base/pull/100#issuecomment-813701869 .  While it didn't have much effect in that highly artificial case with a single client, it might be interesting from the prospective of scalability.  Omitting `dbScanLock()` would allow many clients to run `dbChannel_get_count()` in parallel, potentially mitigating "thundering herd" scenarios following a subscription update.

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -623,6 +623,7 @@ all_done:
 
 long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
 {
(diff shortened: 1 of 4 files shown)


mdavidsaver:
This change turns up a potential race w/ helgrind as `DBR_STATUS` triggers reads `ACKS` and `ACKT`.  These values are not used when `dbr <= oldDBR_TIME_DOUBLE`, so I think this is a false positive.

PR #157, opened 2021-04-16
allow glob pattern in var command

dirk-zimoch:


mdavidsaver:
Seems reasonable.  I think the code could be shortened by combining the two loops by adding `args[0].sval == NULL ||`.

```c
if(args[1].sval == NULL) {
        int found = 0;
        for (v = iocshVariableHead ; v != NULL ; v = v->next)
            if (args[0].sval == NULL || epicsStrGlobMatch(v->pVarDef->name, args[0].sval) != 0) {
                varHandler(v->pVarDef, NULL);
                found = 1;
            }
        if (args[0].sval != NULL && !found)
            fprintf(epicsGetStderr(), "No var matching %s found.\n", args[0].sval);
    }
```

dirk-zimoch:
Only that `var` with no arguments would fail with "No var matching (null) found." if no variables exist (which is unlikely).

mdavidsaver:
Ok, I've edited my previous post to add the condition in two places :)

anjohnson:
I was thinking just yesterday that it would be nice if `epicsEnvShow` could take a glob pattern, but (a) that would be a different PR to this one, and (b) we probably don't have a portable way to iterate through the set of environment variables anyway (I don't mean our list of envParams, which is different), which we'd need to use the `epicsStrGlobMatch()` function. Some platforms provide that ability (e.g. `extern char **environ`), but I haven't come across a standard way of doing it (unless C++ has a suitable API?).

dirk-zimoch:
Well, there is environ[] on Linux and VxWorks and _environ[] on Windows. I guess other OS are similar. I can try in a new PR next week...

dirk-zimoch:
Quickly checked VxWorks: environ is a macro from envLib.h but has basically the same functionality. All other implementations of osdEnv.c imply the existence of the eviron global variable.

anjohnson:
I didn't check before, but `epicsEnvShow()` is already defined per-OS in `libCom/src/osi/os/*/osdEnv.c` and it looks like only the VxWorks implementation isn't already using an `environ` or similar variable. The code for this routine isn't that big but this might be another situation where you could remove duplicate code and move the main part of the routine to a new osiEnv.c file. Note that the macOS implementation actually has to call a routine to get the `environ` pointer, maybe we could do something similar.

mdavidsaver (approved):
Looks ok to me.

mdavidsaver:
Applied as 555e6712467732e5e7a9985093b58738990bad6d

PR #158, opened 2021-04-20
caRepeater fallback to internal for all targets

mdavidsaver:
Instead of complaining if the `caRepeater` executable can't be found, silently fall back to an in-process repeater thread as is done for embedded targets.

Two preparatory changes are also made.

1. Fix the posix/ `osiSpawnDetachedProcess()` so the `execvp()` failure is actually reported.  Currently `osiSpawnDetachedProcessFail` is only returned if `fork()` fails.
2. Allow optional bypass of the warning printed by `osiSpawnDetachedProcess()` by prefixing the process name with a special character.  (I picked `'!'` to start)

The possibility came up in conversation with @thomascobb.

### DIFF
--- a/modules/libcom/src/osi/os/WIN32/osdProcess.c
+++ b/modules/libcom/src/osi/os/WIN32/osdProcess.c
@@ -50,9 +50,13 @@ LIBCOM_API osiGetUserNameReturn epicsStdCall osiGetUserName (char *pBuf, unsigne
 LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess 
     ( const char *pProcessName, const char *pBaseExecutableName )
 {
+    BOOL silent = pProcessName && pProcessName[0]=='!';
     BOOL status;
     STARTUPINFO startupInfo;
     PROCESS_INFORMATION processInfo;
+
+    if(silent)
+        pProcessName++; /* skip '!' */
 
     GetStartupInfo ( &startupInfo );
     startupInfo.lpReserved = NULL;
@@ -115,7 +119,7 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
             /* Free the buffer. */
             LocalFree (errStrMsgBuf);
         }
-        else {
+        else if(!silent) {
             fprintf (stderr, "!!WARNING!!\n");
             fprintf (stderr, "Unable to locate executable \"%s\".\n", pBaseExecutableName);
             fprintf (stderr, "You may need to modify your \"path\" environment variable.\n");
--- a/modules/libcom/src/osi/os/posix/osdProcess.c
+++ b/modules/libcom/src/osi/os/posix/osdProcess.c
@@ -63,7 +63,11 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
     (const char *pProcessName, const char *pBaseExecutableName)
 {
     int status;
+    int silent = pProcessName && pProcessName[0]=='!';
     int fds[2]; /* [reader, writer] */
+
+    if(silent)
+        pProcessName++; /* skip '!' */
 
     if(pipe(fds))
         return osiSpawnDetachedProcessFail;
@@ -137,7 +141,7 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
      * Run the specified executable
      */
     status = execlp (pBaseExecutableName, pBaseExecutableName, (char *)NULL);
-    if ( status < 0 ) {
+    if ( status < 0 && !silent ) {
         fprintf ( stderr, "**** The executable \"%s\" couldn't be located\n", pBaseExecutableName );
         fprintf ( stderr, "**** because of errno = \"%s\".\n", strerror (errno) );
         fprintf ( stderr, "**** You may need to modify your PATH environment variable.\n" );
--- a/modules/libcom/src/osi/os/posix/osdProcess.c
+++ b/modules/libcom/src/osi/os/posix/osdProcess.c
@@ -24,7 +24,9 @@
 #include <unistd.h>
 #include <sched.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <pwd.h>
+#include <fcntl.h>
 
 #include "osiProcess.h"
 #include "errlog.h"
@@ -61,12 +63,18 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
     (const char *pProcessName, const char *pBaseExecutableName)
 {
     int status;
+    int fds[2]; /* [reader, writer] */
+
+    if(pipe(fds))
+        return osiSpawnDetachedProcessFail;
 
     /*
      * create a duplicate process
      */
     status = fork ();
     if (status < 0) {
+        close(fds[0]);
+        close(fds[1]);
         return osiSpawnDetachedProcessFail;
     }
 
@@ -75,11 +83,29 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
      * in the initiating (parent) process
      */
     if (status) {
-        return osiSpawnDetachedProcessSuccess;
-    }
+        osiSpawnDetachedProcessReturn ret = osiSpawnDetachedProcessSuccess;
+        char buf;
+        ssize_t n;
+        close(fds[1]);
 
-    /*
-     * This is executed only by the new child process.
+        n = read(fds[0], &buf, 1);
+        /* Success if child exec'd without sending a '!'.
+         * Of course child may crash soon after, but can't
+         * wait around for this to happen.
+         */
+        if(n!=0) {
+            ret = osiSpawnDetachedProcessFail;
+        }
+
+        close(fds[0]);
+        return ret;
+    }
+    close(fds[0]);
+    (void)fcntl ( fds[1], F_SETFD, FD_CLOEXEC );
+
+    /* This is executed only by the new child process.
+     * Since we may be called from a library, we don't assume that
+     * all other code has set properly set FD_CLOEXEC.
      * Close all open files except for STDIO, so they will not
      * be inherited by the new program.
      */
@@ -89,6 +115,8 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
             if (fd==STDIN_FILENO) continue;
             if (fd==STDOUT_FILENO) continue;
             if (fd==STDERR_FILENO) continue;
+            /* pipe to our parent will be closed automatically via FD_CLOEXEC */
+            if (fd==fds[1]) continue;
             close (fd);
         }
     }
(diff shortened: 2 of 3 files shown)


anjohnson:
This might have undesirable consequences, so we should consider possible side-effects before merging.

Assuming I've read the code properly: If the child process can't find the `caRepeater` executable, the repeater will then be running inside the main client process, and thus will exit when it does. On machines that normally run some combination of long-running GUI screens and cron jobs to run some periodic but short CA activity the repeater might get picked up by the cron clients, which would cause problems with the GUIs that want to see the beacons which the repeater forwards. There could be long periods when there is no repeater running.

I always recommend explicitly running the repeater on any production machine, and presumably nowadays that would be started and monitored by systemd. If the repeater dies systemd should immediately restart it, but the repeater process will immediately exit if some CA client has managed to take ownership of the UDP socket first, which might have interesting effects on the systemd auto-restart. Using auto-restart should prevent the issue in the paragraph above, as long as systemd never gives up retrying.

It might be better if the child process were to run the repeater in the event that it can't find a `caRepeater` executable instead of the main process doing that, so once started it won't die with the client process. If we do that I would also want it to change its process name (as shown by `ps`) if possible, so a user won't think they have a rogue duplicate client process that they might then kill. Both of these would need changes to the libCom APIs.

mdavidsaver:
> This might have undesirable consequences, so we should consider possible side-effects before merging.

Oh yes.  Although I think the worst case is the same as no repeater.  I know I've killed `caRepeater` without apparent ill effects to running CA clients.

> I always recommend explicitly running the repeater on any production machine

The motivating use case a more self contained installation.  Think `$HOME` or some minimal Linux container.

If this change is not acceptable in Base, I can keep it in epicscorelibs.

ralphlange:
I like the idea to remove weight from minimal apps running in containers.

What about making it opt-in (e.g., through setting an env variable)?

mdavidsaver:
I think opt-in would work.

@anjohnson How about checking that `$EPICS_CA_REPEATER_FALLBACK` is `"THREAD"`?

anjohnson:
Adding a knob that can cause bad behaviour if set wrong probably isn't a great idea, depends how bad it is. Running the repeater in the child process if it can't find a program to exec would be pretty close to what we currently do, except it keeps all the FDs and memory inherited from the process it forked from. It could close FDs, but probably can't do anything about memory in that case. I think that ought to be acceptable if it can change its visible process name though.

CA clients can work fine without a repeater when you're doing development and test stuff, but I wouldn't want to run that way in a production control-room setting where clients run for weeks on end and rebooted IOCs should automatically reconnect reasonably quickly, so the clients need to be able to see the beacons from the IOCs.

How do the Java clients behave when there's no repeater running? I don't think they provide one of their own do they? Similarly what about caProto?

mdavidsaver:
> Running the repeater in the child process ...

I guess you mean to just `fork()` w/o an exec?  This did occur to me.  As I understand it, this is usually discouraged when the process might have more than one thread.

> ... a production control-room ...

Remember that we are talking about a fallback here.  If `caRepeater` is already running, as I'm certain is the case for all APS systems, then this is a non-issue.

> How do the Java clients behave when there's no repeater running?

CAJ provides its own repeater implementation, which is started opportunistically.  As do other CA implementations.  The repeater functionality isn't complicated.

```
$ netstat -ulpn|grep 5065
udp        0      0 0.0.0.0:5065            0.0.0.0:*                           13231/java          
$ cat /proc/13231/cmdline 
/usr/lib/jvm/java-11-openjdk-amd64/bin/java-classpath/home/mdavidsaver/source/phoebus/phoebus-product/target/product-4.6.6-SNAPSHOT.jarcom.cosylab.epics.caj.CARepeater-p5065
```

mdavidsaver:
@anjohnson If we can't immediately agree on the entire PR, can you comment on the first commit?  I see this as a bug fix so that `osiSpawnDetachedProcess()` on posix targets will correctly report `execvp()` failure to the caller.

modules/libcom/src/osi/os/posix/osdProcess.c
```diff
@@ -89,6 +113,7 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
             if (fd==STDIN_FILENO) continue;
             if (fd==STDOUT_FILENO) continue;
             if (fd==STDERR_FILENO) continue;
+            if (fd==fds[1]) continue;
```
anjohnson, line 123: Can you update the comment block above to match please, since you're excluding another FD from being closed.

modules/libcom/src/osi/os/posix/osdProcess.c
```diff
@@ -115,6 +140,9 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
         fprintf ( stderr, "**** You may need to modify your PATH environment variable.\n" );
         fprintf ( stderr, "**** Unable to start \"%s\" process.\n", pProcessName);
     }
+    /* signal error to parent */
+    ssize_t ret = write(fds[1], "!", 1);
+    (void)ret; /* not much we could do about this */
```
anjohnson, line 152: Should the child process call `close(fds[1])` here?

modules/libcom/src/osi/os/posix/osdProcess.c
```diff
@@ -75,8 +83,24 @@ LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess
...
+        n = read(fds[0], &buf, 1);
+        /* Success if child exec'd without sending a '!'.
+         * Of course child may crash soon after, but can't
+         * wait around for this to happen.
+         */
+        if(n!=0) {
+            ret = osiSpawnDetachedProcessFail;
+        }
+
+        return ret;
```
anjohnson, line 105: Should the parent process call `close(fds[0])` here?

modules/libcom/src/osi/os/WIN32/osdProcess.c
```diff
@@ -50,10 +50,14 @@ LIBCOM_API osiGetUserNameReturn epicsStdCall osiGetUserName (char *pBuf, unsigne
 LIBCOM_API osiSpawnDetachedProcessReturn epicsStdCall osiSpawnDetachedProcess 
     ( const char *pProcessName, const char *pBaseExecutableName )
 {
+    BOOL silent = pProcessName[0]=='!';
```
anjohnson: Since `osiSpawnDetachedProcess()` is a public API, should this instead be:
```C
BOOL silent = pProcessName && pProcessName[0]=='!';
```
Windows permits a NULL process name [`STARTUPINFO.lpTitle`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-startupinfoa)

anjohnson (changes requested):
How about requiring `$EPICS_CA_REPEATER_AS_THREAD` to be `ALLOWED`? If the variable is created as an `envParam` a site/build would be also able to configure its own compiled-in default without having to set the environment variable, but I'm not sure if that's desirable in the use-case you're thinking about.

A few code questions inline.

Also needs appropriate updates to the CA Reference manual for the new variable and behaviour, at least [here](https://epics.anl.gov/base/R7-0/5-docs/CAref.html#EPICS) and [here](https://epics.anl.gov/base/R7-0/5-docs/CAref.html#Repeater).

modules/libcom/src/osi/os/posix/osdProcess.c
mdavidsaver: oops, yes.

modules/libcom/src/osi/os/posix/osdProcess.c
mdavidsaver: The child process is about to `_exit()`, which will implicitly close all FDs.  I've added an explicit `close()` anyway.

mdavidsaver:
> Also needs appropriate updates to the CA Reference ...

I find myself wondering whether the effort of documenting and educating the epics community would be more or less than the effort needed to convince Andrew than this is a low risk change...

@anjohnson Have you thought any more about what tests I might do to help ease you mind?

modules/libcom/src/osi/os/posix/osdProcess.c
anjohnson: Your new wording is good, but the part I really wanted changing was `Close all open files except for STDIO,` – please add `and the pipe` (or remove that sentence completely).

anjohnson (commented):
Sorry if I wasn't clear: I'll be happy for this to be implemented as an opt-in conditional on the env. variable `EPICS_CA_REPEATER_AS_THREAD` being `ALLOWED`, as long as the use of the variable is described in the CA Reference manual (and mentioned in the Release Notes). As I also said above, if you use an envParam a site can give it a default in configure/CONFIG_SITE_ENV but I'm not sure if that makes sense for this particular option so that's just an idea for you to think about.

mdavidsaver:
Ok, to take a further step back.  The situation which Andrew is worried about, as I understand it, is when the first CA client is short lived, and a second client starts while the first is running.  If the second client remains running after the first has exited, and the repeater is tied to the lifetime of the first client, the second will be left without a repeater.

@anjohnson Is this correct?

anjohnson:
@mdavidsaver That was the simplest case that I came up with.

With the current code when there is no repeater running all subsequent new clients will try to run one, and will warn the user if they can't find it. Assuming the user ignores those warnings, as long as at least one client has the repeater in its path the service will eventually get started as its own process that will out-live the client that started it. Nowadays it will likely get killed by systemd when that user logs out, but until then everything should work fine.

If every new client can run the repeater in a thread the service will bounce between short-lived clients, and the long-lived clients will keep reconnecting to the new services, but unless a long-lived client gets (re)started while no such thread is running there will be no long-running repeater service, and the user will get no feedback about the problem. That could give CA a reputation for having reconnection issues.

mdavidsaver:
Can you elaborate on any other situations?  I can't realistically assuage a vague sense of discomfort, or reputational fear, by technical argument.

I think I can demonstrate that, in the situation I describe, on *nix targets at present there is no practical difference between having caRepeater launched by `osiSpawnDetachedProcess()` vs. a repeater thread.  `osiSpawnDetachedProcess()` on *nix has never lived up to the "Detached" part of the name, so the `caRepeater` process launched by the first short lived client will be killed when that client exits.

anjohnson:
I suspect most users run a lot more clients that exit normally than get killed, and in those cases the Repeater does survive. Even running `caget` starts one, thus I dispute your proposed demonstration:

```sh
woz$ netstat -a -n | grep 5065
woz$ caRepeater &
[1] 39302
woz$ netstat -a -n | grep 5065
udp4       0      0  *.5065                 *.*                               
woz$ kill %1
[1]+  Terminated: 15          caRepeater
woz$ caget hello-world
Channel connect timed out: 'hello-world' not found.
woz$ netstat -a -n | grep 5065
udp4       0      0  *.5065                 *.*                               
woz$ ps -ef | grep caRepeater
 1245 39382     1   0 11:53AM ttys002    0:00.01 caRepeater
 1245 39456   725   0 11:59AM ttys002    0:00.00 grep caRepeater
woz$ 
```

Thus while I'll let you make it, I'm going to insist that this change be documented in the Release Notes since it's an easily observable change. I'm not asking you to give lots of detail in the entry, but there needs to be something marking it so users who might be affected have something they can identify as a change that we made.

mdavidsaver:
Applied as 08b741ed056297d22665d576df6b7a4f0d5c1e8a

PR #159, opened 2021-04-20
Support glob pattern for epicsEnvShow

dirk-zimoch:
Tested on RHEL-7, vxWorks 5.5, vxWorks 6.9 and Windows (using wine).

### DIFF
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -259,35 +259,40 @@ size_t epicsStrnLen(const char *s, size_t maxlen)
     return i;
 }
 
-int epicsStrGlobMatch(const char *str, const char *pattern)
+int epicsStrnGlobMatch(const char *str, size_t len, const char *pattern)
 {
-    const char *cp = NULL, *mp = NULL;
+    const char *mp = NULL;
+    size_t cp = 0, i = 0;
 
-    while ((*str) && (*pattern != '*')) {
-        if ((*pattern != *str) && (*pattern != '?'))
+    while ((i < len) && (str[i]) && (*pattern != '*')) {
+        if ((*pattern != str[i]) && (*pattern != '?'))
             return 0;
         pattern++;
-        str++;
+        i++;
     }
-    while (*str) {
+    while ((i < len) && str[i]) {
         if (*pattern == '*') {
             if (!*++pattern)
                 return 1;
             mp = pattern;
-            cp = str+1;
+            cp = i+1;
         }
-        else if ((*pattern == *str) || (*pattern == '?')) {
+        else if ((*pattern == str[i]) || (*pattern == '?')) {
             pattern++;
-            str++;
+            i++;
         }
         else {
             pattern = mp;
-            str = cp++;
+            i = cp++;
         }
     }
     while (*pattern == '*')
         pattern++;
     return !*pattern;
+}
+
+int epicsStrGlobMatch(const char *str, const char *pattern) {
+    return epicsStrnGlobMatch(str, (size_t)-1, pattern);
 }
 
 char * epicsStrtok_r(char *s, const char *delim, char **lasts)
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -23,6 +23,15 @@
 #include <errno.h>
 #include <ctype.h>
 
+#ifndef vxWorks
+#include <stdint.h>
+#else
+/* VxWorks automaticaly includes stdint.h defining SIZE_MAX in 6.9 but not earlier */
+#ifndef SIZE_MAX
+#define SIZE_MAX (size_t)-1
+#endif
+#endif
+
 #include "epicsAssert.h"
 #include "epicsStdio.h"
 #include "cantProceed.h"
@@ -292,7 +301,7 @@ int epicsStrnGlobMatch(const char *str, size_t len, const char *pattern)
 }
 
 int epicsStrGlobMatch(const char *str, const char *pattern) {
-    return epicsStrnGlobMatch(str, (size_t)-1, pattern);
+    return epicsStrnGlobMatch(str, SIZE_MAX, pattern);
 }
 
 char * epicsStrtok_r(char *s, const char *delim, char **lasts)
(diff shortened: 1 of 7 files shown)


modules/libcom/src/misc/epicsString.h
```diff
@@ -36,6 +36,7 @@ LIBCOM_API char * epicsStrnDup(const char *s, size_t len);
 LIBCOM_API int epicsStrPrintEscaped(FILE *fp, const char *s, size_t n);
 #define epicsStrSnPrintEscaped epicsStrnEscapedFromRaw
 LIBCOM_API size_t epicsStrnLen(const char *s, size_t maxlen);
+LIBCOM_API int epicsStrnGlobMatch(const char *str, size_t len, const char *pattern);
 LIBCOM_API int epicsStrGlobMatch(const char *str, const char *pattern);
```
anjohnson, line 50: Could you add Doxygen comments here for the new routine (see `epicsStrSimilarity()` lower down if you need an example), and maybe for the old one too while you're at it. The `@since` keyword for the new API should say `@since UNRELEASED` and will be populated by whoever makes the next release.

The wording from the AppDevGuide about original routine could be used, it said just:
> `epicsStrGlobMatch` returns non-zero if `str` matches the shell wild-card `pattern`.

modules/libcom/src/misc/epicsString.c
```diff
@@ -259,37 +259,42 @@ size_t epicsStrnLen(const char *s, size_t maxlen)
...
+            i = cp++;
         }
     }
     while (*pattern == '*')
         pattern++;
     return !*pattern;
 }
 
+int epicsStrGlobMatch(const char *str, const char *pattern) {
+    return epicsStrnGlobMatch(str, (size_t)-1, pattern);
```
anjohnson: The `(size_t)-1` here gives off a bad smell.

C99 requires stdint.h to define `SIZE_MAX` which IMHO would be the best solution here, but we never used stdint.h in EPICS because VxWorks didn't support C99 until VxWorks 6. I believe you are the last EPICS developer who has been trying to build the EPICS 7 core on VxWorks 5.5 – are you still doing that, or have you given it up?

modules/libcom/src/osi/os/vxWorks/osdEnv.c
```diff
@@ -86,14 +87,12 @@ LIBCOM_API void epicsStdCall epicsEnvUnset (const char *name)
...
-    if (name == NULL) {
-        envShow (0);
-    }
-    else {
-        const char *cp = getenv (name);
-        if (cp == NULL)
-            printf ("%s is not an environment variable.\n", name);
-        else
-            printf ("%s=%s\n", name, cp);
+    extern char **ppGlobalEnviron;
```
anjohnson: Might be worth a comment: `/* Used in 'environ' macro but not declared in envLib.h */`

anjohnson (changes requested):
Overall this looks very promising, thanks for working on it.

There are a series of tests for the `epicsStrGlobMatch()` routine in modules/libcom/test/epicsStringTest.c which I assume all still pass after your changes. We also need tests for the new API, to show that `len` interacts properly with the `str` and `pattern` parameters. Please extend the tests for that.

Some additional comments and requests inline below.

modules/libcom/src/misc/epicsString.h
dirk-zimoch: Two similar Doxygen comments or one for both, `epicsStrGlobMatch` and `epicsStrnGlobMatch`? (If that is even possible?)

modules/libcom/src/misc/epicsString.c
dirk-zimoch: I am still using vxWorks 5.5.
I could use `~0L` or define SIZE_MAX if it is not available otherwise. Or make a stdint.h header file for vxWorks 5.

dirk-zimoch: Hm... not even VxWorks 6.7 finds stdint.h. It works with 6.9.
VxWorks 6.7 indeed has stdint.h, twice in fact:

- diab/5.7.0.0/include/cnew/stdint.h
- vxworks-6.7/target/usr/h/stdint.h

But I use the gnu compiler, not the diab and target/usr/h is not in the include path.
I tried to modify configure/os/CONFIG.Common.vxWorksCommon but that creates conflicts with other files like vxTypes.h.

I think stdint.h is not usable before vxWorks 6.9 (or maybe 6.8 which I don't have installed).

modules/libcom/src/misc/epicsString.h
mdavidsaver: Please avoid duplication.  eg. associate the bulk of the content with `epicsStrGlobMatch()`, with a short description of the difference and a `@see` reference for `epicsStrnGlobMatch()`.

modules/libcom/src/misc/epicsString.c
mdavidsaver: MSVC has been the other holdout for C99.  I guess vxworks is unlikely to have the POSIX alternative `inttypes.h`.  Personally I still use `(size_t)-1` in C code.  I also see an incorrect definition in `epicsSingletonMutex.cpp`, although present usage is inconsequential.
https://github.com/epics-base/epics-base/blob/b00130493fe695fad623e509bda9d507fccec019/modules/libcom/src/cxxTemplates/epicsSingletonMutex.cpp#L22-L24

anjohnson: Maybe epicsTypes.h should `#include <stdint.h>` except on VxWorks <= 6.7 and before MSVC started, where we would define and document the things that we rely on, initially just `SIZE_MAX`? Eventually I can see us redefining some the epics types in terms of the stdint.h types, once we insist on C99 compiler support (EPICS 7.1).

dirk-zimoch: I found this today:
../../src/factory/FieldCreateFactory.cpp:577: warning: converting of negative value `-0x00000000000000001' to `size_t'
So it is not unusual to cast -1 to size_t. At least I do it explicitly :-)

dirk-zimoch: Turns out that VxWorks 6.9 automatically includes stdint.h via vxWorks.h which is included using a compiler option in all vxWorks builds. Thus checking if SIZE_MAX is defined for vxWorks is sufficient and no complicated version check is required.

dirk-zimoch:
Tests added, doxygen added, `SIZE_MAX` used,  comment added to `ppGlobalEnviron`.

mdavidsaver (approved):
Looks ok to me.

@anjohnson Do you want to do any vxworks testing prior to merge?

anjohnson (approved):
I don't need to test this on VxWorks myself since Dirk has already done that. Approve.

mdavidsaver:
Applied as b777233efb06fa4e988c4f0738b0270dd3d095a3

PR #163, opened 2021-04-30
Improved Validation and Error Messages for RELEASE files.

hhslepicka:
- Added error message for when EPICS_BASE is not set on templates.
- Added error message and validation at convertRelease.pl for empty RELEASE files.

I came across this issue when I tried to build a module which depends only on EPICS_BASE and I didn't have it defined on my RELEASE.local file but it was defined on my environment variables.

### DIFF
--- a/src/template/base/top/configure/CONFIG
+++ b/src/template/base/top/configure/CONFIG
@@ -13,6 +13,11 @@ ifdef T_A
 -include $(TOP)/configure/RELEASE.$(EPICS_HOST_ARCH).$(T_A)
 endif
 
+# Check for proper EPICS_BASE
+ifneq (file,$(origin EPICS_BASE))
+ $(error EPICS_BASE must be defined in configure/RELEASE or a similar file or it will not be correctly read by convertRelease.pl!)
+endif
+
 CONFIG = $(RULES)/configure
 include $(CONFIG)/CONFIG
 
--- a/src/template/base/top/configure/CONFIG
+++ b/src/template/base/top/configure/CONFIG
@@ -8,9 +8,10 @@ RULES = $(EPICS_BASE)
 # RELEASE files point to other application tops
 include $(TOP)/configure/RELEASE
 -include $(TOP)/configure/RELEASE.$(EPICS_HOST_ARCH).Common
+
 ifdef T_A
--include $(TOP)/configure/RELEASE.Common.$(T_A)
--include $(TOP)/configure/RELEASE.$(EPICS_HOST_ARCH).$(T_A)
+  -include $(TOP)/configure/RELEASE.Common.$(T_A)
+  -include $(TOP)/configure/RELEASE.$(EPICS_HOST_ARCH).$(T_A)
 endif
 
 # Check EPICS_BASE is set properly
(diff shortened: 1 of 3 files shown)


anjohnson (approved):
Looks good; I may modify the error messages and Perl code slightly when I merge this but I get what it's doing and why.

Thanks!

hhslepicka:
Awesome! Thanks for the review, @anjohnson.

hhslepicka:
@anjohnson news on this one?

anjohnson:
Hi Hugo, I'm working on testing and merging this right now.

Base itself doesn't have a `configure/RELEASE` file, and when building your branch I saw several `No variables defined in RELEASE*s` errors show up since the Perl script runs `releaseTops()` many times during Base builds. Thus I'm going to remove that check, although doing one in `checkRelease()` still makes sense for modules that don't have the new `configure/CONFIG` file. I'm changing that check to just look for `EPICS_BASE` though, and I'm adding additional checks that the file `$(EPICS_BASE)/configure/CONFIG_BASE` exists too, proving that `EPICS_BASE` points to a real Base installation.

I just pushed my changes back to your branch so the CI-tests can run, and will merge the result once they pass, unless you have any other comments or suggestions.

- Andrew

hhslepicka:
Thank you for looking into this, @anjohnson!
I have no comments or suggestions.

mdavidsaver:
Followup to this morning's discussion with @anjohnson.  I hadn't appreciated that this change was to a template/, although this fact was staring me full in the face.  So this change will take some time to percolate out to user code.

For the record, a related/overlapping change I'm testing which would be more immediate:

```diff
diff --git a/configure/CONFIG b/configure/CONFIG
index fab3c245f..ed7f03151 100644
--- a/configure/CONFIG
+++ b/configure/CONFIG
@@ -30,6 +30,11 @@ ifeq ($(origin EPICS_HOST_ARCH), undefined)
   EHA :=
 endif
 
+ifneq ($(origin EPICS_BASE),file)
+$(error EPICS_BASE must be defined in configure/RELEASE or a similar file or it will \
+ not be correctly read by convertRelease.pl)
+endif
+
 -include $(CONFIG)/RELEASE
 -include $(CONFIG)/RELEASE.$(EPICS_HOST_ARCH)
 -include $(CONFIG)/RELEASE.$(EPICS_HOST_ARCH).Common
```

anjohnson:
Given that the template change isn't functionally necessary it wouldn't surprise me if it _never_ gets into some support modules and IOC applications.

I'll be happy to commit the change below – the error message doesn't need to provide an explanation, but if someone goes to disable the `$(error ...)` they'll see the comment that tells them more:
```patch
diff --git a/configure/CONFIG b/configure/CONFIG
index fab3c245f..a4e5c5b7c 100644
--- a/configure/CONFIG
+++ b/configure/CONFIG
@@ -16,6 +16,10 @@ ifneq ($(wildcard $(TOP)/configure/CONFIG_BASE_VERSION),)
   CONFIG = $(TOP)/configure
   BASE_TOP=YES
 else
+  ifneq ($(origin EPICS_BASE),file)
+    # Essential for the EPICS build system, see convertRelease.pl
+    $(error EPICS_BASE must be set in a configure/RELEASE file)
+  endif
   CONFIG ?= $(EPICS_BASE)/configure
 endif
 
```
The result looks like this:
```
woz$ make
/Users/anj/Software/epics/base-7.0/configure/CONFIG:21: *** EPICS_BASE must be set in a configure/RELEASE file.  Stop.
woz$
```

PR #165, opened 2021-05-03
Proper CRLF handling

tboegi:
Add proper CLRF-LF rules to .gitattributes.
Renormalie the bat files to LF in the repo and CRLF on disk.


mdavidsaver:
cf. https://git-scm.com/docs/gitattributes#_end_of_line_conversion

It looks like only the two .bat files are effected.

.gitattributes
```diff
@@ -4,3 +4,10 @@
 .appveyor/    export-ignore
 .appveyor.yml export-ignore
 README        export-subst
+
+#Which files need CRLF handling
+* text=auto
+*.pl text eol=lf
+*.sh text eol=lf
+*.bat text eol=crlf
+*.cmd -text
```
mdavidsaver: What is the motivation for treating `*.cmd` as binary?  Should this be `eol=lf`?

tboegi: -text stands for "don't treat line endings"
binary means "don't treat line endings - don't try to diff either"

There are unfortunatly 2 types of .cmd files:
a) Windows-ish (like .bat)
b) EPICS-ish (like st.cmd)
Don't mess with the line endings here, just to be sure. 

(This should probably go into the commit message)

mdavidsaver: Also, if `.pl` need special handling, what about `.pm` and `.plt`, or `.pod`?

I understand that `.bat` must be crlf since the windows batch file parser is quite picky.  What about the others?  It would be good to document the motivations.

anjohnson: Any why fix `*.pl` files to use LF (without also fixing `*.pm` and `*.plt` which also contain Perl code)? The Windows versions of Perl properly understands CRLF line-endings. The `*.sh` setting here also seems pointless to me.

mdavidsaver:
```sh
$ git ls-files|grep 'cmd$'
modules/libcom/test/iocshTestBadArg.cmd
modules/libcom/test/iocshTestBadArgIndirect.cmd
modules/libcom/test/iocshTestSuccess.cmd
modules/libcom/test/iocshTestSuccessIndirect.cmd
```

We currently have only iocsh scripts with `.cmd`.

tboegi: OK @mdavidsaver @anjohnson : Thanks for the reviews.
I'll send a new patch tomorrow.

FreddieAkeroyd:
As a windows developer, I would prefer `*.cmd` as `text` so it gets CRLF on windows,  IOCs are happy and it allows the file to work in windows editors that don't like just LF. Restricting `*.sh` to LF could be useful, if you have done a checkout in a folder using windows git and then access that same folder from cygwin/wsl then the `#!` in a shell script doesn't work if it has CRLF  (mingw64 seems more forgiving and looks to work with either line ending).

anjohnson:
@FreddieAkeroyd Thanks for the information about the `*.sh` pattern and the `#!` line.

The `startup/unix.sh` file is the only shell script that might apply to but it doesn't have a `#!` line (it must be sourced anyway, not run), and the 2 other `.sh` files in the epics-base repo are both developer-only scripts. There is a `#!/bin/sh` line in `startup/EpicsHostArch` but that filename doesn't match the pattern. Thus I don't see us needing a `*.sh` pattern in the `.gitattributes` file at all.

FreddieAkeroyd:
Actually, it seem that `sh` on the file also suffers a problem with line endings, I'd normally seen it on the `#!` part but it look it applies to the rest of the file too. So I think an `*.sh` pattern to always use LF would help in a few special cases, and also having `startup/EpicsHostArch` set to LF as you mention

mdavidsaver:
Ah, right.  Bash (or whatever) may be quite happy with CRLF, but the Linux kernel is only [looking for `'\n'`](https://github.com/torvalds/linux/blob/5e321ded302da4d8c5d5dd953423d9b748ab3775/fs/binfmt_script.c#L59).  So this is like the limitation of only one argument on a `#!` line.

```sh
$ hexdump -C test1.sh 
00000000  23 21 2f 62 69 6e 2f 73  68 0d 0a 65 63 68 6f 20  |#!/bin/sh..echo |
00000010  68 65 6c 6c 6f 0d 0a                              |hello..|
00000017
$ ./test1.sh 
bash: ./test1.sh: /bin/sh^M: bad interpreter: No such file or directory
```

Results in an attempt to read `"/bin/sh\r"`.

While strictly speaking it is only executable scripts which need `eol=lf`, it doesn't seem worthwhile trying to special case all of these.  So I'll accept `eol=lf` of all extensions which may potentially be executable scripts (including `.py`).

mdavidsaver:
Finding the extensions of all scripts intended to be executable.

```sh
$ git grep -l '^#!'| grep -o '\.[^.]*$'| sort -u
.cmd@Common
.cmd@Cross
.pl
.plt
.py
.sh
```

So I guess the change to `.gitattributes` would be:

```
# default to automatic
* text=auto

# batch scripts must have windows eol
*.bat text eol=crlf

# scripts which may be executable "#!" must have *nix eol to run on Linux (at least)
*.pl text eol=lf
*.plt text eol=lf
*.sh text eol=lf
*.py text eol=lf
*.cmd text eol=lf
# .cmd in unexpanded templates
*.cmd@* text eol=lf
```

anjohnson:
Your search appears to have missed `/startup/EpicsHostArch` and the `S99*` scripts, which all have no extension.

mdavidsaver:
... and also `modules/ca/src/client/S99caRepeater@`.  I guess these two need to be special cases in `.gitattributes`?

mdavidsaver:
`*.cmd text eol=lf` is missing, but I think this is close enough for such a small change.  I'll add this line and merge.

mdavidsaver:
Applied as: dc03d519fba9d0f4fe252e56547f428ae3aca4ee

@tboegi I see that github now requires that Actions CI builds be explicitly enabled for "new contributors".  I've now done this, so your future Base PRs should be tested automatically.

tboegi:
@mdavidsaver and all: Thanks for the help

PR #168, opened 2021-05-08
Add GitLab CI jobs to compile Base on CentOS 6/7/8

ralphlange:
Instead of using Debian/Ubuntu releases that are "comparable" to specific RHEL versions, GitLab CI can use the official CentOS Docker images.

Some patching is needed for CentOS 6 - push python to 2.7, point the repository URLs to "vault" (CentOS 6 is after EOL).

The failures using clang on CentOS 8 are due to clang not supporting a gcc fortification configuration that the Perl build uses for Cap5. Any hints how to handle that are appreciated. (Disable Cap5?)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,124 @@
+# .gitlab-ci.yml for EPICS Base jobs on different versions of CentOS
+# (see: https://github.com/epics-base/ci-scripts)
+
+variables:
+  GIT_SUBMODULE_STRATEGY: "recursive"
+  SETUP_PATH: ".ci:.ci-local"
+  CMP: "gcc"
+  BGFC: "default"
+  BASE: "SELF"
+  EPICS_TEST_IMPRECISE_TIMING: "YES"
+
+# Install prerequisites: common and release specific parts
+.yum_common: &yum_common |-
+  yum -y install gcc gcc-c++ make git perl epel-release
+  yum -y install clang perl-Test-Simple perl-Test-Harness perl-ExtUtils-ParseXS
+  yum -y install p7zip
+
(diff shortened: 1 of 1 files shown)


ralphlange:
Here's how that looks on GitLab: https://gitlab.com/ralphlange/epics-base/-/pipelines/299625730

mdavidsaver:
This change looks ok to me.  I see that the `rh8_clang*` jobs being added are failing with `unknown argument: '-fstack-clash-protection'`.  I don't see where this is coming from.

Updating `.ci` now causes the `gcc-*` jobs on GHA to fail.  These can either be removed now, or I will fix post merge (by transplanting a change from PVXS).

Also, I did look into container builds on GHA.  It turns out to be [possible](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainerimage), but with an unfortunate limitation that the checkout action doesn't handle submodules when git isn't present in the container image by default (falls back to some REST API).  Maybe someone with more experience with Docker has a way around this?  I don't see this as urgent though, and am happy to try gitlab-ci.

ralphlange:
The failures are coming from the fact that compiling a Perl module (Cap5) always uses the same compiler options that were used with Perl itself, and Perl on RH8 (compiled with gcc) chose to use a hardening option that clang on RH8 does not support.

I have asked @anjohnson for a suggestion on how to proceed, with no answer yet.

mdavidsaver:
Oh, I see.  So inadvertent mixing of gcc and clang.

anjohnson:
Core group 2023-01-04: AI on ANJ: Perl Cap5 try to detect when Perl's CC is different to ours and don't build Cap5 in that case.

PR #170, opened 2021-05-13
`FINAL_LOCATION` in init scripts

### LINKED ISSUE #167: caRepeater init not using `FINAL_LOCATION`
`S99caRepeater@` and `caRepeater.service@` in `modules/ca/src/client/` are being expanded with `INSTALL_BIN` based on `INSTALL_LOCATION` instead of `FINAL_LOCATION`.

mdavidsaver:
Attempt to fix #167

### DIFF
--- a/configure/CONFIG_SITE
+++ b/configure/CONFIG_SITE
@@ -151,6 +151,14 @@ CROSS_WARN=YES
 #  different location then uncomment and set this.
 #INSTALL_LOCATION=<fullpathname>
 
+# The location from which files placed in INSTALL_LOCATION will actually run.
+# This path is compiled into executables, and so should be an absolute.
+#  May be used to achieve the effect of autotools.  eg.
+#   ./configure --prefix=<FINAL_LOCATION>
+#   make DESTDIR=<INSTALL_LOCATION>
+# Defaults to the absolute expansion of $(INSTALL_LOCATION)
+#FINAL_LOCATION=
+
 # Use POSIX thread priority scheduling (if available).
 #  Must be either YES or NO
 USE_POSIX_THREAD_PRIORITY_SCHEDULING = YES
--- a/configure/CONFIG_COMMON
+++ b/configure/CONFIG_COMMON
@@ -71,6 +71,8 @@ INSTALL_DBD                 = $(INSTALL_LOCATION)/dbd
 INSTALL_DB                  = $(INSTALL_LOCATION)/db
 INSTALL_CONFIG              = $(INSTALL_LOCATION)/configure
 
+FINAL_LOCATION = $(shell $(PERL) $(TOOLS)/fullPathName.pl $(INSTALL_LOCATION))
+
 # Directory for OS independant build created files
 COMMON_DIR = ../O.Common
 
--- a/modules/ca/src/client/Makefile
+++ b/modules/ca/src/client/Makefile
@@ -113,7 +113,7 @@ SCRIPTS_Linux = caRepeater.service
 
 EXPAND += S99caRepeater@
 EXPAND += caRepeater.service@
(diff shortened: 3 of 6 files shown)


configure/CONFIG_SITE
```diff
@@ -151,6 +151,11 @@ CROSS_WARN=YES
 #  different location then uncomment and set this.
 #INSTALL_LOCATION=<fullpathname>
 
+# The location from which files placed in INSTALL_LOCATION will actually
+#  be run.  May be different when INSTALL_LOCATION is a temporary staging area.
+#  eg. make INSTALL_LOCATION=/build/1234 FINAL_LOCATION=/usr/lib/epics
+FINAL_LOCATION=$(shell $(PERL) $(TOOLS)/fullPathName.pl $(INSTALL_LOCATION))
```
anjohnson: The `configure/CONFIG_SITE` file contains no other instances of complex default settings, and I'd rather we not introduce any (it's just asking for trouble unnecessarily IMHO). Can you make this a simple commented-out setting like the one above it, and and put the complex version for the default value in the `configure/CONFIG` file using `?=` so that only gets used if the variable hasn't been set here.

The description should say that this must be an absolute path since it won't work otherwise. Using an example of a make run may be confusing, our readers won't all understand the connection between setting variables here and on the `make` command-line, so I'd just leave that line out.

anjohnson (changes requested):
Almost – see inline change request.

configure/CONFIG_SITE
mdavidsaver: > ... put the complex version for the default value in the `configure/CONFIG` file ...

Ok.

> ... the connection between setting variables here and on the `make` command-line ...

I think an example is important.  I guess I need a better one.  I'm hoping to keep it from getting too long, since `CONFIG_SITE` isn't a manual.  How about.

```
# The location from which files placed in INSTALL_LOCATION will actually run.
# This path is compiled into executables, and so should be an absolute.
#  May be used to achieve the effect of autotools.  eg.
#   ./configure --prefix=<FINAL_LOCATION>
#   make DESTDIR=<INSTALL_LOCATION>
# Defaults to the absolute expansion of $(INSTALL_LOCATION)
#FINAL_LOCATION=
```

anjohnson: Mentioning autotools could be even more confusing, you're still thinking like an experienced developer not a graduate student of Physics. Maybe rather than trying to explain why, the comment should just say that this is for building EPICS for a binary distribution package, something like:
```makefile
# Absolute path where EPICS Base will be run from, if external
# software will move it away from INSTALL_LOCATION. This may be
# needed when building packages for binary distributions.
#FINAL_LOCATION = 
```

Related but a different point: Somewhere, maybe in the `configure/CONFIG` file, we ought to have something pointing out that scripts and programs that are used at build-time shouldn't depend on this, since they have to be able to run from `INSTALL_LOCATION` too.

mdavidsaver: I've moved the definition to `CONFIG_COMMON` just after the definition of `INSTALL_LOCATION`.

`FINAL_LOCATION` is a tool for setting up a multi-stage build process.  I know of one graduate student of Physics who did this.  Still, even an expert needs some bread crumbs.  I feel that it is absolutely essential that we provide some translation between our jargon, and something more familiar.  GNU autotools strikes me as being a good common language, and captures the essential information in two lines.

anjohnson: I meant no offense to grad students, but those that do software packaging have different skills than those that are just trying to get EPICS running on their experiment. Anyway I'll accept this, thanks.

anjohnson (approved):

mdavidsaver:
Applied as of 315fb3df693b78ffb13d056d7c35ac6e6a83df29

PR #171, opened 2021-05-13
set SONAME for cross builds

### LINKED ISSUE #169: Cross builds not setting SONAME
While shared libraries are always installed with the `SHRLIB_VERSION` suffix, this is only being used on a couple of host targets where the SONAME is actually being set.

```sh
$ readelf -d lib/linux-x86_64/libCom.so.3.19.0 |grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libCom.so.3.19.0]
$ readelf -d lib/linux-aarch64/libCom.so.3.19.0 |grep SONAME
```

It looks like the `-Wl,-h` is only passed for certain host builds.  Maybe this can move into `configure/os/CONFIG.Common.linuxCommon`?

```make
$ git grep '\-Wl,-h'
configure/os/CONFIG.linux-x86.linux-x86:SHRLIB_LDFLAGS += -Wl,-h$@
configure/os/CONFIG.linux-x86.linux-x86:LOADABLE_SHRLIB_LDFLAGS += -Wl,-h$@
configure/os/CONFIG.solaris-sparc-gnu.solaris-sparc-gnu:SHRLIB_LDFLAGS += -Wl,-z,defs -Wl,-z,text -Wl,-h,$@
configure/os/CONFIG.solaris-sparc-gnu.solaris-sparc-gnu:LOADABLE_SHRLIB_LDFLAGS += -Wl,-z,text -Wl,-h,$@
```

mdavidsaver:
Attempt to fix #169.


anjohnson (approved):
No objections that I can articulate; embedded cross-builds don't create shared libraries anyway, so I can't see where this might cause issues.

PR #172, opened 2021-05-18
Fix issue with very large timeout passed to epicsEventWaitWithTimeout()

FreddieAkeroyd:
A very large timeout was getting converted to a 0 wait and causing some unit tests to fail in strange and random ways. Not trapping large timeouts was an oversight when converting to waitable timers on WIN32

### DIFF
--- a/src/libCom/osi/os/WIN32/osdEvent.c
+++ b/src/libCom/osi/os/WIN32/osdEvent.c
@@ -92,24 +92,27 @@ epicsShareFunc epicsEventStatus epicsEventWait ( epicsEventId pSem )
 epicsShareFunc epicsEventStatus epicsEventWaitWithTimeout (
     epicsEventId pSem, double timeOut )
 { 
-    static const unsigned nSec100PerSec = 10000000u; /* number of 100ns intervals per second */
-    static const unsigned mSecPerSec = 1000u;
+    /* waitable timers use 100 nanosecond intervals, like FILETIME */
+    static const unsigned ivalPerSec = 10000000u; /* number of 100ns intervals per second */
+    static const unsigned mSecPerSec = 1000u;     /* milliseconds per second */
     HANDLE handles[2];
     DWORD status;
     LARGE_INTEGER tmo;
     HANDLE timer;
-    LONGLONG nSec100;
+    LONGLONG nIvals;  /* number of intervals */
 
     if ( timeOut <= 0.0 ) {
         tmo.QuadPart = 0u;
     }
     else if ( timeOut >= INFINITE / mSecPerSec  ) {
-        nSec100 = (LONGLONG)(INFINITE - 1) * (nSec100PerSec / mSecPerSec); /* for compatibility with old approach */
-        tmo.QuadPart = -nSec100;  /* negative value means a relative time offset for timer */
+        /* we need to apply a maximum wait time to stop an overflow. We choose (INFINITE - 1) milliseconds,
+           to be compatible with previous WaitForSingleObject() implementation */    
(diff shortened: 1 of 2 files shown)


src/libCom/osi/os/WIN32/osdEvent.c
```diff
@@ -93,16 +93,23 @@ epicsShareFunc epicsEventStatus epicsEventWaitWithTimeout (
     epicsEventId pSem, double timeOut )
 { 
     static const unsigned nSec100PerSec = 10000000u;
```
mdavidsaver: Probably unrelated, but I happened to notice that this constant is 10 million (7 zeros).

Also, Is the name `nSec100` meant to convey `nSec / 100`?

FreddieAkeroyd: `nSec100` is for the number of `100ns` intervals, which is the unit used by waitable timers, probably because windows `FILETIME` uses this. That then gives 10^7 of these intervals per second. I can add a code comment.

src/libCom/osi/os/WIN32/osdThread.c
```diff
@@ -808,14 +808,21 @@ HANDLE osdThreadGetTimer()
...
     LARGE_INTEGER tmo;
     HANDLE timer;
+    LONGLONG nSec100;
 
     if ( seconds <= 0.0 ) {
         tmo.QuadPart = 0u;
     }
+    else if ( seconds >= INFINITE / mSecPerSec  ) {
+        nSec100 = (LONGLONG)(INFINITE - 1) * (nSec100PerSec / mSecPerSec); /* for compatibility with old approach */
+        tmo.QuadPart = -nSec100;
```
mdavidsaver: Trying to decode these expressions gives me:

`INFINITE / mSecPerSec` -> `0xfffffffe/1000` -> `4294967` seconds as the maximum timeout.

`(LONGLONG)(INFINITE - 1) * (nSec100PerSec / mSecPerSec)` -> `0xfffffffeull * 10000` -> `42949672940000` or `4294967.294 ns`

So this tracks.

`tmo.QuadPart = -nSec100;` as `SetWaitableTimer()`[Negative values indicate relative time](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-setwaitabletimer).

FreddieAkeroyd: I will add a code comment re: negative values

src/libCom/osi/os/WIN32/osdEvent.c
mdavidsaver: Ok, I see.  I'm not asking for a change, just noting that the constant name isn't so helpful to me.  Were it me, I would skip the named constant and instead add a comment.  eg. for `timeOut >= INFINITE / 1000u` something to the effect that the high limit is `INFINITE` milliseconds, which originated from historical use of `WaitForSingleObject()`.

anjohnson: Michael said he wasn't asking for change but I think it may be warranted (since the use of 100ns is confusing) – could you introduce a name for that interval, and adjust the other names to match? We may re-use the term "tick" too much although I wouldn't object if you want to use it, you already called 100ns an "interval" (ival, intv, intvl?), but anything else you come up with might work instead. The easier we can make the code for humans to read the less likelihood of bugs being introduced and the faster it is to change it later.

FreddieAkeroyd: thanks - I chose "ival"

mdavidsaver:
Applied as d1ddbad053cf092e70c593a67fdd41494005eda7.  Thanks!

Issue #175, opened 2021-06-15
RTEMS5 network issues/TODOs


mdavidsaver:
A catalog of issues I'm seeing with an RTEMS5 build.

## FHI defaults

These defaults shouldn't be hidden away in C code.

https://github.com/epics-base/epics-base/blob/256babf9614a199709e2728f0ecd79bf57b668ec/modules/libcom/RTEMS/posix/rtems_init.c#L96-L99

## Timeouts

DHCP has a timeout, but NTP fetch does not.  imo. this should be the other way around.  This is problematic for me since with QEMU 3.1.0 at least, `-net user` the DHCP server doesn't seem to respond to `option ntp_servers`.  (may be related by the next item)

## DHCP option names

There may be an issue with `option ntp_servers` vs. `options ntp-servers`.  cf. `freebsd/sbin/dhclient/tables.c` in the libbsd source, or `/etc/dhcp/dhclient.conf`.

https://github.com/epics-base/epics-base/blob/256babf9614a199709e2728f0ecd79bf57b668ec/modules/libcom/RTEMS/posix/rtems_init.c#L819-L822

## NIC emulation

`-net nic,model=e1000` works for me... eventually.

`-net nic,model=rtl8139` doesn't work for me.  I see `reason=CARRIER`, but never `reason=BOUND`.

mdavidsaver:
wrt. `-net nic,model=rtl8139`, it's not straightforward to troubleshoot, but I don't think the QEMU DHCP server ever sees a request.

```
qemu-system-i386 -m 64 -no-reboot -serial stdio -display none -net nic,model=rtl8139 -net nic,model=ne2k_pci \
 -net user -append --console=/dev/com1 -kernel epicsTypesTest
```

Also, I do eventually see a `reason=IPV4LL` with `new_ip_address=169.254.205.11`.  This looks like dhcpcd timing out and using a auto-config address, which I don't think we ever want.

mdavidsaver:
`-mtune=pentiumpro -march=pentium` from `$(CPU_FLAGS)` in the RTEMS makefiles shows up in `CPPFLAGS` instead of `CFLAGS`/`CXXFLAGS` where I would expect to find them.  I don't think this will break anything though.

```sh
$ make -s -C configure/O.RTEMS-pc686-qemu PRINT.CPPFLAGS
[...]

PR #176, opened 2021-06-15
RTEMS5: Make QEMU tests work

mdavidsaver:
Partially address #175.

Rewrite the DHCP handler to avoid repeated explicit `char*` handling.

Make `rtemsInit_NTP_server_ip` optional, and remove the default.

Change the dhcpcd.conf option name separator from `_` to `-`.  May be necessary.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -85,7 +85,6 @@ jobs:
             configuration: default
             rtems: "5"
             rtems_target: RTEMS-pc686-qemu
-            test: NO
             name: "Ub-20 gcc-9 + RT-5.1 pc686"
 
           - os: ubuntu-20.04
@@ -145,7 +144,6 @@ jobs:
             rtems: "4.9"
             name: "Ub-20 gcc-9 + RT-4.9"
             rtems_target: RTEMS-pc386-qemu
-            test: NO
 
           - os: macos-latest
             cmp: clang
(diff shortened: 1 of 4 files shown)


mdavidsaver:
@hjunkes @brendanchandler fyi.

hjunkes:
Looks ok to me. Thanks
As for the names in dhcpcd.cnf with "_" or "-" I am unsure. In the documentation of FreeBSD is usually worked with "_". 
https://www.freebsd.org/cgi/man.cgi?query=dhcpcd&sektion=8&apropos=0&manpath=FreeBSD+11.0-RELEASE+and+Ports

brendanchandler:
Looks reasonable to me.

mdavidsaver:
I've made some progress towards running RTEMS5 test with QEMU.  I still haven't been able to make `-net nic,model=rtl8139` work for me (with QEMU >=3.1.0).  However, I found an ugly hack to make `-net nic,model=e1000` boot faster by stubbing out the eeprom checksum verification code.  With this change the trivial `epicsTypeTest` will run for me in 14 seconds (vs. 4 seconds with RTEMS 4.x).  The remaining time seems to be mostly taken by 3x duplicate address checks.

```
sending ARP probe (1 of 3), next in %0.1f seconds
```

(yes, the printf spec is being printed without expansion :) )

Adding 10 seconds to every test still more than doubles the total runtime, but it is at least in the realm of the practical for me.

I've also added a change to re-enable logging from libbsd (boy do I wish I'd noticed this sooner...).  The change leaves the default handler (w/ synchronous `printf()`) in place until near the end of `POSIX_Init()`, then sets a handler using `errlogVprintf()`.

PR #177, opened 2021-06-23
Atomic compare and swap

dirk-zimoch:
New atomic compare and swap semantics returning bool (`int` actually becaue `bool` is not C) instead of the old value (because this is the only flavor supported by VxWorks 6). This is related to bug https://bugs.launchpad.net/epics-base/+bug/1932118.

I have renamed the functions `epicsAtomicCmpAndSwap*()` to   #`epicsAtomicCmpAndSwap*()` so that potential externall usage will not show unexpected behavior because of the changed return value (but instead will fail to link).
I have changed the C++ wrapper `compareAndSwap` to return `bool` but I have not changed the name. (Should I ?)

Tested on RHEL7 and vxWorks 6.7.

### DIFF
--- a/modules/database/src/ioc/db/callback.c
+++ b/modules/database/src/ioc/db/callback.c
@@ -233,7 +233,7 @@ void callbackStop(void)
 {
     int i;
 
-    if (epicsAtomicCmpAndSwapIntT(&cbState, cbRun, cbStop)!=cbRun) return;
+    if (!epicsAtomicCompareAndSwapIntT(&cbState, cbRun, cbStop)) return;
 
     for (i = 0; i < NUM_CALLBACK_PRIORITIES; i++) {
         epicsAtomicSetIntT(&callbackQueue[i].shutdown, 1);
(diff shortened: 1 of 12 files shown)


dirk-zimoch:
Patches for pvAccess and pva2pva usage of `compareAndSwap`: [pvaPatches.zip](https://github.com/epics-base/epics-base/files/6702154/pvaPatches.zip)

mdavidsaver:
I can't say I'm thrilled about a breaking API change like this to work around limitations w/ (legacy?) vxWorks.  Wouldn't newer vxWorks use the GCC builtins?

For legacy PPC targets maybe it is reasonable to include some `asm()`?

dirk-zimoch:
I am not thrilled about the way EPICS again and again breaks compatibility with legacy systems. But what can we do? The world is not perfect.

dirk-zimoch:
VxWorks 6.7 (maybe 6.6+) on PPC seems to support `__sync_val_compare_and_swap`. If that works, I will make a new pull request and close this one.

dirk-zimoch:
Can use __sync_val_compare_and_swap in vxWorks 6.6 ... 6.8 (as it was used in 6.9 already) even for PPC.
Thus this request becomes obsolete.

PR #179, opened 2021-06-24
fix missing declaration of 'close' function in vxWorks 6.3 and lower

dirk-zimoch:
VxWorks up to version 6.3 does not have the declaration for `close()` in `unistd.h` but in `ioLib.h`.
This is a problem since `-Werror-implicit-function-declaration` is enabled.
This fix does not harm any other vxWorks version and does not affect ant other OS.


anjohnson (approved):
Looks fine to me.

PR #180, opened 2021-06-24
Com: Adjust epicsAtomic conditionals for GCC

mdavidsaver:
cf. https://bugs.launchpad.net/epics-base/+bug/1932118

@dirk-zimoch fyi.

### DIFF
--- a/modules/libcom/src/osi/compiler/gcc/epicsAtomicCD.h
+++ b/modules/libcom/src/osi/compiler/gcc/epicsAtomicCD.h
@@ -23,50 +23,36 @@
 
 #define EPICS_ATOMIC_CMPLR_NAME "GCC"
 
+/* expands __GCC_HAVE_SYNC_COMPARE_AND_SWAP_ concatentating the numeric value __SIZEOF_*__ */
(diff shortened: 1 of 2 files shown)


anjohnson:
Thanks for the sorely needed explanatory comments.

An aside, but I discovered recently that the preprocessor’s `defined()` function is only actually valid when it appears directly inside a `#if` statement, although gcc and most other compilers have always allowed it in macros that get expanded in one. Recent versions of clang actually warn about UB when they see that, although they do still honor it.

mdavidsaver:
Now that 7.0.6 is out the door, I'd like to merge this.

mdavidsaver:
Core meeting review: acceptable

anjohnson:
Working on CLANG version, race to who gets to merge it first...

anjohnson:
I'm not sure that all the macro magic that sets `GCC_ATOMIC_INTRINSICS_AVAIL_SYNC` is correct here, please don't merge.

I have a version with it fixed though (I think) and with the common-with-clang implementation too. I'm supposed to be on vacation this afternoon though (what's left of it) so it'll be sometime next week before I'm ready to push that.

mdavidsaver:
Superseded by #192

PR #185, opened 2021-07-06
epicsTime: rework

mdavidsaver:
Re-implement around `struct epicsTimeStamp` (C API) with `class epicsTime` becoming a wrapper.  Prefer epicsInt64 arithmetic.

While definitions in `epicsTime.h` are shuffled around to move the C definitions before the C++,
the only API change is the removal of the opaque `struct l_fp` (NTP time conversion).

This should also fix the issue with overflow noted in [lp:1896295](https://bugs.launchpad.net/bugs/1896295).

### DIFF
--- a/modules/libcom/src/osi/epicsTime.cpp
+++ b/modules/libcom/src/osi/epicsTime.cpp
@@ -12,13 +12,6 @@
 /* epicsTime.cpp */
 /* Author Jeffrey O. Hill */
 
-// Notes:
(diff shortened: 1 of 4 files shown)


mdavidsaver:
Core meeting review: acceptable

mdavidsaver:
Applied as cb8c7998b62701a849a6fa9c299cc1613f66a627

PR #189, opened 2021-07-22
Fix a couple memory leaks and a segfault

JJL772:
This fixes some possible memory leaks pointed out by static code analysis, and a segfault I encountered while running `dptpn` without a value parameter.
None of the memory leaks fixed are critical bugs, just small things that occur on certain failures.

I'm not sure what to do here in dptpn, so I just have it pass an empty string in the case where `pvalue` is NULL (Line 616 in dbNotifty.c):
```c
    strncpy(ptpnInfo->buffer, pvalue ? pvalue : "", sizeof(ptpnInfo->buffer));
```

### DIFF
--- a/modules/database/src/ioc/db/dbBkpt.c
+++ b/modules/database/src/ioc/db/dbBkpt.c
@@ -331,6 +331,7 @@ long dbb(const char *record_name)
      if (pnode->ex_sem == NULL) {
         printf("   BKPT> Out of memory\n");
         dbScanUnlock(precord);
+        free(pnode);
         epicsMutexUnlock(bkpt_stack_sem);
         return(1);
      }
--- a/modules/database/src/ioc/db/dbNotify.c
+++ b/modules/database/src/ioc/db/dbNotify.c
@@ -613,8 +613,10 @@ long dbtpn(char *pname, char *pvalue)
     ptpnInfo = dbCalloc(1, sizeof(tpnInfo));
     ptpnInfo->ppn = ppn;
     ptpnInfo->callbackDone = epicsEventCreate(epicsEventEmpty);
-    strncpy(ptpnInfo->buffer, pvalue, 80);
-    ptpnInfo->buffer[79] = 0;
+    if (pvalue) {
+        strncpy(ptpnInfo->buffer, pvalue, sizeof(ptpnInfo->buffer));
+        ptpnInfo->buffer[sizeof(ptpnInfo->buffer)-1] = 0;
+    }
 
     ppn->usrPvt = ptpnInfo;
     epicsThreadCreate("dbtpn", epicsThreadPriorityHigh,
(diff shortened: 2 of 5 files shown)


mdavidsaver (approved):
These all look like sensible changes to me.

mdavidsaver:
> I'm not sure what to do here in dptpn ...

Crashing is clearly no the intended behavior.  So my first thought is to reject the missing argument early.  eg. change:

```c
if (!pname) {
```

to

```c
if (!pname || !pvalue) {
```

JJL772:
Yeah, I can make that change, but that brings into question the purpose of line 607:
```cpp
    ppn->requestType = pvalue ? putProcessRequest : processGetRequest;
```
So it seems like `pvalue` is expected to be null in at least some cases.

mdavidsaver:
Fair point.  I've not used `dbtpn()` myself, and clearly not too many others have either.

Looking more closely,  This buffer is intended for output when `requestType == processGetRequest`.  cf. `getCallback()` above in `dbNotify.c`.  Initializing `buffer` to `""`, while it should not be necessary in this case, seems a reasonable precaution.

anjohnson:
The second argument to a `dbtpn` command is supposed to be optional. If omitted it results in a call to `dbtpn()` with `pvalue` being NULL and the code is then supposed to test the `processGet` functionality. I currently get a segfault from the `strncpy()` if I omit that argument, so thanks for reporting this. However the `struct tpInfo *ptpnInfo` is allocated using `calloc()` so there is no need to initialize the buffer at all when `pvalue` is NULL. I would thus prefer that the fix for the conditional be like this (plus your introductions of `sizeof()` that I didn't include), which works for me:
```diff
diff --git a/modules/database/src/ioc/db/dbNotify.c b/modules/database/src/ioc/db/dbNotify.c
--- a/modules/database/src/ioc/db/dbNotify.c
+++ b/modules/database/src/ioc/db/dbNotify.c
@@ -613,8 +613,10 @@ long dbtpn(char *pname, char *pvalue)
     ptpnInfo = dbCalloc(1, sizeof(tpnInfo));
     ptpnInfo->ppn = ppn;
     ptpnInfo->callbackDone = epicsEventCreate(epicsEventEmpty);
-    strncpy(ptpnInfo->buffer, pvalue, 80);
-    ptpnInfo->buffer[79] = 0;
+    if (pvalue) {
+        strncpy(ptpnInfo->buffer, pvalue, 80);
+        ptpnInfo->buffer[79] = 0;
+    }
 
     ppn->usrPvt = ptpnInfo;
     epicsThreadCreate("dbtpn", epicsThreadPriorityHigh,
```

I approve of all the other changes.

Thanks!

JJL772:
Awesome! I've gone ahead and added a branch like you suggested

anjohnson (approved):
Thanks!

anjohnson:
Core Group 7/28 Approved.

mdavidsaver:
Applied as of 7c991f3f2a804e98df7ab89d83577694c8503a48

PR #191, opened 2021-07-29
int64in: Fix monitor delta test

kasemir:
Only the lower 32 bit used to be compared.
https://bugs.launchpad.net/epics-base/+bug/1938459


anjohnson (approved):

mdavidsaver (approved):
Looks simple enough to me.

PR #192, opened 2021-08-02
Gcc atomic cleanup

anjohnson:
Updated version of #180 with Clang support


anjohnson:
LibCom atomic tests passed on VxWorks 6.9 (GCC 4.3.3), configuration gives this output:
```
# Classify Build conditions
# Compiler dependent impl name GCC
# OS dependent impl name VX-ATOMICLIB
# Use GCC atomic builtin memory barrier
# Use GCC builtins for int
# Use GCC builtins for size_t
# Use default epicsAtomicSetIntT()
# Use default epicsAtomicSetSizeT()
# Use default epicsAtomicSetPtrT()
# Use default epicsAtomicGetIntT()
# Use default epicsAtomicGetSizeT()
# Use default epicsAtomicGetPtrT()
```

mdavidsaver (approved):
Looks fine to me.

For GCC, the only non-comment change I see compared with #180 is the such from `#define ... define(...)` to `#if ...\n#define`.  Is there anything else I'm missing?

anjohnson:
IIRC the main change from your version was to move that `#if` line back below the two `#define GCC_ATOMIC_INTRINSICS_AVAIL_`... macros that it tests, and to not use the `defined()` that you added to their parts of the `#if` condition (they are always defined if that `#if` gets used).

PR #193, opened 2021-08-05
Com: clear IP_MULTICAST_ALL on Linux

mdavidsaver:
Quoting [`man 7 ip`](https://man7.org/linux/man-pages/man7/ip.7.html).

```
IP_MULTICAST_ALL (since Linux 2.6.31)
              This option can be used to modify the delivery policy of
              multicast messages to sockets bound to the wildcard
              INADDR_ANY address.  The argument is a boolean integer
              (defaults to 1).  If set to 1, the socket will receive
              messages from all the groups that have been joined
              globally on the whole system.  Otherwise, it will deliver
              messages only from the groups that have been explicitly
              joined (for example via the IP_ADD_MEMBERSHIP option) on
              this particular socket.
```

Not of the other targets I've tested (OSX and RTEMS+libbsd) behave in this non-compliant way by default.  It seems reasonable to me that, where possible, sockets created by `epicsSocketCreate()` should behave in a uniform way across all supported targets.  So I would like to set `IP_MULTICAST_ALL=0` by default.

If it isn't done for all sockets, it will at least have to by done for the CA UDP sockets.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdSock.c
+++ b/modules/libcom/src/osi/os/posix/osdSock.c
@@ -99,6 +99,22 @@ LIBCOM_API SOCKET epicsStdCall epicsSocketCreate (
             close ( sock );
             sock = INVALID_SOCKET;
         }
+
+#ifdef __linux__
+#  ifndef IP_MULTICAST_ALL
+#    define IP_MULTICAST_ALL		49
+#  endif
+        /* Enable compliant filtering of multicasts on Linux.  cf. 'man 7 ip' */
+        if(domain==AF_INET && type==SOCK_DGRAM){
+            static int logged;
+            int val = 0;
+            if(setsockopt(sock, IPPROTO_IP, IP_MULTICAST_ALL, (char*)&val, sizeof(val)) && !logged) {
+                logged = 1;
+                errlogPrintf("Warning: Unable to clear IP_MULTICAST_ALL (err=%d).  This may cause problems on multi-homed hosts.\n",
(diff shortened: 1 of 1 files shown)


modules/libcom/src/osi/os/posix/osdSock.c
```diff
@@ -99,6 +99,22 @@ LIBCOM_API SOCKET epicsStdCall epicsSocketCreate (
             close ( sock );
             sock = INVALID_SOCKET;
         }
+
+#ifdef __linux__
+#  ifndef IP_MULTICAST_ALL
+#    define IP_MULTICAST_ALL		49
```
anjohnson, line 105: Is it wise to define and use this macro if the C library headers don't actually define it? It has probably been included in glibc for a long time and I'm fine with doing this if it's available, but we don't test on other libc implementations and they might not define it. My first instinct would be that if it isn't defined, we shouldn't try to use it at all.

mdavidsaver: Wisdom is a tricky thing.  My first instinct was the same as yours.  However, I found at least one environment (python manylinux1) still has old Linux 2.6 era headers.  As far as I can tell, IP socket option 49 was unused prior to the introduction of `IP_MULTICAST_ALL`.  Since this is all wrapped by `#ifdef __linux__` it should not effect other OSs.  So anyone still running such an old kernel (circa RHEL5) should only see this new warning.

anjohnson (approved):

PR #195, opened 2021-08-08
Fix long `CALC$`

### LINKED ISSUE #194: Issue with long string `CALC$`
Originating with https://epics.anl.gov/tech-talk/2021/msg01491.php

Test case from https://epics.anl.gov/tech-talk/2021/msg01504.php

```
record(calc, "test_calc")
{
    field(SCAN, "1 second")
    field(CALC, "RNDM*100")
}

$ caget -S test_calc.CALC$
test_calc.CALC$ RNDM*100

$ caput -S test_calc.CALC$ "300"
Old : test_calc.CALC$ RNDM*100
    Warning: "Channel write request failed"
    Context: "op=1, channel=test_calc.CALC$, type=DBR_CHAR, count=4, ctx="test_calc""
    Source File: ../oldChannelNotify.cpp line 159
    Current Time: Sat Aug 07 2021 15:27:54.755166000
New : test_calc.CALC$ 3NDM*100

.. and the softIoc shows this error:

Illegal field value PV: test_calc calc: Illegal CALC field
test_calc.CALC: Syntax error, unknown operator/operand in expression "3NDM*100"
```

mdavidsaver:
Add test and attempt to resolve #194.

@dirk-zimoch Please have a look.

### DIFF
--- a/modules/database/test/std/rec/regressTest.c
+++ b/modules/database/test/std/rec/regressTest.c
@@ -141,13 +141,36 @@ void testCADisconn(void)
     testdbCleanup();
 }
 
+/* https://github.com/epics-base/epics-base/issues/194 */
+static
+void testLongCalc(void)
+{
+    startRegressTestIoc("regressLongCalc.db");
+
+    testdbGetFieldEqual("test_calc.CALC", DBF_STRING, "RNDM*100");
+
+    testdbPutArrFieldOk("test_calc.CALC$", DBF_CHAR, 4, "300\0");
+    testdbGetFieldEqual("test_calc.CALC", DBF_STRING, "300");
+    testdbPutFieldOk("test_calc.PROC", DBF_LONG, 1);
+    testdbGetFieldEqual("test_calc", DBF_DOUBLE, 300.0);
+
+    const char small[] = "0.0000000000000000000000000000000000000001";
+    testdbPutArrFieldOk("test_calc.CALC$", DBF_CHAR, sizeof(small), small);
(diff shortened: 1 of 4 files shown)


modules/database/test/std/rec/regressTest.c
```diff
@@ -134,15 +136,32 @@ void testCADisconn(void)
...
 
+/* https://github.com/epics-base/epics-base/issues/194 */
+static
+void testLongCalc(void)
+{
+    startRegressTestIoc("regressLongCalc.db");
+
+    testdbGetFieldEqual("test_calc.CALC", DBF_STRING, "RNDM*100");
+    testdbPutArrFieldOk("test_calc.CALC$", DBF_CHAR, 4, "300\0");
+    testdbGetFieldEqual("test_calc.CALC", DBF_STRING, "300");
```
anjohnson, line 153: The above checks shorten the field value, could you add another check that extends it beyond its current length? Doing that passes with your fix, but it's the kind of thing which could easily fail in some broken future modifications. I don't think we _need_ to check for writing strings >40 chars here, but it might be a good thing to check that potential limit at the same time, so I suggest writing `"0.0000000000000000000000000000000000000001"` to CALC$ which should set VAL to `1.0e-40`.

anjohnson (approved):
Thanks for picking this up so quickly. I have tested your fix which works for me. There's a suggestion below for an additional test that could catch some other potential issues in future modifications.

mdavidsaver:
I've added the requested additional test case.  I'll merge unless CI turns up something new.

mdavidsaver:
Applied as of 88079c9107403a2de8d225be033c491ce10ddda6

Issue #183, opened 2021-06-30
DB link DBF_MENU to DBF_STRING conversions broken


anjohnson:
From a [bug report](https://epics.anl.gov/tech-talk/2021/msg01253.php) by Bayan Sobhani (Alex) I derived this test database:
```
record(ai, "ai") {}
record(stringin, "si") {
  field(INP, "ai.SEVR")
}
```
This shows the break on the 7.0 branch:
```
woz$ ./base-7.0/bin/darwin-x86/softIoc -d ~/db/test.db
Starting iocInit
############################################################################
## EPICS R7.0.5.1-DEV
## Rev. R7.0.5-264-g7ca25d3c888c56844877-dirty
############################################################################
iocRun: All initialization complete
epics> dbpf si.PROC 1
missing record support entry table Record Support Routine (get_enum_str) Record Type ai PV ai.SEVR  dbGetField

DBF_UCHAR:          1 = 0x1   
epics>
```
The database works properly with 3.15.9, 7.0.2 and 7.0.4.

`SEVR` is a `DBF_MENU` field, so the code shouldn't be trying to call the `RSET::get_enum_str()` routine for it anyway.

FreddieAkeroyd:
Ah, I should have checked here earlier, just been writing an email to report the same thing. In our case it was reading FTVL from a waveform record into a STRING aSub record field.

anjohnson:
@dirk-zimoch It seems likely to me that this bug was introduced by your dbChannelForDBLinks changes from last November, could you find time to take a look at it? We could probably use some additional test cases to match as well.

mdavidsaver:
> ... We could probably use some additional test cases to match as well.

fyi. [dbDbLinkTest.c](https://github.com/epics-base/epics-base/blob/7.0/modules/database/test/ioc/db/dbDbLinkTest.c), and the associated `dbDbLinkTest.db`, would be one obvious place to put additional tests like this.

dirk-zimoch:
I‘ll have a look…

dirk-zimoch:
Database links now use the same code path as CA links. So I tested what happens when the INP link is a CA link. In that case it fails in 3.14.12 as well with the error message  "dbpf: Value conversion from STRING to UCHAR failed". (Totally misleading of course. What happens is that processing the si record fails because reading INP fails and dbpf thinks that writing the string "1" to .PROC (a DBF_UCHAR) went wrong. ) In 7.0.5 this fails silently.

[...]

PR #196, opened 2021-08-12
Test  and fix DB link DBF_MENU to DBF_STRING

mdavidsaver:
Add test to demonstrate #183.  No fix yet.

### DIFF
--- a/modules/database/src/ioc/db/dbDbLink.c
+++ b/modules/database/src/ioc/db/dbDbLink.c
@@ -188,8 +188,11 @@ static long dbDbGetValue(struct link *plink, short dbrType, void *pbuffer,
                 && dbChannelSpecial(chan) != SPC_ATTRIBUTE
                 && ellCount(&chan->filters) == 0)
     {
-        /* simple scalar: set up shortcut */
-        unsigned short dbfType = dbChannelFinalFieldType(chan);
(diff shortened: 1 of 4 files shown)


anjohnson (approved):
Good solution, thanks!

mdavidsaver:
Note that the added test case has a TODO since this change does not effect CA links, where `DBF_MENU -> DBF_STRING` still doesn't work.

mdavidsaver:
Applied as of 7ce69155a9736ab02ea1265901c55d9044a19350

ralphlange:
FYI
We're seeing a test failure of the added regression test:
```
/home/build/workspaces/trunk/m-epics/target/base-7.0.6.1/modules/database/test/std/rec/O.linux-x86_64
...
regressTest.t ...........
not ok 54 - dbGetField("li2", 5) -> 3 == 0
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/54 subtests
```

mdavidsaver:
I take it your test runner isn't setup to capture core dumps?  The only other occurrence I've seen so far (probably) was from a windows build which didn't produce a meaningful stack trace.

cf. https://github.com/epics-base/epics-base/discussions/162#discussioncomment-1614969

ralphlange:
Happened again yesterday. I will try to set up core dump capture.

PR #198, opened 2021-09-01
Fix several compilation errors for -DDEBUG

tboegi:
After Adding "-DDEBUG" to the compiler options some files don't
compile any more.
Fix the compilation errors.

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -1174,7 +1174,7 @@ void db_event_flow_ctrl_mode_on (dbEventCtx ctx)
      * notify the event handler task
      */
     epicsEventSignal(evUser->ppendsem);
-#ifdef DEBUG
+#if defined(DEBUG) && defined(vxWorks)
     printf("fc on %lu\n", tickGet());
 #endif
 }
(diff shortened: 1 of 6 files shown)


modules/database/src/ioc/db/dbEvent.c
```diff
@@ -1174,7 +1174,7 @@ void db_event_flow_ctrl_mode_on (dbEventCtx ctx)
      * notify the event handler task
      */
     epicsEventSignal(evUser->ppendsem);
-#ifdef DEBUG
+#if defined(DEBUG) && defined(vxWorks)
```
anjohnson, line 1177: Better to just delete this and the similar one below, nobody is ever going to debug this on VxWorks again.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -1245,8 +1245,8 @@ static int claim_ciu_action ( caHdrLargeArray *mp,
         return RSRV_OK;
     }
 
-    DLOG ( 2, ("CAS: claim_ciu_action found '%s', type %d, count %d\n",
-        pName, dbChannelCAType(dbch), dbChannelElements(dbch)) );
+    DLOG ( 2, ("CAS: claim_ciu_action found '%s', count %ld\n",
+        pName, dbChannelElements(dbch)) );
```
anjohnson, line 1249: I think `dbChannelCAType(dbch)` should become `dbChannelExportCAType(dbch)`, the type is presumably useful information.

anjohnson (approved):
Mostly looks good, a couple of suggestions inline which I will do while merging if I get to that first.

Thanks!

PR #199, opened 2021-09-01
cleanup mbboDirect bit field handling

mdavidsaver:
Attempt to resolve [lp:1939568](https://bugs.launchpad.net/epics-base/+bug/1939568).  Force B0 -> BF to always follow the corresponding bit in VAL (I _think_ not a change).  Also allow device support (and OMSL close loop) to overwrite VAL.  This removes most special handling for OMSL.

The main added complication is a new `OBIT` field, to handle the differences between `B*` and `VAL` which can't be tracked with `MLST`.  (because `dbPut()` always posts an update for non-VAL fields)

### DIFF
--- a/modules/database/src/std/rec/mbboDirectRecord.c
+++ b/modules/database/src/std/rec/mbboDirectRecord.c
@@ -255,7 +255,6 @@ static long special(DBADDR *paddr, int after)
         /* Adjust VAL corresponding to the bit changed */
         epicsUInt8 *pBn = (epicsUInt8 *) paddr->pfield;
         epicsUInt32 bit = 1 << (pBn - &prec->b0);
-        epicsUInt32 oobit = prec->obit;
 
         /* Because this is !(VAL and PP), dbPut() will always post a monitor on this B* field
          * after we return.  We must keep track of this change separately from MLST to handle
@@ -270,12 +269,9 @@ static long special(DBADDR *paddr, int after)
             prec->val &= ~bit;
             prec->obit &= ~bit;
         }
-        if(oobit!=prec->obit)
-            db_post_events(prec, &prec->obit, DBE_VALUE|DBE_LOG);
 
         prec->udf = FALSE;
         convert(prec);
-
     }
 
     return 0;
@@ -298,7 +294,6 @@ static void monitor(mbboDirectRecord *prec)
     if (prec->mlst != prec->val) {
         events |= DBE_VALUE | DBE_LOG;
         prec->mlst = prec->val;
-        db_post_events(prec, &prec->mlst, events);
     }
     if (events) {
         db_post_events(prec, &prec->val, events);
@@ -314,7 +309,6 @@ static void monitor(mbboDirectRecord *prec)
             }
         }
         prec->obit = prec->val;
-        db_post_events(prec, &prec->obit, events);
     }
 
     events |= DBE_VALUE | DBE_LOG;
--- a/modules/database/src/std/rec/mbboDirectRecord.c
+++ b/modules/database/src/std/rec/mbboDirectRecord.c
@@ -88,6 +88,14 @@ static long writeValue(mbboDirectRecord *);
 
 #define NUM_BITS 32
 
+static
+void bitsFromVAL(mbboDirectRecord *prec)
+{
+    unsigned i;
+    for(i=0; i<NUM_BITS; i++)
+        (&prec->b0)[i] = !!(prec->val&(1u<<i));
+}
+
 static long init_record(struct dbCommon *pcommon, int pass)
 {
     struct mbboDirectRecord *prec = (struct mbboDirectRecord *)pcommon;
@@ -131,19 +139,7 @@ static long init_record(struct dbCommon *pcommon, int pass)
             status = 0;
     }
 
-    if (!prec->udf &&
-        prec->omsl == menuOmslsupervisory) {
-        /* Set initial B0 - B1F from VAL */
-        epicsUInt32 val = prec->val;
-        epicsUInt8 *pBn = &prec->b0;
-        int i;
-
-        for (i = 0; i < NUM_BITS; i++) {
-            *pBn++ = !! (val & 1);
-            val >>= 1;
-        }
-    }
-
+    bitsFromVAL(prec);
     prec->mlst = prec->val;
     prec->oraw = prec->rval;
     prec->orbv = prec->rbv;
@@ -174,24 +170,13 @@ static long process(struct dbCommon *pcommon)
             }
             prec->val = val;
         }
-        else if (prec->omsl == menuOmslsupervisory) {
-            epicsUInt8 *pBn = &prec->b0;
-            epicsUInt32 val = 0;
-            epicsUInt32 bit = 1;
-            int i;
-
-            /* Construct VAL from B0 - B1F */
-            for (i = 0; i < NUM_BITS; i++, bit <<= 1)
-                if (*pBn++)
-                    val |= bit;
-            prec->val = val;
-        }
         else if (prec->udf) {
-            recGblSetSevr(prec, UDF_ALARM, prec->udfs);
+            recGblSetSevrMsg(prec, UDF_ALARM, prec->udfs, "UDFS");
             goto CONTINUE;
         }
 
         prec->udf = FALSE;
+        bitsFromVAL(prec);
         /* Convert VAL to RVAL */
         convert(prec);
 
@@ -234,6 +219,9 @@ CONTINUE:
         recGblGetTimeStampSimm(prec, prec->simm, NULL);
     }
 
+    /* update bits to reflect any change made by dset */
+    bitsFromVAL(prec);
+
     monitor(prec);
 
     /* Wrap up */
@@ -255,60 +243,41 @@ static long special(DBADDR *paddr, int after)
         return 0;
     }
 
-    if (!after)
-        return 0;
-
-    switch (paddr->special) {
-    case SPC_MOD:   /* Bn field modified */
-        if (prec->omsl == menuOmslsupervisory) {
-            /* Adjust VAL corresponding to the bit changed */
-            epicsUInt8 *pBn = (epicsUInt8 *) paddr->pfield;
-            epicsUInt32 bit = 1 << (pBn - &prec->b0);
-
-            if (*pBn)
-                prec->val |= bit;
-            else
-                prec->val &= ~bit;
-
-            prec->udf = FALSE;
-            convert(prec);
+    if(after==0 && fieldIndex >= mbboDirectRecordB0 && fieldIndex <= mbboDirectRecordB1F) {
+        if(prec->omsl == menuOmslclosed_loop) {
+            /* To avoid confusion, reject changes to bit fields while in closed loop.
+             * Not a 100% solution as confusion can still arise if dset overwrites VAL.
+             */
+            return S_db_noMod;
         }
-        break;
 
-    case SPC_RESET: /* OMSL field modified */
-        if (prec->omsl == menuOmslclosed_loop) {
-            /* Construct VAL from B0 - B1F */
-            epicsUInt8 *pBn = &prec->b0;
-            epicsUInt32 val = 0, bit = 1;
-            int i;
+    } else if(after==1 && fieldIndex >= mbboDirectRecordB0 && fieldIndex <= mbboDirectRecordB1F) {
+        /* Adjust VAL corresponding to the bit changed */
+        epicsUInt8 *pBn = (epicsUInt8 *) paddr->pfield;
+        epicsUInt32 bit = 1 << (pBn - &prec->b0);
+        epicsUInt32 oobit = prec->obit;
 
-            for (i = 0; i < NUM_BITS; i++, bit <<= 1)
-                if (*pBn++)
-                    val |= bit;
-            prec->val = val;
+        /* Because this is !(VAL and PP), dbPut() will always post a monitor on this B* field
+         * after we return.  We must keep track of this change separately from MLST to handle
+         * situations where VAL and B* are changed prior to next monitor().  eg. by dset to
+         * reflect bits actually written.  This is the role of OBIT.
+         */
+
+        if (*pBn) {
+            prec->val |= bit;
+            prec->obit |= bit;
+        } else {
+            prec->val &= ~bit;
+            prec->obit &= ~bit;
         }
-        else if (prec->omsl == menuOmslsupervisory) {
-            /* Set B0 - B1F from VAL and post monitors */
-            epicsUInt32 val = prec->val;
-            epicsUInt8 *pBn = &prec->b0;
-            int i;
+        if(oobit!=prec->obit)
+            db_post_events(prec, &prec->obit, DBE_VALUE|DBE_LOG);
 
-            for (i = 0; i < NUM_BITS; i++, pBn++, val >>= 1) {
-                epicsUInt8 oBn = *pBn;
+        prec->udf = FALSE;
+        convert(prec);
 
-                *pBn = !! (val & 1);
-                if (oBn != *pBn)
-                    db_post_events(prec, pBn, DBE_VALUE | DBE_LOG);
-            }
-        }
-        break;
-
-    default:
-        recGblDbaddrError(S_db_badChoice, paddr, "mbboDirect: special");
-        return S_db_badChoice;
     }
 
-    prec->udf = FALSE;
     return 0;
 }
 
@@ -329,9 +298,24 @@ static void monitor(mbboDirectRecord *prec)
     if (prec->mlst != prec->val) {
         events |= DBE_VALUE | DBE_LOG;
         prec->mlst = prec->val;
+        db_post_events(prec, &prec->mlst, events);
     }
-    if (events)
+    if (events) {
         db_post_events(prec, &prec->val, events);
+    }
+    {
+        unsigned i;
+        epicsUInt32 bitsChanged = prec->obit ^ (epicsUInt32)prec->val;
+
+        for(i=0; i<NUM_BITS; i++) {
+            /* post bit when value or alarm severity changes */
+            if((events&~(DBE_VALUE|DBE_LOG)) || (bitsChanged&(1u<<i))) {
+                db_post_events(prec, (&prec->b0)+i, events | DBE_VALUE | DBE_LOG);
+            }
+        }
+        prec->obit = prec->val;
+        db_post_events(prec, &prec->obit, events);
+    }
 
     events |= DBE_VALUE | DBE_LOG;
     if (prec->oraw != prec->rval) {
--- a/modules/database/src/std/rec/mbboDirectRecord.c
+++ b/modules/database/src/std/rec/mbboDirectRecord.c
@@ -139,7 +139,24 @@ static long init_record(struct dbCommon *pcommon, int pass)
             status = 0;
     }
 
-    bitsFromVAL(prec);
+    if (!prec->udf)
+        bitsFromVAL(prec);
+    else {
+        /* Did user set any of the B0-B1F fields? */
+        epicsUInt8 *pBn = &prec->b0;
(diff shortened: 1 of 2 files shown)


mdavidsaver:
The test database and phoebus screen I used for testing.  [mbboDirectTest.db.txt](https://github.com/epics-base/epics-base/files/7093513/mbboDirectTest.db) [mbboDirectTest.bob](https://github.com/epics-base/epics-base/files/7093514/mbboDirectTest.bob.txt)

dirk-zimoch:
Isn‘t ˋOBITˋ the same thing which is called ˋOVALˋ in other record types?

mdavidsaver:
For mbboDirect, the the direct equivalent is `MLST`.

https://github.com/epics-base/epics-base/blob/9c1340d7d507e712501a4c056097280eab8ef200/modules/database/src/std/rec/mbboDirectRecord.c#L298-L299

`OBIT` is almost the same as `MLST`, except that it accounts for the `db_post_event()` happening during `dbPut()`.

https://github.com/epics-base/epics-base/blob/9c1340d7d507e712501a4c056097280eab8ef200/modules/database/src/ioc/db/dbAccess.c#L1393-L1395

This happens after `special(prec, 1)`, but before `dbProcess()` would be called.  (eg. by `dbPutField()`).

modules/database/src/std/rec/mbboDirectRecord.c
```diff
@@ -131,19 +139,7 @@ static long init_record(struct dbCommon *pcommon, int pass)
...
-        epicsUInt8 *pBn = &prec->b0;
-        int i;
-
-        for (i = 0; i < NUM_BITS; i++) {
-            *pBn++ = !! (val & 1);
-            val >>= 1;
-        }
-    }
-
+    bitsFromVAL(prec);
```
anjohnson: The previous code here has conditionals on both !UDF and OMSL == "supervisory" so this is definitely a behavior change which I think is wrong — the .db file could have set the Bn fields, but this call will always overwrite them. This code should probably be constructing VAL from the Bn fields in some configurations.

modules/database/src/std/rec/mbboDirectRecord.c
```diff
@@ -255,60 +243,41 @@ static long special(DBADDR *paddr, int after)
...
-                prec->val &= ~bit;
-
-            prec->udf = FALSE;
-            convert(prec);
+    if(after==0 && fieldIndex >= mbboDirectRecordB0 && fieldIndex <= mbboDirectRecordB1F) {
+        if(prec->omsl == menuOmslclosed_loop) {
+            /* To avoid confusion, reject changes to bit fields while in closed loop.
+             * Not a 100% solution as confusion can still arise if dset overwrites VAL.
+             */
+            return S_db_noMod;
```
anjohnson, line 268: Noting another behavior change, `special()` never rejected puts before.

modules/database/src/std/rec/mbboDirectRecord.c
```diff
@@ -255,60 +243,41 @@ static long special(DBADDR *paddr, int after)
...
-                epicsUInt8 oBn = *pBn;
-
-                *pBn = !! (val & 1);
-                if (oBn != *pBn)
-                    db_post_events(prec, pBn, DBE_VALUE | DBE_LOG);
-            }
-        }
-        break;
+        if(oobit!=prec->obit)
+            db_post_events(prec, &prec->obit, DBE_VALUE|DBE_LOG);
```
anjohnson: Were we trying to avoid calling `db_post_events()` from `special()`?

modules/database/src/std/rec/mbboDirectRecord.c
```diff
@@ -329,9 +298,24 @@ static void monitor(mbboDirectRecord *prec)
     if (prec->mlst != prec->val) {
         events |= DBE_VALUE | DBE_LOG;
         prec->mlst = prec->val;
+        db_post_events(prec, &prec->mlst, events);
```
anjohnson: I don't think we should be posting events on the MLST field (no other records do), and having said that I'm not convinced that doing so on the OBIT field makes much sense either.

modules/database/src/std/rec/mbboDirectRecord.c
```diff
@@ -234,6 +219,9 @@ static long process(struct dbCommon *pcommon)
         recGblGetTimeStampSimm(prec, prec->simm, NULL);
     }
 
+    /* update bits to reflect any change made by dset */
+    bitsFromVAL(prec);
```
anjohnson, line 240: Noting behavior change: Earlier versions didn't set the Bn fields when VAL was read from DOL (with OMSL=closed_loop).

anjohnson (changes requested):
One major issue in `init_record()` needs fixing. Some comments here are just pointing out where the record behavior will change, I'm not necessarily objecting to all of those changes though.

The record reference text also needs updating, it currently only refers to B0-BF (missing B10-B1F completely) and the big block of text needs rewriting. I'll probably work on those changes if you don't want to.

kasemir:
This does allow writing to VAL, and the B* fields update to match VAL.
But there are still some quirks.

Execute just this with a softIoc:
```
record(mbboDirect, "test")
{
   field(SCAN, "10 second")
}
```

.. and then start `camonitor test test.B0 test.B1` as soon as possible.

Initially, it will show
```
test                           <undefined> 0 UDF INVALID
test.B0                        <undefined> 0 UDF INVALID
test.B1                        <undefined> 0 UDF INVALID
```

Before this change, with base 7.0.6, after 10 seconds you see
```
test                           2021-09-02 08:04:34.280210 0  
```

The record processed, has a valid time stamp, no more alarm.

With the proposed update, the result is this:
```
test                           <undefined> 0 UDF INVALID
test.B0                        <undefined> 0 UDF INVALID
test.B1                        <undefined> 0 UDF INVALID
```

No time stamp, still in alarm, now with additional updates for the bit fields.

What I would like to see:
```
test                           2021-09-02 08:04:34.280210 0  
test.B0                           2021-09-02 08:04:34.280210 0  
test.B1                           2021-09-02 08:04:34.280210 0  
```

Record processes, gets time stamp and clears alarm, monitor on VAL updates, _and_ monitors on bit fields update.
Might be debatable if all bit fields should send monitors, since they didn't in the past, so that would be a change.
But not getting a valid time stamp and clearing the UDF when processed seems to be a new bug.

modules/database/src/std/rec/mbboDirectRecord.c
kasemir: I think that's a desirable behavior change. Setting Bn from VAL whenever VAL is set from DOL (closed_loop) or via channel access (supervisory) seems useful. Side effect would be additional monitors, but only if there are network clients, and they are very likely there because some display shows Bx, so they're expected to update.

modules/database/src/std/rec/mbboDirectRecord.c
mdavidsaver: Doesn't !UDF at this point imply that VAL has been initialized by `dbPutString()`?  (or explicitly cleared in a .db)

I'm looking for new cases for mbbioDirectTest.c.  What are the supported ways to initialize mbboDirect prior to a first `process()`?

```
record(mbboDirect, "A") {
    field(VAL, "3")
    field(B0, "1")
}
record(mbboDirect, "B") {
    field(B0, "1")
}
record(mbboDirect, "C") {
    field(VAL, "3")
}
```

Are there others?

modules/database/src/std/rec/mbboDirectRecord.c
mdavidsaver: I did most of my testing without this case.  It can be removed with few consequences.  The result would be that attempts to change B* while in closed loop will "flicker", with two updates in quick succession.  Of course the same would now happen if a device support overwrites VAL, which can't be detected ahead of time.

modules/database/src/std/rec/mbboDirectRecord.c
mdavidsaver: Yes, I said this.  Posting OBIT was primarily a debugging aid, which I can remove here and in process().

modules/database/src/std/rec/mbboDirectRecord.c
dirk-zimoch:
```
field(VAL, "3")
field(B0, "1")
```
is all well and good, but what is
```
field(VAL, "3")
field(B0, "0")
```
supposed to do? First set VAL to 3, then clear B0, resulting in VAL=2?
And what about
```
field(B0, "0")
field(VAL, "3")
```
Overwrite B0 with VAL=3? Having order dependent fields?

modules/database/src/std/rec/mbboDirectRecord.c
mdavidsaver: Again, debugging which can be removed.  Being able to use an OPI is much more convenient than the alternatives.  Lots of `printf()` or parsing frequent `dbgf()`.

mdavidsaver:
> ... Initially, it will show ...

Noted.  I was trying to follow the behavior of aoRecord, but missed where UDF is cleared after `convert()`.  (some confusing indentation...)

modules/database/src/std/rec/mbboDirectRecord.c
dirk-zimoch: +1

modules/database/src/std/rec/mbboDirectRecord.c
mdavidsaver: Right... I had intended to show inconsistency.  Poor choice of example by me.

Ordering at least doesn't matter (I think) as the two fields are independent prior to `init_record()`

mdavidsaver:
Also, to note a path not (yet) taken.  I considered adding an attribute to recordtype definitions to denote a VAL-like field, which would allow the automatic post from dbPut() to be omitted for fields other than VAL.  This would avoid the need to add OBIT.

https://github.com/epics-base/epics-base/blob/9c1340d7d507e712501a4c056097280eab8ef200/modules/database/src/ioc/db/dbAccess.c#L1391-L1395

Digression... It's silly that even a minor addition to the DBD parser is a breaking change.  Some forward compatibility would be a welcomed addition.  (as would a replacement for the VDCT...)

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: The order of setting fields in a DB file can't matter (with the existing IOC code at least) because only record type code could act on fields outside of dbCommon, and `prec->rset` doesn't get set until iocInit.

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: I agree that this change is desirable. It should be documented though.

modules/database/src/std/rec/mbboDirectRecord.c
dirk-zimoch: As all Bx and VAL default to 0, the only initialization that makes sense, in my opinion, is to OR all VAL and Bx fields.
```
field(VAL, "3")
field(B0, "0")
field(B4, "1")
```
Result: VAL=19 (0x13), B0,B1,B4=1

mdavidsaver: @dirk-zimoch How do you intend for this should interact with autosave?  Remember that autosave pass0 happens before `init_record(prec, 0)`.

dirk-zimoch: Having VAL in autosave and Bx in the template will really cause confusion.
So what? Overwrite Bx with VAL but only if VAL!=0?

mdavidsaver: > The order of setting fields in a DB file can't matter ...

wrt. `VAL` vs. `B*` this is true.  Excepting the special cases of `VAL` vs `UDF,` and `DTYP` vs. `INP/OUT` (and maybe others).

> ... template will really cause confusion.

Why yes, I believe it does.

> ... VAL!=0 ...

Looking at VAL alone doesn't give enough information.  eg. having non-zero B* by default, with autosave restoring zeros.

The one example I can find (of my own creation) is:

>    info(autosaveFields_pass0, "VAL")

So I guess the logic in `init_record()` should become:

```
if(!prec->udf) {
    bitsFromVAL(prec);
} else {
    // overwrite VAL from B*
}
```

anjohnson: > `VAL` vs `UDF`

True.

> `DTYP` vs. `INP/OUT`

I believe you fixed that when you rewrote the address parser:
```
woz$ cat dtyp.db 
record(lso, lso1) {
  field(OUT, "@stdout")
  field(DTYP, "stdio")
}
record(lso, lso2) {
  field(DTYP, "stdio")
  field(OUT, "@stdout")
}
woz$ softIoc -d dtyp.db 
Starting iocInit
## EPICS R7.0.6.1-DEV
## Rev. R7.0.6-23-g21e7e4ddfbfbcf8ca5c6-dirty
epics> dbpr lso1 1
AMSG:               APST: On Change     ASG :               BKPT: 00            
DESC:               DISA: 0             DISP: 0             DISS: NO_ALARM      
DISV: 1             DOL : CONSTANT      DTYP: stdio         EVNT:               
FLNK: CONSTANT      LEN : 0             MPST: On Change     NAME: lso1          
NAMSG:              OMSL: supervisory   OUT : INST_IO @stdout                   
PACT: 0             PHAS: 0             PINI: NO            PRIO: LOW           
PUTF: 0             RPRO: 0             SCAN: Passive       SDIS: CONSTANT      
SEVR: INVALID       SIML: CONSTANT      SIMM: NO            SIOL: CONSTANT      
SIZV: 41            SSCN: <nil>         STAT: UDF           TPRO: 0             
TSE : 0             TSEL: CONSTANT      UDF : 1             UDFS: INVALID       
VAL : PTR 0x7fed6065fed0                
epics> dbpr lso2 1
AMSG:               APST: On Change     ASG :               BKPT: 00            
DESC:               DISA: 0             DISP: 0             DISS: NO_ALARM      
DISV: 1             DOL : CONSTANT      DTYP: stdio         EVNT:               
FLNK: CONSTANT      LEN : 0             MPST: On Change     NAME: lso2          
NAMSG:              OMSL: supervisory   OUT : INST_IO @stdout                   
PACT: 0             PHAS: 0             PINI: NO            PRIO: LOW           
PUTF: 0             RPRO: 0             SCAN: Passive       SDIS: CONSTANT      
SEVR: INVALID       SIML: CONSTANT      SIMM: NO            SIOL: CONSTANT      
SIZV: 41            SSCN: <nil>         STAT: UDF           TPRO: 0             
TSE : 0             TSEL: CONSTANT      UDF : 1             UDFS: INVALID       
VAL : PTR 0x7fed6065ffa0                
```

> Looking at VAL alone doesn't give enough information.

The contents of VAL can't tell you what to do. The original code in `init_record()` uses both UDF and OMSL to determine whether VAL or Bn should take precedence. If VAL was set either from a constant in DOL or by the device support's `init_record()` the UDF field will be FALSE and VAL takes precedence, so in `supervisory` mode it initializes Bn from VAL. Note that in this code `process()` constructs VAL from Bn when in `supervisory` mode.

In your new version the `process()` routine _always_ sets Bn from VAL, so your `init_record()` needs to initialize VAL from Bn when OMSL is `closed_loop` (unless UDF is FALSE, which indicates that a constant in DOL or the device support provided a value so it must then call `bitsFromVAL()`).

That's the change I was referring to in my original comment on this piece of code.

anjohnson:
Question for @dirk-zimoch and @kasemir since you both reported issues here — should we back-port the fix to the 3.15 branch, or is a fix on just 7.0 acceptable?

kasemir:
Just the latest 7... is fine with me.

dirk-zimoch:
I am currently not using 3.15. But if 7 keeps losing compatibility with VxWorks and if we keep using VxWorks, I may have to use 3.15. In that case I would prefer to have this change ported back.

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: Removing as agreed.

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: Removing...

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: The situation that originally concerned me here was where someone had set OMSL to `closed_loop` but hadn't actually set DOL. As long as UDF is FALSE writing to the B* fields previously succeeded, but didn't make VAL change at all, then later flipping OMSL rewrote the B* fields from VAL. I agree the old behavior is confusing, so I will document the change, although I can see us getting complaints since rejecting a put (instead of just ignoring it) may trigger a pop-up on a GUI display whenever a user flips a bit for a record that's in the wrong mode, and we aren't providing any way to prevent that.

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: Documenting.

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: I withdraw my suggestion above that OMSL matters in `init_record()` and instead agree with Michael's proposal (took me a while, I know...). I'm making the call to `bitsFromVAL(prec)` conditional on `!prec->udf`, and adding an `else` clause which converts the B* fields to an `epicsUInt32`. If the result is non-zero it writes it into `VAL` and clears `UDF`.

This means that setting VAL either directly, through DOL, or by device support read-back would take precedence over anything set in the B* fields. If anything writes to the VAL field (and clears UDF) before a record is initialized, its B* fields will be updated from VAL by `init_record()`.

We don't have any way to find out if any of the the B* fields were actually set if they were all zero, and I think it's important that records should come up as UDF if they haven't actually been given a value.

Did I miss anything?

anjohnson (commented):
I'm about to push 2 commits implementing the changes discussed here, I hope they are non-controversial. I have some more mostly cosmetic changes including updates to the record reference in the .dbd.pod which I haven't committed yet, and I haven't done anything about adding test code for the new initialization cases (help welcomed!).

Need sleep...

modules/database/src/std/rec/mbboDirectRecord.c
```diff
@@ -131,16 +139,21 @@ static long init_record(struct dbCommon *pcommon, int pass)
             status = 0;
     }
 
-    if (!prec->udf &&
-        prec->omsl == menuOmslsupervisory) {
-        /* Set initial B0 - B1F from VAL */
-        epicsUInt32 val = prec->val;
+    if (!prec->udf)
+        bitsFromVAL(prec);
+    else {
```
mdavidsaver, line 144: I am slightly surprised by the absence of brackets for the `if()`.

mdavidsaver (commented):
@anjohnson Your changes are ok by me.

modules/database/src/std/rec/mbboDirectRecord.c
anjohnson: What rule do you think we should follow? Your `bitsFromVAL()` routine defines a `for ()` loop without using braces so you appear to allow some exceptions to a "braces everywhere" rule.

ralphlange: Personally, I just keep if and else parts the same.
I do both, but either all without or all with braces.

mdavidsaver: Ditto.

You know I'm not a fan of strict style rules.  I commented because I'd not previously perceived or perpetrated this particular permutation.  (that I know of)

anjohnson: Look just below the `CONTINUE:` label in the `process()` routine for a somewhat older example. There's also a similar one where there are braces after the `if ()` but not the following `else if ()` clause immediately above this line.

PR #200, opened 2021-09-11
Fix issue when osdThreadGetTimer() called from non EPICS threads

FreddieAkeroyd:
This fixes the issue reported in https://epics.anl.gov/tech-talk/2021/msg01691.php

The function need to use epicsThreadImplicitCreate() to handle the case of if it is called from a non EPICS thread - this approach is already used in epicsThreadSuspendSelf, epicsThreadGetPrioritySelf, epicsThreadGetIdSelf and epicsThreadGetNameSelf.
This repeated logic should probably be moved into a separate function, but maybe that is for a separate PR?


FreddieAkeroyd:
Maybe I do a separate PR for the bigger refactor and branch that PR off 3.15 for it to be upmerged later?

anjohnson:
Closing this, I'll be merging up the refactor PR from 3.15 instead shortly.

PR #201, opened 2021-09-11
Refactor to use common epicsThreadGetParamWIN32() function

FreddieAkeroyd:
This is the refactor mentioned in #200 and also corrects the issue mentioned there

### DIFF
--- a/src/libCom/osi/os/WIN32/osdThread.c
+++ b/src/libCom/osi/os/WIN32/osdThread.c
@@ -780,10 +780,10 @@ epicsShareFunc int epicsShareAPI epicsThreadIsSuspended ( epicsThreadId id )
 HANDLE osdThreadGetTimer()
 {
     win32ThreadParam * pParm = epicsThreadGetParamWIN32();
-    if (pParam) {
+    if (pParm) {
         return pParm->timer;
     } else {
-	return NULL;
+        return NULL;
     }
 }
 
--- a/src/libCom/osi/os/WIN32/osdThread.c
+++ b/src/libCom/osi/os/WIN32/osdThread.c
@@ -477,21 +477,6 @@ epicsShareFunc unsigned int epicsShareAPI
     return stackSizeTable[stackSizeClass];
 }
 
-void epicsThreadCleanupWIN32 ()
-{
-    win32ThreadGlobal * pGbl = fetchWin32ThreadGlobal ();
-    win32ThreadParam * pParm;
-
-    if ( ! pGbl )  {
-        fprintf ( stderr, "epicsThreadCleanupWIN32: unable to find ctx\n" );
-        return;
-    }
-
-    pParm = ( win32ThreadParam * ) 
-        TlsGetValue ( pGbl->tlsIndexThreadLibraryEPICS );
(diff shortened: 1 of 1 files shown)


anjohnson:
This looks good on a brief visual inspection.

I noticed that on the 7.0 branch libcom/src/osi/os/WIN32/osdFindSymbol.c is using `static __declspec(thread)` (`static __thread` for GCC) — should we be replacing calls to the MS-specific APIs for thread-local storage with this decorator (on 7.0 at least)?

It would be nice to have some unit tests that start non-EPICS threads and prove that any routines like `epicsThreadSleep()` which now rely on internal thread-local variables do actually work properly in that context. Obviously it needs OS-specific code to start the non-EPICS threads, but that shouldn't be too hard, and it would give us somewhere to keep track of all the APIs that actually rely on thread-local storage internally.

mdavidsaver:
> ... should we be replacing calls to the MS-specific APIs ...

imo. this would be a step backwards.  `__declspec(thread)` is the MS backwards compatible spelling of the C++11 `thread_local` qualifier.  This works and is easier to use...

FreddieAkeroyd:
Microsoft actually recommend `thread_local` now, but it does require compiler support - I recall we still need to support VS2010 and that doesn't appear to support `thread_local`.  There used to issues with `__declspec(thread)` and using `LoadLibrary`  https://docs.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=msvc-160  but that seems to have been fixed in Vista and above. We could create our own `EPICS_THREAD_LOCAL` that gets set to the right thing depending on os and compiler, or just wait until we require  C++11 / C11 and then change to `thread_local`.

Yes, some unit tests using non EPICS threads would be a good idea - do we need an `epicsThreadRawCreate` or similar such private helper function? Sorry for introducing a bug, I didn't appreciate the importance of calling `epicsThreadImplicitCreate`  - hopefully the addition of a dedicated `epicsThreadGetParamWIN32` will stop this happening again if a different function wants to access tls parameters

mdavidsaver:
> ... `epicsThreadRawCreate` ...

Consider me skeptical of this idea.  Certainly I don't see a function like this becoming part of libCom.  Adding as a test code helper would need to be weighed against explicitly coding w/ `#ifdef` for the one test case where it would be used.

Remember that tests have ready access to one non-epics thread... `main()`.

mdavidsaver (approved):
This change looks fine to me.

FreddieAkeroyd:
Yes, I wouldn't see  `epicsThreadRawCreate` as a general part of libcom, but you are right - if all such thread tests are together then the code could be `#ifdef` in that particular test main file. I was thinking it might be useful to be able to create raw threads for testing in other modules, but that would not be base unit tests so probably not.

anjohnson:
The main thread runs the platform's epicsThread initialization code, so I don't think that really helps with testing our APIs on other non-EPICS threads. A single test file with OS-specific `#if` sections that define a thread creation routine for the tests to call should be sufficient (with a `#else` section that triggers `testSkip()` or `testTodoBegin() .. testTodoEnd()` for OS arch's that haven't been implemented there yet).

PR #203, opened 2021-09-24
IPv6: part 1, Linux and MacOs

tboegi:
Prepare EPICS Base to handle IPv6:

The socket interface got a new structure, "osiSockAddr46"
There is a compiler flag, EPICS_HAS_IPV6, which must be set to 1,
if IPv6 is enabled at all.
osiSock.c got new functions, that deal with IPv6:

epicsSocket46GetDefaultAddressFamily()
epicsSocket46Create()
epicsSocket46Recvfrom()
epicsSocket46Sendto()
epicsSocket46Connect()
epicsSocket46Accept()

They all work with osiSockAddr46.
Unless -DEPICS_HAS_IPV6=1 is used, there is no change (yet).

Backward compatiblity:
The "union osiSockAddr" is left unchanged, so that code relying
on its structure is not affected.
However, osiSockAddr is too short to hold IPv6 addresses.
Code that that needs IPv6 can do it's own business, or use use the
epicsSocket46xxx functions mentioned above.

For EPICS base this has been done. Most variables using osiSockAddr46
have been renamed. E.g. from addr into addr46. Not all of them.

Start testing, using IPv6:
  caget, camonitor can work via IPV6 like this:
  $ export EPICS_CA_AUTO_ADDR_LIST=NO
  $ export EPICS_CA_ADDR_LIST='[2001:db8:1:0:0:1:2:3]'
  cainfo IOC:m1

  Using link local addresses:
    Find out the link-local address of your IOC.
    e.g. fe80::3958:418:65b8:230c
    Find out the interface of your client, sitting on the same network
    switch as the IOC.
    $ ifconfig
    # or
    $ ip addr show
    in my case it is "en0"
    # ping the IOC:
    $ ping6 fe80::3958:418:65b8:230c
      ping6: sendmsg: No route to host
    # That didn't work, we new this.
    # Compose a proper adress, combining the remote IPv6 with
    # our local interface name:
    $ ping6 fe80::3958:418:65b8:230c%en0
    16 bytes from fe80::3958:418:65b8:230c%en0
    # with this knowledge, we can use cainfo:
    $ export EPICS_CA_AUTO_ADDR_LIST=NO
    $ export EPICS_CA_ADDR_LIST='[fe80::3958:418:65b8:230c%en0]'
    $ cainfo IOC:m1

TODO, unordered:
- Some cleanup...
- Check what ipAddrToA() and sockAddrToA() are doing. Do we need both ?
- Find out how IPv6 multicast is working
- Get an official multicast address from IANA ?
- Do we need to change the repeater ? Or is multicast workig "as is" ?
- VxWorks, RTEMS, Windows, other OS ?
- EPICS Base 7 only, or 3.15 as well ?
- Define new EPICS variables.
- pvAccess
- gateways
- Check access control (based on IP addresses)
- asyn (I have made a "prof of concept", code on request)

### DIFF
--- a/modules/libcom/src/osi/osiSock.h
+++ b/modules/libcom/src/osi/osiSock.h
@@ -20,13 +20,40 @@
 #include "osdSock.h"
 #include "ellLib.h"
 
+/* Enable it for debugging */
+/* #define NETDEBUG */
+
+/* We assume that darwin and linux can handle IPv6 with the current code base */
+#if defined ( darwin ) || defined ( linux )
+#ifndef RTEMS_LEGACY_STACK
+
+#ifndef EPICS_HAS_IPV6
+#define EPICS_HAS_IPV6 1
+#endif
+
+#endif
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define EPICSSOCKET_CONNECT_IPV4            (1<<4)
+#define EPICSSOCKET_CONNECT_IPV6            (1<<6)
+
+
 struct sockaddr;
 struct sockaddr_in;
 struct in_addr;
+
+typedef union osiSockAddr46 {
+    struct sockaddr_in  ia;
+    struct sockaddr     sa;
+#if EPICS_HAS_IPV6
+    struct sockaddr_in6 in6 /* in6 for IPv6 */;
+#endif
+} osiSockAddr46;
+
 
 LIBCOM_API SOCKET epicsStdCall epicsSocketCreate ( 
     int domain, int type, int protocol );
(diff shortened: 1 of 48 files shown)


mdavidsaver:
It would be helpful for me if you could describe what changes are made in terms of what sockets are created.  eg. what does `netstat -tulpn` show a running `softIoc`?

>     * Check what ipAddrToA() and sockAddrToA() are doing. Do we need both ?

`sockAddrToA()` is calling `ipAddrToA()`.  If this were changed, then `ipAddrToA()` probably could become an inline or macro shim calling `sockAddrToA()`.

>     * Find out how IPv6 multicast is working

I've been looking at ipv6 mcast in the context of PVA, and it is a __major__ pain.  fyi. so far I haven't been able to make use ipv6 multicast through the loopback interface.  This makes (reliable) unit test difficult, and probably spells the end of the PVA local multicast hack...

Some helpful commands for exploring mcast routing on Linux...

```sh
$ ip route show type multicast table all
multicast ff00::/8 dev br0 table local proto kernel metric 256 pref medium
multicast ff00::/8 dev wlan0 table local proto kernel metric 256 pref medium
$ ip route get 224.0.0.128
multicast 224.0.0.128 dev wlan0 src 192.168.1.2 uid 1000 
    cache <mc>
$ ip route get oif lo 224.0.0.128
multicast 224.0.0.128 dev lo src 192.168.1.2 uid 1000 
    cache <mc>
$ ip route get ff01::1:ff00:1
multicast ff01::1:ff00:1 from :: dev br0 table local proto kernel src fe80::XXXX:XXXX:XXXX:XXXX metric 256 pref medium
$ ip route get oif lo ff01::1:ff00:1
RTNETLINK answers: Network is unreachable
```

> $ export EPICS_CA_ADDR_LIST='[fe80::3958:418:65b8:230c%en0]'

Noticing the `en0` here.  One part of the pain of IPv6 is the usage of interface indices with relation to the address `sin6_scope_id`.  Based on my experience so far, I see numeric interface indices as too volatile to be treated as long term configuration.  eg. sometimes (but not always) my laptop comes out of suspend with different indices for the wifi adapter (a USB device).

wrt. PVXS I'm planning to keep a time based cache of the interface index, name, and address(es) list (cf. https://github.com/mdavidsaver/pvxs/pull/21)

modules/libcom/src/osi/osiSock.h
```diff
@@ -20,14 +20,43 @@
...
 struct in_addr;
 
+typedef union osiSockAddr46 {
+    struct sockaddr_in  ia;
+    struct sockaddr     sa;
+#if EPICS_HAS_IPV6
+    struct sockaddr_in6 in6 /* in6 for IPv6 */;
+    uint8_t             raw_bytes[28]; /* 2 fam, 2 port, 4 flowinfo, 16 Ipv6, 4 scope */
+#endif
+} osiSockAddr46;
```
mdavidsaver, line 55: Maybe better to use `union sockaddr_storage` when available, and define it when not.  Also, `raw_bytes` looks like an invitation for portability bugs.

modules/libcom/src/osi/osiSock.h
```diff
@@ -20,14 +20,43 @@
...
+
+
 struct sockaddr;
 struct sockaddr_in;
 struct in_addr;
 
+typedef union osiSockAddr46 {
+    struct sockaddr_in  ia;
+    struct sockaddr     sa;
+#if EPICS_HAS_IPV6
```
mdavidsaver, line 52: Try replacing `#if EPICS_HAS_IPV6` with `#ifdef AF_INET6`.  This should avoid the need for makefile games.

tboegi:
Thanks for the comments so far -
I'll address them in the next round - in 1-2 weeks or so

tboegi:
The Linux box shows this:
sudo netstat -tulpn | grep 506
tcp6       0      0 :::5064                 :::*                    LISTEN      29074/../../bin/lin 
udp6       0      0 :::5064                 :::*                                29074/../../bin/lin 
udp6       0      0 :::5065                 :::*                                28871/caRepeater    

(And tcp6 will accept IPv4 as well)
The Mac is slighty different:
netstat -p udp -a -n | grep 506
udp46       0      0 *:5064                *.*
udp46       0      0 *:5065                *.*
(And similar for tcp)

tboegi:
Some basic multicasting is found here:
https://github.com/epics-base/epics-base/pull/205

tboegi:
Re-open this PR: 
This is a forced push with a rebase onto the latest origin/7.0 branch

The incremental development with all commits, merges and all forth-and-back can be seen in #205

tboegi:
Basically we can IPv6 on Mac/Linux.
The repeater does not handle IPv6 at all: 
Should we extend the message type for CA_PROTO_RSRV_IS_UP to carry
an optional IPv6 address ?

Otherwise I would like to pause here for a while, get feedback, and ask people to test.

tboegi:
This is not the way to go - see #205 for the full development

PR #204, opened 2021-10-03
dbnd filter pass through DBE_ALARM|DBE_PROPERTY

mdavidsaver:
From https://epics.anl.gov/tech-talk/2021/msg01761.php

```
one gets the following after ioc startup.

ACM:TEST:AO.{dbnd:{abs:1.5}}   <undefined> 0 UDF INVALID

one gets nothing after writing  the value 0 to ACM:TEST:A0

one gets nothing after writing  the value 1 to ACM:TEST:A0

one gets a monitor when writing the value 2 to ACM:TEST:A0

ACM:TEST:AO.{dbnd:{abs:1.5}}   2021-10-03 05:17:38.406470 2

What does that mean for a display manager? Should not the real value 0 then be given later from the monitor when 0 is written?
```

It occurs to me that with 235f8ed2fb85270a1b9edddbff6a1c5b10f484b9 the dbnd filter now has enough information to distinguish, and pass through, events with an alarm state change.

### DIFF
--- a/modules/database/src/std/filters/dbnd.c
+++ b/modules/database/src/std/filters/dbnd.c
@@ -20,7 +20,7 @@
 #include <chfPlugin.h>
 #include <recGbl.h>
 #include <epicsExit.h>
-#include <db_field_log.h>
+#include <dbAccess.h>
 #include <epicsExport.h>
 
 typedef struct myStruct {
@@ -81,8 +81,8 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
         status = dbFastGetConvertRoutine[pfl->field_type][DBR_DOUBLE]
                  (localAddr.pfield, (void*) &val, &localAddr);
         if (!status) {
-            send = 0;
-            recGblCheckDeadband(&my->last, val, my->hyst, &send, 1);
+            send = pfl->mask & ~(DBE_VALUE|DBE_LOG);
+            recGblCheckDeadband(&my->last, val, my->hyst, &send, pfl->mask & (DBE_VALUE|DBE_LOG));
             if (send && my->mode == 1) {
                 my->hyst = val * my->cval/100.;
             }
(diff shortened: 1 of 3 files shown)


mdavidsaver:
I've realized that 235f8ed2fb85270a1b9edddbff6a1c5b10f484b9 was initializing `mask` with the client provided event mask of all interesting events.  So it is necessary in `db_post_events()` to overwrite this with the actual event mask.

Note that I considered adding an argument to `db_create_event_log()` for the mask, but decided against it atm. to keep this change small.

modules/database/src/std/filters/dbnd.c
```diff
@@ -81,8 +81,8 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl) {
         status = dbFastGetConvertRoutine[pfl->field_type][DBR_DOUBLE]
                  (localAddr.pfield, (void*) &val, &localAddr);
         if (!status) {
-            send = 0;
-            recGblCheckDeadband(&my->last, val, my->hyst, &send, 1);
+            send = pfl->mask & ~(DBE_VALUE|DBE_LOG);
+            recGblCheckDeadband(&my->last, val, my->hyst, &send, pfl->mask & (DBE_VALUE|DBE_LOG));
             if (send && my->mode == 1) {
```
mdavidsaver, line 86: @ralphlange Should this conditional change as well?  eg. into
```c
if ((send & (DBE_VALUE|DBE_LOG)) && my->mode == 1) {
```

mdavidsaver:
imo. This seems like a low risk change.  Any objection to my merging this week?

anjohnson:
I had to build this version and play with it to see how the filter behaves differently. It seems to be doing what I expect it to, but the change in behavior should be described in the documentation. I'm not sure about the additional conditional, it may make sense but I don't have as much of a feel for relative mode myself.

I don't think this should go into 7.0.6.1 (which I will release after merging #199). We should look at the other filters too since they might need similar changes, or even additional parameters to control how they filter alarms.

modules/database/src/std/filters/dbnd.c
AntonMezger: Hi,

Could you test once more, because I thought it worked fine, but it seems that the dbnd does not work any more. I get now all changes, when I replace dbnd.c with this patch

mdavidsaver:
Merged as of b1d9c57101557a41df4a41ec02e4bc8c3ca65266

PR #205, opened 2021-10-06
IPv6: part 1

tboegi:
Prepare EPICS Base to handle IPv6:

The socket interface got a new structure, "osiSockAddr46"
There is a compiler flag, EPICS_HAS_IPV6, which must be set to 1,
if IPv6 is enabled at all.
osiSock.c got new functions, that deal with IPv6:

epicsSocket46GetDefaultAddressFamily()
epicsSocket46Create()
epicsSocket46Recvfrom()
epicsSocket46Sendto()
epicsSocket46Connect()
epicsSocket46Accept()

They all work with osiSockAddr46.
Unless -DEPICS_HAS_IPV6=1 is used, there is no change (yet).

Backward compatiblity:
The "union osiSockAddr" is left unchanged, so that code relying
on its structure is not affected.
However, osiSockAddr is too short to hold IPv6 addresses.
Code that that needs IPv6 can do it's own business, or use use the
epicsSocket46xxx functions mentioned above.

For EPICS base this has been done.
To make things more clear, most variables using osiSockAddr46 have been
renamed, e.g. from addr into addr46.
(Probably I missed them them)

A first version of a IPv6 howtow is here:
documentation/ipv6-howto.txt
More information will be added, while we go ahead and learn

TODO, unordered:
- Some cleanup...
- sockAddrToA() does not use reverse DNS for IPv6
- The repeater can not repeat IPv6 becons (casw)
- VxWorks, RTEMS, Windows, other OS ?
- EPICS Base 7 only, or 3.15 as well ?
- pvAccess
- gateways
- Check access control (based on IP addresses)
- asyn (I have made a "prof of concept", code on request)


mdavidsaver:
@tboegi fyi, you could have force push to #203 instead of switching to a new PR.  Reusing a PR is preferable as it results in fewer automatic emails for those who are spectating.

anjohnson:
Hi Torsten,

We discussed this PR in today's EPICS Core Developers meeting. Technically your work looks very interesting, and with my APS hat on I'm personally very happy to see that running CA over IPv6 is feasible, instead of PVA being the only IPv6-capable protocol that EPICS might support in the future. What are the limitations that you have discovered about using the result? How easily could a site switch over to IPv6? Would this permit parallel operations over both IP versions during any such change-over? Are there any fundamental changes related to network topology, name broadcasts or beacons which newcomers to IPv6 would need to understand?

In our meeting we agreed that to be mergeable an IPv6 extension will need to provide full IPv4 backwards compatibility for application code that calls the existing APIs, and should document how to migrate external modules (such as Asyn) that call the existing APIs and headers from libCom over to the new API to use IPv6. Will you be able to continue working on this? It obviously isn't needed urgently (and I'm sure you're as busy as we are) so even if it takes a year or more to complete we would welcome your continued efforts.

For ease of code review when you get to that stage we would like to see it split it into multiple PRs in much smaller logical chunks which can be merged in some defined order, maybe starting with one that just adds the new libCom API, or even just some well-defined part of it, with tests for the functionality provided within each PR. This way we could release Base versions that might allow say Asyn to connect to devices over tcp6 before the rest of the IOC is ready to support it.

As the EPICS core group's "naming czar" I would like to see some changes to the names for the new API though — I have never liked our use of the "osi" prefix (short for Operating System Independent) in public API names, I've always thought that prefix should for internal interfaces only, not for data types and routines that would be used by code outside of Base itself.  My preference for new API names would thus be to use "epicsSock" as the basic name prefix for a new sockets API, and to only have the digits "4" and "6" used in identifiers that are specific to a particular IP version.

Thanks,

- Andrew

tboegi:
>Hi Torsten,

Hej Andrew
> ...instead of PVA being the only IPv6-capable protocol that EPICS might support in the future.

My idea is that the code for CA could be a source of inspiration,
may be a blueprint for other IPv6 work, like PVA.

>What are the limitations that you have discovered about using the result?

As of today, we send out beacons via IPv6, but they are not received (yet)
Work is ongoing (but not yet pushed to github)

>How easily could a site switch over to IPv6?

How many "network nodes" do we need to migrate ?
How much time does it need to compile and deploy a new version of EPICS base ?
How much time does it need to switch from CA to PVA ?

If we assume a well working IPv6 network:
Do we have firewalls ?
Are they well configured ?
Do we need a PV gateway ?
It could take everything between 4 hours and 4 weeks for one beam line.


>Would this permit parallel operations over both IP versions during any such change-over?

Yes

>Are there any fundamental changes related to network topology,
>name broadcasts or beacons which newcomers to IPv6 would need to understand?

Yes, definitively. IPv6 does not use broadcasts. We have link-local addresses


>In our meeting we agreed that to be mergeable an IPv6 extension will need to provide
>full IPv4 backwards compatibility for application code that calls the existing APIs,
>and should document how to migrate external modules (such as Asyn)
asyn needs more and different APIs then what EPICS base/CA needs.

I have done a prototype to use asyn with IPv6, that is on hold for the moment.

>that call the existing APIs and headers from libCom over to the new API to use IPv6.

OK

>Will you be able to continue working on this?

I think so, somewhat in the background.

>It obviously isn't needed urgently (and I'm sure you're as busy as we are)
>so even if it takes a year or more to complete we would welcome your continued efforts.

OK, that sounds like a plan.

>For ease of code review when you get to that stage we would like
>to see it split it into multiple PRs in much smaller logical
>chunks which can be merged in some defined order,
>maybe starting with one that just adds the new libCom API,
>or even just some well-defined part of it,
>with tests for the functionality provided within each PR.

That is a good idea, however:
I would like that someone else test the latest IPv6 branch "as is",
verifies that this is working, and then we can nice-ify the commits.
And did we agree, who is writing the tests ?


>This way we could release Base versions that might allow say Asyn to connect
>to devices over tcp6 before the rest of the IOC is ready to support it.

Let my think about this.


>As the EPICS core group's "naming czar"
>I would like to see some changes to the names for the new API though
>— I have never liked our use of the "osi" prefix (short for Operating System Independent)
>in public API names,
>I've always thought that prefix should for internal interfaces only,
>not for data types and routines that would be used by code outside of Base itself.
>My preference for new API names would thus be to use "epicsSock"
>as the basic name prefix for a new sockets API, and to only have the digits "4" and "6"
>used in identifiers that are specific to a particular IP version.

Agreed.
Let me finish the beacon-not-received problem.
After that, we can easily do a "bulk-rename" of functions.
And may be, a kind of online presentation ?

tboegi:
@anjohnson @ralphlange @mdavidsaver 
I started with a very small preparation step.
Feedback about names is welcome

https://github.com/epics-base/epics-base/pull/221

PR #206, opened 2021-10-12
Added Heinz's new osdEvent.c to RTEMS-posix

### LINKED ISSUE #202: Posix implementation of epicsEvent is unsuitable for RTEMS-posix targets
The code in libCom/osi/os/posix/osdEvent.c was written for workstation OSs that don't support true ISRs. The `struct epicsEventOSD` in that file which implements the epicsEvent object contains both a `pthread_mutex_t` and a `pthread_cond_t`. Unfortunately neither of these can be used in ISR context on an RTEMS-posix target. The `epicsEventTrigger()` routine must be callable from ISR context when the target supports them.

Discovered by @hjunkes and @mdavidsaver.

anjohnson:
Fixes #202

### DIFF
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
@@ -62,7 +62,7 @@ epicsEventWait(epicsEventId pSem)
 }
 
 LIBCOM_API epicsEventStatus
-epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
+epicsEventWaitWithTimeout(epicsEventId pSem, double timeout)
 {
     int sc;
     rtems_interval delay;
@@ -71,22 +71,22 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
     if (!rate)
         return epicsEventError;
 
-    if (timeOut <= 0.0) {
+    if (timeout <= 0.0) {
         sc = rtems_binary_semaphore_try_wait(&pSem->rbs);
         if (!sc)
             return epicsEventOK;
         else
             return epicsEventWaitTimeout;
     }
-    else if (timeOut < (double) UINT32_MAX / rate) {
-        delay = timeOut * rate;
+    else if (timeout < (double) UINT32_MAX / rate) {
+        delay = timeout * rate;
         if (delay == 0) {
-            /* 0 < timeOut < 1/rate; round up */
+            /* 0 < timeout < 1/rate; round up */
             delay = 1;
         }
     }
     else {
-        /* timeOut is NaN or too big to represent; wait forever */
+        /* timeout is NaN or too big to represent; wait forever */
         delay = RTEMS_NO_TIMEOUT;
     }
 
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
@@ -12,7 +12,7 @@
 
 #include <stdio.h>
 #include <malloc.h>
-#include <limits.h>
+#include <stdint.h>
 #include <rtems.h>
 #include <rtems/error.h>
 
@@ -66,7 +66,7 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
 {
     int sc;
     rtems_interval delay;
-    int rate = rtems_clock_get_ticks_per_second();
+    rtems_interval rate = rtems_clock_get_ticks_per_second();
 
     if (!rate)
         return epicsEventError;
@@ -78,11 +78,16 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
         else
             return epicsEventWaitTimeout;
     }
-    else if (timeOut >= (double) INT_MAX / rate) {
-        delay = 0;
+    else if (timeOut < (double) UINT32_MAX / rate) {
+        delay = timeOut * rate;
+        if (delay == 0) {
+            /* 0 < timeOut < 1/rate; round up */
+            delay = 1;
+        }
     }
     else {
-        delay = timeOut * rate;
+        /* timeOut is NaN or too big to represent; wait forever */
+        delay = RTEMS_NO_TIMEOUT;
     }
 
     sc = rtems_binary_semaphore_wait_timed_ticks(&pSem->rbs, delay);
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
@@ -1,0 +1,111 @@
+/*************************************************************************\
+* Copyright (c) 2021 Fritz Haber Institute, Berlin
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+/*
+ * RTEMS osdEvent.c
+ *      Author: H. Junkes
+ *              junkes@fhi.mpg.de
+ */
+
+#include <stdio.h>
+#include <malloc.h>
+#include <limits.h>
+#include <rtems.h>
+#include <rtems/error.h>
+
+#include <rtems/thread.h>
+
+#include "libComAPI.h"
+#include "epicsEvent.h"
+
+typedef struct epicsEventOSD {
+    rtems_binary_semaphore rbs;
+} epicsEventOSD;
+
+/*
+ * Create a simple binary semaphore
+ */
+LIBCOM_API epicsEventId
+epicsEventCreate(epicsEventInitialState initialState)
+{
+    epicsEventOSD *pSem = malloc (sizeof(*pSem));
+
+    if (pSem) {
+        rtems_binary_semaphore_init(&pSem->rbs, NULL);
+        if (initialState)
+            rtems_binary_semaphore_post(&pSem->rbs);
+    }
+    return pSem;
+}
+
+LIBCOM_API void
+epicsEventDestroy(epicsEventId pSem)
+{
+    rtems_binary_semaphore_destroy(&pSem->rbs);
+}
+
+LIBCOM_API epicsEventStatus
+epicsEventTrigger(epicsEventId pSem)
+{
+    rtems_binary_semaphore_post(&pSem->rbs);
+    return epicsEventOK;
+}
+
+LIBCOM_API epicsEventStatus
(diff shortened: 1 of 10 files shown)


modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
```diff
@@ -0,0 +1,111 @@
...
+        if (!sc)
+            return epicsEventOK;
+        else
+            return epicsEventWaitTimeout;
+    }
+    else if (timeOut >= (double) INT_MAX / rate) {
+        delay = 0;
+    }
+    else {
+        delay = timeOut * rate;
```
mdavidsaver: Probably not so important, but also probably not intended.  As I read it, a `NaN` would fall through to the `else` and result in UB when converted to `int`.

modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
```diff
@@ -0,0 +1,111 @@
...
+    rtems_binary_semaphore_wait(&pSem->rbs);
+    return epicsEventOK;
+}
+
+LIBCOM_API epicsEventStatus
+epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
+{
+    int sc;
+    rtems_interval delay;
+    int rate = rtems_clock_get_ticks_per_second();
```
mdavidsaver: As I read it, `rtems_clock_get_ticks_per_second()` will evaluate as `uint32_t`.  The `tick` argument of `rtems_binary_semaphore_wait_timed_ticks()` is also `uint32_t`.  I don't think the mixing of signed int here would cause a problem, but it certainly doesn't help.

mdavidsaver (commented):
I have a couple of comments about `epicsEventWaitWithTimeout()` since we've had, and still have, problems with converting time to ticks.  None of these are critical, and this change is probably fine as is.

mdavidsaver:
A digression wrt. converting time to ticks.

I tried to track through what `pthread_cond_timedwait()` does.  I _think_ this happens in `_Watchdog_Ticks_from_timespec()` were it looks like the tick counter is actually seconds and nanoseconds packed as a uint64_t.  So I wondered if we could use this conversion which is probably faster and more accurate then out floating point arithmetic.

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/posix/src/condtimedwait.c#L38

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/posix/src/condwaitsupp.c#L112

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/include/rtems/score/threadqimpl.h#L217

Now back out to `_POSIX_Condition_variables_Wait_support()`, which later calls `_Thread_queue_Context_set_enqueue_callout()`.  I'm assuming the `CLOCK_REALTIME` code path.  Which comes eventually to

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/score/src/threadqtimeout.c#L120

Here I see a call to `_Timecounter_Getnanotime()` for which by some slight of hand is actually defined as `getnanotime()`.

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/score/src/kern_tc.c#L619-L622

The value being used isn't simply incremented, so I'll assume this is the same packed sec+ns.

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/score/src/kern_tc.c#L2159

Anyway... now for `rtems_binary_semaphore_try_wait()`

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/include/rtems/thread.h#L263

https://github.com/RTEMS/rtems/blob/c8a10609883c67629ea029be617a42b75d168f0f/cpukit/score/src/semaphore.c#L98

Which uses `_Thread_queue_Context_set_enqueue_timeout_ticks()` instead of `_Thread_queue_Context_set_enqueue_timeout_realtime_timespec()`, so sadly we can't just use `_Watchdog_Ticks_from_timespec()` as the 'Ticks' in this name refers to a different unit of "tick".

If any of the rtems devs have lurked through all of this.  Would an alternative to `rtems_binary_semaphore_try_wait()` which takes a `struct timespec*` as input be a reasonable addition?

mdavidsaver:
As discussed during today's core meeting...

I suggest reordering the tests of `timeOut` such than `NaN` falls through to an `else{}` where it is treated like `Inf`.

Andrew asks for a look at bounds checking after casting to integer to ensure that a delay of `0.0000000000001` doesn't become infinite.

modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
anjohnson: Applies to the VxWorks implementation too, I'm fixing it there (and in os/vxWorks/osdMessageQueue.cpp).

anjohnson:
@hjunkes if you like I can make and push the changes we're asking for, since I have this branch checked out and am making similar VxWorks changes.

hjunkes:
Unfortunately I can't push my changes to the PR:
rtems@rtems-dev:~/EPICS_BASE/epics-base$ git push
remote: Permission to anjohnson/epics-base.git denied to hjunkes.
fatal: unable to access 'https://github.com/anjohnson/epics-base.git/': The requested URL returned error: 403

```
rtems@rtems-dev:~/EPICS_BASE/epics-base$ git diff 6063de9a8bc0bc5209a4e3024c44bde08a7f35ec f7de8b972e3b2bfdc37c77ee76527340a9ecf461
diff --git a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
index 54dfccfff..09fa310c4 100644
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
@@ -20,6 +20,7 @@

 #include "libComAPI.h"
 #include "epicsEvent.h"
+#include "epicsMath.h"

 typedef struct epicsEventOSD {
     rtems_binary_semaphore rbs;
@@ -66,7 +67,7 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
 {
     int sc;
     rtems_interval delay;
-    int rate = rtems_clock_get_ticks_per_second();
+    rtems_interval rate = rtems_clock_get_ticks_per_second();

     if (!rate)
         return epicsEventError;
@@ -78,11 +79,12 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
         else
             return epicsEventWaitTimeout;
     }
-    else if (timeOut >= (double) INT_MAX / rate) {
+    else if (timeOut >= (double) INT_MAX / rate || isnan(timeOut)) {
         delay = 0;
     }
     else {
         delay = timeOut * rate;
+       if (delay <= 0) delay = 1;
     }

     sc = rtems_binary_semaphore_wait_timed_ticks(&pSem->rbs, delay);
diff --git a/modules/libcom/src/osi/os/vxWorks/osdEvent.c b/modules/libcom/src/osi/os/vxWorks/osdEvent.c
index 27b271e51..2d2b3f044 100644
--- a/modules/libcom/src/osi/os/vxWorks/osdEvent.c
+++ b/modules/libcom/src/osi/os/vxWorks/osdEvent.c
@@ -19,6 +19,7 @@
 #include <limits.h>

 #include "epicsEvent.h"
+#include "epicsMath.h"

 /* The following not defined in any vxWorks header */
 int sysClkRateGet(void);
@@ -42,7 +43,7 @@ epicsEventStatus epicsEventWaitWithTimeout(epicsEventId id, double timeOut)

     if (timeOut <= 0.0) {
         ticks = 0;
-    } else if (timeOut >= (double) INT_MAX / rate) {
+    } else if (timeOut >= (double) INT_MAX / rate || isnan(timeOut)) {
         ticks = WAIT_FOREVER;
     } else {
         ticks = timeOut * rate;
```

anjohnson:
Hello Heinz,

It's possible to code this without explicitly using `isnan()` at all, making use of the fact that all numeric comparisons involving a NaN value will return false. By adjusting the comparisons and their order, the NaN case can be made to fall out in the right place. For VxWorks I'm going to use this code everywhere that we have to convert a delay into ticks:
```C
    if (timeOut <= 0.0) {
        ticks = 0;
    }
    else if (timeOut < (double) INT_MAX / rate) {
        ticks = timeOut * rate;
        if (ticks == 0) {
            /* 0 < timeOut < 1/rate, round up */
            ticks = 1;
        }
    }
    else {
        /* timeOut is NaN or too big to represent in ticks */
        ticks = WAIT_FOREVER;
    }
```
I propose these changes for your code, which I have just pushed to this branch for testing:
```patch
diff --git a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
index 54dfccfff..abd9a8e9a 100644
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdEvent.c
@@ -12,7 +12,7 @@
 
 #include <stdio.h>
 #include <malloc.h>
-#include <limits.h>
+#include <stdint.h>
 #include <rtems.h>
 #include <rtems/error.h>
 
@@ -66,7 +66,7 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
 {
     int sc;
     rtems_interval delay;
-    int rate = rtems_clock_get_ticks_per_second();
+    rtems_interval rate = rtems_clock_get_ticks_per_second();
 
     if (!rate)
         return epicsEventError;
@@ -78,11 +78,16 @@ epicsEventWaitWithTimeout(epicsEventId pSem, double timeOut)
         else
             return epicsEventWaitTimeout;
     }
-    else if (timeOut >= (double) INT_MAX / rate) {
-        delay = 0;
+    else if (timeOut < (double) UINT32_MAX / rate) {
+        delay = timeOut * rate;
+        if (delay == 0) {
+            /* 0 < timeOut < 1/rate; round up */
+            delay = 1;
+        }
     }
     else {
-        delay = timeOut * rate;
+        /* timeOut is NaN or too big to represent; wait forever */
+        delay = RTEMS_NO_TIMEOUT;
     }
 
     sc = rtems_binary_semaphore_wait_timed_ticks(&pSem->rbs, delay);
```
I also just added a similar fix to the RTEMS-score/osdEvent.c version.

anjohnson:
Approved at Core Group meeting today.

PR #208, opened 2021-10-14
epicsSingleton.h: Fix inline mismatch for pInstance()

tboegi:
Fix a compilation warning under Mingw:

```
g++ -D_MINGW -D__USE_MINGW_ANSI_STDIO -DBUILDING_libCa_API -O3 -Wall -m64 -DEPICS_BUILD_DLL -DEPICS_CALL_DLL -I. -I../O.Common -I. -I. -I.. -I../test -I../../../../../include/compiler/gcc -I../../../../../include/os/WIN32 -I../../../../../include -o netSubscription.obj -c ../netSubscription.cpp
In file included from ../localHostName.h:27,
from ../cac.h:32,
from ../netSubscription.cpp:30:
../../../../../include/epicsSingleton.h:164:15: warning: 'void* SingletonUntyped::pInstance() const' redeclared without dllimport attribute after being referenced with dll linkage
164 | inline void * SingletonUntyped :: pInstance () const
| ^~~~~~~~~~~~~~~~
----------------
```

Solution:
"void * pInstance () const;" is changed into
"inline void * pInstance () const;"

### DIFF
--- a/modules/libcom/src/cxxTemplates/epicsSingleton.h
+++ b/modules/libcom/src/cxxTemplates/epicsSingleton.h
@@ -29,7 +29,7 @@ public:
     void incrRefCount ( PBuild );
     typedef void ( * PDestroy ) ( void * );
     void decrRefCount ( PDestroy );
-    void * pInstance () const;
+    inline void * pInstance () const;
 private:
     void * _pInstance;
     std :: size_t _refCount;


anjohnson:
Any ideas why that's the only method which needs the `inline` keyword there? I'm not objecting to the change, just wondering whether we should be proactive and add it to any of the others too. Another other solution might be to just inline the trivial method(s) into the class declaration.

FreddieAkeroyd:
It is probably because this method is returning a pointer to an object unlike the others, so returning data across a dll boundary. The method returning the object is decorated dllExport (due to the class being dllexport), but i'm guessing when it is inlined the decoration may get lost and it looks like it is returning an undecorated pointer `_pInstance`? The pointer may be undecorated as it is private, would be interesting to see if it picked up the class dllexport if it was a public data item instead.  I expect moving the implementation to inside the class declaration as Andrew mentions would also remove the warning. This may be an compiler issue as it seems it can handle inline across dll boundaries, but you need to specify it in both place. When inline is specified later, it doesn't go back and fix up the dll export - maybe visual studio does that, or ignores the inline as it is dllexport, or just doesn't notice...

mdavidsaver:
While I may come to regret touching this code, I've gone ahead and switched all of the template method definitions in `epicsSingleton.h` to be inline.  cf. 2b3c6f2e262a7dfc506cce66547ad2dc74d8a80b

PR #210, opened 2021-11-02
Fix regressTest regression

mdavidsaver:
Attempt to fix race condition in regressTest.  cf. https://github.com/epics-base/epics-base/discussions/162#discussioncomment-1460745

Removes the now unused private API `dbCaGetUpdateCount()` in favor of a new `testdbCaWaitForUpdateCount()` which covers the original use case in `dbCaLinkTest.c` as well as a new one in `regressTest.c`.

Changes `dbCaSync()` to wait for the dbCa worker queue to be empty.  Usage of `dbCaSync()` remains correct in the original case of synchronizing `dbPutField()` to an `OUT_LINK`.  However, it should not be used with an `IN_LINK`.

### DIFF
--- a/modules/database/src/ioc/db/dbCa.c
+++ b/modules/database/src/ioc/db/dbCa.c
@@ -196,6 +196,53 @@ static void caLinkDec(caLink *pca)
     if (callback) callback(userPvt);
 }
 
+struct waitPvt {
+    caLink *pca;
+    epicsEventId evt;
+};
+
+static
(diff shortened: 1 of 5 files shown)


mdavidsaver:
4 successful GHA runs and 2 appveyor.  This seems fixed, or enough of an improvement to merge.

anjohnson:
Agreed, please merge. I saw one possibly similar failure of asyncSoftTest.c on Appveyor the other day, we'll see if that crops up once this one is no longer causing failures.

Thanks!

PR #213, opened 2021-11-10
Hex in hw links

dirk-zimoch:
Re-allow hex numbers in hardware links


anjohnson (approved):
This looks good to me. I don't see any DB files in the APS Accelerator 3.14.12 IOC applications which would need to be changed to avoid interpretation as octal numbers, or which have been using `0x` in the address fields (other than in device-support parsed `@parm` strings).

mdavidsaver (approved):
The code change looks fine to me.

For some reason appveyor hasn't built this PR.  I don't see any hints in the appveyor event log.  I don't see this as a blocker since the GHA windows builds pass.

Issue #211, opened 2021-11-03
Lack of `fork()` safety


mdavidsaver:
Documenting the current state of things.  As it stands now, there are a number places where the Base APIs may not safely be used by a child process after a `fork()`.  Adventurous users who investigate the safety of particular APIs are encouraged to note this, either in documentation, or on this issue.

Less adventurous users should assume that no Base API is safe to call/access from a child process after `fork()`.

The general [recommendation](https://www.google.com/search?q=mixing+fork+and+pthread) is to avoid usage of `fork()` when multiple threads exist in the parent process.  A newly `fork()`d child process consists of only a single thread, and a copy of parent memory.  So any other threads in the parent process effectively "disappear", leaving memory in an indeterminate state (eg. with a mutex locked).  Further, some, but not all, parent file descriptors are inherited by the child.  Unlike memory, these are not copies, so changes by the child may effect the parent.

Known safety issues in libCom

1. The `errlog` worker thread
2. The `ipAddrToAsciiGlobal` worker thread (emulation of asynchronous DNS)
3. Free lists

mdavidsaver:
References

- https://github.com/epics-base/epics-base/issues/83#issuecomment-667919934
- "Re: Pure Python IOC (CAProcess issue)" https://epics.anl.gov/tech-talk/2020/msg01358.php

anjohnson:
Can this issue be closed now that #214 has been merged?

mdavidsaver:
https://github.com/epics-base/epics-base/issues/286#issuecomment-1261203282 has a survey of global c++ constructors/destructors currently in libCom.

mdavidsaver:
Update circa 7.0.8.  Three remaining global constructors in libCom on linux-x86_64.

```
$ nm lib/linux-x86_64/libCom.a |grep GLOBAL__sub
0000000000000000 t _GLOBAL__sub_I_fdManager.cpp
0000000000000000 t _GLOBAL__sub_I_epicsThread.cpp
0000000000000000 t _GLOBAL__sub_I_osdTime.cpp
```

#451 would remove the ctor in epicsThread.cpp.

mdavidsaver:
Status update as of 7.0.9:  https://github.com/epics-base/epics-base/issues/286#issuecomment-2646472250

PR #214, opened 2021-11-13
Warn of unsafe use of epicsThread after `fork()`

mdavidsaver:
```
Warning: Undefined Behavior!
         Detected use of epicsThread from child process after fork()
```

Add a warning about UB (cf. #211) if `epicsThreadInit()` is called from a child process after `fork()`.  Many, but not all unsafe code paths should trip this.  eg. `epicsThreadOnce()` and by extension `errlogPrintf()`, but not `epicsMutexLock()`.  My choice is mainly driven by an assumption that calls to `epicsThreadInit()` are not happening in "hot" code paths.

`pthread_atfork()` is used to detect a `fork()`, however a warning is only emitted if the child calls `epicsThreadInit()`, so it should not be triggered by eg. `osiSpawnDetachedProcess()` doing a `fork()` and then `execv()`.

This PR also changes `posix/osdThread.c` to avoid implicit calls to the epicsStdio redirection calls since these use `epicsThreadOnce()`.  I switched the only immediately obvious (to me) safe usage in `epicsThreadShow()` to explicitly call `epicsStdoutPrintf()`.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -330,12 +330,29 @@ int          status;
     a_p->usePolicy = arg.ok;
 }
 #endif
-
+
+/* 0 - In the process which loads libCom.
+ * 1 - In a newly fork()'d child process
+ * 2 - In a child which has been warned
+ */
+static int childAfterFork;
+
+static void childHook(void)
+{
+    epicsAtomicSetIntT(&childAfterFork, 1);
+}
 
 static void once(void)
 {
     epicsThreadOSD *pthreadInfo;
     int status;
+
+#ifdef __rtems__
+    (void)childHook;
+#else
+    status = pthread_atfork(NULL, NULL, &childHook);
+    checkStatusOnce(status, "pthread_atfork");
+#endif
 
     pthread_key_create(&getpthreadInfo,0);
     status = osdPosixMutexInit(&onceLock,PTHREAD_MUTEX_DEFAULT);
@@ -431,6 +448,11 @@ static void epicsThreadInit(void)
     static pthread_once_t once_control = PTHREAD_ONCE_INIT;
     int status = pthread_once(&once_control,once);
     checkStatusQuit(status,"pthread_once","epicsThreadInit");
+
+    if(epicsAtomicGetIntT(&childAfterFork)==1 &&  epicsAtomicCmpAndSwapIntT(&childAfterFork, 1, 2)==1) {
+        fprintf(stderr, "Warning: Undefined Behavior!\n"
+                        "         Detected use of epicsThread from child process after fork()\n");
+    }
 }
 
 LIBCOM_API
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -34,6 +34,10 @@
 #include <sys/mman.h> 
 #endif
 
+/* epicsStdio uses epicsThreadOnce(), require explicit use to avoid unexpected recursion */
+#define epicsStdioStdStreams
+#define epicsStdioStdPrintfEtc
+
 #include "epicsStdio.h"
 #include "ellLib.h"
 #include "epicsEvent.h"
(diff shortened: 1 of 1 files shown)


modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -326,13 +330,22 @@ int          status;
     a_p->usePolicy = arg.ok;
 }
 #endif
-
+
+static int childAfterFork;
```
anjohnson, line 338: There could be an enum to declare the magic numbers that this variable will hold (currently 0, 1 and 2), making them self-documenting (e.g. `forkNone`, `forkDone`, `forkWarned`).

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -326,13 +330,22 @@ int          status;
...
+    epicsAtomicSetIntT(&childAfterFork, 1);
+}
 
 static void once(void)
 {
     epicsThreadOSD *pthreadInfo;
     int status;
 
+    status = pthread_atfork(NULL, NULL, &childHook);
+    checkStatusOnce(status, "pthread_atfork");
```
anjohnson, line 354: According to [this documentation](https://docs.rtems.org/branches/master/posix-users/process_creation_and_execution.html#pthread-atfork) RTEMS will return `ENOSYS` from this `pthread_atfork()` call, which will trigger a warning message.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -427,6 +440,11 @@ static void epicsThreadInit(void)
     static pthread_once_t once_control = PTHREAD_ONCE_INIT;
     int status = pthread_once(&once_control,once);
     checkStatusQuit(status,"pthread_once","epicsThreadInit");
+
+    if(epicsAtomicGetIntT(&childAfterFork)==1 &&  epicsAtomicCmpAndSwapIntT(&childAfterFork, 1, 2)==1) {
```
anjohnson, line 452: The call to `epicsAtomicGetIntT()` can't be necessary here (if it were there would be a race in the time between the two atomic calls). If the value isn't `forkDone` at that time the call to `epicsAtomicCmpAndSwapIntT()` shouldn't succeed, which gives the same outcome with fewer atomic operations.

anjohnson (commented):
Minor issues below, no objection to the principle of this at all.

modules/libcom/src/osi/os/posix/osdThread.c
mdavidsaver: Ok, I expected I'd have to `#ifdef` this for one reason or another.

modules/libcom/src/osi/os/posix/osdThread.c
anjohnson: Correction it's not actually a race, but the warning could get missed.

mdavidsaver: Nope, not necessary.  This is an optimization (perhaps premature) to avoid the cost of a more expensive atomic CAS.  C++ `std::atomic` forced me into this pattern.  The first argument of [`compare_exchange_strong()`](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange) is a non-const reference.  I think I see, and agree, with the logic though.

modules/libcom/src/osi/os/posix/osdThread.c
mdavidsaver: How about a comment here listing states and meanings?

Switching to an enum requires casting back to `int*` for the epicsAtomic calls.  eg. 

```c
epicsAtomicSetIntT(&childAfterFork, 1);
// becomes
epicsAtomicSetIntT((int*)&childAfterFork, forkChild);
```

anjohnson: My intention was to just use the enum to declare the magic number values (you could use `#define`s instead but I think an enum is slightly a better solution), but leave the `childAfterFork` variable as an `int` (with a comment that it should contain one of those enum values) — that seems to be the recommended way to do this kind of thing in C without having to do the casting of pointers to the variable. IIRC you can't guarantee what width of integer the compiler will actually use if you make it an `enum`.

modules/libcom/src/osi/os/posix/osdThread.c
anjohnson: You could avoid using an `#ifdef` just for RTEMS by accepting a status value of `ENOSYS` from any OS since that tells us whatever OS it is probably doesn't support forking, or at least can't do `pthread_atfork()`.

mdavidsaver: It's a conditional one way or the other.  I think `#ifdef __rtems__` is clearer.

modules/libcom/src/osi/os/posix/osdThread.c
anjohnson: My reason for suggesting the enum was to document the magic number states lower down in the code. `epicsAtomicCmpAndSwapIntT(&childAfterFork, forkDone, forkWarned) == forkDone` is quicker to read and understand than `epicsAtomicCmpAndSwapIntT(&childAfterFork, 1, 2) == 1`.

anjohnson:
Approved group 12/15/21.

PR #215, opened 2021-11-19
It is not recommended to mix putchar with printf

dirk-zimoch:
At least on vxWorks 6.9 the newlines are not printed in the test results. The (Linux) `putchar` man page says:

> **BUGS**
       It  is  not advisable to mix calls to output functions from the stdio library with low-level calls to write(2)
       for the file descriptor associated with the same output stream; the results will be undefined and very  proba‐
       bly not what you want.

On the other hand, the vxWorks manual claims:
> For example, printf( ) is a wrapper that calls the system call write( ).

This problem may affect more systems than just vxWorks. Generally better not mix `putchar()`, `puts()` and the like with the buffered `printf()` family of functions on the same file (including `stdout`).

BTW: For similar reasons do not mix `printf()` with `cout <<`.

Anyway, this patch fixes the problem.

### DIFF
--- a/modules/libcom/src/misc/epicsUnitTest.c
+++ b/modules/libcom/src/misc/epicsUnitTest.c
@@ -163,7 +163,7 @@ int testOkV(int pass, const char *fmt, va_list pvar) {
     vprintf(fmt, pvar);
     if (todo)
         printf(" # TODO %s", todo);
-    putchar('\n');
+    printf("\n");
     fflush(stdout);
     epicsMutexUnlock(testLock);
     return pass;
@@ -219,7 +219,7 @@ int testDiag(const char *fmt, ...) {
     epicsMutexMustLock(testLock);
     printf("# ");
     vprintf(fmt, pvar);
-    putchar('\n');
+    printf("\n");
     fflush(stdout);
     epicsMutexUnlock(testLock);
     va_end(pvar);
@@ -231,7 +231,7 @@ void testAbort(const char *fmt, ...) {
     va_start(pvar, fmt);
     printf("Bail out! ");
     vprintf(fmt, pvar);
-    putchar('\n');
+    printf("\n");
     fflush(stdout);
     va_end(pvar);
     abort();


anjohnson (changes requested):
Were you actually seeing a situation where the `putchar()` and `printf()` calls from a unit test were not synchronized? Please copy & paste an example of the actual output that you're seeing on VxWorks. The BUGS section of the Linux manpage is describing a different situation since `write()` does not make use of the stdout stream.

The short descriptions on the VxWorks 6.9 manpages describe the routines as:
```
     putchar() - write a character to the standard output stream (ANSI)
     printf() - write a formatted string to the standard output stream (ANSI)
```
so both are documented as writing to the same output stream, and `printf()` is a _much_ less efficient routine. The underlying `putc()` routine that `putchar()` calls is simple enough that it's frequently implemented as a macro, and the code in epicsUnitTest.c always calls `fflush(stdout)` afterwards.

The epicsStdio.h header does define both `putchar` and `printf` as macros in order to support redirection of the stdout stream in iocsh commands, and also makes `stdout` a macro so the `fflush(stdout)` will always apply to the same stream. I admit that we don't currently define `vprintf` as a macro there which is a bug, but that shouldn't be causing this issue because neither the epicsUnitTest.c file nor any of the local header files which it includes pull in epicsStdio.h, so the code in epicsUnitTest.c should be calling the original OS routines directly.

I need a much stronger justification before I can accept these changes.

mdavidsaver:
> ... It is not advisable to mix calls to ...

It is my understanding that this refers to mixing calls to buffered I/O functions accepting `FILE*` with unbuffered accepting an `int` file descriptor.

I am under the impression that `printf(...)` is expected to be shorthand for `fprintf(stdout, ...)`, which is a buffered operation (uses `FILE*`).  As well that `putchar(c)` is shorthand for `putc(c,stdout)`.

https://linux.die.net/man/3/putchar

Of course it is entirely possible that some vxworks versions/configurations are not conformant.  Does vxworks have any build/runtime settings which might be relevant?

dirk-zimoch:
The output looks like this snippet:

ok 215 - PVStructure copyok 216 - PVStructure copyUncheckedok 217 - PVStructure copyUnchecked w/ cleared maskok 218 - PVStructure copyUnchecked w/ value mask onlyok 219 - PVStructure copyUnchecked w/ value+timeStamp maskok 220 - null PVUnion copyok 221 - PVUnion scalar copy, to null PVUnionok 222 - PVUnion scalar copy, to different type PVUnionok 223 - PVUnion PVStructure copy, to different type PVUnion# Check methods for accessing structure fieldsok 224 - a.get() != NULLok 225 - 0x6bfec0 == 0x6bfec0ok 226 - a.get() != NULLok 227 - 0x543160 == 0x543160ok 228 - fld->getSubField<PVInt>("invalid").get()==NULLok 229 - fld->getSubField<PVDouble>("test").get()==NULLok 230 - fld->getSubField<PVDouble>("helo.world").get()==NULLok 231 - fld->getSubField<PVDouble>("hello.").get()==NULLok 232 - fld->getSubField<PVDouble>("hello..world").get()==NULLok 233 - fld->getSubField<PVDouble>(".").get()==NULLok 234 - fld->getSubField<PVInt>(" test").get()==NULLok 235 - fld->getSubField<PVInt>("hello.world.invalid").get()==NULLok 236 - caught expected exception: Failed to get field: (Field name is NULL string)ok 237 - caught expected exception: Failed to get field: invalid (invalid not found)ok 238 - caught expected exception: Failed to get field: test (Field has wrong type)ok 239 - caught expected exception: Failed to get field: hello. (Zero-length field name encountered)ok 240 - caught expected exception: Failed to get field: hello..world (Zero-length field name encountered)ok 241 - caught expected exception: Failed to get field: . (Zero-length field name encountered)ok 242 - caught expected exception: Failed to get field:  test (No spaces allowed in field name)

No newlines.
Replacing putchar with printf solves the problem. and performance is not really an issue here. We are just printing test results.

dirk-zimoch:
> Does vxworks have any build/runtime settings which might be relevant?

Not that I am aware of. On Monday I will check how the earlier vxWorks versions behave…

dirk-zimoch:
This is strange. I repeated all tests today with VxWorks 6.7 and 6.9 and EPICS 7.0.5  and 7.0.6.1.
The missing newline is gone!
Maybe something had happened to my terminal? Not sure.
(Even more strange is that now a test fails where I was sure it succeeded before. )
Closing this PR for now...

PR #216, opened 2021-12-07
Fix gdd string to double conversion

ralphlange:
This is an alternative fix for the issue described in https://github.com/epics-modules/pcas/issues/4 and https://github.com/epics-extensions/ca-gateway/issues/37
It keeps the call to `epicsScanDouble()` before trying `sscanf()` first with `"%lf"` (ignoring extra characters) then with `"%x"` (to catch hex numbers).

### DIFF
--- a/src/ca/legacy/gdd/aitConvert.cc
+++ b/src/ca/legacy/gdd/aitConvert.cc
@@ -50,14 +50,16 @@ bool getStringAsDouble ( const char * pString,
         ftmp = itmp;
     }
     else {
-	    int j = epicsScanDouble( pString, &ftmp );
-	    if ( j != 1 ) {
-		    j = sscanf ( pString,"%x", &itmp );
-            if ( j == 1 ) {
-                ftmp = itmp;
-            }
-            else {
-                return false;
+        int j = epicsScanDouble ( pString, &ftmp );
+        if ( j != 1 ) {
+            j = sscanf ( pString, "%lf", &ftmp );
+            if ( j != 1 ) {
+                j = sscanf ( pString, "%x", &itmp );
+                if ( j == 1 ) {
+                    ftmp = itmp;
+                } else {
+                    return false;
+                }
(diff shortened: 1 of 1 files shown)


anjohnson (approved):
Seems fine by me, thanks!

anjohnson:
I'm confused; Ralph's commit 7c168f20f4e38be8d578c96a32358ef97fe282aa got merged 4 hours ago and GitHub shows it appearing on the 3.15 branch, but it doesn't seem to have recognized yet that this PR should be closed as a result. I have a feeling that's happened before though, I will wait a few hours to see if they just have a slow processing pipeline.

anjohnson:
Ahh, no Ralph's original commit in this PR was 56aae9d3d60da9d96d18b0f90d1aca0eff4a117d so where did 7c168f20f4e38be8d578c96a32358ef97fe282aa come from? @ralphlange I think you may have got confused by Freddie's PR that he meant to point to the ISIS Base repo and you merged the wrong PR. I will close this one.

ralphlange:
I merged outside of GitHub as s fast-forward (no merge commit) and updated the commit before pushing.
Sorry that I confused GitHub and you by doing so...
AFAIK, auto-closing works when the commit is merged into the default branch of the repo, which in this case will never happen.

Freddie's PR was in the PCAS module, not in Base. Unconfuseable.

anjohnson:
My bad, thanks for doing this. FYI all those Jenkins build failures were caused by all the 3.15 jobs having lost their settings for the post-build warning/error analysis of the build log, probably as a result of multiple Jenkins updates without building the job.

ralphlange:
I see. Yes, slow-moving branches are prone to that sort of bitrot.
Can't you just build them periodically, once week or once a month, to check if they are still valid?

anjohnson:
Good idea, I just set the 3.15 branch to build monthly.

PR #218, opened 2021-12-17
Make sure epicsInt8 is signed on all architectures

dirk-zimoch:
epicsTypes.h used to define `epicsInt8` as `char` since the very first version. Is is and has always been wrong on architectures where `char` is signed, including some of the earliest architrectures supported by EPICS (PPC on vxWorks). `epicsInt8` is supposed to be signed on all architectures.
This has led to faulty conversions from `DBF_CHAR` fields to other field types on those architectures in cases where the high bit of the char was set. It affects other software, e.g. device support and drivers, as well if they convert `epicsInt8` to signed datatypes like `int`. 

A simple test case is:
```
record(longin, "init") {
  field(VAL, "-1")
  field(PINI, "YES")
  field(FLNK, "wf")
}
record (waveform, "wf") {
  field(FTVL, "CHAR")
  field(NELM, "1")
  field(INP, "init")
  field(FLNK, "longin")
}
record (longin, "longin") {
  field(INP, "wf")
  field(FLNK, "ai")
}
record (ai, "ai") {
  field(INP, "wf")
}
```

On architectures with signed `char` (x86 by default), the records "longin" and "ai" contain the value -1. On architectures with unsigned char (e.g. by default all PPC I checked), they contain the value 255 as if the waveform had the type `UCHAR`. This is wrong. 

This fix corrects the conversion of DBF_CHAR fields to signed types.

**However, some existing databases running on affected architectures (PPC) may expect or rely on this buggy behavior!**
Those databases should be modified to use `UCHAR` waveforms.
Luckily, no standard EPICS base record type has any `DBF_CHAR` fields. Thus, only `CHAR` waveform, aai, and aao records are affected.

Totally unrelated and not fixed here is the fact that the `caget` command line tool prints `DBF_CHAR` fields (and `DBF_UCHAR` - CA makes no difference) using a cast to `char`. This results in architecture dependent habavior on the client side. Most clients are probably x86 systems, and print those fields as signed numbers, while PPC clients print them as unsigned numbers. This is unrelated to the IOC or the transport layer. Internally, CA uses `dbr_char_t`, which is unsigned, for both, `DBF_CHAR` and `DBF_UCHAR`. Thus on the majority of systems, `caget` prints wrong results, but the impact of fixing it may be huge and the CA documentation does not specify the signedness of `dbr_char_t`.

### DIFF
--- a/modules/ca/src/client/comBuf.h
+++ b/modules/ca/src/client/comBuf.h
@@ -88,6 +88,7 @@ public:
     bool push ( const T & value );
     template < class T >
     unsigned push ( const T * pValue, unsigned nElem );
+    unsigned push ( const char * pValue, unsigned nElem );
     unsigned push ( const epicsInt8 * pValue, unsigned nElem );
     unsigned push ( const epicsUInt8 * pValue, unsigned nElem );
     unsigned push ( const epicsOldString * pValue, unsigned nElem );
@@ -204,6 +205,11 @@ inline unsigned comBuf :: push ( const epicsInt8 *pValue, unsigned nElem )
 }
 
 inline unsigned comBuf :: push ( const epicsUInt8 *pValue, unsigned nElem )
+{
+    return copyInBytes ( pValue, nElem );
+}
+
+inline unsigned comBuf :: push ( const char *pValue, unsigned nElem )
 {
     return copyInBytes ( pValue, nElem );
 }
--- a/modules/libcom/src/misc/epicsTypes.h
+++ b/modules/libcom/src/misc/epicsTypes.h
@@ -41,7 +41,7 @@ typedef enum {
  * These are sufficient for all our current archs
  * @{
  */
-typedef char            epicsInt8;
+typedef signed char     epicsInt8;
 typedef unsigned char   epicsUInt8;
 typedef short           epicsInt16;
 typedef unsigned short  epicsUInt16;
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -16,6 +16,17 @@ should also be read to understand what has changed since earlier releases.
 
 <!-- Insert new items immediately below here ... -->
 
+### Make sure epicsInt8 is signed on all architectures
+
+So far, `epicsInt8` and thus `DBF_CHAR` used to be unsigned on architectures
+where `char` is unsigned, for example on many PPC architectures.
+This had led to different behavior between architectures when converting
+`DBF_CHAR` to signed integer or to floating point types.
+
+WARNING: This fix may change behavior of existing databases on on architectures
+with unsigned `char` (many PPC) when using input links to read from `CHAR`
+waveforms. Architectures with signed `char` (usually x86) are unaffected.
+
 ### Fix for `undefined` in configure/RELEASE files
 
 Prevents `Use of uninitialized value` warnings from convertRelease.pl.
--- a/modules/ca/src/client/comQueRecv.cpp
+++ b/modules/ca/src/client/comQueRecv.cpp
@@ -46,7 +46,7 @@ void comQueRecv::clear ()
     this->nBytesPending = 0u;
 }
 
-unsigned comQueRecv::copyOutBytes ( epicsInt8 *pBuf, unsigned nBytes )
+unsigned comQueRecv::copyOutBytes ( char *pBuf, unsigned nBytes )
 {
     unsigned totalBytes = 0u;
     do {
--- a/modules/ca/src/client/comQueRecv.h
+++ b/modules/ca/src/client/comQueRecv.h
@@ -33,7 +33,7 @@ public:
     comQueRecv ( comBufMemoryManager & );
     ~comQueRecv ();
     unsigned occupiedBytes () const;
-    unsigned copyOutBytes ( epicsInt8 *pBuf, unsigned nBytes );
+    unsigned copyOutBytes ( char *pBuf, unsigned nBytes );
     unsigned removeBytes ( unsigned nBytes );
     void pushLastComBufReceived ( comBuf & );
     void clear ();


modules/libcom/src/misc/epicsTypes.h
```diff
@@ -41,7 +41,7 @@ typedef enum {
  * These are sufficient for all our current archs
  * @{
  */
-typedef char            epicsInt8;
+typedef signed char     epicsInt8;
```
mdavidsaver, line 44: I doing this unconditionally the best solution?

It seems to me like this will maximize any negative impact.

modules/ca/src/client/comBuf.h
```diff
@@ -208,6 +209,11 @@ inline unsigned comBuf :: push ( const epicsUInt8 *pValue, unsigned nElem )
     return copyInBytes ( pValue, nElem );
 }
 
+inline unsigned comBuf :: push ( const char *pValue, unsigned nElem )
+{
+    return copyInBytes ( pValue, nElem );
+}
+
```
mdavidsaver, line 216: I don't like adding a third specialization here.  A simpler solution would be to combine the `epicsInt8` and `epicsUInt8` specializations into the `T` definition with a condition on `sizeof(T)==1`.

```c++
template < class T >
inline bool comBuf :: push ( const T & value )
{
    if(sizeof(T)==1)
        return copyInBytes ( pValue, nElem );
```

If adding a third specialization is somehow necessary, then the argument types should be the three plain C types.

mdavidsaver:
Also, it seems like step one would be to add a test like:

```c
STATIC_ASSERT(-1==(int)(epicsInt8)-1);
```

modules/libcom/src/misc/epicsTypes.h
dirk-zimoch: I think it will maximize consistency. Compare with stdint.h:
typedef signed char             int8_t;

Also, you pointed out earlier that for C++ `char`, `signed char` and `unsigned char` are three different things, not two. Conditional definition will increase the number of cross platform inconsistencies.

modules/ca/src/client/comBuf.h
dirk-zimoch: Maybe. I tried what works. As the function is `inline`, the run-time overhead is zero.
Not sure if your solution adds run-time overhead. And branching on types does not seem like good C++ style to me.

dirk-zimoch:
The `STATIC_ASSERT` looks very promising.

dirk-zimoch:
Do you want to add a similar `STATIC_ASSERT` to PVA after including stdint.h to check `int8_t`? That would catch problems like I had with vxWorks 6.9.0 early.

modules/libcom/src/misc/epicsTypes.h
mdavidsaver: I say this as a practical consideration.  It might help with concerns by @anjohnson about breaking downstream code on targets where `char` is and always has been, signed.

Personally, I wish that 5e0bf53fecb332a53656ca5e68a62823a6893d59 had gone the other way, and defined `epicsInt*` in terms of `int*_t`, but I understand what the motivations where (little usage of C99).

mdavidsaver:
> Do you want to add a similar `STATIC_ASSERT` to PVA after including stdint.h to check `int8_t`?

Our policy has been to avoid/remove tests of system libraries as "out of scope" for Base.  However, it seems reasonable to make an exception in this case as a specific regression is known to impact Base users.

mdavidsaver:
Taking a step back...

I agree with Dirk that it is deceptive for a type named `epicsInt8` to ever be unsigned.  (This isn't something which I always remember)  I know there is history here which may make a change painful.  Still, in my mind this needs to be done.  So for me the question is how to minimize the impact.

The options I see are:

1) Mandating `-fsigned-char`.  As (sort of) an ABI change which will effect code external to Base.

2) Changing `epicsInt8` to (sometimes?) be `signed char`.

3) Add (or rename to) `epicsSInt8`.

Are there others options?

mdavidsaver:
GCC has a [predefined macro](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) `__CHAR_UNSIGNED__`, although the recommendation is to use `limits.h`.  So how about something like this?

```c
#include <limits.h>
...
#if CHAR_MAX==UCHAR_MAX
typedef signed char epicsInt8;
#else
typedef char epicsInt8;
#endif
```

modules/ca/src/client/comBuf.h
mdavidsaver: The "ideal" solution would be one partial specialization which covers the three `sizeof(T)==1` types.  cf. [SFINAE](https://en.cppreference.com/w/cpp/language/sfinae) and [`std::enable_if`](https://en.cppreference.com/w/cpp/types/enable_if).  I'm not going to insist that you figure out this fairly tricky pattern just for this change.  Unless optimization is completely disabled (`-O0`), branching on a constant has no runtime cost.  I've verified this on various GCC versions.

modules/libcom/src/misc/epicsTypes.h
anjohnson: I hope it's unlikely that someone would use `epicsInt8` in their code as a synonym for `char` and pass such an array to OS routines like `strlen()` and `strcpy()` which take a `char *`. If they did in C code this change should only result in build warnings on arch's where `char` is signed, but C++ code could fail to compile — however I guess that may be what we would want to happen in that case?

anjohnson:
Checking the synApps modules the asynRecord has a couple of scalar `DBF_CHAR` fields which have been there since 4-13 or earlier:
```
tux$ grep DBF_CHAR synApps_6_2_1/support/*/dbd/*Record.dbd
synApps_6_2_1/support/asyn-R4-42/dbd/asynRecord.dbd:    field(BOUT,DBF_CHAR) {
synApps_6_2_1/support/asyn-R4-42/dbd/asynRecord.dbd:    field(BINP,DBF_CHAR) {
```

However they are both marked `special(SPC_DBADDR)` and their `cvt_dbaddr()` routine actually maps accesses to the record's input and output data arrays, so they are never used directly for `epicsInt8` data.

anjohnson:
> The options I see are:
> 
> 1. Mandating -fsigned-char. As (sort of) an ABI change which will effect code external to Base.
> 2. Changing epicsInt8 to (sometimes?) be signed char.

Choice 1 seems like a non-starter to me because we're altering the OS's ABI.

If we adopt choice 2 (making the next release 7.0.7), when might `epicsInt8` not become `signed char`, and why do that?

mdavidsaver:
> when might `epicsInt8` not become `signed char`,

x86.

fyi. Checking with an RTEMS5 toolchain, both arm32 and ppc32 do have `char` unsigned.

> and why do that?

It would remove the possibility of any breakage on what is probably the most popular architecture.  That said, I you (@anjohnson ) are ok with an unconditional change, then I'm happy to drop this idea.

modules/libcom/src/misc/epicsTypes.h
mdavidsaver: With this change, the situation you describe would certainly become a compilation error.  I don't see any way to avoid a breakage in this situation though.  The fix (an explicit cast) is simple and backwards compatible.

The potentially nastier sort of bug w/ c++ is when the type of a virtual method or template changes as these might not result in a compilation error.  I've been bitten by this when porting code for a library which switched a base class method from `long` -> `size_t` so that my overloaded method no longer matched.  Still, this possibility isn't enough to deter me.

mdavidsaver:
> Are there others options?

I guess another possibility would be to rename `epicsInt8` or add a new `epicsSInt8`.  Although this would result in a large amount of imo. mostly unnecessary churn.

anjohnson:
If we're going to change it, I think it makes more sense to do it on all target arch's. Doing it for only some could make mitigation in downstream modules even more tricky and hard to explain, so I'd rather this be all or nothing.

Dirk's test case at the top of this PR shows the issues we get at the moment. The important question is whether the problems due to the change would be any worse than those we're solving, because a `CHAR` array/waveform would become an `INT8` waveform instead and would have a different signedness than an actual `char` on some targets.

Dirk did miss out the aSub record type from Base which could also be affected, and from synApps possibly only the mcaRecord. I added the lsi and lso records to try to persuade people to stop using waveforms to hold string data, and we could point to these as alternatives if we do get any complaints.

There seem to be some uses of `epicsInt8` in the code (particularly in modules/database/src) that should probably be looked at carefully as well, e.g. these two results from a `git grep` seem inconsistent. They may have been mitigating the earlier inconsistency so perhaps they could now be made uniform:
```
modules/database/src/ioc/dbStatic/dbStaticRun.c:        case DBF_CHAR:  *(epicsInt8 *)pfield = (epicsInt8) i64; break;
modules/database/src/ioc/dbStatic/dbStaticRun.c:        case DBF_UCHAR:  *(epicsUInt8 *)pfield = (epicsInt8) u64; break;
```
The code for `epicsParseInt8()` in modules/libcom/src/misc/epicsStdlib.c actually looks right now...

dirk-zimoch:
I prefer to fix `epicsInt8` on all architectures. Anything else make the code more intransparent. On caritectures where `char` is unsigned, the change is needed and on the others it make no difference (expect for C++ type matching).
I would vote against a new `epicsSInt8`. It would only add complications.
I tested `-fsigned char` on vxWorks PPC but that leads to other problems (crashes) in the OS. (I had *not* rebuilt all vxWorks code.)

Checking several common drivers which use `epicsInt8`:

- Asyn seems to be fine as it just copies data but does not try to convert.
- Motor is fine. No conversions from/to `epicsInt8`.
- StreamDevice will change (i.e. correct) behavior on PPC when outputting `CHAR` waveforms as signed integers with formats like `"%i"`. If that is a problem, the user can simply use (should have used already) `UCHAR` waveforms.
- AreaDetector will change (i.e. correct) behavior on PPC when converting `epicsInt8`/`CHAR` input data to signed data types like `double`. The FFT plugin seems to be the only code affected by this.  AreaDetector assumes `epicsInt8` to be signed and thus currently works wrong on PPC.

anjohnson:
Group meeting 12-Jan-22: Approved.

PR #221, opened 2022-01-17
Add epicsBaseDebugLog

tboegi:
Add a debug print, inspired by asynPrint():
epicsBaseDebugLog("fmt", ...)

Example: A print in source code like this:
epicsBaseDebugLog ("idx=%u socket=%d revents=0x%x\n",
                    idx, (int)pPollFds[idx].fd, pPollFds[idx].revents);

will give a print like this:
2022/01/17 08:41:07.042  repeater.cpp:662 idx=1 socket=4 revents=0x1

As the printing is heavily needed in the changes in libcom for IPv6,
put the source into libcom.
But anybody linked against libcom can use it.

There are 2 versions:

  epicsBaseDebugLog(fmt, ...)
    Date, time __FILE__ and __LINE__ are inserted into the print.

   epicsBaseDebugLogFL(fmt, fileName, lineNo, ...)
    Date, and are inserted into the print.
    __FILE__ and __LINE__ are provided by the caller.

### DIFF
--- a/modules/libcom/src/Makefile
+++ b/modules/libcom/src/Makefile
@@ -32,6 +32,7 @@ include $(LIBCOM)/cvtFast/Makefile
 include $(LIBCOM)/cppStd/Makefile
 include $(LIBCOM)/cxxTemplates/Makefile
 include $(LIBCOM)/dbmf/Makefile
+include $(LIBCOM)/debuglog/Makefile
 include $(LIBCOM)/ellLib/Makefile
 include $(LIBCOM)/env/Makefile
 include $(LIBCOM)/error/Makefile
--- a/modules/libcom/src/debuglog/Makefile
+++ b/modules/libcom/src/debuglog/Makefile
@@ -1,0 +1,14 @@
+#*************************************************************************
+# Copyright (c) 2010 UChicago Argonne LLC, as Operator of Argonne
+#     National Laboratory.
+# EPICS BASE is distributed subject to a Software License Agreement found
+# in file LICENSE that is included with this distribution. 
+#*************************************************************************
+
+# This is a Makefile fragment, see src/libCom/Makefile.
+
+SRC_DIRS += $(LIBCOM)/debuglog
+
+INC += epicsBaseDebugLog.h
+
+Com_SRCS += epicsBaseDebugLog.c
--- a/modules/libcom/src/debuglog/epicsBaseDebugLog.c
+++ b/modules/libcom/src/debuglog/epicsBaseDebugLog.c
@@ -1,0 +1,32 @@
+/*************************************************************************\
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+#ifndef epicsBaseDebugLogh
+#define epicsBaseDebugLogh
+
+#include  <string.h>
+#include  <epicsStdio.h>
+#include  <stdarg.h>
+#include  "epicsTime.h"
+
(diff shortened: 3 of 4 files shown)


mdavidsaver:
Is there a compelling need for another logging API?  Could functionality like this instead be integrated into `errlogPrintf()`?

wrt. implementation.  I would like to see less complicated inlined code.  Also, are you aware of the portability issues with `__VA_ARGS__` and MSVC?  While this may finally be [ending](https://docs.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview), it will remain a concern for some time.

A relevant example of the logging front end in [pvxs](https://github.com/mdavidsaver/pvxs/blob/f7e5f3c127b3b89be029672db6ee3f7a573c6458/src/pvxs/log.h#L78-L81), which is [backed](https://github.com/mdavidsaver/pvxs/blob/f7e5f3c127b3b89be029672db6ee3f7a573c6458/src/log.cpp#L51) by errlog.  Where:

```c
log_info_printf(io, "Searching to %s%s\n", ...
```

Log a line like:

```
2022-01-17T08:17:50.588655144 INFO pvxs.client.io Searching to 192.168.1.255:5076
```

tboegi:
Thanks @mdavidsaver for the fast feedback.
What I was asking for is a logging function, that does NOT loose messages and adds a timestamp.
The existing errlogPrintf() does loose debug-messages,` exactly when you need them most.
After some digging I think that the reason is the non-buffered IO on stderr.
This is probably the right thing to do on a server:
Put out some messages here and there, and if they are too many in a short amount of time,
simply drop them.
For a client (or a server) in debugging mode, it is sometimes better to slow down the business
when using a blocking fprintf(stdout) but do not loose logs.
And then we need the timestamps....
How would your suggestion look like for logging:
- Do not loose message, wait if needed
-  Can be called from C-code
- Adds a timestamp

mdavidsaver:
> * Do not loose message, wait if needed

To my mind, a critical feature of a debug logging API is to (ideally) have no impact on execution of the calling code.  In practice of course "no impact" is impossible, so we settle for "minimally invasive".  Which doesn't describe blocking `printf()`.

If I find that errlog is dropping messages an inconvenient times I start by increasing the buffer depth, or with PVXS use filtering to only print those messages I think are relevant.

As I see it, "minimally invasive" and "lose-less" are mutually exclusive.  If I only get one, I will choose minimally invasive.


> * Can be called from C-code

`pvxs/log.h` can be translated into portable C code.  There is no global constructor code, only static initial values.  The rest consists of atomic integer operations, and a thread local `char[]` for the timestamp string.  Both can also be accomplished with libCom APIs.

> * Adds a timestamp

Perhaps the following would be a better least common denominator for Base?

```c
/** Return current time printed to a thread local buffer.
 * The buffer is overwritten/invalidated by subsequent calls from the same thread.
 * Will never return NULL.
 * If not NULL, the "fmt" argument is passed through to epicsTimeToStrftime().
 * If NULL, a default is used.
 */
LIBCOM_API
const char* errlogCurrentTime(const char* fmt);
```

This could be used in conjunction with `printf()` or `errlogPrintf()`.

eg.

```c
printf("%s hello wrold\n", errlogCurrentTime());
```

tboegi:
@mdavidsaver  The errlogCurrentTime() could be a way forward.
Let's hope that I can spend some time on this the next days/week.

anjohnson:
Core Group 2022-10-27: Not something we want to see as is, adopting @mdavidsaver's suggestions above would be a suitable way forward.

tboegi:
Thanks for the feedback - I just close this one

PR #222, opened 2022-02-01
EXT link modifier

mdavidsaver:
A re-design of #111 which ~flips the logic.  This maps better to the use case I had in mind.  I try to remember to check to `dbcar()` frequently when editing a .db file, but often slip up.  Typos become broken links.

Adds three new link modifiers: `INT`, `AUTO`, and `EXT`.

- `AUTO` make the current default explicit.  DB and CA links may target any PV, local or remote.
- `INT` requires that a DB or CA link target a local record.
- `EXT` requires that a DB or CA link must not target a local record.  Effectively prevents DB links.

In addition to appearing explicitly as link modifiers, the implied default may be set a new .db file `set()` statement.

For example:

```
# default default is AUTO
set("link:scope","INT")

record(longin, "my:dev:sigA") {
    field(INP, "my:def:sigB CPP")
    field(TPRO, "1")
}

record(longin, "my:dev:sigB") {
    field(INP, "some:other EXT")
    field(TPRO, "1")
}

# optional, restore default for any subsequent records
set("link:scope","AUTO")
set("link:scope") # equivalent
```

image

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -24,6 +24,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#define EPICS_PRIVATE_API
+
 #include "alarm.h"
 #include "cantProceed.h"
 #include "cvtFast.h"
--- a/modules/database/src/ioc/db/dbLink.c
+++ b/modules/database/src/ioc/db/dbLink.c
@@ -67,6 +67,27 @@ const char * dbLinkFieldName(const struct link *plink)
     return "????";
 }
 
+const char *dbLinkSrcName(const struct link* plink)
+{
+    const char *ret = "";
+
+    if(plink->type == DB_LINK || plink->type == CA_LINK || plink->type == PV_LINK) {
+        unsigned srcMask = plink->value.pv_link.pvlMask & pvlOptSrcMask;
+
+        if(srcMask == pvlOptSrcUnDef)
+            srcMask = plink->flags & pvlOptSrcMask;
+
+        ret = " INVALID"; /* should not be possible */
+        switch(srcMask) {
+        case pvlOptSrcUnDef:ret = " UNDEF"; break; /* should also not be possible */
+        case pvlOptSrcAuto: ret = ""; break;
+        case pvlOptSrcInt:  ret = " INT"; break;
+        case pvlOptSrcExt:  ret = " EXT"; break;
+        }
+    }
+    return ret;
+}
+
 /* Special TSEL handler for PV links */
 /* FIXME: Generalize for new link types... */
 static void TSEL_modified(struct link *plink)
--- a/modules/database/src/ioc/db/dbLink.c
+++ b/modules/database/src/ioc/db/dbLink.c
@@ -89,28 +89,28 @@ static void TSEL_modified(struct link *plink)
 
 /***************************** Generic Link API *****************************/
 
-void dbInitLink(struct link *plink, short dbfType)
+long dbInitLink(struct link *plink, short dbfType)
 {
     struct dbCommon *precord = plink->precord;
 
     /* Only initialize link once */
     if (plink->flags & DBLINK_FLAG_INITIALIZED)
-        return;
+        return 0;
     else
         plink->flags |= DBLINK_FLAG_INITIALIZED;
 
     if (plink->type == CONSTANT) {
         dbConstInitLink(plink);
-        return;
+        return 0;
     }
 
     if (plink->type == JSON_LINK) {
         dbJLinkInit(plink);
-        return;
+        return 0;
     }
 
     if (plink->type != PV_LINK)
-        return;
+        return 0;
 
     if (plink == &precord->tsel)
         TSEL_modified(plink);
@@ -118,7 +118,7 @@ void dbInitLink(struct link *plink, short dbfType)
     if (!(plink->value.pv_link.pvlMask & (pvlOptCA | pvlOptCP | pvlOptCPP))) {
         /* Make it a DB link if possible */
         if (!dbDbInitLink(plink, dbfType))
-            return;
+            return 0;
     }
 
     /* Make it a CA link */
@@ -142,6 +142,8 @@ void dbInitLink(struct link *plink, short dbfType)
                 plink->value.pv_link.pvname);
         }
     }
+
+    return 0;
 }
 
 void dbAddLink(struct dbLocker *locker, struct link *plink, short dbfType,
--- a/modules/database/src/ioc/db/dbLink.c
+++ b/modules/database/src/ioc/db/dbLink.c
@@ -41,8 +41,9 @@
 #include "dbLink.h"
 #include "dbLock.h"
 #include "dbScan.h"
-#include "dbStaticLib.h"
+#include "dbStaticPvt.h"
 #include "devSup.h"
+#include "epicsString.h"
 #include "link.h"
 #include "recGbl.h"
 #include "recSup.h"
@@ -86,18 +87,67 @@ static void TSEL_modified(struct link *plink)
     }
 }
 
+typedef struct {
+    jlink j;
+} dummyLink;
+
+static
+void dummyJlifFree(jlink *ptr) {}
+
+static
+jlif dummyJlif = {
+    "dummy",
+    NULL,
+    &dummyJlifFree,
+};
+
+static
+void dummyLsetRemove(struct dbLocker *locker, struct link *plink) {}
+
+static
+int dummyLsetConnected(const struct link *plink)
+{
+    return 0;
+}
+
+static
+lset dummyLset = {
+    0,
+    1,
+    NULL,
+    &dummyLsetRemove,
+    NULL,
+    NULL,
+    NULL,
+    &dummyLsetConnected,
+};
+
+static
+dummyLink dummyL = {
+    {
+        &dummyJlif,
+        NULL,
+        0,
+        0,
+    }
+};
 
 /***************************** Generic Link API *****************************/
 
 long dbInitLink(struct link *plink, short dbfType)
 {
     struct dbCommon *precord = plink->precord;
+    unsigned srcMask;
 
     /* Only initialize link once */
-    if (plink->flags & DBLINK_FLAG_INITIALIZED)
+    if (plink->flags & DBLINK_FLAG_INITIALIZED) {
         return 0;
-    else
+    } else {
         plink->flags |= DBLINK_FLAG_INITIALIZED;
+        if((plink->flags & pvlOptSrcMask)==pvlOptSrcUnDef) {
+            plink->flags |= pvlOptSrcAuto; /* link field never set from any .db file */
+        }
+    }
 
     if (plink->type == CONSTANT) {
         dbConstInitLink(plink);
@@ -111,6 +161,60 @@ long dbInitLink(struct link *plink, short dbfType)
 
     if (plink->type != PV_LINK)
         return 0;
+    /* From this point, must change link type to something else on success or error */
+
+    srcMask = plink->value.pv_link.pvlMask & pvlOptSrcMask;
+
+    if(srcMask == pvlOptSrcUnDef) {
+        /* apply default from set("link:scope", ...) */
+        srcMask = plink->flags & pvlOptSrcMask;
+        assert(srcMask != pvlOptSrcUnDef); /* file scope is always defined */
+    }
+
+    int isLocal = dbChannelTest(plink->value.pv_link.pvname)==0;
+    if(srcMask != pvlOptSrcAuto) {
+        int allow = 1;
+
+        if(srcMask == pvlOptSrcInt && !isLocal) {
+            /* local link to non-local target */
+            DBENTRY closest;
+            const char *fldname = dbLinkFieldName(plink);
+            fprintf(stderr, "%s.%s " ERL_ERROR ": Unable to create INT link to \"" ANSI_BOLD("%s") "\".\n",
+                    precord->name, fldname, plink->value.pv_link.pvname);
+
+            dbInitEntry(pdbbase, &closest);
+            if(!dbFindRecordSimilar(&closest, plink->value.pv_link.pvname, NULL)) {
+                unsigned indent = strlen(precord->name) + strlen(fldname) + 23;
+                epicsPrintf("%*s  Did you mean \"" ANSI_BOLD("%s.%s") "\" ?\n",
+                            indent, "",
+                            closest.precnode->recordname,
+                            closest.pflddes->name);
+            }
+            dbFinishEntry(&closest);
+
+            allow = 0;
+
+        } else if(srcMask == pvlOptSrcExt && isLocal) {
+            /* external link to local target */
+            const char *fldname = dbLinkFieldName(plink);
+
+            fprintf(stderr, "%s.%s " ERL_ERROR ": Unable to create EXT link to \"" ANSI_BOLD("%s") "\".\n",
+                    precord->name, fldname, plink->value.pv_link.pvname);
+
+            allow = 0;
+        }
+
+        if(!allow) {
+            /* We must replace PV_LINK with something else to allow for correct shutdown.
+             */
+            plink->type = JSON_LINK;
+            plink->lset = &dummyLset;
+            plink->value.json.jlink = &dummyL.j;
+            plink->value.json.string = "{}";
+
+            return S_dbLib_badLink;
+        }
+    }
 
     if (plink == &precord->tsel)
         TSEL_modified(plink);
@@ -124,8 +228,6 @@ long dbInitLink(struct link *plink, short dbfType)
     /* Make it a CA link */
     if (dbfType == DBF_INLINK)
         plink->value.pv_link.pvlMask |= pvlOptInpNative;
-
-    int isLocal = dbChannelTest(plink->value.pv_link.pvname)==0;
 
     dbCaAddLinkCallbackOpt(NULL, plink, NULL, NULL, NULL, isLocal ? DBCA_CALLBACK_INIT_WAIT : 0);
     if (dbfType == DBF_FWDLINK) {
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1519,6 +1519,29 @@ static void dbAlias(char *name, char *alias)
     dbFinishEntry(pdbEntry);
 }
 
+static void warnSetUnknown(const char *name)
+{
+    fprintf(stderr, ERL_WARNING " Unknown name in set(\"%s\", ...)\n", name);
+    dbIncludePrint();
+}
+
+static void dbSet(const char *name)
+{
+    warnSetUnknown(name);
+
+    /* clean up unused arguments */
+    while(ellCount(&tempList))
+        dbmfFree(popFirstTemp());
+}
+
+unsigned dbLinkScopeDefault(void)
+{
+    int ret = pvlOptSrcAuto;
+    if(pinputFileNow)
+        ret = pinputFileNow->linkDefLoc;
+    return ret;
+}
+
 static void dbRecordBody(void)
 {
     DBENTRY *pdbentry;
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -61,6 +61,9 @@ epicsExportAddress(int,dbQuietMacroWarnings);
 int dbRecordsAbcSorted=0;
 epicsExportAddress(int,dbRecordsAbcSorted);
 
+int dbLinkDefaultINT=0;
+epicsExportAddress(int,dbLinkDefaultINT);
+
 /*private routines */
 static void yyerrorAbort(char *str);
 static void allocTemp(void *pvoid);
@@ -110,6 +113,7 @@ typedef struct inputFile{
     const char  *filename;
     FILE        *fp;
     int         line_num;
+    unsigned linkDefLoc;
 }inputFile;
 static ELLLIST inputFileList = ELLLIST_INIT;
 
@@ -293,6 +297,7 @@ static long dbReadCOM(DBBASE **ppdbbase,const char *filename, FILE *fp,
         pinputFile->fp = fp;
         fp = NULL;
     }
+    pinputFile->linkDefLoc = dbLinkDefaultINT ? pvlOptSrcInt : pvlOptSrcAuto;
     pinputFile->line_num = 0;
     pinputFileNow = pinputFile;
     my_buffer[0] = '\0';
@@ -1527,7 +1532,26 @@ static void warnSetUnknown(const char *name)
 
 static void dbSet(const char *name)
 {
-    warnSetUnknown(name);
+    if(strcmp(name, "link:scope")==0) {
+        char* val = popFirstTemp();
+
+        if(!val || strcmp(val, "AUTO")==0) {
+            pinputFileNow->linkDefLoc = pvlOptSrcAuto;
+
+        } else if(strcmp(val, "EXT")==0) {
+            pinputFileNow->linkDefLoc = pvlOptSrcExt;
+
+        } else if(strcmp(val, "INT")==0) {
+            pinputFileNow->linkDefLoc = pvlOptSrcInt;
+
+        } else {
+            fprintf(stderr, ERL_WARNING " Invalid value in set(\"%s\", \"%s\")\n", name, val);
+        }
+        dbmfFree(val);
+
+    } else {
+        warnSetUnknown(name);
+    }
 
     /* clean up unused arguments */
     while(ellCount(&tempList))
@@ -1536,7 +1560,7 @@ static void dbSet(const char *name)
 
 unsigned dbLinkScopeDefault(void)
 {
-    int ret = pvlOptSrcAuto;
+    int ret = dbLinkDefaultINT ? pvlOptSrcInt : pvlOptSrcAuto;
     if(pinputFileNow)
         ret = pinputFileNow->linkDefLoc;
     return ret;
--- a/modules/database/src/ioc/dbStatic/link.h
+++ b/modules/database/src/ioc/dbStatic/link.h
@@ -68,6 +68,23 @@ DBCORE_API extern const maplinkType pamaplinkType[LINK_NTYPES];
 #define pvlOptInpString  0x100  /*Input as string*/
 #define pvlOptOutNative  0x200  /*Output native*/
 #define pvlOptOutString  0x400  /*Output as string*/
+/* DB/CA link target mask.
+ * If unset, then the implied default (file scope) is used.
+ * - INT:  link must target local record
+ * - EXT:  link must not target local record
+ * - AUTO: link may target any PV
+ *
+ * Used in two places:
+ * - dbLink::flags
+ *   Reflects default if no modifier explicitly provided.  Will never be UnDef.
+ * - dbLink::value::pv_link::pvlMask
+ *   Flags parsed from link string.  May be UnDef.
+ */
+#define pvlOptSrcMask   0x3000
+#define pvlOptSrcUnDef  0x0000
+#define pvlOptSrcInt    0x1000
+#define pvlOptSrcExt    0x2000
+#define pvlOptSrcAuto   0x3000
 
 /* DBLINK Flag bits */
 #define DBLINK_FLAG_INITIALIZED    1 /* dbInitLink() called */
--- a/modules/database/src/ioc/dbStatic/link.h
+++ b/modules/database/src/ioc/dbStatic/link.h
@@ -73,6 +73,7 @@ DBCORE_API extern const maplinkType pamaplinkType[LINK_NTYPES];
 #define DBLINK_FLAG_INITIALIZED    1 /* dbInitLink() called */
 #define DBLINK_FLAG_TSELisTIME     2 /* Use TSEL to get timeStamp */
 #define DBLINK_FLAG_VISITED        4 /* Used in loop detection */
+/* pvlOptSrc* bits also used to indicate default from file parsing. */
 
 struct macro_link {
     char *macroStr;
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -23,6 +23,8 @@
 #include <errno.h>
 #include <limits.h>
 
+#define EPICS_PRIVATE_API
+
 #include "dbDefs.h"
 #include "ellLib.h"
 #include "envDefs.h"
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -551,11 +551,13 @@ static long doResolveLinks(dbRecordType *pdbRecordType, dbCommon *precord,
     dbFldDes **papFldDes = pdbRecordType->papFldDes;
     short *link_ind = pdbRecordType->link_ind;
     int j;
+    long ret = 0;
 
     /* For all the links in the record type... */
     for (j = 0; j < pdbRecordType->no_links; j++) {
         dbFldDes *pdbFldDes = papFldDes[link_ind[j]];
         DBLINK *plink = (DBLINK*)((char*)precord + pdbFldDes->offset);
+        long status;
 
         if (ellCount(&precord->rdes->devList) > 0 && pdbFldDes->isDevLink) {
             devSup *pdevSup = dbDTYPtoDevSup(pdbRecordType, precord->dtyp);
@@ -568,9 +570,11 @@ static long doResolveLinks(dbRecordType *pdbRecordType, dbCommon *precord,
             }
         }
 
-        dbInitLink(plink, pdbFldDes->field_type);
+        status = dbInitLink(plink, pdbFldDes->field_type);
+        if(!ret)
+            ret = status; /* latch first error */
     }
-    return 0;
+    return ret;
 }
 
 static long doInitRecord1(dbRecordType *pdbRecordType, dbCommon *precord, void *user)
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -85,7 +85,7 @@ static void initDrvSup(void);
 static void initRecSup(void);
 static void initDevSup(void);
 static void finishDevSup(void);
-static void initDatabase(void);
+static long initDatabase(void);
 static void initialProcess(void);
 static void exitDatabase(void *dummy);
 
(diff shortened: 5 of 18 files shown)


mdavidsaver:
Open questions:

* Syntax
  * `REM` or `EXT` or ???
  * `set("linkScopeDefault","LOC")` is a bit word-y.
* Behavior
  * Should runtime re-target really apply the same restriction as during `iocInit()`?

ralphlange:
Interesting! Great use of `epicsStrSimilarity()`.

What about a third option `AUTO` that implements the current behavior and is the default - to ensure compatibility for existing applications? I don't see how else existing databases could be run without causing a lot of messages and failing the bootup.

I could live with lots-of-warnings for existing databases if there was a one-line way to suppress them in cases where you just want to deal with the necessary changes later. But "IOC not running!" for an existing database should really be avoided.

mdavidsaver:
> ... to ensure compatibility for existing applications?

To be clear.  This PR is intended to be a backwards compatible change.  The current behavior remains the (implicit) default.

mdavidsaver:
> ... But "IOC not running!" for an existing database should really be avoided.

It is.  Individual .db files must explicitly opt in this new behavior.  So all existing .db should load exactly as before.

mdavidsaver:
> ... So all existing .db should load exactly as before.

Actually.  I should qualify this.  All existing _correctly formed_  databases should load as before.  I'm adding more error propagation in `iocInit()`.  So it is possible that an IOC loading some weirdly constructed link could now fail completely, instead of only partly.  I personally see this as acceptable, even desirable, change.

ralphlange:
> > ... But "IOC not running!" for an existing database should really be avoided.
> 
> It is. Individual .db files must explicitly opt in this new behavior. So all existing .db should load exactly as before.

Ok. I didn't get that.

Re: Names.
I find `EXT` a lot more obvious than `REM` (which leads a comment in Windows batch files, right?).
At the same time it is obvious that local/remote are an antonym pair, not local/external.
(Newspeak would use local/unlocal. Just saying. Would be a doubleplusungood choice.)

modules/database/src/ioc/misc/iocInit.c
```diff
@@ -546,28 +557,34 @@ static void doResolveLinks(dbRecordType *pdbRecordType, dbCommon *precord,
...
+static long doInitRecord1(dbRecordType *pdbRecordType, dbCommon *precord,
     void *user)
 {
     rset *prset = pdbRecordType->prset;
 
-    if (!prset) return;         /* unlikely */
+    if (!prset) return S_dbLib_noRecSup;         /* unlikely */
 
     if (prset->init_record)
         prset->init_record(precord, 1);
```
anjohnson, line 587: Most `rset::init_record()` routines already return an error status (mostly for missing or incomplete device support, or whatever `dset::init_record()` returns). We currently ignore that status, but this would give us the ability to make use of it if we wanted to.

However given my next comment below, that might not be such a wonderful idea...

modules/database/src/ioc/misc/iocInit.c
```diff
@@ -546,28 +557,34 @@ static void doResolveLinks(dbRecordType *pdbRecordType, dbCommon *precord,
...
-    return;
+    ret = iterateRecords(doInitRecord0, NULL);
+    if(!ret)
+        ret = iterateRecords(doResolveLinks, NULL);
+    if(!ret)
+        ret = iterateRecords(doInitRecord1, NULL);
+
+    if(!ret)
+        ret = epicsAtExit(exitDatabase, NULL);
+    return ret;
```
anjohnson, line 607: The major behavior change with this commit is that now any failure in any of the routines that get called by `initDatabase()` will stop any later routines from being called and `iocInit()` will exit immediately, whereas previously those errors were ignored and everything got run anyway. I'm not necessarily saying that we shouldn't make this change, I am pointing out the effect so we're all aware of it.

Accepting this commit will thus prevent IOCs that had these kinds of errors but used to run anyway from being able to be started at all after upgrading to this release. Some sites might be unhappy with that, since the unaffected parts of their IOCs used to work despite the errors — that's called graceful degradation, and it might actually be an important characteristic of an EPICS control system that we weren't previously aware of.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1265,6 +1266,29 @@ static void dbAlias(char *name, char *alias)
     dbFinishEntry(pdbEntry);
 }
 
+static void dbScopeSet(const char *name, const char *val)
+{
+    if(strcmp(name, "linkScopeDefault")==0) {
+        if(strcmp(val, "LOC")==0) {
+            pinputFileNow->linkDefLoc = 1;
+        } else if(strcmp(val, "REM")==0) {
+            pinputFileNow->linkDefLoc = 0;
```
anjohnson: My preferred alternative to `LOC`/`REM` (local/remote) would be `INT`/`EXT` (internal/external) which might be a bit more obvious to some people — I see Ralph already suggested the same thing. My concern about using "local" is local to what? Someone might think this means the link targets must be local to this .db file (since that's the scope to the "set" command).

I agree that `linkScopeDefault` seems a bit wordy, I don't have a good replacement yet but I'll sleep on it.

modules/libcom/src/error/errlog.h
```diff
@@ -275,13 +275,15 @@ LIBCOM_API void errSymLookup(long status, char *pBuf, size_t bufLength);
...
 #define ANSI_ESC_CYAN "\033[36;1m"
+#define ANSI_ESC_BOLD "\033[1m"
 #define ANSI_ESC_RESET "\033[0m"
 #define ANSI_RED(STR)     ANSI_ESC_RED     STR ANSI_ESC_RESET
 #define ANSI_GREEN(STR)   ANSI_ESC_GREEN   STR ANSI_ESC_RESET
 #define ANSI_YELLOW(STR)  ANSI_ESC_YELLOW  STR ANSI_ESC_RESET
 #define ANSI_BLUE(STR)    ANSI_ESC_BLUE    STR ANSI_ESC_RESET
 #define ANSI_MAGENTA(STR) ANSI_ESC_MAGENTA STR ANSI_ESC_RESET
 #define ANSI_CYAN(STR)    ANSI_ESC_CYAN    STR ANSI_ESC_RESET
+#define ANSI_BOLD(STR)    ANSI_ESC_BOLD    STR ANSI_ESC_RESET
```
anjohnson: Future enhancement: There is another code `"\033[22m"` which clears bold mode without also affecting the foreground color. Using that instead of the reset sequence would allow nesting of `ANSI_BOLD()` with the colors.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -1101,6 +1104,17 @@ static long dbPutFieldLink(DBADDR *paddr,
     if (status)
         return status;
 
+    /* do not apply per-file default from parser for changes after iocInit() */
```
anjohnson: I think this described behavior is right, we ran a test at APS just before Christmas where I redirected an internal link in our main timing IOC to an external IOC. I don't think that's an unusual thing to want to do.

anjohnson (commented):
Various suggestions and comments. In general this seems like a good idea although the scope of the `set()` DB statement would extend further if someone uses it in a template.db that gets used at build-time by msi.

modules/database/src/ioc/db/dbAccess.c
mdavidsaver: I should note that even it will always be possible to retarget a link as CA.  Applying the per-file default would only required the user making the change know to append " REM" (or " EXT").

I chose not to required this as it seems to me like more annoyance than added value.  An interactive change will anyway provide immediate feedback.  (eg.  caput will show a new link value with "CA")

modules/libcom/src/error/errlog.h
mdavidsaver: Nesting seems interesting, but something I'll leave for the future.  Note that the `ANSI_ESC_*` color code macros currently set both foreground color and bold (eg. `"\033[31;1m"`).  So as is, nesting an SGR 22 would have the effect to un-bold the following text.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
mdavidsaver: > My preferred alternative to LOC/REM (local/remote) would be INT/EXT ...

ok.

modules/database/src/ioc/misc/iocInit.c
mdavidsaver: > The major behavior change with this commit is ...

Avoiding potential for change is why I explicitly didn't add a test for eg. `init_record()` returning an error.  So I believe the only previously "soft" errors which would now become "hard" errors have to do with link initialization and missing device/driver support tables (imo. a serious error).

> ... that's called graceful degradation ...

Also a trap for the unwary...  Our default is like [`On Error Resume Next`](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/on-error-statement).

ralphlange:
Don't we need an explicit `AUTO` keyword too, to allow specifying the classical behavior in a context that has set the default to `LOC` or `REM`?

mdavidsaver:
I've renamed `LOC` -> `INT` and `REM` -> `EXT`.

> Don't we need an explicit `AUTO` keyword too

The current default can be made explicit by including:

```
set("linkScopeDefault", "AUTO")
```

I'm undecided as to whether a `INT` link modifier is worth the effort to maintain.  It may be my lack of imagination, but I don't see it becoming common to add `INT` to an existing .db file.  Rather, I see the migration path being to add `set("linkScopeDefault", "INT")` and then fix any errors which appear on startup.

mdavidsaver:
Notes on paths not (yet) taken.  I chose `set("linkScopeDefault", "INT")` instead of something shorter like `linkScopeDefault("INT")` because I intend for `set()` to be a hook for future forward compatible changes.  `set()` will only warn for unknown key/value.

> ... Someone might think this means the link targets must be local to this .db file ...

I've considered adding a fully per-file restriction, but decided to pass for the present.  eg. `set("linkScopeDefault", "FILE")`  I think it could be accomplished without much effort, or overhead, by keeping a counter of files opened (incremented in `dbOpenFile()`), then storing the current count in `dbRecord` and `DBLINK`.  The check would then be to require that source and target have the same count.

> ... someone uses it in a template.db that gets used at build-time by msi.

wrt. MSI.  I could see adding something like [`#line`](https://en.cppreference.com/w/cpp/preprocessor/line) to annotate generated .db files.  eg.

```
set("source", "mydev.substitutions:15:VAR=X,OTHER=Y")
set("source", "piece.template:42")
... some record()
set("source", "") # reset's the stack of extra sources
```

Definitely out of scope for this PR though.

mdavidsaver:
Possible alternative syntax:

```
set("linkScopeDefault", "AUTO")

set("link.scope", "AUTO")  # does "default" add much info?

set("link:scope", "AUTO")  # ':' separator following the developing convention for info()

flag("link.scope", "AUTO") # is "set" too overloaded?

use("link.scope", "AUTO") # nod to perl's "use strict" and similar.
```

ralphlange:
> I've renamed `LOC` -> `INT` and `REM` -> `EXT`.
> 
> > Don't we need an explicit `AUTO` keyword too
> 
> The current default can be made explicit by including:
> 
> ```
> set("linkScopeDefault", "AUTO")
> ```
> 
> I'm undecided as to whether a `INT` link modifier is worth the effort to maintain. It may be my lack of imagination, but I don't see it becoming common to add `INT` to an existing .db file. Rather, I see the migration path being to add `set("linkScopeDefault", "INT")` and then fix any errors which appear on startup.

Let me add a use case:
I am doing a Device Support module that provides some post-processing functionality. Something like - what I did recently - a support for FFT of a sampled signal.
Part of that is a template database containing the whole set of records to configure and run one FFT instance. All the links between those records should be `INT`, as doing them across CA would add loads of overhead and jitter. Except for the one input link that points to the time domain input array (often a ring buffer of the sampled value). I want my template to work with the FFT done on the same IOC as the sampling or on a separate IOC, depending on the CPU power and configuration and number of FFT instances.
Wouldn't the most straightforward approach be declaring `INT` as default and setting the input link explicitly to `AUTO`?

mdavidsaver:
> Wouldn't the most straightforward approach be declaring `INT` as default and setting the input link explicitly to `AUTO`?

I think you are describing is the use case / migration path I tried to articulate in https://github.com/epics-base/epics-base/pull/222#issuecomment-1049074006.  Setting `linkScopeDefault` to `INT`, and marking individual links with `EXT`.

> ... Part of that is a template database ...

hmmm.  I usually think of .db files installed by my modules as complete, not as fragments for external inclusion.  Although I know the later happens sometimes.

As I think about it.  The `set()` toggles `pinputFileNow->linkDefLoc` which is latched in `dbPutString()` for a `DBF_*LINK`.  So a `set("linkScopeDefault", ...` can meaningfully appear multiple times in a single file.  eg. fragment could switch to INT, and then back to AUTO.

```
set("linkScopeDefault", "INT")
... some records()
set("linkScopeDefault", "AUTO")
... other records()
```

The lack of lexical scoping makes it easy to imagine forgetting the second `set()`.  Although the consequences of omission would be immediate and loud (if there are any external links).

anjohnson:
> I usually think of .db files installed by my modules as complete, not as fragments for external inclusion.

I don't think you should assume that, every site seems to configure IOCs slightly differently, and at some sites the order of records in a support module's .db file might _not_ be the same as the order that the records actually get loaded into their IOCs. I'm slightly concerned that this PR will be adding an ordering requirement on the handling of records that has never been present before, so it could be incompatible with how some sites currently create and load their IOC databases. A support module would have to jump through a few (probably fairly easy) hoops to use this while still keeping compatibility with older Base versions.

It's not that common at APS, but some of our engineers here have used msi with a .substitutions file to combine some or all of the original .db files for an IOC into one (or a few) fully- or partly-expanded .db files that the IOC loads. In that case any `set()` statements would leak beyond the original .db file in the expanded version unless `set("linkScopeDefault", "AUTO")` appears at the end of any .db file which sets it to something else.

mdavidsaver:
> I don't think you should assume that ...

I guess the pensive nature of "hmmm." didn't carry through.  So I am agreeing complete .db files are a bad assumption.  This is what motivates my discussion about using multiple `set("linkScopeDefault", ...`.

ralphlange:
Let me argue from a different angle, then.

We have three distinct modes of operation: `INT`, `EXT` and `AUTO`.
Why can I only select two of those modes explicitly and have to choose the third one by resetting the default and not marking the link explicitly?
Setting the mode through the default is a brittle thing with a tendency to have unintended side effects.

anjohnson:
Having an equivalent to the [`inherit` keyword](https://www.w3schools.com/cssref/css_inherit.asp) for CSS (web) might make sense if we introduced the ability to scoping of sections of the database. I didn't mention this idea before because the new `set()` statement is the only use we currently have for that, but something like this might be useful if the idea expands in the future:

```py
set("link.visibility", "INT") {
  record(ai, "$(P):In") {
    field(INP, "... INH")
  }
  ...
}
```

It might even be useful to set the visibility inside an individual record if we wanted to do that, but I'm probably getting ahead of myself with that idea, maybe that's a bit too fine-grained.

mdavidsaver:
> We have three distinct modes of operation: `INT`, `EXT` and `AUTO`.

Ah.  Maybe my use of jargon is getting in the way?  There are only two modes.  `AUTO` is the current behavior, and `INT` which would be the new behavior.  `EXT` only appears in link strings while (atm.) `INT` and `AUTO` never appear in link strings.

mdavidsaver:
> ... I'm probably getting ahead of myself with that idea, maybe that's a bit too fine-grained.

My thoughts went in this direction as well.  I like the idea.  Though I decided that, for the moment, it would be too much effort.

ralphlange:
> > We have three distinct modes of operation: `INT`, `EXT` and `AUTO`.
> 
> Ah. Maybe my use of jargon is getting in the way? There are only two modes. `AUTO` is the current behavior, and `INT` which would be the new behavior. `EXT` only appears in link strings while (atm.) `INT` and `AUTO` never appear in link strings.

Ah. My bad. I understood `EXT` as "_must_ connect to an external PV", but you meant it as "_may_ connect to an external PV". I rest my use case.

mdavidsaver:
Also.  wrt. https://github.com/epics-base/epics-base/pull/222#issuecomment-1049093135  I'm leaning towards:

> set("link:scope", "AUTO")

mdavidsaver:
I've changed `linkScopeDefault` -> `link:scope`.  Also, `set()` can now accept a variable argument list (the first arg. is mandatory).  I'm treating the following as equivalent.

```
set("link:scope", "AUTO")
set("link:scope")
```

I've also added an entry to RELEASE_NOTES.md

anjohnson:
Regarding the name `link:scope` the term "scope" may not have much meaning for some end-users, and could cause confusion if you try to talk about the scope of the setting (as against the scope of the link searches themselves). That's why I suggested using the word "visibility" earlier, although some other similar term might be a bit more meaningful depending on exactly what it does.

@ralphlange By

> I rest my use case.

did you mean you are withdrawing your objection, or that you're saying "this demonstrates why I was confused and I still want it to be changed"? I could read that either way...

mdavidsaver:
> ... I suggested using the word "visibility" earlier ...

This didn't escape my notice.  My initial reaction was negative as "visibility" means something else to me (and it's longer).  I'm going to think about it though.

ralphlange:
> @ralphlange By
> 
> > I rest my use case.
> 
> did you mean you are withdrawing your objection, or that you're saying "this demonstrates why I was confused and I still want it to be changed"? I could read that either way...

`EXT` is doing exactly what I wanted to: explicitly setting a link so that it works externally or internally with no warnings.

modules/database/src/ioc/misc/iocInit.c
exzombie: From what I see, the only non-success return values come from the new EXT addition and from missing `prset`s, so I don't expect breakage of databases that are not already broken.

exzombie:
Overall, I'm very happy to see this feature. Based on our discussion, I think it is quite compatible; in summary, because unknown link modifiers are ignored, it is safe to add EXT to database templates distributed with modules without breaking compatibility with older versions ob epics-base. It is then up to the IOC developer to enable the new behavior. Perhaps it would be a good idea to add this recommendation to the release notes.

Now, for the bikeshedding part ;) . I think
- INT and EXT are quite clear
- "link:scope" is less than perfect, but sufficient, I find it clear enough.
- "set" could perhaps be renamed to "option", inspired by the sequencer which uses it to specify per-file and per-scope settings.

documentation/RELEASE_NOTES.md
```diff
@@ -19,50 +19,50 @@ should also be read to understand what has changed since earlier releases.
...
+The default behavior of DB/CA links makes it easy to target links between IOCs.
+It also makes it easy for typos to result in dangling CA links. This can be
+detected with caget/dbgf/dbcar, but all of these depend on explicit extra action
+being taken on the part of whoever sets up, or changes an IOC.
 
 To preserve compatibility, the default behavior of link does not change.
 
-This behavior/mode, now retroactively called `AUTO` may now be express explicitly with:
+This behavior/mode, now retroactively called `AUTO`, may now be expressed
+explicitly with:
```
anjohnson: Can you add words something like "in a .db file" here, just to make that explicit to readers.

documentation/RELEASE_NOTES.md
```diff
@@ -19,50 +19,50 @@ should also be read to understand what has changed since earlier releases.
...
 
 ```
 set("link:scope","AUTO")
 ```
 
-A new `INT` behavior/mode is added where links will be default be required to target
-record within the same IOC.  Links which may target an external PV must be marked
-with a new `EXT` link modifier.
+A new `INT` behavior/mode is added where links will by default be required to
```
anjohnson: Maybe say "A new behavior/mode configured by setting `link:scope` to `INT` (internal) is added ..."?

documentation/RELEASE_NOTES.md
```diff
@@ -19,50 +19,50 @@ should also be read to understand what has changed since earlier releases.
...
 
 This new mode is an attempt to reduce the possibility of typos in links going
-unnoticed by allowing the a **per-file** change to opt-in to a default.
+unnoticed by allowing a **per-file** change to opt-in to the new behavior.
 
 This allows the author of a .db file to explicitly mark those (usually few)
-links which are intended to become external `CA` links.
-Links not marked as `EXT` will trigger an error during `iocInit()` if the target
-is not a local record.
+links which are intended to become external `CA` links. Links not marked as
```
anjohnson: "With `link:scope` set to `INT`, links not marked ..." maybe?

documentation/RELEASE_NOTES.md
```diff
@@ -19,50 +19,50 @@ should also be read to understand what has changed since earlier releases.
...
 record(longin, "my:dev:sigB") {
     field(INP, "some:other EXT") # allowed
-    field(TPRO, "1")
 }
 ```
 
-In this example the first record has a like with a typo (`def` vs. `dev`).
+In this example the first record has a link with a typo (`def` vs. `dev`).
 With the `link:scope` set to `INT`, and no `EXT` modifier, loading this
 .db will now result in an error.
```
anjohnson: Needs words about the difference between what to do in the different kinds of .db files (templates intended for expansion etc.).

modules/database/src/ioc/misc/iocInit.c
anjohnson: I just noticed this again while re-reviewing the code. Should we add something like a "failFast" flag, which would tell the IOC whether to stop initialization on the first error it finds or to continue anyway? Operational IOCs might have that cleared, but that statement would have to be explicitly added to the st.cmd file, and by default it would be set. Unfortunately that would require more tests to check that things still run okay with it cleared.

anjohnson (commented):
One more idea/suggestion/question, and Release notes suggestions.

modules/database/src/ioc/misc/iocInit.c
exzombie: To elaborate on my previous comment: I have looked at `dbInitLink()`, which IIUC is the function you are worried about, and it returns success in all cases except in the case of a missing EXT modifier added by this PR. So while I think a "failFast" flag would be useful if new checks are added, it is not needed for this PR because it is an opt-in change already.

mdavidsaver:
I've updated the release notes

> ... Should we add something like a "failFast" flag ...

imo. no, nothing so generic.  A flag to demote a specific error to a warning might be acceptable to me.

> ... Operational IOCs ...

I'll repeat my skepticism of the idea that any IOC benefits from ignoring errors detectable during initialization.  imo. this only invites time to be lost in troubleshooting more complicated symptoms which arise later.  To me such laxness is justified only as a legacy behavior.  One which we are incrementally migrating away from.  This PR is one step.

anjohnson:
Core Group 7/20: ANJ and RL to review again, hopefully approve next time.

anjohnson:
Core Group 2022-10-27: Now acceptable, the Perl parser may need extending to add support for the `set(NAME, ...)` syntax in .db files.

AI on ANJ to implement this.

anjohnson:
On trying this out and looking at it in more detail while thinking about how to implement support for it in `dbExpand.pl`, I'm concerned that after a database has been loaded into the IOC there is currently no way to see what the scope setting is for a link that inherited it from a `set("link:scope", ...)` statement.

The IOC's `dbDumpRecord pdbase *` command can no longer be used to generate a .db file from all the records that the IOC loaded, because the inherited link scope setting isn't output by that command.

Actually to make both `dbDumpRecord` and `dbExpand.pl` work properly they might have to explicitly output/add an `INT` flag to every link that has that flag set, since we can load different link fields of the same record from different `record(){}` entries that have different scope settings. Currently `dbGetString()` and `dbParseLink()` only support putting an `EXT` flag in the field.

I also don't particularly like the fact that a scope error is only detectable by running `iocInit` which only reports the first error it sees and then stops — this is not very developer friendly. IMO `iocInit` should report all the links that have errors before stopping, or there could be another command the user can run to find all of the problem links at once. The iocInit error could then suggest the user run that command.

I now think this design needs a bit more thought.

mdavidsaver:
Core Group 2023-03-06: Introduce `INT` link modifier to allow explicit specification of both behaviors.  Tools processing `.db` files should either pass through any `set()` blocks in order, or insert `EXT` or `INT` modifiers into all link fields.

mdavidsaver:
A significant rework...  Three link modifiers are recognized when settings links `INT`, `EXT`, and `AUTO`.  The same three may be passed to `set(link:scope, ...)`, although `AUTO` is (initially?) an alias for `EXT`.  When reading, `INT` or `EXT` is always printed as `AUTO` is always resolved.

> I also don't particularly like the fact that a scope error is only detectable by running `iocInit` which only reports the first error it sees and then stops

This "local" check can only be performed on a complete database (aka. after all possible `dbLoadRecords()`).  This is `iocInit()`.  wrt. multiple errors.  I have adjusted the "initDatabase() error plumbing" change to attempt all possible initialization, then fail with the status of the first error.

> ... they might have to explicitly output/add an `INT` flag to every link that has that flag set ...

This is now done... although I don't like the necessary side effects of printing INT or EXT in every link.  This changes the string representation of every link.  In epics-base this broke some tests.  I am not sure what the possible effects for dependent modules would be.   Is the `dbDumpRecord()` regurgitation use case worth this churn?

edit: It breaks QSRV2 unit tests as well

ralphlange:
Just occurred to me that #222 is a rework of #111. Cool!
Just too bad that we're ~ three weeks past 444 in terms of numbering...

modules/database/src/ioc/dbStatic/link.h
```diff
@@ -68,10 +68,15 @@ DBCORE_API extern maplinkType pamaplinkType[];
 #define pvlOptInpString  0x100  /*Input as string*/
 #define pvlOptOutNative  0x200  /*Output native*/
 #define pvlOptOutString  0x400  /*Output as string*/
+#define pvlOptSrcMask   0x3000
+#define pvlOptSrcAuto   0x0000
+#define pvlOptSrcInt    0x1000
+#define pvlOptSrcExt    0x2000
```
exzombie, line 86: These need to be described.

modules/database/src/ioc/db/dbLink.c
```diff
@@ -86,39 +107,64 @@ static void TSEL_modified(struct link *plink)
...
+        if(!dbFindRecordSimilar(&closest, plink->value.pv_link.pvname, NULL)) {
+            unsigned indent = strlen(precord->name) + strlen(fldname) + 23;
+            epicsPrintf("%*s  Did you mean \"" ANSI_BOLD("%s.%s") "\" ?\n",
+                        indent, "",
+                        closest.precnode->recordname,
+                        closest.pflddes->name);
+        }
+        dbFinishEntry(&closest);
+        return S_dbLib_badLink;
+    }
```
exzombie: When the IOC cannot create an INT link, I get a nice error message. But trying to stop the IOC using the `exit` commands throws an assertion error (quoted below) and hangs.

BTW, I'm curious how the check works. I've never used `dbChannelTest()`, so I'm relying on its doc comment. I would read this check as "if source is INT and if you can find the name over CA". Which is weird, I'd expect the check to be "if source is INT and if you can't find the name in the local database".

```
A call to 'assert(plink->type == PN_LINK)'
    by thread '_main_' failed in ../dev/devAiSoftCallback.c line 137.
Dumping a stack trace of thread '_main_':
[    0x7b0853334712]: libCom.so.3.22.1(epicsBackTrace+0x9)
[    0x7b0853334664]: libCom.so.3.22.1(epicsStackTrace+0x61)
[    0x7b085332e202]: libCom.so.3.22.1(epicsAssert+0x4d)
[    0x7b08534a436f]: libdbRecStd.so.3.22.1(del_record+0x83)
[    0x7b085342e687]: libdbCore.so.3.22.1(doCloseLinks+0xd2)
[    0x7b085342dbd7]: libdbCore.so.3.22.1(iterateRecords+0x44)
[    0x7b085342e905]: libdbCore.so.3.22.1(iocShutdown+0x2e)
[    0x7b085342e9d0]: libdbCore.so.3.22.1(exitDatabase+0x9)
[    0x7b08533267b4]: libCom.so.3.22.1(epicsExitCallAtExitsPvt+0x16)
[    0x7b0853326986]: libCom.so.3.22.1(epicsExitCallAtExits+0x46)
[    0x7b0853326ae8]: libCom.so.3.22.1(epicsExit+0x8)
[          0x407bcc]: softIoc(main+0x7c3)
[    0x7b0852c2924e]: libc.so.6(__libc_start_call_main+0x7e)
[    0x7b0852c29309]: libc.so.6(__libc_start_main+0x89)
[          0x406395]: softIoc(_start+0x25)
EPICS Release EPICS R7.0.7.1-DEV.
Local time is 2024-06-12 14:37:37.707133513 EDT
Please E-mail this message to the author or to tech-talk@aps.anl.gov
Calling epicsThreadSuspendSelf()

```

exzombie:
> This is now done... although I don't like the necessary side effects of printing INT or EXT in every link. This changes the string representation of every link. In epics-base this broke some tests. I am not sure what the possible effects for dependent modules would be. Is the dbDumpRecord() regurgitation use case worth this churn?

I haven't looked at which tests are failing and what you have changed since the last time I looked at this, so I don't know whether you are referring only to dumping the database, or to printing INT or EXT in general. If the latter, I have to say that I'm in favor of always printing it, meaning, when reading links with CA/PVA/dbgf, when using dbpr or capr.pl, and so on. I thought this already happens for NMS and NPP, which are also optional, causing string representation to change; how are INT and EXT different in this regard?

mdavidsaver:
> I haven't looked at which tests are failing ...

I have rebase'd as is.  Let's see what happens.

> ... I have to say that I'm in favor of always printing it ...

Strictly speaking, I would agree this is the "correct" behavior.  My argument against is entirely practical.  It will break cases where link strings are being compared verbatim.  As is encouraged by `testdbGetFieldEqual()`.  (I guess I have to blame my lack of foresight?)

One [example](https://github.com/mdavidsaver/pvxs/blob/57f79ce747f062a32eede1a630d33a5234e0808f/test/testqsingle.cpp#L464) from PVXS unit tests.

```c++
testdbGetFieldEqual("test:ai2.INP", DBF_STRING, "test:ai NPP NMS");
```

Of course, I would change this test if the link string representation changes.  It would just be annoying to have to test for epics-base version.

modules/database/src/ioc/db/dbLink.c
mdavidsaver: wrt. `dbChannelTest()`.  I think of this as a "short-circuit" version of `dbChannelCreate()`.  It does the existence test, but does not allocate.  It is also the test CA and PVA servers use to determine whether to respond to a search.

`dbContext::createChannel()` (aka. local CA) actually does a more involved test w/ `dbChannel_create()`.  So I suppose there is a gap here which could see a non-local CA link created.  Maybe with a PV name with a filter expression which is invalid on the local IOC, but valid on some remote server?  I'll have to think about this.

mdavidsaver: > ... trying to stop the IOC ...

Right, `dbInitLink()` must change the type from `PV_LINK` to something else...

exzombie:
I noticed that the link scope has no effect at runtime. Using `caput` to change an INP field will successfully connect to a PV in a different IOC even with the INT attribute. While it would be nice if INT could be enforced at runtime, it's just that: a nic-to-have. It can always be added later, link scope at init time is a valuable feature by itself. However, until INT can be enforced at runtime, the link text should return EXT so that it's clear that the link is not limited to internal scope.

exzombie:
> Of course, I would change this test if the link string representation changes. It would just be annoying to have to test for epics-base version.

Agreed, but what can you do? The only suggestion I can come up with is a dedicated function for link comparison; at least the test for base version would  then be done in one place.

mdavidsaver:
> Agreed, but what can you do?

Omit printing `" EXT"` for DB and CA links.

mdavidsaver:
I may have found a way out of this unittest vs. dbDumpRecords conundrum, with another redesign...

Now the link modifier may have one of four distinct value: `INT`, `EXT`, `AUTO`, or undefined.  This last will then use the value from the file scope where the link was set (which will be one of the three defined values).

The `EXT` modifier changes behavior.  Now it will _require_ a non-local target.  This prevents creation of a DB link.

When rendering link as a string, `INT` and `EXT` are printed explicitly.  However, `AUTO` is not shown, so the default representations will not change.

In `struct link`, the value from the file scope is stored in `link::flags` (default `pvlOptSrcAuto`).  The value from the link parsing is stored in `link::value::pv_link::pvlMask` (default `pvlOptSrcUnDef`).

mdavidsaver:
> I noticed that the link scope has no effect at runtime.  ...

This was a deliberate choice on my part.  My motivating use case is typos in .db files, which is only relevant at startup.  However, I don't have a strong opinion about what should happen afterwards, and could be persuaded to reconsider.

One of my reservations had been in not knowing what to do with invalid links.  Now I think I know what to do, have json link support which fails all operations and raises INVALID_ALARM.

exzombie:
The meaning of keywords in the new design makes a lot of sense, it's both simpler and clearer. It's one of those "how come we haven't come up with this eariler‽" moments. And thanks for updating the PR description, this discussion has gotten too long to follow :)

As for the runtime behavior, I have the same usecase as you and don't really care much about honoring these at runtime, not yet at least. But I *do* consider it important not to introduce confusion (which I explain below), and to prevent ossification. Right now, what happens is that setting a link at runtime is equivalent to AUTO, right? Which I think is fine, but I believe trying to specify EXT or INT should be rejected. And that's because accepting these attributes and silently doing something undefined will lead to problems down the road when you try to introduce well-defined behavior: someone somewhere will have "working" scripts doing pvputs to link fields using EXT and INT attributes in the false belief that they make a difference, and whatever change you make will break them. Hence, ossification. If, on the other hand, you reject invalid attributes, people won't try to set the links using them, and you are free to enable them later on.

Now, for the confusion I mentioned above. Here's the current behavior. Suppose I have an IOC with records ai1 and ai2. Then, in another IOC, I have a record called junk with the following definition:
```
record(ai, "junk") {
    field(INP, "ai1 PP EXT")
}
```

Doing `pvget junk.INP` yields `"ai1 PP NMS EXT"`, fine. Then I do `pvput junk.INP ai2`. Repeating the `pvget` yields me `"ai2 NPP NMS INT"`. I think that's no good. Setting the internal flag to AUTO instead should do the trick, it guess.

ralphlange:
Thanks @exzombie for your thoughts - I never went through such general considerations, but they seem to make a lot of sense.

I agree that the three-keyword approach looks strikingly obvious.

However, I don't see why this should not be treated the same at runtime.
Assuming that changing a link at runtime means:
- disconnecting the old link
- parsing the new link
- resolving the new link.

I don't understand why resolving the new link at runtime would or should be any different from resolving it at start-up.

Resolving as pseudo-code:
```
if (not EXT): done = resolve_local()
if (not (INT or done)): resolve_remote()
```

Why would that differ between initial resolve and run-time resolve?

mdavidsaver:
Rebased.

exzombie:
[Here's a patch](https://github.com/exzombie/epics-base/commit/26cc3c8dff71416b3288bb4be2f1f75ebba6bc74) that forbids use of these modifiers at runtime (I can't push to this PR myself). It also suppresses printing them, both for pvget/dbgf and for dbDumpRecord. Thus, these link modifiers are restricted solely to db file. Which, for now, is sufficient functionality, I think.

PR #224, opened 2022-03-09
caservertask: Remove new local variable conf shadows...

tboegi:
There is a global variable "conf" which is hidden by a local
variable "conf" in req_server().
Since the latter is a pointer, rename it into "pConf".

Noticed by the lgtm bot.


mdavidsaver:
Did you locate the/a `conf` global?  From a quick look, I don't see it.  imo. `conf` is too generic to be a global, and invites this sort of accidental "collision".  Case in point, the `rsrv_init()` just below also has a `conf` local.  So maybe better to rename the global.

tboegi:
Oh dear, 
the global "conf" was added in my branch for IPv6
Sorry wasting your time @mdavidsaver

PR #225, opened 2022-03-09
repeater: Do not loose a socket in register_new_client()

tboegi:
When a client registers with the repeater, a check is done that
the client is on the same host.
If a host has e.g. an ethernet interface with address 1.2.3.4
both registrations from 1.2.3.4 or 127.0.0.1 are accepted.
All others are ignored.
The check is done by creating a temporary socket and bind() it to
the IP address to be checked, 1.2.3.4 in the example, with port = 0

When this succeeds, the registration can continue.
The current code has possible improvements:
- The temporary socket is not closed (and kind of lost)
- When there are 2 or more ethernet interfaces, only one is handled.
  The one where the first registration is received.

Solution:
Close the socket after bind() and simplify the rest of the code

### DIFF
--- a/modules/ca/src/client/repeater.cpp
+++ b/modules/ca/src/client/repeater.cpp
@@ -367,24 +367,21 @@ static void register_new_client ( osiSockAddr & from,
 
     /*
      * the repeater and its clients must be on the same host
+     * If we receive a registration send to 1.2.3.4
+     * thatis OK, if, and only if, we have 1.2.3.4 on one interface
+     * Try to bind to this interface to check this.
+     * This is IPv4 only. Add a check here in preparation for IPv6
      */
-    if ( INADDR_LOOPBACK != ntohl ( from.ia.sin_addr.s_addr ) ) {
-        static SOCKET testSock = INVALID_SOCKET;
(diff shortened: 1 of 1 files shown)


tboegi:
I may have mis-understood the whole thing.
Probably someone with more insights should make a code review.

mdavidsaver:
The `testSock` variable is static, so only a single socket is kept.  I think this was probably intended behavior.  Maybe as a sort of reservation?

It also seems wasteful to me as new repeater clients are a relatively rare occurrence.

imo. not a bug, but an aspect which could reasonably be changed.

https://github.com/epics-base/epics-base/blob/91941af992f6c32ef4f4ef55a2952f581aeee36a/modules/ca/src/client/repeater.cpp#L372-L388

tboegi:
Now I see (my misunderstanding)
The idea is to keep one socket around, which can be bind() to different
addresses.
My assumption was, that after the first bind(), we can not do another bind().
But the second bind() is possible - the socket will bind() to the new address.
So not a bug - just a misunderstanding.
I'll close this PR as well, thanks for the review

PR #228, opened 2022-03-30
Allow override of relative softIoc dbd path

daykin:
See https://epics.anl.gov/tech-talk/2022/msg00405.php. Default behavior does not change, but if ABSOLUTE_DBD is set to YES, softIoc will always use EPICS_BASE/dbd/softIoc.dbd and EPICS_BASE/db/softIocExit.db, as in base <=3.16.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -37,7 +37,11 @@
 #include "special.h"
 #include "iocInit.h"
 
-
+/* This file is included from dbYacc.y
+ * Duplicate some declarations to avoid warnings from analysis tools which don't know about this.
+ */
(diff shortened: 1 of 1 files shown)


mdavidsaver:
I've no objection to making this controllable.  Although a runtime search path might be preferable.  Using the second argument of `dbLoadDatabase()` to search first relative, then absolute, directory paths.

However, I dislike adding yet another build time option.  Our CI testing matrix is already very large, and yet not large enough.  This change would be adding more untested combinations.

If a build time option is unavoidable for some reason, then maybe `LINKER_USE_RPATH == ORIGIN` could be used as a condition.

daykin:
I agree with your reasoning that a build-time option is not ideal. I will revise to first try the 'relative' path, then fall back to the absolute path as defined in <= 3.16.

PR #230, opened 2022-04-07
fall back to EPICS_BASE/dbd/softIoc.dbd

daykin:
As discussed in #228 with @mdavidsaver, allow softIoc to fall back on EPICS_BASE/softIoc.dbd if relative path resolution fails, instead of adding yet another build flag.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -159,12 +159,13 @@ static void *getLastTemp(void)
     return(ptempListNode->item);
 }
 
-const char *dbOpenFile(DBBASE *pdbbase,const char *filename,FILE **fp)
+const char *dbOpenFile(DBBASE **ppdbbase,const char *filename,FILE **fp)
 {
-    ELLLIST     *ppathList = (ELLLIST *)pdbbase->pathPvt;
+    ELLLIST     *ppathList;
     dbPathNode  *pdbPathNode;
     char        *fullfilename;
-
+    if (*ppdbbase == 0) *ppdbbase = dbAllocBase();
+    ppathList = (ELLLIST *)(*ppdbbase)->pathPvt;
     *fp = 0;
     if (!filename) return 0;
     if (!ppathList || ellCount(ppathList) == 0 ||
@@ -271,7 +272,7 @@ static long dbReadCOM(DBBASE **ppdbbase,const char *filename, FILE *fp,
         FILE *fp1 = 0;
 
         if (pinputFile->filename)
-            pinputFile->path = dbOpenFile(pdbbase, pinputFile->filename, &fp1);
+            pinputFile->path = dbOpenFile(&pdbbase, pinputFile->filename, &fp1);
         if (!pinputFile->filename || !fp1) {
             errPrintf(0, __FILE__, __LINE__,
                 "dbRead opening file %s\n",pinputFile->filename);
@@ -431,7 +432,7 @@ static void dbIncludeNew(char *filename)
 
     pinputFile = dbCalloc(1,sizeof(inputFile));
     pinputFile->filename = macEnvExpand(filename);
-    pinputFile->path = dbOpenFile(pdbbase, pinputFile->filename, &fp);
+    pinputFile->path = dbOpenFile(&pdbbase, pinputFile->filename, &fp);
     if (!fp) {
         epicsPrintf("Can't open include file \"%s\"\n", filename);
         yyerror(NULL);
--- a/modules/database/src/std/softIoc/softMain.cpp
+++ b/modules/database/src/std/softIoc/softMain.cpp
@@ -29,6 +29,7 @@
 #include "iocsh.h"
 #include "osiFileName.h"
 #include "epicsInstallDir.h"
+#include "dbStaticPvt.h"
 
 extern "C" int softIoc_registerRecordDeviceDriver(struct dbBase *pdbbase);
 
@@ -37,6 +38,8 @@ extern "C" int softIoc_registerRecordDeviceDriver(struct dbBase *pdbbase);
 #  define EPICS_BASE "/"
 #  error -DEPICS_BASE required
 #endif
+
+
 
 #define DBD_BASE "dbd" OSI_PATH_SEPARATOR "softIoc.dbd"
 #define EXIT_BASE "db" OSI_PATH_SEPARATOR "softIocExit.db"
@@ -53,14 +56,15 @@ static void exitSubroutine(subRecord *precord) {
     epicsExitLater((precord->a == 0.0) ? EXIT_SUCCESS : EXIT_FAILURE);
 }
 
-void usage(const char *arg0, const std::string& base_dbd) {
+void usage(const char *arg0, const std::string& base_dbd, const std::string& prefix) {
     std::cout<<"Usage: "<<arg0<<
                " [-D softIoc.dbd] [-h] [-S] [-s] [-v] [-a ascf]\n"
                "[-m macro=value,macro2=value2] [-d file.db]\n"
                "[-x prefix] [st.cmd]\n"
                "\n"
-               "    -D <dbd>  If used, must come first. Specify the path to the softIoc.dbdfile."
+               "    -D <dbd>  If used, must come first. Specify the path to the softIoc.dbdfile.\n"
                "        The compile-time install location is saved in the binary as a default.\n"
+               "        If the compile-time default is not found, EPICS_BASE/dbd will also be tried.\n"
                "\n"
                "    -h  Print this mesage and exit.\n"
                "\n"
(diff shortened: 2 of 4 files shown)


anjohnson:
Hi Evan,

The softIoc's `-v` ⇒ `verbose` flag is supposed to display the iocsh equivalent to the commands that it's running internally, so the lines displayed could be copy/pasted into a new st.cmd file to replicate the automation that happens inside the softMain.c file. The iocsh command should always be printed before the command is run, so that any error messages from it are output just as they would if the command came from an st.cmd file. Setting `verbose` shouldn't otherwise change the output though, which this version currently does — you're using it to suppress warning/error messages in the non-verbose case.

Also, I don't like using the return status from `dbLoadDatabase()` to determine whether to try the alternative path. That status could indicate an error for other reasons such as a syntax error in the file, which shouldn't trigger a read of the other path. I would prefer that it try to `fopen()` or `stat()` the first filename, and if that fails try the other file, but there are complications to doing that. The user might have set the `EPICS_DB_INCLUDE_PATH` environment variable and provided the DBD filename they want in a `-D <dbdfile>` option, which is currently a legal way to specify what DBD file to load and would break if a simple `stat()` test were added.

However that does suggest a possible solution: Use the `EPICS_DB_INCLUDE_PATH` environment variable to provide the list of directories to check, leaving the single call to `dbLoadDatabase()` with the bare filename (or whatever the user provided in their `-D` option. If the environment variable is already set just append to it (remembering that on Windows the path separator is different, use the `OSI_PATH_LIST_SEPARATOR` macro to get the right one for this architecture).

I'm also concerned that running `softIoc -h` ⇒ `usage()` is supposed to print the dbd_file path, but I'm not sure that it would output the right thing in your case. Maybe it could show the search path instead?


Finally, I'm not a C++ expert, but doesn't this construct:
```c++
std::cout << "something " + var + "\n";
```
require the compiler to create a temporary string to hold the result of concatenating the 3 parts, whereas this
```c++
std::cout << "something " << var << "\n";
```
would just send them off the the output stream one after the other with no memory allocation needed?

- Andrew

modules/database/src/std/softIoc/softMain.cpp
```diff
@@ -109,8 +109,20 @@ void lazy_dbd(const std::string& dbd_file) {
...
         std::cout<<"dbLoadDatabase(\""<<dbd_file<<"\")\n";
-    errIf(dbLoadDatabase(dbd_file.c_str(), NULL, NULL),
-          std::string("Failed to load DBD file: ")+dbd_file);
+
+    if (dbLoadDatabase(dbd_file.c_str(), NULL, NULL)){
+        if(verbose)
+            std::cout<<"Failed to load DBD file: "+dbd_file+"\n";
+        std::cout<<"Trying to load relative to EPICS_BASE instead.\n";
+        std::string dbd_base(DBD_FILE);
+        if(dbLoadDatabase(dbd_base.c_str(), NULL, NULL)){
```
mdavidsaver: I had been imagining concatenating the paths and using the existing dbd search path, but I see now that this would be making an assumption about the file name.

I'm also wondering if silently falling back on any failed load will cause confusion if eg. someone runs `softIoc -D /path/with/typo.dbd`.  Seems like this should be an error.  So the fallback would only be if `dbd_file==DBD_FILE_REL`.

As for how to fall back reasonably.  @anjohnson is correct that calling `dbLoadDatabase()` twice is a bad idea.  This will have unfortunate consequences if the first failure was due to a syntax error.  (eg. someone messing with `menuScan`)

It occurs to me though that [`dbOpenFile()`](https://github.com/epics-base/epics-base/commit/62092d4399e67eefee4089892cc1cd29395d79fe#diff-0465e0aca1de11876beea5b8eef174b4ed00bf0d89b66d1f3c2f2b0406fbb671R306-R313) now offers a way to try opening a file w/o parsing it.

So the logic could be something like:

```c++
FILE* fp = NULL;
if(!dbOpenFile(pdbbase, dbd_file.c_str(), &fp) && dbd_file==DBD_FILE_REL) {
    dbOpenFile(pdbbase, DBD_FILE, &fp)
}
if(!fp) { ... error, unable to open file ... }
else if(dbLoadDatabaseFP(&pdbbase, fp, NULL, NULL)) {
   ... syntax error or unable to open include'd ...
```

daykin:
- [x]  Change 'verbose' output to the convention of displaying the internal commands
- [x] use `dbOpenFile` to determine what to do instead of checking status of `dbLoadDatabase`. 
- [x] C++ stream output - ~~@anjohnson how do you prefer when stdout is flushed? `cout<<"blah"<<"\n"` or `cout<<"blah"<<endl`?~~ Using `cerr` for errors, since this is before iocInit. using `endl` to flush buffers end lines.
- [x] get rid of multiple calls to dbLoadDatabase - ~~Append EPICS_BASE to EPICS_DB_INCLUDE path if it isn't there, and only plug the bare filename into the first argument, with EPICS_DB_INCLUDE_PATH as the second argument (unless -D is specified, which overrides any of this cleverness)~~ Accomplished with the dbOpenFile logic 
- [x] Document this behavior in `softIoc -h`

Does this cover everything?

mdavidsaver:
> C++ stream output - @anjohnson how do you prefer when stdout is flushed? `cout<<"blah"<<"\n"` or `cout<<"blah"<<endl`?

To work correctly on windows, each "batch" of output needs to end with `endl` or an explicit `cout.flush()`.  Windows line buffering is weird, and output not flushed can be delayed, or lost on exit.

> * Append EPICS_BASE to EPICS_DB_INCLUDE path if it isn't there, and only plug the bare filename into the first argument

Wouldn't this be redundant to use of `dbOpenFile()`?  If needed, it is better to avoid modifying `$EPICS_DB_INCLUDE` in favor of `dbPath()`.

anjohnson:
IIRC `dbOpenFile()` needs `pdbbase` to have been allocated, and it doesn't convert the `EPICS_DB_INCLUDE` variable into the list of paths inside that. I don't think using that may be quite as easy as you think, unfortunately. I suspect setting the environment variable directly would avoid having to do lots of fiddly initialization which `dbLoadDatabase()` does for us (maybe in `dbReadCOM()` too), but I don't really mind which way you go if I'm wrong about that.

mdavidsaver:
> IIRC `dbOpenFile()` needs `pdbbase` to have been allocated, ...

To state the hopefully obvious, `dbOpenFile()` along with `dbReadCOM`() are internal functions.  Code can be freely moved between the two so long as the behavior of the public `dbRead*()` functions does not change.

modules/database/src/std/softIoc/softMain.cpp
```diff
@@ -103,17 +110,33 @@ void errIf(int ret, const std::string& msg)
...
-
+    std::string loaded_file;
+    FILE *fp = NULL;
+    std::string dbd_relative_to_epics_base(DBD_FILE);
+    loaded_file = dbd_file;
+    dbOpenFile(&pdbbase, dbd_file.c_str(), &fp);
+    if((!fp) && dbd_file==prefix+DBD_FILE_REL){
+        loaded_file=dbd_relative_to_epics_base;
+        std::cerr<<"Failed to open DBD file: "<<dbd_file<<std::endl;
+        std::cerr<<"Trying to open it relative to EPICS_BASE instead..."<<std::endl;
```
mdavidsaver: This feels a bit like "Failed to find command 'bash' in '/usr/local/bin', trying '/usr/bin'".  Useful when troubleshooting this PR, but otherwise a distraction.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -159,9 +159,10 @@ static void *getLastTemp(void)
     return(ptempListNode->item);
 }
 
-const char *dbOpenFile(DBBASE *pdbbase,const char *filename,FILE **fp)
+const char *dbOpenFile(DBBASE **ppdbbase,const char *filename,FILE **fp)
 {
-    ELLLIST     *ppathList = (ELLLIST *)pdbbase->pathPvt;
+    if (*ppdbbase == 0) *ppdbbase = dbAllocBase();
+    ELLLIST     *ppathList = (ELLLIST *)(*ppdbbase)->pathPvt;
```
mdavidsaver: In plain C code, some older compilers still that variable definitions be grouped at the top of each block.  So in this case:

```c
    ELLLIST     *ppathList;
    if (*ppdbbase == 0) *ppdbbase = dbAllocBase();
    ppathList = (ELLLIST *)(*ppdbbase)->pathPvt;
```

mdavidsaver (changes requested):

mdavidsaver:
Looking more closely, I've noticed the search `path` manipulations in `dbReadCom()` which happen before `dbOpenFile()`.  As it stands this PR would effectively skip these, and uses whatever the (persistent) search path left over from any previous `dbLoad*/dbRead*()`.  I'm trying to think through the implications in terms of which paths will actually be tried...

PR #231, opened 2022-04-17
make epicsNAN and epicsINF constants really constant

mdavidsaver:
This seems like a simple change.  But I also wonder why there constants weren't actually `const`.  Am I missing something?

### DIFF
--- a/modules/libcom/src/osi/epicsMath.cpp
+++ b/modules/libcom/src/osi/epicsMath.cpp
@@ -33,8 +33,8 @@ static float makeINF ( void )
 #endif
 
 extern "C" {
-float epicsNAN = NAN;
-float epicsINF = INFINITY;
+const float epicsNAN = NAN;
+const float epicsINF = INFINITY;
 }
 
 #ifdef _MSC_VER
--- a/modules/libcom/src/osi/os/Darwin/epicsMath.h
+++ b/modules/libcom/src/osi/os/Darwin/epicsMath.h
@@ -20,8 +20,8 @@
 extern "C" {
 #endif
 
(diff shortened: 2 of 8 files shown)


anjohnson:
I believe it was an MSVC optimizer problem, we had to rely on doing the actual FP calculation at run-time to get the right results as the optimizer was making incorrect assumptions, or giving errors or something. The problems might have been inside epicsCalcTest.cpp rather than in the actual evaluation of the two values though, see [this MR](https://code.launchpad.net/~freddie-akeroyd/epics-base/+git/epics-base/+merge/373425) for some earlier discussion. We need to make sure all the Appveyor tests using older compilers all pass with this change.

mdavidsaver:
> ... see [this MR](https://code.launchpad.net/~freddie-akeroyd/epics-base/+git/epics-base/+merge/373425) for some earlier discussion ...

Ah, I was explicitly searching for mentions of `const`, so I didn't find this.


> ... We need to make sure all the Appveyor tests using older compilers all pass with this change.

The run under mdavidsaver has [passed](https://ci.appveyor.com/project/mdavidsaver/epics-base/builds/43255837), and the run for [this PR](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/43264842) appears likely to if/when that last job starts.

mdavidsaver:
epicsCalcTest passed in all jobs, so I take it that this change at least doesn't make any MSVC optimizer weirdness any weirder.

mdavidsaver:
Merged as 5485adacb9b407d8cbcf6ffbb6e41a6690f9025b

PR #232, opened 2022-04-17
calc: add FMOD

mdavidsaver:
I sometimes find myself doing RF-y calculations with calc records, where I would like to have floating point modulo.  I can mostly get by with a ternary, but have been known to get this wrong on the first try.  So here is my attempt to add `FMOD(x, y)` to the calc/postfix engine.

### DIFF
--- a/modules/libcom/test/epicsCalcTest.cpp
+++ b/modules/libcom/test/epicsCalcTest.cpp
@@ -298,7 +298,7 @@ MAIN(epicsCalcTest)
     const double a=1.0, b=2.0, c=3.0, d=4.0, e=5.0, f=6.0,
                  g=7.0, h=8.0, i=9.0, j=10.0, k=11.0, l=12.0;
 
-    testPlan(630);
+    testPlan(635);
 
     /* LITERAL_OPERAND elements */
     testExpr(0);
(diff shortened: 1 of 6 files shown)


modules/libcom/test/epicsCalcTest.cpp
```diff
@@ -370,6 +370,7 @@ MAIN(epicsCalcTest)
     testExpr(cosh(0.5));
     testExpr(exp(1.));
     testExpr(floor(1.5));
+    testExpr(fmod(1.5, 1.0));
```
anjohnson, line 373: Check for -ve arguments as well? Not sure how C defines that.

anjohnson (changes requested):
Look at the -ve argument issue, add tests.

modules/libcom/test/epicsCalcTest.cpp
mdavidsaver: I've added some tests covering negative values.

fyi. [`man fmod`](https://linux.die.net/man/3/fmod) states that for `fmod(x,y)`, "... the returned value has the same sign as x ...".

mdavidsaver:
Applied as a6afef4850c3357819173acbc37ffd880bddfe61

PR #233, opened 2022-04-17
Embedded DBD file contents in `*_registerRecordDeviceDriver.cpp`

mdavidsaver:
I'm experimenting with including the content of the `.dbd` file in the generated `*_registerRecordDeviceDriver.cpp` in a way which allows this to be used if `*_registerRecordDeviceDriver(pdbbase)` is called without a preceeding `dbReadDatabase("*.dbd")`.

My motivation is having one less file to move around, and confuse, while trying to coordinate testing by someone not so familiar with linux.

As it stands, this is a minimal demonstration of an idea.  It will only work on POSIX targets where `fmemopen()` is available.  It also adds ~500kB in size to every IOC executable.  Having such a large string constant actually crashes MSVC sometimes!  (`fatal error C1060: compiler is out of heap space`)

Compression would be the obvious way to deal with this.  Even simple gzip compacts by an order of magnitude.  Of course, this raises the issue of adding compression library(ies) as dependencies to Base.  Also, rather than `fmemopen()` or similar, a better solution would be adapt the db parser as is already done for the acf parse (cf. `asInitMem()`).

### DIFF
--- a/configure/RULES.Db
+++ b/configure/RULES.Db
@@ -527,15 +527,15 @@ $(foreach file, $(DB_INSTALLS), $(eval $(call DB_INSTALLS_template, $(file))))
 #---------------------------------------------------------------
 # register record,device,driver support
 
-%_registerRecordDeviceDriver.cpp: $(COMMON_DIR)/%.dbd
+%_registerRecordDeviceDriver.cpp: $(COMMON_DIR)/%.dbd $(EPICS_BASE_HOST_BIN)/registerRecordDeviceDriver.pl
 	@$(RM) $@
 	$(REGISTERRECORDDEVICEDRIVER) $(REGRDDFLAGS) -o $@ $< $(basename $@) $(IOCS_APPL_TOP)
 
-%_registerRecordDeviceDriver.cpp: %.dbd
+%_registerRecordDeviceDriver.cpp: %.dbd $(EPICS_BASE_HOST_BIN)/registerRecordDeviceDriver.pl
 	@$(RM) $@
 	$(REGISTERRECORDDEVICEDRIVER) $(REGRDDFLAGS) -o $@ $< $(basename $@) $(IOCS_APPL_TOP)
 
-%_registerRecordDeviceDriver.cpp: ../%.dbd
+%_registerRecordDeviceDriver.cpp: ../%.dbd $(EPICS_BASE_HOST_BIN)/registerRecordDeviceDriver.pl
 	@$(RM) $@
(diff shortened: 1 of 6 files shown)


mdavidsaver:
Some notes on the portability of `fmemopen()` and ~equivalents which can yield a `FILE*` which does not leave debris on a disk after `fclose()`.  (so an OSI API would only have to wrap `fopen()`)

* [`fmemopen()`](https://man7.org/linux/man-pages/man3/fmemopen.3.html) is part of POSIX.1-2008.  Found references for Linux/glibc, RTEMS/newlib, and freebsd.  No reference for OSX.  Windows doesn't have this.
* [`fopencookie()`](https://man7.org/linux/man-pages/man3/fopencookie.3.html) is non-standard.  Found references for Linux/glibc, RTEMS/newlib, and freebsd.  No reference for OSX.  Windows doesn't have this.
* [`funopen()`](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/funopen.3.html) is non-standard.  Found references for OSX, RTEMS/newlib, and freebsd.  Windows doesn't have this.
* [`open()`](https://man7.org/linux/man-pages/man2/open.2.html) with `O_TMPFILE` is Linux specific.
* [`_open()`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-170) with `_O_TEMPORARY` is Windows specific.
* `fdopen()` / `_fdopen()` is C89, and uniformly available.

`fmemopen()`, `fopencookie()`, and `funopen()` are all variations on a `FILE*` in memory (no associated file descriptor).  `fmemopen()` is simpler to use, but requires a contiguous allocation.  `fopencookie()` and `funopen()` take callbacks to hook into `fread()` and friends.

`O_TMPFILE` and `_O_TEMPORARY` allow creation of a file descriptor which is either anonymous (`O_TMPFILE`) or automatically deleted on close (`_O_TEMPORARY`).

It seems to me that `fopencookie()` and `funopen()` are similar enough that one piece of code could be made to use either.  This covers all but vxWorks and Windows.  With Windows, `_O_TEMPORARY`, or it's winapi equivalent, seem to be the only option.  I can't find reference to anything equivalent on vxWorks.

anjohnson:
OSX provides `fmemopen()` since it's required by POSIX.1-2008. It looks like binary mode might be worth avoiding with it as the man-pages may describe slightly different behavior to Linux (I don't have time to read them both in detail). Make sure there's space for a terminating NUL byte in the text-mode buffer.

mdavidsaver:
Abandoned in favor of #248

PR #234, opened 2022-04-19
Add OMSL and DOL to aaoRecord

mdavidsaver:
Add `OMSL` and `DOL` to `aaoRecord` as is done by `aoRecord`.

Simplifies somewhat priming an `aao` or output-like `waveform` with a constant value.

```
record(aao, "$(P=)strcpy_") {
    field(FTVL, "CHAR")
    field(NELM, "256")

    field(PINI, "YES")
    field(DOL , {const:"I am a really really very long string which just doesn't seem to end when it should!"})
    field(OMSL, "closed_loop")
    field(OUT , "$(P=)send:2")
}
```

### DIFF
--- a/modules/database/src/std/rec/aaoRecord.c
+++ b/modules/database/src/std/rec/aaoRecord.c
@@ -44,6 +44,7 @@
 #include "special.h"
 #include "cantProceed.h"
 #include "menuYesNo.h"
+#include "menuOmsl.h"
 
 #define GEN_SIZE_OFFSET
 #include "aaoRecord.h"
@@ -92,6 +93,7 @@ rset aaoRSET={
 epicsExportAddress(rset,aaoRSET);
 
 static void monitor(aaoRecord *);
+static long fetchValue(aaoRecord *, int);
 static long writeValue(aaoRecord *);
 
 static long init_record(struct dbCommon *pcommon, int pass)
@@ -142,7 +144,7 @@ static long init_record(struct dbCommon *pcommon, int pass)
         recGblRecordError(S_dev_missingSup, prec, "aao: init_record");
         return S_dev_missingSup;
     }
-    return 0;
+    return fetchValue(prec, 1);
 }
 
 static long process(struct dbCommon *pcommon)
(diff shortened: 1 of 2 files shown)


mdavidsaver:
Need to investigate interaction with bptr swapping.

mdavidsaver:
> Need to investigate interaction with bptr swapping.

@anjohnson I don't think there is any issue here.  BPTR swap only happens in device support, and at all other times `prec->bptr` must point to valid storage for at least NELM*sizeof(FTVL) .  Otherwise a `dbPut()` by external code could cause an overrun.

If I'm somehow wrong about this, there is little chance of a regression with a default of OMSL="supervisory".

ralphlange:
That sounds correct and matches my experience.

In my first (and so far only) [real-life project](https://github.com/ralphlange/fftw/tree/master/fftwDevSup) involving BPTR swapping, I was interfacing C++ smart pointers (with Michael's help) and eventually decided to plainly copy-out the array in the aao record case. (Shared pointers with custom allocators don't go too well with our "pointer to memory" interface.)
In the aai/waveform case, BPTR swapping works fine using an intermediate object that keeps the shared pointer and manages the swapping. Multiple "region of interest" records can have their BPTRs independently point to different parts of the same array, which is wrapped in a shared pointer.

anjohnson:
Okay, just wanted to make sure we hadn't missed anything given the differences that Dirk included.

I just found the Launchpad bug [#1903924](https://bugs.launchpad.net/epics-base/+bug/1903924) where I asked for this on 2020-11-11. @dirk-zimoch requested in a comment to the bug that new fields to the end of the record, to maintain ABI compatibility. I see that I had also asked for a test to prove that this doesn't change the behavior of existing records which don't set either field.

dirk-zimoch:
When .dbd files are converted to dbd.pot, ABI compatibility is often broken anyway because of reordering. Please at least spend a new release number when doing so. I always try to avoid ABI breaking changes in minor releases, let alone patch level releases. In EPICS 7 context I would change to 7.(n+1).0.0 for such a change, not 7.0.(n+1).0 and definitely not 7.0.6.(n+1)
 (and collect as many ABI breaking changes as possible before releasing).

mdavidsaver:
> ... I always try to avoid ABI breaking changes in minor releases, ...

ABI changes will (or should) be signaled by an SONAME change.  (imo. sites override `SHRLIB_VERSION` at their peril)

mdavidsaver:
Also wrt. "minor releases" I think you mean something different to "EPICS_REVISION".  For historical reasons Base does not, and likely never will, follow semantic version.

mdavidsaver:
Merged as 0552f2bb721a9796d23edf86f1a5a92c54120939

PR #235, opened 2022-04-25
ioc template avoid passing "0"

mdavidsaver:
A small and seemingly simple change, but high visibility...

At present `st.cmd@Cross` contains:

> dbLoadDatabase("../../dbd/_APPNAME_.dbd",0,0)

which iocsh interprets like:

> dbLoadDatabase "../../dbd/_APPNAME_.dbd" "0" "0"

The logic in `dbOpenFile(pdbbase, "my.dbd", fp)` which skips `open()`ing plain `"my.dbd"` if it doesn't contain `/` or `\` results in an extra confusing situation where only `"0/my.dbd"` is tried.  cf.

https://github.com/epics-base/epics-base/blob/e329fa3296bf587a1c2159a48da9d826f1310afc/modules/database/src/ioc/dbStatic/dbLexRoutines.c#L170-L176

(Was there ever a reason for this condition?)

### DIFF
--- a/modules/database/src/template/top/iocBoot/ioc/st.cmd@Cross
+++ b/modules/database/src/template/top/iocBoot/ioc/st.cmd@Cross
@@ -6,7 +6,7 @@
 #< envPaths
 
 ## Register all support components
-dbLoadDatabase("../../dbd/_APPNAME_.dbd",0,0)
+dbLoadDatabase "../../dbd/_APPNAME_.dbd"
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Based on (my memory of) a conversation last week, this change is acceptable as it doesn't effect the RTEMS or vxWorks templates.

mdavidsaver:
Merged as 19daf6136f73fa29a9a5be2460cf7c6ea9e9c37a

Issue #238, opened 2022-04-27
Recently-deprecated epicsThreadExitMain() is used by i-Tech Libera IOC code


anjohnson:
We just rebuilt the IOC for one or our Libera devices, updating from Base 7.0.4 to 7.0.6.1 which broke the IOC (running on Ubuntu 18.04, x86_64).

Their software calls `epicsThreadExitMain()` at the end of some code that I suspect is processing records in their own threads, the C++ symbol `istd::thread::thread_fnc(void*)` appears below but I believe there's tail-call optimization going on, I have the source for the call which appears at the end of a different more application-specific function.

We can change their code, but my guess is that they are calling this to free up thread-specific resources that the IOC is allocating. Given the frequency of the message repeats I suspect their threads are short-lived, so they need some call to release the stuff we allocate in `createImplicit()`, and I don't see any other API we provide which can do that.

Can we undeprecate this, or provide some other way of undoing `createImplicit()`? I do see that as a legitimate need.

```
iocInit
Starting iocInit
############################################################################
## EPICS R7.0.6.1-C2-1
## Rev. 7.0.6.1-C2-1
############################################################################
iocRun: All initialization complete
## Do iocStd post-initialization
< "/C2/ioc-modules/build/IOCSTD/4.1.0-C2-1+LIB-003/bin/linux-x86_64/postInit.iocsh"
# iocStdPostInit
#<aoi name="aoi_site_controls_iocstd_ioc-data-dump">
# Record our boot parameters
stdPrtBootParam "/C2/iocinfo/bootparams/sioc2k2llrf"
# Log the reboot time and reason
rebootLog "/C2/iocinfo/bootlog/sioc2k2llrf",1
epicsThreadExitMain() has been deprecated for lack of usage.  Please report if you see this message.
Thread non-EPICS_140536976410368 (0x7fd0fc002020) can't proceed, suspending.
Dumping a stack trace of thread 'non-EPICS_140536976410368':
[...]

PR #240, opened 2022-05-04
cleanup epicsThreadOSD for any thread, epicsThread or otherwise

mdavidsaver:
Ensure that cleanup happens for all threads, including implicitly created.

### DIFF
--- a/modules/libcom/src/osi/os/WIN32/osdThread.c
+++ b/modules/libcom/src/osi/os/WIN32/osdThread.c
@@ -260,8 +260,7 @@ static void epicsParmCleanupWIN32 ( win32ThreadParam * pParm )
  */
 LIBCOM_API void epicsStdCall epicsThreadExitMain ( void )
 {
-    cantProceed("epicsThreadExitMain() has been deprecated for lack of usage."
-                "  Please report if you see this message.");
+    cantProceed("epicsThreadExitMain() must no longer be used.\n");
 }
 
 /*
--- a/modules/libcom/src/osi/os/WIN32/osdThread.c
+++ b/modules/libcom/src/osi/os/WIN32/osdThread.c
@@ -12,6 +12,16 @@
  * Author: Jeff Hill
  */
 
+/* pull in _WIN32_WINNT* definitions, and hopefully nothing else */
+#include <sdkddkver.h>
+
+/* When not specified, prefer SDK circa Windows 8 when available,
+ * to pick up support for FlsAlloc().
+ */
+#if !defined(_WIN32_WINNT) && defined(_WIN32_WINNT_WIN8)
+#  define _WIN32_WINNT _WIN32_WINNT_WIN8
+#endif
+
 #include <string.h>
 #include <stdlib.h>
 #include <stddef.h>
@@ -32,14 +42,39 @@
 #include "epicsAtomic.h"
 #include "osdThreadPvt.h"
 
+/* When SDK >= 8 is unavailable, or unselected */
+#if !defined(_WIN32_WINNT_WIN8) || _WIN32_WINNT < _WIN32_WINNT_WIN8
+/* Fiber local storage is documented as requiring >= Vista (SDK 7).
+ * However, it has been found that the declarations are missing until (SDK 8).
+ *
+ * To accommodate < SDK 8, fall back to Tls*() which will build and run
+ * correctly for epicsThread s, but will leak TLS allocations from
+ * epicsThreadImplicitCreate().  aka non-EPICS threads.
+ *
+ * Also, WINE circa 5.0.3 provides the FLS storage functions, but doesn't
+ * actually run the dtor function.
+ */
+typedef void (WINAPI *xPFLS_CALLBACK_FUNCTION) (void*);
+static
+DWORD xFlsAlloc(xPFLS_CALLBACK_FUNCTION *dtor) {
+    (void)dtor;
+    return TlsAlloc();
+}
+#define FlsAlloc xFlsAlloc
+#define FlsSetValue TlsSetValue
+#define FlsGetValue TlsGetValue
+#define USE_TLSALLOC_FALLBACK
+#endif
+
 LIBCOM_API void osdThreadHooksRun(epicsThreadId id);
 
 void setThreadName ( DWORD dwThreadID, LPCSTR szThreadName );
+static void WINAPI epicsParmCleanupWIN32 ( void * praw );
 
 typedef struct win32ThreadGlobal {
     CRITICAL_SECTION mutex;
     ELLLIST threadList;
-    DWORD tlsIndexThreadLibraryEPICS;
+    DWORD flsIndexThreadLibraryEPICS;
 } win32ThreadGlobal;
 
 typedef struct epicsThreadOSD {
@@ -207,8 +242,8 @@ static win32ThreadGlobal * fetchWin32ThreadGlobal ( void )
 
     InitializeCriticalSection ( & pWin32ThreadGlobal->mutex );
     ellInit ( & pWin32ThreadGlobal->threadList );
-    pWin32ThreadGlobal->tlsIndexThreadLibraryEPICS = TlsAlloc();
-    if ( pWin32ThreadGlobal->tlsIndexThreadLibraryEPICS == 0xFFFFFFFF ) {
+    pWin32ThreadGlobal->flsIndexThreadLibraryEPICS = FlsAlloc(&epicsParmCleanupWIN32);
+    if ( pWin32ThreadGlobal->flsIndexThreadLibraryEPICS == FLS_OUT_OF_INDEXES ) {
         DeleteCriticalSection ( & pWin32ThreadGlobal->mutex );
         free ( pWin32ThreadGlobal );
         pWin32ThreadGlobal = 0;
@@ -233,10 +268,14 @@ static void epicsParmCleanupDataWIN32 ( win32ThreadParam * pParm )
     }
 }
 
-static void epicsParmCleanupWIN32 ( win32ThreadParam * pParm )
+static void NTAPI epicsParmCleanupWIN32 ( void * praw )
 {
-    win32ThreadGlobal * pGbl = fetchWin32ThreadGlobal ();
-    if ( ! pGbl )  {
+    win32ThreadParam * pParm = praw;
+    win32ThreadGlobal * pGbl;
+    if(!praw)
+        return;
+
+    if ( ! (pGbl = fetchWin32ThreadGlobal ()) )  {
         fprintf ( stderr, "epicsParmCleanupWIN32: unable to find ctx\n" );
         return;
     }
@@ -464,7 +503,7 @@ static unsigned WINAPI epicsWin32ThreadEntry ( LPVOID lpParameter )
     if ( pGbl )  {
         setThreadName ( pParm->id, pParm->pName );
 
-        success = TlsSetValue ( pGbl->tlsIndexThreadLibraryEPICS, pParm );
+        success = FlsSetValue ( pGbl->flsIndexThreadLibraryEPICS, pParm );
         if ( success ) {
             osdThreadHooksRun ( ( epicsThreadId ) pParm );
             /* printf ( "starting thread %d\n", pParm->id ); */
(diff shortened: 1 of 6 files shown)


mdavidsaver:
So far I've checked epicsThreadTest on Linux with valgrind, which shows no leaks (with my usual suppressions).  I also checked RTEMS5/pc686 with gdb to ensure that the destructor function is called.

anjohnson:
This approach looks fine on Posix and RTEMS-posix, thanks.

Can we replicate these semantics for the Windows implementation? I think we need to ensure that we preemptively fix the issue there too, since it does have the `epicsThreadImplicitCreate()` routine which needs `epicsParmCleanupWIN32()` to be called for cleanup. I'd rather that we not provide different APIs on different OSs.

I don't think the VxWorks implementation needs any changes.

I'm not concerned about fixing RTEMS-score, there isn't anything like as `epicsThreadImplicitCreate()` routine here and there are NULL checks missing from the code that would cause faults if routines like `epicsThreadGetNameSelf()` are ever called from a non-epics thread on that platform. It would be nice though to document somewhere the fact that this platform doesn't support calling some of our APIs from native threads.

I would also like to see tests for this, but I'm not sure how we'd tell whether the cleanup has actually happened, and most of the whole test code would probably have to be OS-specific.

mdavidsaver:
> ... I think we need to ensure that we preemptively fix the issue there too ...

No argument there, assuming you don't mean on the PR specifically.  I'll create a separate ticket to track this.

> I would also like to see tests for this, but I'm not sure how we'd tell whether the cleanup has actually happened, and most of the whole test code would probably have to be OS-specific.

As mentioned about, I did manual tests with Linux and RTEMS5 to confirm that the destructor is called.  With Linux/valgrind I also tried commenting out the `free()` which confirmed that the resulting leak is identified.  In keeping with our policy to avoid testing dependencies, I'm willing to accept that if the destructor function is called for one pthread, that it will be called for any pthread.

Of course there are "oddball" cases to consider.  Like RTEMS5 with a non-posix and non-epics thread, or Linux directly calling `clone()` w/o the pthread wrappers.  I think we'll need to draw a line somewhere, I'm just not sure where.

Random thought, maybe a test could use `epicsThreadMap()` to see if an `epicsThread*` has disappeared?

mdavidsaver:
> Random thought, maybe a test could use `epicsThreadMap()` to see if an `epicsThread*` has disappeared?

The implicitly created threads aren't seen by `epicsThreadMap()`...

mdavidsaver:
todo:
- [x] add release notes
- [x]  change message printed from epicsThreadExitMain

mdavidsaver:
I've added use of `FlsAlloc()` to address the issue on windows as well.  I've also extended `epicsThreadTest` to use `epicsThreadMap()` to detect if `osdThread*` is being leaked for EPICS threads.  This is now (almost) the same code path as for a non-EPICS thread, so I think it gives some added coverage.  At least it correctly detected that WINE doesn't fully implement `FlsAlloc()`.

modules/libcom/src/osi/os/WIN32/osdThread.c
```diff
@@ -12,6 +12,12 @@
  * Author: Jeff Hill
  */
 
+#ifndef _WIN32_WINNT
+#   define _WIN32_WINNT 0x0600
+#elif _WIN32_WINNT < 0x0600
+#   error Minimum supported is Windows Vista
+#endif
```
mdavidsaver, line 23: Also worth highlighting this change to explicitly require >= Vista.  I'm not certain what the current (implied) requirement is, so this may not be a practical change.

mdavidsaver:
@FreddieAkeroyd fyi.

mdavidsaver:
```
error C2440: 'function': cannot convert from 'void (__cdecl *)(void *)' to 'PFLS_CALLBACK_FUNCTION'
```

I'm making another attempt to put `WINAPI` in the right place(s).  Which I think is the issue.

FreddieAkeroyd:
I expect it is a missing WINAP as it is only failing on 32bit builds which suggests a missing `__stdcall`, which is what WINAPI on those platforms. On x64 there is no `__stdcall` and WINAPI is effectively `__cdecl`

mdavidsaver:
```
c:\projects\epics-base\modules\libcom\src\osi\os\win32\osdthread.c(30) : fatal error C1083: Cannot open include file: 'fibersapi.h': No such file or directory
```

The next issue.  It looks like vs2010 doesn't include the `fibersapi.h` header.

mdavidsaver:
Since I'm already working around the incomplete `FlsAlloc()` implementation of WINE, I'll try to extend this to vs2010

FreddieAkeroyd:
That is curious - the function is in the windows API and so the header should be part of the windows sdk. According to https://en.wikipedia.org/wiki/Microsoft_Windows_SDK VS2010 ships with SDK 7.0a which is for windows 7, as fibres came in at Vista i think it should be present in that SDK.  Is there any possibility the build is targeting Windows XP?

mdavidsaver:
Maybe only the `fibersapi.h` header is missing?  I notice that with mingw 8, this header is unconditionally included by `windows.h`.

FreddieAkeroyd:
I've just looked at a machine with vs2010 / 7.0a sdk and it indeed seems to missing, however it is present in the 8.1 sdk. If we are targeting vista or higher then you could just assume flsAlloc is present anyway with vs2010 and include the correct definitions from fibersapi.h as part of your #ifdef

mdavidsaver:
I've finally succeeded in getting the fallback to `TlsAlloc()` fallback working w/ vs2010.

> ... If we are targeting vista or higher ...

If occurs to me that, with the `#if ... _MSC_VER>=1700` condition, this PR no longer requires >= Vista.

Should I revert the `WIN32: Drop support for < Vista` commit?

anjohnson:
AI on ANJ: I would like to test this with the original code that triggered issue #238 to ensure that it does fix it, but I can't do that until after I get back. I'm not insisting that we delay merging this, but it would be nice if that's possible.

FreddieAkeroyd:
@mdavidsaver Re _WIN32_WINNT v _MSC_VER: strictly, you should test on _WIN32_WINNT as you are checking an operating system rather than compiler/library feature. It happens that VS2010 ships a windows SDK (7.0a) that is missing the required header, whereas SDK 8 (shipped with VS2012, _MSC_VER 1700) includes this. However you could install the 8.1 SDK and use it with VS2010, and hence _WIN32_WINNT is a better check as it would pick up fibres in this case but not if checking for _MSC_VER. Also VS2012 let you target windows XP for your resulting executable which does not have fibres, checking _WIN32_WINNT would pick this up whereas checking _MSC_VER would probably result in a linker error as the fibre function is missing on XP. 

So if you have found that _WIN32_WINNT is vista, then that still leaves the question of what to do about the the 7.0a SDK that is missing the definition. You can `#ifdef NTDDI_WIN8` to see if you are compiling with windows sdk 8 or higher, this will tell you if you will have an `fibersapi.h` header with VS2010 for example. If you are using the 7.0 sdk then you can eitehr take the hit and not use fibres (as you would do if  _WIN32_WINNT was not vista, or hard code your own prototype for flsAlloc as we know it must be there on vista.

#if (_WIN32_WINNT >= 0x0600) && (!defined(_MSC_VER) || (defined(_MSC_VER) && defined(NTDDI_WIN8)))
#define USE_FIBRES /* safe to include fibresapi.h */
#endif 

or

#if (_WIN32_WINNT >= 0x0600) 
#if  (!defined(_MSC_VER) || (defined(_MSC_VER) && defined(NTDDI_WIN8)))
#include <fibresapi.h>
#else
/* add own prototype for flsAlloc etc */
#endif 
#define USE_FIBRES
#endif

mdavidsaver:
> _WIN32_WINNT v _MSC_VER

Good point.

> /* add own prototype for flsAlloc etc */

So this would be?

```c
typedef void (WINAPI *PFLS_CALLBACK_FUNCTION) (void*);
DWORD FlsAlloc(PFLS_CALLBACK_FUNCTION);
BOOL FlsFree(DWORD);
BOOL FlsSetValue(DWORD, PVOID);
PVOID FlsGetValue(DOWRD);
```

What gives me pause is `PFLS_CALLBACK_FUNCTION`, which isn't defined in any online MS documentation I can find.  The definition above is based on what mingw has with `WINAPI` instead of `NTAPI`.  Were these different for any target circa <=vs2010 ?

FreddieAkeroyd:
`PFLS_CALLBACK_FUNCTION` looks to be defined in `winnt.h` https://docs.microsoft.com/en-us/windows/win32/api/winnt/nc-winnt-pfls_callback_function   Though the 7.0a SDK does not include  `fibersapi.h` it does define  PFLS_CALLBACK_FUNCTION in its `winnt.h` so you should just be able to include this header and not need a typedef (or it may already be there from a  `windows.h` include)  

The `fibersapi.h` from the 8.1 SDK has the following definitions

```
WINBASEAPI DWORD WINAPI FlsAlloc(PFLS_CALLBACK_FUNCTION lpCallback);

WINBASEAPI PVOID WINAPI FlsGetValue(DWORD dwFlsIndex);

WINBASEAPI BOOL WINAPI FlsSetValue( DWORD dwFlsIndex, PVOID lpFlsData);

WINBASEAPI BOOL WINAPI FlsFree(DWORD dwFlsIndex);
```

mdavidsaver:
@FreddieAkeroyd I've switch to testing `_WIN32_WINNT < _WIN32_WINNT_WIN8` (which is I think equivalent to `NTDDI_WIN8`).

I made a couple of attempts to add the prototypes as you suggest, but wasn't successful.  Development w/ appveyor is even more frustrating with an older compiler which doesn't give much context about what is going on.  Unless you think it is essential, I'm going to defer the question of vs2010 to a later PR.  There should be no regression with vs2010, and plugging a potential leak with newer toolchains is a step forward.

FreddieAkeroyd:
Hi, its not strictly a VS2010 questions. The currently logic will work fine for most situations, however we currently build on Window 10 with visual studio 2019 but target windows 7 via _WIN32_WINNT so the the fibre functions will not get activated. I've cloned your branch so i'll just see if i can make a small mod to the logic to cover our use case too.

FreddieAkeroyd:
@mdavidsaver I've ran some tests on https://github.com/FreddieAkeroyd/epics-base/commits/createImplicit-leak

249b25f looks to have passed all tests, that is the commit that would be most useful as it would cover our use case of targeting W7 from W10

I am just trying a commit to cover vista with missing sdk definitions, it works with VS2010/SDK7.0 locally but may throw oddities with mingw/wine. I'll have a brief go at fixing if it does (it might be mingw always has the definitions if you target vista, so I can just limit the additional prototypes to MSVC). Happy if you don't want to wait, 249b25f is the most important and anything extra is just me being curious at the moment...

modules/libcom/src/osi/os/WIN32/osdThread.c
```diff
@@ -12,6 +12,16 @@
  * Author: Jeff Hill
  */
 
+/* pull in _WIN32_WINNT* definitions, and hopefully nothing else */
+#include <sdkddkver.h>
+
```
FreddieAkeroyd, line 17: `sdkddkver.h` will define _WIN32_WINNT if it is not already defined on both Visual studio and mingw, though they differ in their choice of value.  So I believe the code below preferring sdk8 will never get executed

mdavidsaver: @FreddieAkeroyd You are of course correct...

mdavidsaver:
I'm thinking about a partial merge as I don't see a reason to hold up the POSIX fix while we try to get the WIN32 fix right.

wrt. the WIN32 change.  There are clearly subtleties of compatibility that I'm not used to thinking about, and which to be honest don't concern me.  I do almost no windows development, so I don't have skin in this game.  @FreddieAkeroyd since you have both experience and compatibility requirements, do you want to take the lead on this fix?

anjohnson:
Core group 8-Jun Approving the design changes for Posix only (+ RTEMS and VxWorks which don't suffer the leak), awaiting a separate PR from @FreddieAkeroyd for equivalent Windows changes.

FreddieAkeroyd:
@mdavidsaver @anjohnson  sure - so you will remove the windows changes from this PR and I will make a separate PR for WIN32? @mdavidsaver What you have done with  `FlsAlloc` is all great, its just a few oddities with preprocessor defines that needs fixing, not helped by MinGW and Microsoft assuming a different default and the windows 7 sdk missing the fibres header...

mdavidsaver:
> ... so you will remove the windows changes from this PR and I will make a separate PR for WIN32?

@FreddieAkeroyd  I have merged the POSIX changes separately (2f8272d3a17065693b2964e0975427c335b5a691) and am planning to leave the content of this PR as is for your reference.

anjohnson:
Core Group 9/28: @FreddieAkeroyd Any progress on this?

FreddieAkeroyd:
@anjohnson @mdavidsaver I will look at it during codeathon

PR #242, opened 2022-05-09
treewide: add usage help to various iocsh functions

minijackson:
From my understanding of the various functions.

I'm not particularly sure of the correctness for:

- acctst (the iocsh function seems only used for vxworks?)
- dbReportDeviceConfig
- registerAllRecordDeviceDrivers (if it really is equivalent to calling every `<name>_registerRecordDeviceDriver`, the implementation doesn't seem to use that)
- the `registry<thing>Find` functions

### DIFF
--- a/modules/database/src/ioc/db/dbIocRegister.c
+++ b/modules/database/src/ioc/db/dbIocRegister.c
@@ -35,8 +35,14 @@ static const iocshArg * const dbLoadDatabaseArgs[3] =
 {
     &dbLoadDatabaseArg0,&dbLoadDatabaseArg1,&dbLoadDatabaseArg2
 };
-static const iocshFuncDef dbLoadDatabaseFuncDef =
-    {"dbLoadDatabase",3,dbLoadDatabaseArgs};
+static const iocshFuncDef dbLoadDatabaseFuncDef = {
+    "dbLoadDatabase",
+    3,
+    dbLoadDatabaseArgs,
+    "Load the given .dbd file, with 'path' added as a search path, with the given substitutions.\n\n"
+    "Substitutions are usually not needed for .dbd files.\n\n"
+    "Example: dbLoadDatabase dbd/my.dbd\n",
+};
 static void dbLoadDatabaseCallFunc(const iocshArgBuf *args)
 {
     iocshSetError(dbLoadDatabase(args[0].sval,args[1].sval,args[2].sval));
@@ -46,7 +52,14 @@ static void dbLoadDatabaseCallFunc(const iocshArgBuf *args)
 static const iocshArg dbLoadRecordsArg0 = { "file name",iocshArgString};
 static const iocshArg dbLoadRecordsArg1 = { "substitutions",iocshArgString};
 static const iocshArg * const dbLoadRecordsArgs[2] = {&dbLoadRecordsArg0,&dbLoadRecordsArg1};
-static const iocshFuncDef dbLoadRecordsFuncDef = {"dbLoadRecords",2,dbLoadRecordsArgs};
+static const iocshFuncDef dbLoadRecordsFuncDef = {
+    "dbLoadRecords",
+    2,
+    dbLoadRecordsArgs,
+    "Load the given .db file, with the given substitutions.\n\n"
+    "Substitutions should be given in the format 'var1=value1,var2=value2'.\n\n"
+    "Example: dbLoadRecords db/myRecords.db 'user=myself,host=myhost'\n",
+};
 static void dbLoadRecordsCallFunc(const iocshArgBuf *args)
 {
     iocshSetError(dbLoadRecords(args[0].sval,args[1].sval));
--- a/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
@@ -106,7 +106,13 @@ static void dbDumpDriverCallFunc(const iocshArgBuf *args)
 
 /* dbDumpLink */
 static const iocshArg * const dbDumpLinkArgs[] = { &argPdbbase};
-static const iocshFuncDef dbDumpLinkFuncDef = {"dbDumpLink",1,dbDumpLinkArgs};
+static const iocshFuncDef dbDumpLinkFuncDef = {
+    "dbDumpLink",
+    1,
+    dbDumpLinkArgs,
+    "Dump list of registered links\n"
+    "Example: dbDumpLink pdbbase\n",
+};
 static void dbDumpLinkCallFunc(const iocshArgBuf *args)
 {
     dbDumpLink(*iocshPpdbbase);
@@ -147,7 +153,13 @@ static void dbDumpVariableCallFunc(const iocshArgBuf *args)
 static const iocshArg dbDumpBreaktableArg1 = { "tableName",iocshArgString};
 static const iocshArg * const dbDumpBreaktableArgs[] =
     {&argPdbbase,&dbDumpBreaktableArg1};
-static const iocshFuncDef dbDumpBreaktableFuncDef = {"dbDumpBreaktable",2,dbDumpBreaktableArgs};
+static const iocshFuncDef dbDumpBreaktableFuncDef = {
+    "dbDumpBreaktable",
+    2,
+    dbDumpBreaktableArgs,
+    "Dump the given break table\n"
+    "Example: dbDumpBreaktable pdbbase typeKdegC\n",
+};
 static void dbDumpBreaktableCallFunc(const iocshArgBuf *args)
 {
     dbDumpBreaktable(*iocshPpdbbase,args[1].sval);
@@ -157,7 +169,14 @@ static void dbDumpBreaktableCallFunc(const iocshArgBuf *args)
 static const iocshArg dbPvdDumpArg1 = { "verbose",iocshArgInt};
 static const iocshArg * const dbPvdDumpArgs[] = {
     &argPdbbase,&dbPvdDumpArg1};
-static const iocshFuncDef dbPvdDumpFuncDef = {"dbPvdDump",2,dbPvdDumpArgs};
+static const iocshFuncDef dbPvdDumpFuncDef = {
+    "dbPvdDump",
+    2,
+    dbPvdDumpArgs,
+    "Dump the various buckets of the process variable directory.\n"
+    "If verbose is greater than 0, also print the process variables in each bucket.\n"
+    "Example: dbPvdDump pdbbase 1\n",
+};
 static void dbPvdDumpCallFunc(const iocshArgBuf *args)
 {
     dbPvdDump(*iocshPpdbbase,args[1].ival);
@@ -167,7 +186,16 @@ static void dbPvdDumpCallFunc(const iocshArgBuf *args)
 static const iocshArg dbPvdTableSizeArg0 = { "size",iocshArgInt};
 static const iocshArg * const dbPvdTableSizeArgs[1] =
     {&dbPvdTableSizeArg0};
-static const iocshFuncDef dbPvdTableSizeFuncDef = {"dbPvdTableSize",1,dbPvdTableSizeArgs};
+static const iocshFuncDef dbPvdTableSizeFuncDef = {
+    "dbPvdTableSize",
+    1,
+    dbPvdTableSizeArgs,
+    "Change the number of buckets in the process variable directory.\n\n"
+    "The process variable directory size should be set before loading the database.\n"
+    "The size of the process variable directory can automatically grow.\n"
+    "The size must be a power of 2.\n\n"
+    "Example: dbPvdTableSize 1024\n",
+};
 static void dbPvdTableSizeCallFunc(const iocshArgBuf *args)
 {
     dbPvdTableSize(args[0].ival);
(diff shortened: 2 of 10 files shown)


modules/database/src/ioc/db/dbIocRegister.c
```diff
@@ -35,8 +35,12 @@ static const iocshArg * const dbLoadDatabaseArgs[3] =
 {
     &dbLoadDatabaseArg0,&dbLoadDatabaseArg1,&dbLoadDatabaseArg2
 };
-static const iocshFuncDef dbLoadDatabaseFuncDef =
-    {"dbLoadDatabase",3,dbLoadDatabaseArgs};
+static const iocshFuncDef dbLoadDatabaseFuncDef = {
+    "dbLoadDatabase",
+    3,
+    dbLoadDatabaseArgs,
+    "Load the given .dbd file, with 'path' added as a search path, with the given substitutions.\n",
```
exzombie: It would be helpful to show an example of substitutions, like it's done for `dbLoadRecords`.

modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
```diff
@@ -167,15 +186,30 @@ static void dbPvdDumpCallFunc(const iocshArgBuf *args)
...
 static const iocshArg * const dbReportDeviceConfigArgs[] = {&argPdbbase};
-static const iocshFuncDef dbReportDeviceConfigFuncDef = {"dbReportDeviceConfig",1,dbReportDeviceConfigArgs};
+static const iocshFuncDef dbReportDeviceConfigFuncDef = {
+    "dbReportDeviceConfig",
+    1,
+    dbReportDeviceConfigArgs,
+    "Print the link type, link value, device type, record name,\n"
+    "and linearisation info (if applicable),\n"
+    "for every record using a specific device type.\n",
+};
```
exzombie, line 214: Please add an example.

modules/database/src/ioc/dbtemplate/dbtoolsIocRegister.c
```diff
@@ -14,12 +14,20 @@
...
-static const iocshFuncDef dbLoadTemplateFuncDef =
-    {"dbLoadTemplate", 2, dbLoadTemplateArgs};
+static const iocshFuncDef dbLoadTemplateFuncDef = {
+    "dbLoadTemplate",
+    2,
+    dbLoadTemplateArgs,
+    "Load the substitution file given as first argument, apply the substitutions\n"
+    "for each template in the substitution file, and load them using 'dbLoadRecords'.\n\n"
+    "The second argument provides extra variables to substitute in the\n"
+    "template files (not the substitution file).\n\n"
```
exzombie, line 28: Nice, I'm glad this is spelled out :+1:

modules/database/src/ioc/misc/iocshRegisterCommon.c
```diff
@@ -28,8 +28,15 @@
...
-    {"registerAllRecordDeviceDrivers", 1, rrddArgs};
+static const iocshFuncDef rrddFuncDef = {
+    "registerAllRecordDeviceDrivers",
+    1,
+    rrddArgs,
+    "Register all records, devices, from all DBD available.\n\n"
+    "Calling this function is equivalent to calling every\n"
+    "'<name>_registerRecordDeviceDriver'.\n\n"
+    "These are registered into the database given as first argument,\n"
+    "which should be 'pdbbase' in most cases.\n"};
```
exzombie: We now understand that this argument always needs to be `"pdbbase"`, please replace this explanation with an example call.

modules/database/src/ioc/registry/registryIocRegister.c
```diff
@@ -21,28 +21,44 @@ static const iocshArg * const registryXxxFindArgs[1] = {&registryXxxFindArg0};
 
 /* registryRecordTypeFind */
 static const iocshFuncDef registryRecordTypeFindFuncDef = {
-    "registryRecordTypeFind",1,registryXxxFindArgs};
+    "registryRecordTypeFind",
+    1,
+    registryXxxFindArgs,
+    "Prints the registry address of the record type given as first argument.\n",
```
exzombie: Please add an example. I know it's trivial, but it would be nice for consistency with other `registry*Find` commands where it's more important.

modules/database/src/ioc/registry/registryIocRegister.c
```diff
@@ -21,28 +21,44 @@ static const iocshArg * const registryXxxFindArgs[1] = {&registryXxxFindArg0};
...
     printf("%p\n", (void*) registryRecordTypeFind(args[0].sval));
 }
 
 /* registryDeviceSupportFind */
 static const iocshFuncDef registryDeviceSupportFindFuncDef = {
-    "registryDeviceSupportFind",1,registryXxxFindArgs};
+    "registryDeviceSupportFind",
+    1,
+    registryXxxFindArgs,
+    "Prints the registry address of the device support given as first argument.\n",
```
exzombie: Please add an example. It's not clear in what format the argument should be, i.e. prefixed with `dev` or not, capitalization ...

modules/database/src/ioc/registry/registryIocRegister.c
```diff
@@ -21,28 +21,44 @@ static const iocshArg * const registryXxxFindArgs[1] = {&registryXxxFindArg0};
...
     printf("%p\n", (void*) registryDeviceSupportFind(args[0].sval));
 }
 
 /* registryDriverSupportFind */
 static const iocshFuncDef registryDriverSupportFindFuncDef = {
-    "registryDriverSupportFind",1,registryXxxFindArgs};
+    "registryDriverSupportFind",
+    1,
+    registryXxxFindArgs,
+    "Prints the registry address of the driver support given as first argument.\n",
```
exzombie: Please add an example, for the same reason as with `registryDeviceSupportFind`

modules/database/src/ioc/registry/registryIocRegister.c
```diff
@@ -21,28 +21,44 @@ static const iocshArg * const registryXxxFindArgs[1] = {&registryXxxFindArg0};
...
     printf("%p\n", (void*) registryDriverSupportFind(args[0].sval));
 }
 
 /* registryFunctionFind */
 static const iocshFuncDef registryFunctionFindFuncDef = {
-    "registryFunctionFind",1,registryXxxFindArgs};
+    "registryFunctionFind",
+    1,
+    registryXxxFindArgs,
+    "Prints the registry address of the iocsh function given as first argument.\n",
```
exzombie: This is not correct, it's not only for iocsh functions, but all registered functions. Using an iocsh function in an example would be fine, though.

modules/database/src/tools/registerRecordDeviceDriver.pl
```diff
@@ -276,8 +276,13 @@ END
...
 static const iocshArg *rrddArgs[] = {&rrddArg0};
-static const iocshFuncDef rrddFuncDef =
-    {"$subname", 1, rrddArgs};
+static const iocshFuncDef rrddFuncDef = {
+    "$subname",
+    1,
+    rrddArgs,
+    "Register the various records, devices, for this DBD.\\n\\n"
+    "These are registered into the database given as first argument,\\n"
+    "which should be 'pdbbase' in most cases.\\n"};
```
exzombie: Same as before, no other string than `"pdbbase"` is allowed.

modules/database/src/ioc/db/dbIocRegister.c
```diff
@@ -46,7 +50,14 @@ static void dbLoadDatabaseCallFunc(const iocshArgBuf *args)
...
 static const iocshArg dbLoadRecordsArg1 = { "substitutions",iocshArgString};
 static const iocshArg * const dbLoadRecordsArgs[2] = {&dbLoadRecordsArg0,&dbLoadRecordsArg1};
-static const iocshFuncDef dbLoadRecordsFuncDef = {"dbLoadRecords",2,dbLoadRecordsArgs};
+static const iocshFuncDef dbLoadRecordsFuncDef = {
+    "dbLoadRecords",
+    2,
+    dbLoadRecordsArgs,
+    "Load the given .db file, with the given substitutions.\n\n"
+    "Substitutions should be given in the format 'var1=value1,var2=value2'.\n"
+    "For more information about substitutions, see the 'macro substitution' documentation.\n",
```
exzombie: Please refer to the AppDevGuide for macro substitution docs.

modules/database/src/ioc/dbtemplate/dbtoolsIocRegister.c
```diff
@@ -14,12 +14,20 @@
...
+    "See 'help dbLoadRecords' and the 'macro substitution' documentation for more information.\n",
```
exzombie: Please refer to the AppDevGuide for macro substitution docs.

exzombie (commented):
I've only checked the changes you made, I haven't gone through all the commands to see if you missed anything.

modules/database/src/ioc/db/dbIocRegister.c
minijackson: I don't think substitutions are quite used for .dbd files. But I've since commited changes that add Example to all the commands I touched (sorry, I should have warned you).

modules/database/src/ioc/db/dbIocRegister.c
minijackson: To my understanding, the AppDevGuide is to be deprecated in favor of a common ReadTheDocs documentation? I tried to be generic to prevent this documentation to be outdated.

modules/database/src/ioc/db/dbIocRegister.c
anjohnson: They are (almost?) never used when loading DBD files, but as the argument shows they are available if someone finds a use for them. The `dbLoadDatabase` and `dbLoadRecords` commands are actually interchangeable in that both can be used to load either `.dbd` or `.db` files. Even the `path` argument to `dbLoadDatabase` isn't really used any more, although it was needed in older versions (3.13 and earlier).

modules/database/src/ioc/db/dbIocRegister.c
anjohnson: We have been moving away from and stopped updating the AppDevGuide several years ago (3.16.2 was the last version). There is a [Read-the-Docs version](https://docs.epics-controls.org/en/latest/appdevguide/AppDevGuide.html) of it, but the developers haven't been updating that either. I've been trying to find another reference for variable substitutions but can't fine anything obvious, other than the internals of macLib which does the substitution work. I would probably drop the last line since we can't point to anywhere obvious.

modules/database/src/ioc/db/dbIocRegister.c
minijackson: > The dbLoadDatabase and dbLoadRecords commands are actually interchangeable in that both can be used to load either `.dbd` or `.db` files

Do you feel the need to document that in the `usage` section of either commands? The latest version of this PR added the text:

> Substitutions are usually not needed for .dbd files

Do you feel that it is sufficient?

anjohnson: No need to document that. This looks fine to me now, thanks!

modules/database/src/ioc/misc/iocshRegisterCommon.c
```diff
@@ -28,8 +28,17 @@
...
+static const iocshFuncDef rrddFuncDef = {
+    "registerAllRecordDeviceDrivers",
+    1,
+    rrddArgs,
+    "Register all records, devices, from all DBD available.\n\n"
+    "Calling this function is equivalent to calling every\n"
+    "'<name>_registerRecordDeviceDriver'.\n\n"
+    "These are registered into the database given as first argument,\n"
+    "which should always be 'pdbbase'.\n\n"
+    "Example: registerAllRecordDeviceDrivers pdbbase\n",
```
anjohnson: Needs something about dynamic linking only.

modules/database/src/template/top/exampleApp/src/_APPNAME_Hello.c
```diff
@@ -17,7 +17,15 @@ void hello(const char *name) {
...
 static const iocshArg    *helloArgs[] = {&helloArg0};
-static const iocshFuncDef helloFuncDef = {"hello", 1, helloArgs};
+static const iocshFuncDef helloFuncDef = {
+    "hello",
+    1,
+    helloArgs,
+    "Greet the person whose name is given as first argument.\n"
+    "If 'name' is unspecified, greet the world in general.\n\n"
+    "Example: hello\n"
+    "Example: hello stranger\n",
```
anjohnson, line 28: This won't build against earlier Base versions that lack the extended help message, needs wrapping in the `IOCSHFUNCDEF_HAS_USAGE` macro which permits that.

anjohnson (commented):
Approve with these modifications.

modules/database/src/ioc/misc/iocshRegisterCommon.c
minijackson: I'm not sure I understand what you mean. Do you mean we should document the fact that it will only find records/shell functions/etc. for Apps that were dynamically linked into the IOC?

mdavidsaver: The essential point is that generated `*_registerRecordDeviceDriver()` code serves two purposes, one at build time and one at runtime.  `registerAllRecordDeviceDrivers()` is a alternative for the second purpose only.

When linking an executable statically, omitting generated `*_registerRecordDeviceDriver.cpp` from the build can silently result in a non-functional executable as the linker may (and likely will) discard some or all of the code being registered as unused.

I think that @anjohnson would like this to be mentioned in some way.  Maybe something like:

> Calling this function is equivalent to calling every '<name>_registerRecordDeviceDriver()' which has been linked into the process.  eg. by dynamic loading, or generated '<name>_registerRecordDeviceDriver.cpp' file(s).

minijackson:
I think this PR is ready

anjohnson:
Core group 8 June - Approved, thanks!

mdavidsaver:
Merged as a722bb9289c73e41d3010c986ed5f300398cc959

PR #243, opened 2022-05-10
make Perl hash iteration reproducible

minijackson:
From my limited testing, with a deterministic version header, this makes epics-base's linux-x86_64 build 100% reproducible. This also should make downstream modules much more deterministic.

---

By default Perl doesn't iterate over hashes in a reproducible manner,
to avoid DDoS. Since we aren't providing untrusted inputs, it is fine to
disable this behaviour.

Among other things, this makes the .dbd and some .h generation reproducible.

More information here:

- https://reproducible-builds.org/docs/stable-outputs/
- https://perldoc.perl.org/perlrun#PERL_HASH_SEED

### DIFF
--- a/configure/CONFIG_SITE
+++ b/configure/CONFIG_SITE
@@ -178,5 +178,10 @@ LINKER_USE_RPATH = YES
 # All root directories are considered to be the same.
 LINKER_ORIGIN_ROOT = $(INSTALL_LOCATION)
 
+# Make Perl hash iteration reproducible.
+#  Uncomment to enable reproducible builds.
+#  See: https://reproducible-builds.org/docs/stable-outputs/
+#export PERL_HASH_SEED = 0
+
 # Overrides for the settings above may appear in a CONFIG_SITE.local file
 -include $(CONFIG)/CONFIG_SITE.local


anjohnson:
Please move this to the `configure/CONFIG_SITE` file, and comment it out by default. Users who want reproducible builds can see the comment and enable it either by modifying that file or by copying it to a `CONFIG_SITE.local` file instead.

mdavidsaver:
@anjohnson Is there some reason why this aspect of reproducability shouldn't be the default?

minijackson:
I was wondering the same thing. We had a discussion about wanting tests to be executed in a non-deterministic order, but I feel it would be better to explicitly iterate over them randomly instead of relying on Perl's implicit DDoS protection behavior.

mdavidsaver:
> We had a discussion about wanting tests to be executed in a non-deterministic order ...

Perl 5.32.1 has `prove --shuffle`.

anjohnson:
The Perl community added this to permute the order that hashes get iterated, to tickle any code that assumes a specific order (which is explicitly not guaranteed by Perl). The order of running tests was the first thing that came to my mind but we don't use `prove` to run them when using `make tapfiles` anyway. I'd like to keep the permuting behavior by default.

anjohnson (approved):
Great, thanks!

mdavidsaver:
Merged as 6fbf95ab1889463e81db6f511fcbcde7b9ed8480

PR #244, opened 2022-05-10
Added doxygen comments

echandler-anl:
For Codeathon 2022

### DIFF
--- a/modules/libcom/src/misc/epicsString.h
+++ b/modules/libcom/src/misc/epicsString.h
@@ -34,11 +34,11 @@ extern "C" {
  * 
  * Copies characters from string to an output buffer and converts C-style escape sequences to 
  * their binary form.  Since the output string can never be longer than the source, it is legal for 
- * @p inbuf and @p outbuf to point to the same buffer and @p outsize and @p inlen
+ * \p inbuf and \p outbuf to point to the same buffer and \p outsize and \p inlen
  * be equal, thus performing the character translation in-place.
  * 
  * \param outbuf  buffer to copy string to. The resulting string will be zero-terminated as long as 
- *                @p outsize is non-zero. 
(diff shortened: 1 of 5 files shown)


modules/libcom/src/misc/epicsString.h
```diff
@@ -24,20 +30,107 @@
...
                                            const char *inbuf, size_t inlen);
+
+/** \brief Converts non-printable characters into C-style escape sequences
+ *
+ * Copies characters from the string into a output buffer converting non-printable characters into C-style 
+ * escape sequences.   In-place translations are not allowed since the escaped results will usually be larger
+ * than the input string.
+ *
+ * The following escaped character constants will be used in the output:
+ * \a \b \f \n \r \t \v \\ \’ \" \0
```
anjohnson: Hi Elaine,

Please confirm that the escaped characters above do actually appear in the Doxygen output properly when viewed in a browser. You might need to wrap them in some kind of a code block, but I forget how to do that in Doxygen right now, there are examples in other headers though.

Thanks!

minijackson: Some characters will need to be escaped as they will be interpreted as a Doxygen command (for example `\a` is the Doxygen command for [italics](https://www.doxygen.nl/manual/commands.html#cmda)). It should be possible to escape via the backslash via `\\`, or by surrounding the section with `\verbatim`, `\endverbatim` ([doc](https://www.doxygen.nl/manual/commands.html#cmdverbatim)). The last option is better in my opinion since it allows reading the source code as is.

echandler-anl: Doxygen comment has been fixed.

anjohnson:
Merged at 50f9b12e7d12228c449d23c3d004990ec7b27211 after rebasing.

PR #245, opened 2022-05-10
Codeathon 2022 Header annotations project dbChannel.h

oksanagit:


modules/database/src/ioc/db/dbChannel.h
```diff
@@ -9,14 +9,16 @@
 * in file LICENSE that is included with this distribution.
 \*************************************************************************/
 
-/*
+#ifndef INC_dbChannel_H
+#define INC_dbChannel_H
+
+/** \file dbChannel.h
+ * @brief Declarations fir the @ref dbChannel "dbChannel" record type
```
minijackson: From a quick look at the code, it seems that Doxygen `\cmd` instead of `@cmd` are preferred. It might need official confirmation, though.

anjohnson: Some editors may recognize one or the other, some files come from other projects and already had annotations, so we have both `@file` and `\file` in the tree and don't really care which get used since Doxygen understands both.

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -62,7 +64,8 @@ typedef struct dbChannel {
     ELLLIST post_chain;       /* list of filters to be called post-event-queue */
 } dbChannel;
 
-/* Prototype for the channel event function that is called in filter stacks
+/**
+ * Prototype for the channel event function that is called in filter stacks
```
minijackson, line 68: For the brief description, I think with the way the Doxyfile is configured, you have to prepend it with the `\brief` command. Else the brief description will not show up in indexes and such.

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -74,29 +77,30 @@ typedef struct dbChannel {
  */
 typedef db_field_log* (chPostEventFunc)(void *pvt, dbChannel *chan, db_field_log *pLog);
 
-/* Return values from chFilterIf->parse_* routines: */
+/** Return values from chFilterIf->parse_*- routines: */
```
minijackson, line 80: For documentation which only has a brief description, you can you the `//!` comment type to avoid having to input `\brief`

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -74,29 +77,30 @@ typedef struct dbChannel {
...
-/* These routines must be implemented by each filter plug-in */
+/** These routines must be implemented by each filter plug-in */
 typedef struct chFilterIf {
-    /* cleanup pointer passed to dbRegisterFilter().
-     * Called during DB shutdown
-     */
+/** 
+ * cleanup pointer passed to ::dbRegisterFilter ().
+ * Called during DB shutdown
+ */
```
minijackson, line 90: I think it would be nice to have a description of each of the parameters using the `\param` command

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -42,8 +42,8 @@ typedef struct evSubscrip {
     void                    *user_arg;
     struct event_que        *ev_que;
     db_field_log            **pLastLog;
-    unsigned long           npend;  /* n times this event is on the queue */
-    unsigned long           nreplace;  /* n times replacing event on the queue */
+    unsigned long           npend;  /** \brief n times this event is on the queue */
```
minijackson, line 45: This will actually comment the item after the comment. To comment `npend` you could either put the doc comment above, or use `/**<` or `//!<` doc comments

modules/libcom/src/iocsh/initHooks.h
```diff
@@ -36,34 +40,63 @@ typedef enum {
...
+ *      ... \n
+ *      break; \n
+ *   case initHookAfterInterruptAccept: \n
+ *      ... \n
+ *      break; \n
+ *    default: \n
+ *      break; \n
+ *  } \n
+ * } \n
+ * An arbitrary number of functions can be registered.
```
minijackson: If you want to input code in Doxygen comments, you put the code between `\code{.cpp}` and `\endcode` (no `\n` needed). Awesome example, in any case.

modules/libcom/src/iocsh/initHooks.h
```diff
@@ -36,33 +45,57 @@ typedef enum {
...
+ *      break; 
+ *   case initHookAfterInterruptAccept: 
+ *      ... 
+ *      break; 
+ *    default: 
+ *      break; 
+ *  } 
+ * } 
+ * An arbitrary number of functions can be registered. 
+ * \endcode
```
minijackson: If done this way, the english paragraph would end up in the code snippet in the generated documentation. You can try this instead:

```suggestion
/**
 * \brief Any functions that are registered before iocInit reaches the desired state will be called when it reaches that state. 
 * The initHookName function returns a static string representation of the state passed into it which is intended for printing. The following skeleton code shows how to use this facility:
 *
 * \code{.cpp}
 * static initHookFunction myHookFunction; 
 *
 * int myHookInit(void) 
 * { 
 *   return(initHookRegister(myHookFunction)); 
 * } 
 * 
 * static void myHookFunction(initHookState state) 
 * { 
 *  switch(state) { 
 *    case initHookAfterInitRecSup: 
 *      ... 
 *      break; 
 *   case initHookAfterInterruptAccept: 
 *      ... 
 *      break; 
 *    default: 
 *      break; 
 *  } 
 * } 
 * \endcode 
 *
 * An arbitrary number of functions can be registered. 
```

oksanagit:
Well, this meant as an example.  And the text indicates so. Is there any drawback to leave this as is?

modules/libcom/src/iocsh/initHooks.h
oksanagit: Also, the suggested format will not show new lines. This is why I added \n in my first version to make the example readable an d useful .

minijackson: I'm not sure I see the problem you are having. Could you share a screenshot?

I tried it on my computer and here is the result before:

before

and after:

after

anjohnson:
Rebased to 759a7379836630cc1ce460f4d4e070e0028cd4f1 and merged, then added more detail.

oksanagit:
Hi @anjohnson , I see you merged initHooks, thanks. There is another changed earlier in this PR for modules/database/src/ioc/db/dbChannel.h, I know I should of done it separately. Would you like me to create another PR for this file to be reviewed?

anjohnson:
Hi Oksana, if you look at the `dbChannel.h` file in the 7.0.7 release you’ll see that I took your changes to it from this PR and added some more of my own. I originally created that file myself before we were working on the Doxygen annotations so I have a deep understanding of the API, which wasn’t documented in the AppDevGuide. Thanks for getting the work on it started.

PR #246, opened 2022-05-10
Define Clang version variables for use in Makefiles

anjohnson:
Disable specific warning triggered frequently by Perl's XS API.

Question: I included variables for `CLANG_13` and `CLANG_14` but I only actually need `CLANG_12`, should I leave the others out until they are actually needed? I already decided to not define the earlier version var's at all.

### DIFF
--- a/configure/toolchain.c
+++ b/configure/toolchain.c
@@ -23,6 +23,34 @@ CLANG_MAJOR = __clang_major__
 CLANG_MINOR = __clang_minor__
 CLANG_PATCH = __clang_patchlevel__
 
+#ifdef _COMMENT_
+/* CLANG version detection, similar scheme as used for BASE_3_15 etc.
+ * Needed to disable warnings added in later Clang versions.
+ * In a Makefile
+ *   ifdef CLANG_12
+ *     means the compiler is Clang 12.x or later,
+ *   ifeq ($(CLANG_12),1)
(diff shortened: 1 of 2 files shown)


mdavidsaver:
It turns out clang provides a way to directly test whether a warning is supported.  Amazingly this is [`__has_warning`](https://releases.llvm.org/4.0.1/tools/clang/LanguageExtensions.html#has-warning).  This seems preferable to our trying to map from compiler version.  Maybe toolchain.c should provide a new `XSUBPP_FLAGS`?

```c
#ifdef __has_warning
#  if __has_warning("-Wcompound-token-split-by-macro")
XSUBPP_FLAGS += -Wno-compound-token-split-by-macro
#  endif
#endif
```

anjohnson:
AI on ANJ: Replace my solution with the above.

mdavidsaver:
clang can disable warnings directly from C code, which avoids the need for a build system change at all.  eg.  (untested)

```c
#ifdef __has_warning
#  if __has_warning("-Wcompound-token-split-by-macro")
#    pragma clang diagnostic ignored "-Wcompound-token-split-by-macro"
#  endif
#endif
```

https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

anjohnson:
Adding the code from @mdavidsaver directly to Cap5.xs stops these warnings.

PR #248, opened 2022-05-10
In-memory filesystem

mdavidsaver:
Adds the ability for compiling some static file content into an executable.  Patterned on the [Qt resource system](https://doc.qt.io/qt-5/resources.html).  The main entry point is `epicsFopen()` which forwards to `fopen()` unless the provided path is prefixed by `app://`.  Also `epicsMemMount()`.

Current status:

- Reuses the `epicsMakeMemFs.pl` script, with added arguments.
- Supports compression w/ zlib on supported targets.  (automatically probed in `configure/toolchain.c`
- On POSIX targets, uses `fopencookie()` or `funopen()` to stay entirely in-memory.
- On Windows, writes a temporary file using the `O_TEMPORARY` flag for automatic cleanup.
- On other targets (vxWorks): not supported

Open Questions:

- Require absolute (unix) style path.  eg. `app:///dir/file`.
- API names
- Overlap with RTEMS specific `epicsMemFs.h`
- How to express a more general mapping for `epicsMakeMemFs.pl` between disk and IMF names?

TODO items:

- [X] vxWorks, `ioslib.h` isn't adaquate
- [ ] Options for zlib headers/library (eg. LIB vs. SYS_LIB)
- [X] Some IOCsh functions to manipulate files.  (`cp`, `cat`, etc.)

### DIFF
--- a/modules/libcom/RTEMS/epicsMakeMemFs.pl
+++ b/modules/libcom/RTEMS/epicsMakeMemFs.pl
@@ -2,18 +2,85 @@
 #
 
 use File::Basename;
+use File::Spec;
 use Text::Wrap;
 
 use strict;
+use warnings;
 
-my $outfile = shift;
-my $varname = shift;
+use Getopt::Std;
+$Getopt::Std::STANDARD_HELP_VERSION = 1;
 
+use Pod::Usage;
+
+=head1 NAME
+
+epicsMakeMemFs.pl - Generate C code file containing in-memory filesystem
+
+=head1 SYNOPSIS
+
+B<epicsMakeMemFs.pl> [B<-h>] [B<-z>] [B<-T> dir] [B<-P> dir] [B<-f> I<memfs> | I<imf>] output.c varname file1 file2 ...
+
+=head1 OPTIONS
+
+B<epicsMakeMemFs.pl> understands the following options:
+
+=over 4
+
+=item B<-h>
+
+Help, display this document as text.
+
+=item B<-z>
+
+Enable use of zlib compression.
+
+=item B<-T> dir
+
+Store paths relative to this directory.  (eg. "$(TOP)")
+
+=item B<-f> I<memfs> | I<imf>
+
+Selects output format.  I<memfs> is the default.
+
+=item B<-P> dir
+
+With I<imf> format, prepend directory to all files.
+
+=back
+
+=cut
+
+our ($opt_z, $opt_T, $opt_P, $opt_f, $opt_h);
+
+$opt_T = File::Spec->curdir();
+$opt_f = "memfs";
+
+sub HELP_MESSAGE {
+    pod2usage(-exitval => 2, -verbose => $opt_h ? 2 : 0);
+}
+
+HELP_MESSAGE() if !getopts('hzT:P:f:') || $opt_h || @ARGV <= 2;
+
+if($opt_z && !eval "require Compress::Zlib; 1") {
+    $opt_z = 0;
+}
+
+my $outfile = shift @ARGV;
+my $varname = shift @ARGV;
+
+# TODO: output to .tmp, then rename
 open(my $DST, '>', $outfile)
   or die "Failed to open $outfile";
 
-my $inputs = join "\n *    ", @ARGV;
-print $DST <<EOF;
+$Text::Wrap::break = ',';
+$Text::Wrap::columns = 78;
+$Text::Wrap::separator = ",\n";
+
+if($opt_f eq "memfs") { # ================ Format 1
+
+    my $inputs = join "\n *    ", @ARGV;
+    print $DST <<EOF;
 /* $outfile containing
  *    $inputs
  */
@@ -22,41 +89,37 @@ print $DST <<EOF;
 
 EOF
 
-my $N = 0;
+    my $N = 0;
 
-$Text::Wrap::break = ',';
-$Text::Wrap::columns = 78;
-$Text::Wrap::separator = ",\n";
+    for my $fname (@ARGV) {
+        my $realfname = $fname;
 
-for my $fname (@ARGV) {
-  my $realfname = $fname;
+        # strip leading "../" "./" or "/"
+        $fname =~ s(^\.{0,2}/)()g;
 
-  # strip leading "../" "./" or "/"
-  $fname =~ s(^\.{0,2}/)()g;
+        my $file = basename($fname);
+        my @dirs  = split('/', dirname($fname));
 
-  my $file = basename($fname);
-  my @dirs  = split('/', dirname($fname));
+        print $DST "/* $realfname */\n",
+            "static const char * const file_${N}_dir[] = {",
+            map("\"$_\", ", @dirs), "NULL};\n",
+            "static const char file_${N}_data[] = {\n",
+            "  ";
 
-  print $DST "/* $realfname */\n",
-    "static const char * const file_${N}_dir[] = {",
-    map("\"$_\", ", @dirs), "NULL};\n",
-    "static const char file_${N}_data[] = {\n",
-    "  ";
+        open(my $SRC, '<', $realfname)
+            or die "Failed to open $realfname";
+        binmode $SRC;
 
-  open(my $SRC, '<', $realfname)
-    or die "Failed to open $realfname";
-  binmode $SRC;
+        my ($buf, @bufs);
+        while (read($SRC, $buf, 4096)) {
+            @bufs[-1] .= ',' if @bufs;  # Need ',' between buffers
+            push @bufs, join(",", map(ord, split(//, $buf)));
+        }
+        print $DST wrap('', '  ', @bufs);
 
-  my ($buf, @bufs);
-  while (read($SRC, $buf, 4096)) {
-    @bufs[-1] .= ',' if @bufs;  # Need ',' between buffers
-    push @bufs, join(",", map(ord, split(//, $buf)));
-  }
-  print $DST wrap('', '  ', @bufs);
+        close $SRC;
 
-  close $SRC;
-
-  print $DST <<EOF;
+        print $DST <<EOF;
 
 };
 static const epicsMemFile file_${N} = {
@@ -67,19 +130,124 @@ static const epicsMemFile file_${N} = {
 };
 
 EOF
-  $N++;
-}
+        $N++;
+    }
 
-my $files = join ', ', map "&file_${_}", (0 .. $N-1);
+    my $files = join ', ', map "&file_${_}", (0 .. $N-1);
 
-print $DST <<EOF;
+    print $DST <<EOF;
 static const epicsMemFile* files[] = {
-  $files, NULL
+$files, NULL
 };
 
 static
 const epicsMemFS ${varname}_image = {&files[0]};
 const epicsMemFS * $varname = &${varname}_image;
 EOF
+
+
+} elsif($opt_f eq "imf") { # ================ Format 2
+
+    my $inputs = join "\n *    ", @ARGV;
+    print $DST <<EOF;
+/* $outfile containing
+ *    $inputs
+ */
+
+#include <epicsStdio.h>
+EOF
+    my $N = 0;
+    my (@fencode, @alens, @clens);
+
+    for my $fname (@ARGV) {
+        my $realfname = $fname;
+
+        $fname = File::Spec->canonpath($fname);
+        $fname = File::Spec->rel2abs($fname);
+        $fname = File::Spec->abs2rel($fname, $opt_T);
+
+        # attempt Windows to unix path convert
+        $fname =~ s(\\)(/)g;
+
+        # strip leading "../" "./" or "/"
+        $fname =~ s[^(\.{0,2}/)+][];
+
+        # strip leading O.Common/
+        $fname =~ s(^O\.Common/)();
+
+        if($opt_P) {
+            $fname = "${opt_P}${fname}";
+        }
+
+        open(my $SRC, '<', $realfname)
+            or die "Failed to open $realfname";
+        binmode $SRC;
+
+        read $SRC, my $actual, -s $SRC;
+
+        close $SRC;
+
+        my $actualLen = length($actual);
+
+        my $actualz;
+        if($opt_z) {
+            $actualz = Compress::Zlib::compress($actual, 9);
+        }
+
+        our ($encode, $content);
+
+        if(defined($actualz) && (length($actualz) < $actualLen)) {
+            $encode = "epicsIMFZ";
+            $content = $actualz;
+
+        } else {
+            $encode = "epicsIMFRaw";
+            $content = $actual;
+        }
+
+        my $contentLen = length($content);
+
+        $content = wrap('', '        ', join(",", map(ord, split(//, $content))));
+
+        push @fencode, $encode;
+        push @alens, $actualLen;
+        push @clens, $contentLen;
+
+        print $DST <<EOF;
+static const char file_${N}_name[] = "/$fname";
+static const char file_${N}_content[] = {
+        $content
+};
+EOF
+        $N++;
+    }
+    print $DST <<EOF;
+const epicsIMF ${varname}_imf[] = {
+EOF
+
+    $N = 0;
+    for my $fname (@ARGV) {
+
+        print $DST <<EOF;
+    {
+        file_${N}_name,
+        file_${N}_content,
+        $clens[$N],
+        $alens[$N],
+        $fencode[$N],
+        1,
+    },
+EOF
+        $N++;
+    }
+
+    print $DST <<EOF;
+    {NULL, NULL, 0, 0, epicsIMFRaw, 0}
+};
+EOF
+
+} else { # ================ Format Unsupported
+  die "Unsupported output format $opt_f";
+}
 
 close $DST;
--- a/modules/libcom/src/osi/epicsMemFile.c
+++ b/modules/libcom/src/osi/epicsMemFile.c
@@ -1,0 +1,382 @@
+/*************************************************************************\
+* Copyright (c) 2022 Michael Davidsaver
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+#include <stdio.h>
+#include <errno.h>
+#include <string.h>
+
+#define EPICS_PRIVATE_API
+
+#include <dbDefs.h>
+#include <epicsString.h>
+#include <epicsMutex.h>
+#include <epicsThread.h>
+#include <cantProceed.h>
+#include <errlog.h>
+#include <gpHash.h>
+#include <iocsh.h>
+
+#include "epicsStdio.h"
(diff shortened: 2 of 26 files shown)


modules/libcom/src/osi/epicsStdio.h
```diff
@@ -183,6 +184,140 @@ LIBCOM_API int epicsStdCall epicsStdoutVPrintf(
...
+ * \param pathname Path of file to open.
+ * \param mode Open Mode.  For a memory file only "r" or "rb" are accepted.
+ * \returns NULL on error and sets `errno`
+ *
+ * If `pathname` is prefixed with `emem://`, then an entry
+ * previously passed to epicsFMount() may be returned.
+ * Otherwise, the call is forwarded to the OS fopen().
+ *
+ * \note `emem://` files can only be opened for read-only access,
+ *       and may not have an associated file descriptor.
```
exzombie, line 208: `s/may/will/`

modules/libcom/src/osi/epicsStdio.h
```diff
@@ -183,6 +184,140 @@ LIBCOM_API int epicsStdCall epicsStdoutVPrintf(
...
+ */
+#define EPICS_MEM_FILE_NOCOPY (0x100)
+
+/** \brief Information about a memory file
+ */
+typedef struct epicsIMF {
+    /** bitwise "or" of one or more of `EPICS_MEM_FILE_*` */
+    epicsUInt64 flags;
+    /** Absolute path with which this file should be associated
+     *  under the `erf://` in-memory file system.
```
exzombie: Reminder: there are several occurrences of `erf://` in lieu of `emem://`.

modules/libcom/src/osi/epicsStdio.h
```diff
@@ -183,6 +184,140 @@ LIBCOM_API int epicsStdCall epicsStdoutVPrintf(
...
+ * \param files Base of an array of `epicsIMF`.
+ *              The last element must have a NULL path.
+ * \param flags bitwise "or" of zero or more of `EPICS_MEM_MOUNT_*`.
+ * \return zero if all files were successfully mounted.
+ *         non-zero if any file could not be mounted.
+ *         (pass `flags = EPICS_MEM_MOUNT_VERBOSE` for details)
+ *
+ * Attach some files to the in-memory `erf://` file system.
+ *
+ * The `epicsIMF` entries will always be copied, and need
```
exzombie: `s/always/by default/`

modules/libcom/src/osi/epicsStdio.h
```diff
@@ -183,6 +184,140 @@ LIBCOM_API int epicsStdCall epicsStdoutVPrintf(
...
+ *     printf("%s\n", cur->path);
+ * }
+ * \endcode
+ *
+ * This function marked private until an external user is identified.
+ *
+ * \since UNRELEASED
+ */
+LIBCOM_API
+const epicsIMF* epicsMemNextFile(size_t index);
```
exzombie: This isn't really iteration, is it? I'd rename this function to `epicsMemGetFile()` and rewrite the docs: the indices are assigned by `epicsMemMount()`, they are consecutive, and the function returns `NULL` for an out-of-range index.

modules/libcom/src/osi/epicsMemFile.c
```diff
@@ -0,0 +1,260 @@
...
+
+    epicsThreadOnce(&eimfsOnce, &epicsMemInit, NULL);
+
+    for(n=0u; n<nfiles; n++) {
+        size_t pathlen = strlen(files[n].path);
+        unsigned algo = files[n].flags & EPICS_MEM_FILE_ALGO_MASK;
+
+        if(files[n].path[0]!='/' || files[n].path[pathlen-1u]=='/') {
+            if(verb)
+                fprintf(stderr, ERL_WARNING " math must begin, but not end, with '/' \"%s\"\n",
```
exzombie: I wish my teacher had told me that ;)

On a more serious note, is there are reason why these errors are not printed by default? Are you envisioning a situation where one would not be interested in *why* their stuff doesn't work?

For this error specifically, I imagine this facility would noramally be used via Makefile magic and the `epicsMakeMemFs.pl` script which should check for this (I haven't looked at it yet), so it would be very strange to have this error occur, and I don't think we should be silent about it.

modules/libcom/src/osi/epicsStdio.h
```diff
@@ -183,6 +184,140 @@ LIBCOM_API int epicsStdCall epicsStdoutVPrintf(
...
+FILE* epicsMemOpen(const char *pathname, const char *mode);
+
+/** bit mask of possible compression types */
+#define EPICS_MEM_FILE_ALGO_MASK (0xff)
+
+/** Content array is raw bytes (uncompressed) */
+#define EPICS_MEM_FILE_RAW (0x1)
+
+/** Content array is zlib compressed */
+#define EPICS_MEM_FILE_ZLIB (0x2)
```
exzombie: I don't think this is a good idea, unless you plan to make algorithms composable or something. Algo choice should be an enum.

modules/libcom/src/osi/epicsMemFile.c
```diff
@@ -0,0 +1,260 @@
...
+
+                pvt[n].imf.path = NULL;
+                if(!ret)
+                    ret = ENOMEM;
+                continue; /* skip */
+            }
+            memcpy(buf, files[n].content, files[n].contentLen);
+            pvt[n].imf.content = buf;
+        }
+    }
```
exzombie: I would avoid interleaving sanity checks and data copying. I would split copying into a separate function that can do either a deep or shallow copy. Moreover, I don't see `pvt[i].imf.path` being copied, and it should be; perhaps I missed it, and if so, it demonstrates why I don't find this particularly readable. But I'm pretty sure the path is not copied as you also don't free it later on. I think we need both an `epicsIMFCopy` and `epicsIMFFree`.

modules/libcom/src/osi/os/posix/osdMemFile.c
```diff
@@ -0,0 +1,125 @@
...
+#endif
+
+typedef struct {
+    const epicsIMF* mf;
+    size_t pos;
+    const char* actual;
+} mfCookieActual;
+
+static
+int mfCloseActual(void *raw)
```
exzombie: I don't get this naming. Why not just `mfCookie`, `mfClose` etc?

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: @anjohnson says "app://".

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: "may" is intentional.  I can't hide the fact that the WIN32 implementation will have a real FD.  So this note is to make clear that this isn't universal.

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: Would this be an improvement?

```c
// return 1 if *ppmf updated
//        0 on normal end of iteration
//       -1 on error (errno set)
int epicsMemIter(const epicsIMF** ppmf);

void func() {
    int ret;
    const epicsIMF* file = NULL;
    while((ret=epicsMemIter(&file))>0) {
        assert(file);
        ...
    }
    if(ret==-1) {
        // print error using `errno`
    }
}
```

mdavidsaver: Also, note that I'm wrapping this function in `#ifdef EPICS_PRIVATE_API` for the time being.

modules/libcom/src/osi/epicsMemFile.c
mdavidsaver: > On a more serious note, is there are reason why these errors are not printed by default?

As I think about it, I don't have a good reason beyond not wanting to clutter up the console with warnings.  Although, as you point out this is just as likely to result in later surprise.

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: Choice of algorithm is an enumuration using the first 8 bits.  Thus the `*ALGO_MASK` macro, and composition with the `*NOCOPY` bit (`0x100`).  So I guess the question is really whether the `flags` member should be overloaded in this way vs. individual member fields.

I'm slightly inclined towards the combined `flags` as it seems somewhat easier to accommodate future additions.

Although, it might be a good idea to include a "canary" field at the end to ensure that callers zero initialize the `epicsIMF` struct.

modules/libcom/src/osi/epicsMemFile.c
mdavidsaver: > ... I don't see pvt[i].imf.path being copied ...

Good catch.  In an earlier iteration I had a `char[MAX_PATH]`, but now the `char*` needs to be to be covered by the `NOCOPY` logic.

modules/libcom/src/osi/os/posix/osdMemFile.c
mdavidsaver: "Actual" refers to `mfCookieActual::actual` pointing to uncompressed data.  At one point I was thinking about having a different set of callback functions to handle uncompressed vs. compressed.  With the compressed case doing incremental de-compression.  For example, to avoid needing to allocate a 512kB temporary buffer for `softIoc.dbd`.  So far, I've talked myself out of what seems like a premature optimization.

mdavidsaver:
@anjohnson The references I find for [`ioslib.h`](https://docs.windriver.com/bundle/vxworks_7_application_core_os_sr0630-enus/page/CORE/iosLib.html#iosDrvInstall) don't give the prototype(s) for the callback functions.

modules/libcom/src/osi/epicsStdio.h
exzombie: A definite improvement :+1:

modules/libcom/src/osi/os/posix/osdMemFile.c
exzombie: Regarding incremental decompression, I agree.

As for the naming, there's two separate issues. One is that the *Actual* part of `mfCloseActual` and `mfCookieActual` doesn't add anything of value; a cookie is a cookie, there is nothing "actual" about it. The other issue is that "actual" in `epicsIMF::actualLen` *does* make sense in that context as it contrasts with `epicsIMF::contentLen`. Here, not so much. I think it would be better to replace "actual" with "uncompressed", both here and in `epicsIMF`. `EPICS_MEM_FILE_RAW` is just a trivial compression algo, after all.

modules/libcom/src/osi/epicsStdio.h
exzombie: Even if you overload `flags`, algorithms can still be an enum which makes it clear they are not flags. Although, I don't see what you gain with the overloaded approach. If you split the 64-bit `flags` field into 32-bit `flags` and 32-bit `algorithm`, the struct will be no bigger and you will still have 32 flags; if you run out of those, it means the API is getting too complicated.

mdavidsaver:
Release note...

## In memory filesystem and built-in `softIoc.dbd`

A new `epicsFOpen()` is introduced, which wraps `fopen()`, with the added ability to access file stored in a new in-memory file system structure.  This is used to compile a copy of `softIoc.dbd`  into the `softIoc` executable as `app:///softIoc/softIoc.dbd`, which will be used as a fallback should `../../bin/dbd/softIoc.dbd` not be available at runtime.

The script `epicsMakeMemFs.pl` (previously RTEMS-only) now understands a new `-f imf` format to generate C code containing a set of files.  `epicsMemMount()` may be used to add an the result into the in memory filesystem.

modules/libcom/src/osi/os/posix/osdMemFile.c
mdavidsaver: I've dropped "Actual" from these names.

modules/libcom/src/osi/epicsMemFile.c
mdavidsaver: I've reworked the "mount" process in a number of ways.  The path string is now copied.

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: I decided instead to switch to:

```c
void epicsMemFileForEach(void (*func)(void *arg, const epicsIMF* mf),
                         void *arg)
```

Which better hides the fact that `epicsIMF*` is now reference counted.  Calls to (un)mount are not allowed from a callback.  I've also added `epicsMemFileShow()` for the common "ls" case.

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: I've tried to clarify which parts of the `epicsIMF` struct are copied, and when.

modules/libcom/src/osi/epicsStdio.h
mdavidsaver: I've split `epicsIMF::flags` up into an enum and a C bit field.

mdavidsaver:
Updated with a number of changes.

* The `app://` prefix.
* `struct epicsIMF` is reference counted, allowing for `free()` when replaced (by a later `epicsMemMount()`), or removed (with the new `epicsMemUnmount()`).
* "iteration" is now accomplished via a callback using `epicsMemFileForEach()`.  This is also wrapped in `EPICS_PRIVATE_API`.
* A public `epicsMemFileShow()` is added which prints info about all in-memory files.  This seems like the most common, and so far only, use for `epicsMemFileForEach()`.

In trying to make use of this feature, I've had some slight annoyance trying to specify the paths which I want.  This is a limitation of the argument parsing of `epicsMakeMemFs.pl` (and my lack of perl).

eg. the build tree from `modules/database/src/std/O.linux-x86_64` appears as:

```
../O.Common/softIoc.dbd
../softIoc/softIocExit.db
```
I would like these to both appear side by side with the prefix `app:///softIoc/...`.  At the moment this is achieved by passing `-T ../softIoc` to strip off the prefix from `softIocExit.db`, implicitly striping the `../`, and a special case to then remove a leading `O.Common/`.

It seems like there should be a more general way to specify this, while also being able to express the Make dependencies.

mdavidsaver:
> It seems like there should be a more general way to specify this, while also being able to express the Make dependencies.

One way would be (yet another) config file, analogous to [`gen_init_cpio`](https://github.com/torvalds/linux/blob/master/usr/gen_init_cpio.c).  I'm not sure doing something similar with `epicsMakeMemFs.pl` would be worth the added complexity though.

mdavidsaver:
> @anjohnson The references I find for [ioslib.h](https://docs.windriver.com/bundle/vxworks_7_application_core_os_sr0630-enus/page/CORE/iosLib.html#iosDrvInstall) don't give the prototype(s) for the callback functions.

Looking into `iosDrvInstall()` again, I was able to find [an example](https://studies.ac.upc.edu/doctorat/ENGRAP/VxWorks-device-drivers.htm) which I think answers my questions.  It looks like this API is about creating specific character device files, but isn't flexible enough to expose a general purpose filesystem.  It does not seem to provide a way to eg. pass per-FD information from `open()` to `read()` and `close()`.  So eg. there is no way to track per-FD file offsets (or per-FD file name).

The only way I can see that `iosDrvInstall()` might be used would be to call `iosDevAdd()` for every in-memory file, and prevent concurrent `open()`.  For the use cases I have in mind during IOC startup, this would be acceptable.  However, I don't like this limitation, and I also wonder whether registering potentially dozens of character devices would hit a system resource limit.

Unless some other API can be found, I'm going to conclude that `epicsIMF` can't be supported on vxWorks.

modules/libcom/RTEMS/epicsMakeMemFs.pl
```diff
@@ -2,18 +2,85 @@
...
+
+=item B<-T> dir
+
+Store paths relative to this directory.  (eg. "$(TOP)")
+
+=item B<-f> I<memfs> | I<imf>
+
+Selects output format.  I<memfs> is the default.
+
+=item B<P> dir
```
anjohnson: Missing `-`.

modules/libcom/RTEMS/epicsMakeMemFs.pl
```diff
@@ -22,41 +89,37 @@
...
 
-  my ($buf, @bufs);
-  while (read($SRC, $buf, 4096)) {
-    @bufs[-1] .= ',' if @bufs;  # Need ',' between buffers
-    push @bufs, join(",", map(ord, split(//, $buf)));
-  }
-  print $DST wrap('', '  ', @bufs);
+        my ($buf, @bufs);
+        while (read($SRC, $buf, 4096)) {
+            $bufs[-1] .= ',' if @bufs;  # Need ',' between buffers
```
anjohnson: `@bufs` => `$bufs` may be wrong.

modules/libcom/RTEMS/epicsMakeMemFs.pl
```diff
@@ -67,19 +130,124 @@
...
+
+        $fname = File::Spec->canonpath($fname);
+        $fname = File::Spec->rel2abs($fname);
+        $fname = File::Spec->abs2rel($fname, $opt_T);
+
+        # attempt Windows to unix path convert
+        $fname =~ s(\\)(/)g;
+
+        # strip leading "../" "./" or "/"
+        $fname =~ s(^\.{0,2}/)()g;
```
anjohnson: If you want to drop _all_ leading `/`, `./` and/or `../` components, adjust the comment and use this regexp:

```perl
$fname =~ s[^(\.{0,2}/)+][];
```

modules/libcom/src/osi/epicsStdio.h
```diff
@@ -192,6 +193,134 @@ LIBCOM_API int epicsStdCall epicsStdoutVPrintf(
...
+    /** \brief File contents */
+    const char *content;
+    /** \brief Length of file contents in bytes */
+    size_t contentLen;
+    /** \brief Actual/uncompressed content length in bytes.
+     *
+     * If `(flags&EPICS_MEM_FILE_ALGO_MASK)==EPICS_MEM_FILE_RAW`
+     * then `contentLen==actualLen`.
+     * Otherwise, `contentLen` is the compressed size,
+     * and `actualLen` is the uncompressed size.
```
anjohnson: Remove/update

anjohnson:
Core Group 2022-10-27: Approved after the minor cleanup in comments above. VxWorks support isn't a big need, not a blocker for this.

mdavidsaver:
I've made the requested changes.

anjohnson:
Core Group: AI on ANJ, examine possible conflicts with existing external ZLIB modules & AreaDetector.

mdavidsaver:
The GHA linux cross builds are failing because of the Debian packaged cross-compilers have `/usr/include` in their default search paths.  So `/usr/include/zlib.h` is found even though libz is not available.

The only ways I can see to resolve this are to 1) figure out how to do a test link of an executable or 2) default to not even probing for `zlib.h`.

Anticipating that I will encounter a similar problem with libssl and PVXS, I would very much like to find a recipe for test linking...

mdavidsaver:
On hold pending a Makefile recipe for test linking.

PR #249, opened 2022-05-11
Use GNU Readline if the header file is installed for linux-aarch64

ffeldbauer:
For the targets linux-arm, -microblaze, -x86, x86_64 and xscale_be the corresponding CONFIG_SITE.Common.$(T_A) files contain the following lines:
```
# Use GNU Readline if the header file is installed
COMMANDLINE_LIBRARY = $(strip $(if $(wildcard \
    $(firstword $(READLINE_DIR) $(GNU_DIR))/include/readline/readline.h), \
  READLINE, EPICS))
```
to automatically include libreadline if the header files are installed on the system.
But these lines are missing for linux-aarch64.

### DIFF
--- a/configure/os/CONFIG_SITE.Common.linux-aarch64
+++ b/configure/os/CONFIG_SITE.Common.linux-aarch64
@@ -21,6 +21,11 @@
 # build doesn't fail to link the readline library.  If none of them work,
 # comment them all out to build without readline support.
 
+# Use GNU Readline if the header file is installed
+COMMANDLINE_LIBRARY = $(strip $(if $(wildcard \
+    $(firstword $(READLINE_DIR) $(GNU_DIR))/include/readline/readline.h), \
+  READLINE, EPICS))
+
 # No other libraries needed (recent Fedora, Ubuntu etc.):
 #COMMANDLINE_LIBRARY = READLINE
 


simon-ess:
Is there perhaps some value in moving this to CONFIG_SITE.Common.linuxCommon? If nearly all of these have the same code, that might be a better location.

ffeldbauer:
It is not exactly the same.
For arm and xscale_be it looks for either READLINE_DIR or the header in GNU_DIR (as in the code snippet above), for the others it only checks the header in GNU_DIR. But I guess also on x86(_64) you could have your own libreadline installed somewhere else

simon-ess:
It's not the same, true, but I wonder if a default should be set and only those that don't match the default should have theirs modified?

till-s:
I think Simon's suggestion makes sense. I have seen many cases where people create a new T_A just for the purpose of having slightly different files or because they want it to be named a certain way. E.g., at PSI we use RHEL7-x86_64 and I have seen similar tweaks at SLAC. If there is a default in Common.linuxCommon then these kind of variants would pick it up automatically.

mdavidsaver:
Something which occurred to me.  This may, or may not, achieve the effect you are looking for.  Though only for recent compilers.  Older compilers would still need CONFIG_SITE overrides, which should continue to take precedence.

https://github.com/epics-base/epics-base/pull/254

anjohnson:
Core developers 8 June: We prefer to use @mdavidsaver #254 approach for all target arch's, which he will work on.

PR #250, opened 2022-05-11
Fix compiler warnings

dirk-zimoch:
Fix a number of compiler warnings.

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -756,7 +756,7 @@ long dbValueSize(short dbr_type)
         sizeof(epicsFloat64),        /* DOUBLE       */
         sizeof(epicsEnum16)};        /* ENUM         */
 
-    if(dbr_type>=NELEMENTS(size))
+    if((size_t)dbr_type>=NELEMENTS(size))
         return 0;
     return(size[dbr_type]);
 }
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -363,12 +363,12 @@ static void getOptions(DBADDR *paddr, char **poriginal, long *options,
         if( (*options) & DBR_AMSG ) {
             if (!pfl) {
                 STATIC_ASSERT(sizeof(pcommon->amsg)==sizeof(pfl->amsg));
-                strncpy(pbuffer, pcommon->amsg, sizeof(pcommon->amsg)-1);
+                strncpy(pbuffer, pcommon->amsg, DB_AMSG_SIZE);
             } else {
-                strncpy(pbuffer, pfl->amsg,sizeof(pfl->amsg)-1);
+                strncpy(pbuffer, pfl->amsg, DB_AMSG_SIZE);
             }
-            pbuffer[sizeof(pcommon->amsg)-1] = '\0';
-            pbuffer += sizeof(pcommon->amsg);
+            pbuffer[DB_AMSG_SIZE-1] = '\0';
+            pbuffer += DB_AMSG_SIZE;
         }
         if( (*options) & DBR_UNITS ) {
             memset(pbuffer,'\0',dbr_units_size);
@@ -853,7 +853,7 @@ static long getLinkValue(DBADDR *paddr, short dbrType,
     {
         const char *rtnString = dbGetString(&dbEntry);
 
-        strncpy(pbuf, rtnString, maxlen-1);
+        strncpy(pbuf, rtnString, maxlen);
         pbuf[maxlen-1] = 0;
         if(dbrType!=DBR_STRING)
             nReq = strlen(pbuf)+1;
--- a/modules/database/src/ioc/db/dbConstLink.c
+++ b/modules/database/src/ioc/db/dbConstLink.c
@@ -168,7 +168,7 @@ static long dbConstLoadScalar(struct link *plink, short dbrType, void *pbuffer)
         return dbPutConvertJSON(pstr, dbrType, pbuffer, &nReq);
     }
 
-    if(dbrType>=NELEMENTS(convert))
+    if((size_t)dbrType>=NELEMENTS(convert))
         return S_db_badDbrtype;
 
     return convert[dbrType](pstr, pbuffer, NULL);
--- a/modules/database/src/ioc/db/dbConstLink.c
+++ b/modules/database/src/ioc/db/dbConstLink.c
@@ -93,7 +93,7 @@ static long cvt_st_UInt32(const char *from, void *pfield, const dbAddr *paddr)
         status = epicsParseFloat64(from, &dval, &end);
         if (!status &&
             dval >=0 &&
-            dval <= ULONG_MAX)
+            dval <= 0xffffffffUL)
             *to = dval;
     }
     return status;
--- a/modules/database/src/ioc/db/dbLock.c
+++ b/modules/database/src/ioc/db/dbLock.c
@@ -724,7 +724,7 @@ void dbLockSetSplit(dbLocker *locker, dbCommon *pfirst, dbCommon *psecond)
             lockRecord *lr=CONTAINER(cur,lockRecord,compnode);
             dbCommon *prec=lr->precord;
             dbRecordType *rtype = prec->rdes;
-            size_t i;
+            int i;
             ELLNODE *bcur;
 
             ellAdd(&newLS, cur);
--- a/modules/database/src/ioc/db/dbLock.c
+++ b/modules/database/src/ioc/db/dbLock.c
@@ -121,8 +121,7 @@ void dbLockIncRef(lockSet* ls)
 {
     int cnt = epicsAtomicIncrIntT(&ls->refcount);
     if(cnt<=1) {
-        errlogPrintf("dbLockIncRef(%p) on dead lockSet. refs: %d\n", ls, cnt);
-        cantProceed(NULL);
+        cantProceed("dbLockIncRef(%p) on dead lockSet. refs: %d\n", ls, cnt);
     }
 }
 
@@ -145,9 +144,8 @@ void dbLockDecRef(lockSet *ls)
     epicsMutexMustLock(ls->lock);
 
     if(ellCount(&ls->lockRecordList)!=0) {
-        errlogPrintf("dbLockDecRef(%p) would free lockSet with %d records\n",
+        cantProceed("dbLockDecRef(%p) would free lockSet with %d records\n",
                      ls, ellCount(&ls->lockRecordList));
-        cantProceed(NULL);
     }
 
     epicsMutexUnlock(ls->lock);
@@ -421,9 +419,8 @@ retry:
 #ifdef LOCKSET_DEBUG
         if(plock->owner) {
             if(plock->owner!=myself || plock->ownercount<1) {
-                errlogPrintf("dbScanLockMany(%p) ownership violation %p (%p) %u\n",
+                cantProceed("dbScanLockMany(%p) ownership violation %p (%p) %u\n",
                              locker, plock->owner, myself, plock->ownercount);
-                cantProceed(NULL);
             }
             plock->ownercount++;
         } else {
@@ -444,8 +441,7 @@ retry:
         /* if we have at least one lockRecord, then we will always lock
          * at least its present lockSet
          */
-        errlogPrintf("dbScanLockMany(%p) didn't lock anything\n", locker);
-        cantProceed(NULL);
+        cantProceed("dbScanLockMany(%p) didn't lock anything\n", locker);
     }
 }
 
@@ -602,17 +598,15 @@ void dbLockSetMerge(dbLocker *locker, dbCommon *pfirst, dbCommon *psecond)
 
 #ifdef LOCKSET_DEBUG
     if(locker && (A->owner!=myself || B->owner!=myself)) {
-        errlogPrintf("dbLockSetMerge(%p,\"%s\",\"%s\") ownership violation %p %p (%p)\n",
+        cantProceed("dbLockSetMerge(%p,\"%s\",\"%s\") ownership violation %p %p (%p)\n",
                      locker, pfirst->name, psecond->name,
                      A->owner, B->owner, myself);
-        cantProceed(NULL);
     }
 #endif
     if(locker && (A->ownerlocker!=locker || B->ownerlocker!=locker)) {
-        errlogPrintf("dbLockSetMerge(%p,\"%s\",\"%s\") locker ownership violation %p %p (%p)\n",
+        cantProceed("dbLockSetMerge(%p,\"%s\",\"%s\") locker ownership violation %p %p (%p)\n",
                      locker, pfirst->name, psecond->name,
                      A->ownerlocker, B->ownerlocker, locker);
-        cantProceed(NULL);
     }
 
     if(A==B)
@@ -688,19 +682,17 @@ void dbLockSetSplit(dbLocker *locker, dbCommon *pfirst, dbCommon *psecond)
 
 #ifdef LOCKSET_DEBUG
     if(ls->owner!=myself || psecond->lset->plockSet->owner!=myself) {
-        errlogPrintf("dbLockSetSplit(%p,\"%s\",\"%s\") ownership violation %p %p (%p)\n",
+        cantProceed("dbLockSetSplit(%p,\"%s\",\"%s\") ownership violation %p %p (%p)\n",
                      locker, pfirst->name, psecond->name,
                      ls->owner, psecond->lset->plockSet->owner, myself);
-        cantProceed(NULL);
     }
 #endif
 
     /* lockset consistency violation */
     if(ls!=psecond->lset->plockSet) {
-        errlogPrintf("dbLockSetSplit(%p,\"%s\",\"%s\") consistency violation %p %p\n",
+        cantProceed("dbLockSetSplit(%p,\"%s\",\"%s\") consistency violation %p %p\n",
                      locker, pfirst->name, psecond->name,
                      pfirst->lset->plockSet, psecond->lset->plockSet);
-        cantProceed(NULL);
     }
 
 
--- a/modules/database/src/ioc/db/dbTest.c
+++ b/modules/database/src/ioc/db/dbTest.c
@@ -339,7 +339,7 @@ long dbgf(const char *pname)
         return -1;
     }
 
-    no_elements = MIN(addr.no_elements, sizeof(buffer)/addr.field_size);
+    no_elements = MIN((size_t)addr.no_elements, sizeof(buffer)/addr.field_size);
     if (addr.dbr_field_type == DBR_ENUM) {
         long status = dbGetField(&addr, DBR_STRING, pbuffer,
             &options, &no_elements, NULL);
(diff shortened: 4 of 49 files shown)


dirk-zimoch:
Not finished yet...

modules/libcom/src/osi/os/vxWorks/vxComLibrary.c
```diff
@@ -17,6 +17,11 @@
  * modules top be loaded
  */
 
+/* avoid "deprecated" warnings for this internal use of symFindByNameEPICS */
+#include "compilerDependencies.h"
+#undef EPICS_DEPRECATED
+#define EPICS_DEPRECATED
+
```
dirk-zimoch, line 24: Not sure if this whole file is still needed.

dirk-zimoch:
Windows not done yet.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -853,7 +853,7 @@ static long getLinkValue(DBADDR *paddr, short dbrType,
     {
         const char *rtnString = dbGetString(&dbEntry);
 
-        strncpy(pbuf, rtnString, maxlen-1);
+        strncpy(pbuf, rtnString, maxlen);
```
exzombie, line 856: Interesting, I can't imagine which warning this would fix. What happened here, which compiler complained? I can't reproduce it with gcc 10.3.

mdavidsaver:
@dirk-zimoch, would you list which version(s) of which compiler(s) you're using?  (since warnings are so specific to compiler version)

dirk-zimoch:
Copied from https://github.com/epics-base/epics-base/wiki/Codeathon-project-ideas#cleanup-compiler-warnings:

Used complilers (for various 64 and 32 bit x86, ppc and arm Linux, 32 bit ppc vxWorks 6, [and 32 and 64 bit Window] targets):

gcc 3.4.3, 4.2.2, 4.3.3, 4.4.2, 4.4.7, 4.6.1, 4.6.4, 4.7.2, 4.8.3, 4.8.5, 7.2.0, 11.2.0
clang 3.4.2, 12.0.0

Tested C++ standards (using gcc): c++98, c++0x, c++11, c++17, c++20, c++23

GNU compilers could use -Wall -Werror now.

dirk-zimoch:
See also https://github.com/dirk-zimoch/epics-base/pull/4 for fixing unused parameters.

modules/database/src/ioc/db/dbTest.c
```diff
@@ -506,62 +506,62 @@ long dbtgf(const char *pname)
     ret_options=0;
 
     dbr_type = DBR_STRING;
-    no_elements = MIN(addr.no_elements,((sizeof(buffer))/MAX_STRING_SIZE));
+    no_elements = MIN(addr.no_elements,sizeof(buffer)/MAX_STRING_SIZE);
```
exzombie: What kind of warning does this fix? I'm still getting a warning about comparing numbers of different sign because `addr.no_elements` is signed while the result of `sizeof` isn't.

mdavidsaver:
> GNU compilers could use -Wall -Werror now.

Neat.  Then to make sure this stays true, I think that this PR should change some of the CI jobs to pass (at least) `-Werror` by adding `-Werror` to `CMD_CPPFLAGS`.

https://github.com/epics-base/epics-base/blob/9ebd84173803e4f83434dcee7a8e08457d03e07c/.github/workflows/ci-scripts-build.yml#L68-L72

eg.

```
extra: "CMD_CXXFLAGS=-std=c++11 -Werror" 
```

dirk-zimoch:
Just don‘t put it into CONFIG.gnuCommon or many EPICS users will be very unhappy with a lot of 3rd party driver code.  ;-)

dirk-zimoch:
Before I forget: There is some RTEMS code that i have not yet fixed.

modules/database/src/ioc/db/recGbl.c
```diff
@@ -44,8 +44,11 @@
 #include "dbScan.h"
 #include "devSup.h"
 #include "link.h"
-#include "recGbl.h"
 
+/* avoid warning about NULL format argument to recGblSetSevrMsg */
+#undef EPICS_PRINTF_STYLE
+#define EPICS_PRINTF_STYLE(a,b)
+#include "recGbl.h"
```
exzombie, line 51: I don't quite understand how this helps. It applies to this file, but there are no calls to `recGblSetSevrMsg` in here. The warning would still be generated by the call sites in other files, no? Or did I misunderstand the cause of the warning?

modules/libcom/test/epicsTimerTest.cpp
```diff
@@ -128,7 +129,7 @@ inline void delayVerify::start ( const epicsTime &expireTime )
 epicsTimerNotify::expireStatus delayVerify::expire ( const epicsTime &currentTime )
 {
     this->expireStamp = currentTime;
-    if ( --expireCount == 0u ) {
+    if ( epics::atomic::decrement(expireCount) == 0u ) {
```
dirk-zimoch, line 132: Since c++14, gcc warns about using `--` on volatile variables.

modules/libcom/src/osi/os/vxWorks/veclist.c
```diff
@@ -73,13 +77,23 @@ int veclist(int all)
...
-        if(status<0 || value != (int)proutine){
+#else
+        symDesc.mask = SYM_FIND_BY_VALUE;
+        symDesc.name = name;
+        symDesc.nameLen = sizeof(name);
+        symDesc.value = (SYM_VALUE) proutine;
+
+        status = symFind(sysSymTbl, &symDesc);
+        value = (FUNCPTR)symDesc.value;
+#endif
```
dirk-zimoch, line 95: Need to test this…

modules/database/src/ioc/db/dbAccess.c
dirk-zimoch: As the commit message says: “**strange** stringop-truncation warnings”. New in gcc 11.
The warning is:
```
In function ‘getLinkValue’,
    inlined from ‘dbGet’ at ../db/dbAccess.c:947:18:
../db/dbAccess.c:856:9: warning: ‘strncpy’ destination unchanged after copying no bytes [-Wstringop-truncation]
  856 |         strncpy(pbuf, rtnString, maxlen-1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

modules/database/src/std/rec/biRecord.c
```diff
@@ -181,11 +181,9 @@ static long get_enum_str(const DBADDR *paddr, char *pstring)
     if(index!=biRecordVAL) {
         strcpy(pstring,"Illegal_Value");
     } else if(*pfield==0) {
-        strncpy(pstring,prec->znam,sizeof(prec->znam));
-        pstring[sizeof(prec->znam)] = 0;
+        sprintf(pstring, "%.*s", MAX_STRING_SIZE-1, prec->znam);
     } else if(*pfield==1) {
-        strncpy(pstring,prec->onam,sizeof(prec->onam));
-        pstring[sizeof(prec->onam)] = 0;
+        sprintf(pstring, "%.*s", MAX_STRING_SIZE-1, prec->onam);
```
exzombie, line 186: What was the complaint here? Replacing a simple byte-copy with string formatting seems a bit heavy-handed to me, although I don't feel like doing performance measurements.

Same for the bo and mbb{i,o} records.

modules/database/src/ioc/db/recGbl.c
dirk-zimoch: The problem is that NULL is not a valid format string in the eyes of some compilers and There is no way to avoid the NULL in the wrapper function recGblSetSevr

modules/database/src/std/rec/biRecord.c
dirk-zimoch: strncpy(dest,src,sizeof(src)) looks buggy to the compiler. Strncpy should protect against dest overflow, not src.

modules/libcom/src/as/asLibRoutines.c
```diff
@@ -29,6 +29,8 @@
 #include "postfix.h"
 #include "asLib.h"
 
+#undef ECHO /* from termios.h */
```
exzombie, line 32: What changes here? I don't see this macro used anywhere.

modules/database/src/std/rec/biRecord.c
dirk-zimoch: Memcpy would do too but printf does the null termination for us.

modules/database/src/ioc/db/recGbl.c
exzombie: Ah, found it, my bad. Rather than `undef`ing the macro, I suggest replacing `NULL` with an empty string.

modules/libcom/src/osi/epicsTime.cpp
```diff
@@ -250,25 +250,13 @@ size_t epicsStdCall epicsTimeToStrftime (char *pBuff, size_t bufLength, const ch
...
             }
             else {
-                static const char pDoesntFit [] = "************";
-                size_t tmpLen = sizeof ( pDoesntFit ) - 1;
-                if ( tmpLen >= bufLenLeft ) {
-                    tmpLen = bufLenLeft - 1;
-                }
-                strncpy ( pBufCur, pDoesntFit, tmpLen );
-                pBufCur[tmpLen] = '\0';
+                size_t tmpLen = sprintf ( pBufCur, "%.*s", (int)(bufLenLeft-1), "************");
```
exzombie, line 259: I like how much simpler this is, it makes it worth using `sprintf`. I'm just wondering why the copy was done "manually" before, was it just "because", or does it affect performance? Personally, I doubt it, this is pretty much an error path.

modules/libcom/src/error/errlog.c
```diff
@@ -680,7 +680,7 @@ static void errlogThread(void)
             print->pos = 0u;
 
             if(nLost && console)
-                fprintf(console, "errlog: lost %zu messages\n", nLost);
+                fprintf(console, "errlog: lost %lu messages\n", (unsigned long) nLost);
```
exzombie, line 683: I see that you replaced all `%zu` specifiers to `%lu`. Is the `z` modifier unsupported on some compilers?

dirk-zimoch: As the commit message says: It is unsupported on vxWorks, even version 6.9. (On older Windows compilers too.)

modules/database/src/ioc/db/recGbl.c
dirk-zimoch: Gives another warning: “empty format string”

modules/database/src/ioc/db/dbConstLink.c
```diff
@@ -93,7 +93,7 @@ static long cvt_st_UInt32(const char *from, void *pfield, const dbAddr *paddr)
         status = epicsParseFloat64(from, &dval, &end);
         if (!status &&
             dval >=0 &&
-            dval <= ULONG_MAX)
+            dval <= 0xffffffffUL)
```
dirk-zimoch, line 96: Note: On most (except Windows) 64 bit machines, ULONG_MAX has 64 bits, but we want to check for 32 bit overflow.

modules/database/src/ioc/db/dbLock.c
```diff
@@ -121,8 +121,7 @@ void dbLockIncRef(lockSet* ls)
 {
     int cnt = epicsAtomicIncrIntT(&ls->refcount);
     if(cnt<=1) {
-        errlogPrintf("dbLockIncRef(%p) on dead lockSet. refs: %d\n", ls, cnt);
-        cantProceed(NULL);
+        cantProceed("dbLockIncRef(%p) on dead lockSet. refs: %d\n", ls, cnt);
     }
```
dirk-zimoch, line 125: Printf format checker complains about passing NULL as format string to cantProceed():
```
../db/dbLock.c: In function `dbLockIncRef':
../db/dbLock.c:125: warning: null format string
```
(Only the VxWorks 5.5 compiler gcc 2.96, but anyway...)

modules/database/src/ioc/db/dbUnitTest.c
```diff
@@ -339,7 +339,7 @@ void testdbGetArrFieldEqual(const char* pv, short dbfType, long nRequest, unsign
                 break;
             }
 #define OP(DBR,Type,pat) case DBR: {Type expect = *(Type*)pbuf, actual = *(Type*)gbuf; assert(vSize==sizeof(Type)); match &= expect==actual; \
-    if(expect!=actual) testDiag("[%lu] expected=" pat " actual=" pat, n, expect, actual); break;}
+    if(expect!=actual) {testDiag("[%lu] expected=" pat " actual=" pat, n, expect, actual);} break;}
```
dirk-zimoch, line 342: Compiler warns "if clause does not guard" because it thinks the author may have meant to have the `break` as part of the `if` branch, because it is on the same line. Braces make the intention clear.
```
../db/dbUnitTest.c: In function ‘testdbGetArrFieldEqual’:
../db/dbUnitTest.c:342:5: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
  342 |     if(expect!=actual) testDiag("[%lu] expected=" pat " actual=" pat, n, expect, actual); break;}
      |     ^~
../db/dbUnitTest.c:344:13: note: in expansion of macro ‘OP’
  344 |             OP(DBR_CHAR, char, "%c");
      |             ^~
../db/dbUnitTest.c:342:91: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  342 |     if(expect!=actual) testDiag("[%lu] expected=" pat " actual=" pat, n, expect, actual); break;}
      |                                                                                           ^~~~~
../db/dbUnitTest.c:344:13: note: in expansion of macro ‘OP’
  344 |             OP(DBR_CHAR, char, "%c");
      |             ^~
```

modules/libcom/src/as/asLibRoutines.c
dirk-zimoch: In modules/libcom/src/flex/flex.skel.static (which is automatically "included"), we have:
`#define ECHO (void) fwrite( (char *) yytext, yyleng, 1, yyout )`
In modules/libcom/src/flex/scan.c too.

modules/database/src/ioc/db/dbTest.c
dirk-zimoch: This line did not actually have a warning. I changed it for consistency with the lines below.
Below, the gcc 11.2.0 wants to have `()` around the divisor to silence a warning.
``` ../db/dbTest.c: In function ‘dbtgf’:
../db/dbTest.c:514:57: warning: expression does not compute the number of elements in this array; element type is ‘long int’, not ‘epicsInt8’ {aka ‘signed char’} [-Wsizeof-array-div]
  514 |     no_elements = MIN(addr.no_elements,((sizeof(buffer))/sizeof(epicsInt8)));
      |                                                         ^
../db/dbTest.c:57:31: note: in definition of macro ‘MIN’
   57 | #   define MIN(x,y)  (((x) < (y)) ? (x) : (y))
      |                               ^
../db/dbTest.c:514:57: note: add parentheses around the second ‘sizeof’ to silence this warning
  514 |     no_elements = MIN(addr.no_elements,((sizeof(buffer))/sizeof(epicsInt8)));
      |                                                         ^
../db/dbTest.c:57:31: note: in definition of macro ‘MIN’
   57 | #   define MIN(x,y)  (((x) < (y)) ? (x) : (y))
      |                               ^
../db/dbTest.c:472:16: note: array ‘buffer’ declared here
  472 |     long       buffer[400];
      |                ^~~~~~
```
The compiler sees that the size of an array is divided by the size of a datatype and knows that usually people do this to caculate the number of elements. But that requires to devide by the size of the type of an array element, which is not the case here. So the compiler thinks this may be a mistake.

The existing parentheses around the dividend and the whole quotient are useless. I removed them in order to focus on the important parentheses.

dirk-zimoch: I only get signedness warnings on Windows.

modules/database/src/ioc/db/recGbl.c
dirk-zimoch: Originally, gcc 2.96 complained:
```
../db/recGbl.c:256: warning: null format string
```
Passing an empty string warns on several other compilers (gcc 4.2.2 and higher):
```
../db/recGbl.c:256:5: warning: zero-length gnu_printf format string [-Wformat-zero-length]
     return recGblSetSevrMsg(precord, new_stat, new_sevr, "");
```

modules/database/src/std/rec/biRecord.c
dirk-zimoch: The compiler warning from gcc 11.2.0 was:
```
../rec/biRecord.c: In function ‘get_enum_str’:
../rec/biRecord.c:184:42: warning: argument to ‘sizeof’ in ‘strncpy’ call is the same expression as the source; did you mean to use the size of the destination? [-Wsizeof-pointer-memaccess]
  184 |         strncpy(pstring,prec->znam,sizeof(prec->znam));
      |                                          ^
../rec/biRecord.c:187:42: warning: argument to ‘sizeof’ in ‘strncpy’ call is the same expression as the source; did you mean to use the size of the destination? [-Wsizeof-pointer-memaccess]
  187 |         strncpy(pstring,prec->onam,sizeof(prec->onam));
      |                                          ^
```
Using `strncopy` here is kind of pointless because we (that is this function and the compiler) do not know how big the buffer `pstring` actually is (a weakness in the API). By browsing the code for calls to this function, I found it is always called with a buffer `MAX_STRING_SIZE` bytes long. But I hesitate to simply write a `0` into  `pstring[MAX_STRING_SIZE-1]` without need.

Maybe the original reason for `strncopy` was to deal with cases where `znam` and `onam` were not 0 terminated? But at least in EPICS 7, this is always the case. Maybe 3.13 was different.

I could simply change this to `strcpy` with no length check at all but I feel at least a bit safer with the `printf` version.

dirk-zimoch:
> `extra: "CMD_CXXFLAGS=-std=c++11 -Werror" `

I tried something like this, actually `extra: "CMD_CXXFLAGS=-std=c++11 CMD_CPPFLAGS=-Werror"`
but that causes strange errors in clang and also fails for mingw.
I am not patient enough for a development cycle that requires me to wait that the github workers  complete after a push. Thus, I leave this to someone who can actually run those tests locally first.

modules/database/src/std/rec/biRecord.c
exzombie: Ok, I can see why the compiler is not too happy with `strncpy(pstring,prec->onam,sizeof(prec->onam));`, I am not exactly thrilled either :) . But what about `strncpy(pstring,prec->onam,MAX_STRING_SIZE);`?

In any case, it seems to me what is really needed is `strlcpy()`. Given how many times you've used `snprintf()` only to ensure nul termination, this may be a good opportunity to add it to epics-base. Perhaps under a different name so as not to conflict with BSD.

modules/libcom/src/as/asLibRoutines.c
exzombie: Ok, but why is it a problem here? I don't see it used in this file. Unless emacs end grep are making fun of me.

modules/database/src/std/rec/biRecord.c
dirk-zimoch: Could be worse if ever called with a buffer `<MAX_STRING_SIZE` but `>=sizeof(prec->onam)`. (Note that `ONAM` is currently in fact shorter.) Be aware that `strncpy` always writes the full buffer, padding it with zeros!
What we really need is a better API with a buffer size argument. Or an API that simply returns a `const char *` and lets the caller copy the data. I think as it is now, `strcpy` is the best option.

modules/libcom/src/as/asLibRoutines.c
dirk-zimoch: I would say flex is making fun of you. The warning I get is:
```
/opt/eldk-5.1/powerpc-4xx-softfloat/sysroots/i686-eldk-linux/usr/bin/ppc405-linux/powerpc-4xx-softfloat-gcc  -D_GNU_SOURCE -D_DEFAULT_SOURCE            -DUNIX  -Dlinux  -fno-strict-aliasing -DBUILDING_libCom_API   -O3   -Wall -Werror-implicit-function-declaration             -fPIC -I. -I../O.Common -I. -I../osi/compiler/gcc -I../osi/compiler/default -I. -I../osi/os/Linux -I../osi/os/posix -I../osi/os/default -I.. -I../as -I../bucketLib -I../calc -I../cvtFast -I../cppStd -I../cxxTemplates -I../dbmf -I../ellLib -I../env -I../error -I../fdmgr -I../flex -I../freeList -I../gpHash -I../iocsh -I../log -I../macLib -I../misc -I../osi -I../pool -I../ring -I../taskwd -I../timer -I../yacc -I../yacc -I../yajl -I../../../../include/compiler/gcc -I../../../../include/os/Linux -I../../../../include         -c asLib.c
In file included from ../as/asLib.y:205:0:
asLib_lex.c:26:0: warning: "ECHO" redefined [enabled by default]
/opt/eldk-5.1/powerpc-4xx-softfloat/sysroots/ppc405-linux/usr/include/bits/termios.h:192:0: note: this is the location of the previous definition
```
asLib_lex.c  is a generated file made from asLib.y and modules/libcom/src/flex/flex.skel.static. It is included by also generated asLib.c which also includes asLibRoutines.c.

modules/libcom/src/osi/epicsTime.cpp
dirk-zimoch: The code had been written by Jeff Hill in 2006. Many he can answer that question?

modules/database/src/ioc/db/dbAccess.c
dirk-zimoch: Maybe the compiler thinks that `maxlen-1` may become negative (if `*nRequest == 0`) but that is actually taken care of in line 830. And it would be a problem in like 857 too.

dirk-zimoch: I am a bit surprised that the  compiler does not complain about line 890. It looks similar to me.

modules/database/src/std/rec/biRecord.c
exzombie: Huh, you're right. I never paid much attention to string field sizes, and I see now that these are mere 26 chars long :/ . That ... can break expectations.

mdavidsaver:
> Originally, gcc 2.96 complained ...

@dirk-zimoch Of course I know why you are still working with a 20 year old compiler.  It still strikes me that attempting to address every warning shown by every compiler version seems like a losing proposition.  Still, it is your time to spend.

From the prospective of PR review though, I've been thinking about a possible "policy" wrt. quieting compiler warnings.  So far this has led me to:

- Where possible, warnings with few false positives should be promoted to errors.  (eg. `-Werror=implicit-function-declaration`)
  - These could internal to the Base build if there are worries about effecting downstream code.
- A disabled warning (aka. `-Wno-*`) needs an accompanying rational.
  - Enough so that the motivation can be understood and reconsidered in future.
- Warnings should only be "quieted" if they are understood/believed to be false positives.
- Priority should be given to warnings shown by newer compilers
- Priority should be given to warnings coming from runtime code (eg. excluding build tools like our flex/yacc forks)
- Fixes for warnings shown only be older compilers (by some arbitrary cutoff) are acceptable if a specific reason can be articulated, and/or if no functional change results.

@anjohnson @ralphlange Thoughts?


> I am not patient enough for a development cycle that requires me to wait that the github workers complete after a push. Thus, I leave this to someone who can actually run those tests locally first.

I hope you understand that unless some of the CI jobs are adjusted to have `-Werror` (or several `-Werror=`) it is likely the future commits by others will unintentionally (re)introduce warnings.  By investing your time now, you are helping the rest of us to help you towards your goal of zero warnings.  Still, I understand why you may not want to do so now.  Though if you don't, I will remind you the next time you complain ;)

anjohnson:
Core Group: Some issues here have been covered in smaller PRs in March 2023, others may follow.

PR #252, opened 2022-05-11
Improve compress record

simon-ess:
This pull request accomplishes two main points:

1. Add testing to the compress record. This record had only very limited testing of two of the algorithms that are used. This pull request adds in more testing for the existing algorithms.
2. This also adds a new field (PBUF) which allows a user to select if partially filled buffers should be nevertheless calculated. For example, if you have a periodically scanned ai record that is being averaged, at the moment it waits until the entire buffer is full before it produces an average. With this change and with PBUF set to YES, then the average will be calculated based on the running tally.

See https://epics.anl.gov/tech-talk/2021/msg00239.php for some context.

Note that there are some other potential improvements to this record: for example, if you have a waveform record as input to a compress record, and you fill up the compress record and process it, then an update to the waveform record that does not fill it completely followed by another process of the compress record will yield no obvious clue that the compress record was not processed.

We could set the compress record in such a case to UDF or some other alarm state. That said, that feels out of scope for the suggested change.

### DIFF
--- a/modules/database/src/std/rec/compressRecord.dbd.pod
+++ b/modules/database/src/std/rec/compressRecord.dbd.pod
@@ -393,7 +393,15 @@ Scan forward link if necessary, set PACT FALSE, and return.
 		interest(1)
 		menu(compressALG)
 	}
- 	field(BALG,DBF_MENU) {
+	field(PBUF,DBF_MENU) {
+		prompt("Use Partial buffers")
+		promptgroup("30 - Action")
+		special(SPC_RESET)
+		interest(1)
+		menu(menuYesNo)
+		initial("NO")
+	}
+	field(BALG,DBF_MENU) {
 		prompt("Buffering Algorithm")
 		promptgroup("30 - Action")
 		special(SPC_RESET)
--- a/modules/database/src/std/rec/compressRecord.dbd.pod
+++ b/modules/database/src/std/rec/compressRecord.dbd.pod
@@ -175,6 +175,11 @@ of all four records will be calculated and placed into the VAL field.
 If PBUF is set to YES, then after each process the average of the first several
 elements will be calculated.
 
+Note that PBUF has no impact on the C<<< Average >>> method. If one wishes to have a
+rolling average computed, then the best way to achieve that is with two compress
+records: a C<<< Circular buffer >>> which is linked to an C<<< N to 1 Average >>>
+record with PBUF set to YES.
+
 The compression record keeps NSAM data samples.
 
 The field N determines the number of elements to compress into each result.
--- a/modules/database/src/std/rec/compressRecord.dbd.pod
+++ b/modules/database/src/std/rec/compressRecord.dbd.pod
@@ -40,7 +40,7 @@ the beginning or the end of the VAL array.
 
 =head2 Parameter Fields
 
-The record-specific fields are described below.
+The record-specific fields are described below, grouped by functionality.
 
 =recordtype compress
 
@@ -59,10 +59,6 @@ menu(bufferingALG) {
 	choice(bufferingALG_LIFO, "LIFO Buffer")
 }
 recordtype(compress) {
-
-=head2 Parameter Fields
-
-The record-specific fields are described below, grouped by functionality.
 
 =head3 Scanning Parameters
 
@@ -85,7 +81,7 @@ algorithms which can be specified as follows:
 
 The following fields determine what channel to read and how to compress the data:
 
-=fields ALG, INP, NSAM, N, ILIL, IHIL, OFF, RES
+=fields ALG, INP, NSAM, N, ILIL, IHIL, OFF, RES, PBUF
 
 As stated above, the ALG field specifies which algorithm to be performed on the data.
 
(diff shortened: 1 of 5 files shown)


modules/database/test/std/rec/compressTest.c
```diff
@@ -346,10 +348,292 @@ void testLIFOCirc(void)
...
+    va_end(args);
+
+    dbScanLock(addr->precord);
+    testOk1(dbPut(addr, DBR_DOUBLE, values, count)==0);
+    dbScanUnlock(addr->precord);
+}
+
+void testArrayAverage(void) {
+    DBADDR wfaddr, caddr;
+
```
exzombie: Extraneous whitespace. Very important ;)

modules/database/src/std/rec/compressRecord.dbd.pod
```diff
@@ -385,7 +385,15 @@ Scan forward link if necessary, set PACT FALSE, and return.
...
 	}
- 	field(BALG,DBF_MENU) {
+	field(PBUF,DBF_MENU) {
+		prompt("Use Partial buffers")
+		promptgroup("30 - Action")
+		special(SPC_RESET)
+		interest(1)
+		menu(menuYesNo)
+		initial("NO")
+	}
```
exzombie, line 416: The new behavior still needs to be documented, both in release notes and here. I know you know this, just adding a reminder so that we don't forget about it.

modules/database/test/std/rec/compressTest.c
```diff
@@ -346,10 +348,292 @@ void testLIFOCirc(void)
...
+    double values[count];
+
+    va_start(args, count);
+    for (i=0; i< count; i++) {
+        values[i] = va_arg(args, double);
+    }
+    va_end(args);
+
+    dbScanLock(addr->precord);
+    testOk1(dbPut(addr, DBR_DOUBLE, values, count)==0);
```
exzombie: As I see it, `writeToWaveform` is a utility function, and it's only used to write to the waveform record. Thus, I don't think this `dbPut` needs to be checked because it can only fail if the test itself is broken, which has nothing to do with the code under test. Testing it only complicates counting your tests.

modules/database/test/std/rec/compressTest.c
```diff
@@ -346,10 +348,292 @@ void testLIFOCirc(void)
...
+    testDEq(buf, 5.0, 0.01);
+    dbScanUnlock(caddr.precord);
+
+
+    testIocShutdownOk();
+    testdbCleanup();
+}
+
+void
+testAIPartialAverage(void) {
```
exzombie: I'd name the function `testAIAveragePartial` to maintain consistency with other tests.

modules/database/test/std/rec/compressTest.c
```diff
@@ -346,10 +348,292 @@ void testLIFOCirc(void)
...
+    double buf = 0.;
+    double data[4] = {1., 2., 3., 4.};
+    double expected[4] = {1., 1.5, 2., 2.5};
+    long nReq = 1;
+    int i;
+    DBADDR aiaddr, caddr;
+
+    testDiag("Test 'N to 1 Low Value'");
```
exzombie: The string naming this test is from a previous test.

modules/database/test/std/rec/compressTest.c
```diff
@@ -346,10 +348,292 @@ void testLIFOCirc(void)
...
+    eltc(0);
+    testIocInitOk();
+    eltc(1);
+
+    fetchRecordOrDie("ai", aiaddr);
+    fetchRecordOrDie("comp", caddr);
+
+    for (i = 0; i < 4; i++) {
+        dbScanLock(aiaddr.precord);
+        testOk1(dbPut(&aiaddr, DBR_DOUBLE, &data[i], 1) == 0);
```
exzombie: See comment on `writeToWaveform`, same considerations apply.

modules/database/test/std/rec/compressTest.c
```diff
@@ -346,10 +348,292 @@ void testLIFOCirc(void)
...
+    if (dbGet(&caddr, DBR_DOUBLE, &buf, NULL, &nReq, NULL))
+        testAbort("dbGet failed on compress record");
+
+    // We confirm that this hasn't changed i.e. the dbProcess above did nothing
+    testDEq(buf, 1.0, 0.01);
+    dbScanUnlock(caddr.precord);
+
+    testDiag("Test partial data with PBUF set to YES");
+
+    ((compressRecord *)caddr.precord)->pbuf = menuYesNoYES;
```
exzombie, line 576: Record fields need to be accessed with the record locked.

exzombie:
I found no issues with the changes to the record.

As for tests, it's great that you added some, and I only have minor comments. As this is a strict improvement over the previous state, I think it could be merged as-is. That said, the tests could be improved further by making them less repetitive. We can discuss how it could be done if there is interest in doing so.

I agree that putting the record to UDF is out of scope. But for the record (har, har) and future discussion, I wonder how much of a compatibility issue such a change would introduce. On one hand, it would make things more "poper", but it's not worth introducing issues over that. On the other hand, I can't imagine a use case where compress record's STAT becoming UDF would cause any problems, apart from being indicated on an operator screen when it wasn't before. But I guess that just means I'm not imaginative enough :)

simon-ess:
I have updated the documentation, let me know if this seems sufficient.

modules/database/src/std/rec/compressRecord.dbd.pod
simon-ess: I have updated the documentation, let me know if this seems sufficient.

exzombie:
As a user looking at the docs of the PBUF field, I'd intuitively expect it to also work with the `Average` algorithm, making it calculate a moving average. This didn't occur to me yesterday because the use case described on the mailing list calculates exactly that. But without that context, it feels like an omission. I'm not necessarily suggesting implementing a moving average, it feels a bit like scope creep. But perhaps it could be clearly stated in the docs that PBUF does not affect the `Average` algorithm and that one should instead combine a `Circular Buffer` with `N to 1 Average` to achieve the desired effect.

simon-ess:
@exzombie - I had thought the same thing as well. The "Average" algorithm is a bit strange to me in that I would have expected it to behave exactly as you said. However, experimentation and a closer reading showed a totally different behaviour...

Anyhow, I will update the documentation as you suggest. I think that is a good addition.

anjohnson:
Group 2023-03-06: Approved subject to adding a Release Notes entry.

anjohnson:
LGTM, thanks.

mdavidsaver:
Applied as of e5ad12e638629ad814c1eb158e742951c6347db1

PR #253, opened 2022-05-11
Codeathon22 deadline sleep

till-s:
Prepared for using deadline sleep in `periodicTask()` (`dbScan.c`). To that end I introduced `epicsEventWaitUntil()` to the _epicsEvent_ facility. Under the hood, most platforms use `pthread_cond_timedwait()` which has relatively wide support (when using CLOCK_REALTIME, that is;
CLOCK_MONOTONIC is mostly unsupported).

All platforms but WIN32 and RTEMS-score now use posix/osdEvent; vxWorks/osdEvent and RTEMS-posix/osdEvent have been removed. VxWorks < tornado 2.2 is no longer supported.

Windows also supports an absolute timeout which is used to implement `epicsEventWaitUntil()`.

The RTEMS-4 (RTEMS-score) implementation of `epicsEventWaitUntil()` simply converts to a relative timeout because the RTEMS-4 kernel has no support for timers with absolute deadlines (RTEMS-4 `pthread_cond_timedwait()` also converts the absolute timeout to a relative one under the hood, so nothing would be gained by using the posix version under RTEMS-4).

I have completed the task by adding tests for `epicsEventWaitUntil()` and converting the `periodicTask()` to use deadline sleeps via `epicsEventWaitUntil()`. The deadline is updated in integer arithmetic so there are no round-off errors.
There is a caveat: the clock used for sleeping had to be moved from a monotonic clock to the real-time/epicsTime clock because most platforms do not support using a monotonic clock. Consequences:

- the deadline is passed as an `epicsTimeStamp` but the underlying OS timer is using the OS' real-time clock. Thus, discrepancies could arise if epics time and the system time diverge (handling of leap seconds, different synchronization mechanisms, ...)
- if the system clock is stepped backwards by a large amount while a scanning task is sleeping then the sleeping task still waits for the original timeout and may not wake up for a long time.


modules/libcom/src/osi/os/WIN32/osdTimePvt.h
```diff
@@ -1,21 +1,24 @@
 /*************************************************************************\
-* Copyright (c) 2011 UChicago Argonne LLC, as Operator of Argonne
+* Copyright (c) 2002 The University of Chicago, as Operator of Argonne
```
simon-ess, line 2: This seems an accidental change...

till-s: It seems to be a git/github artifact. This is a new file that was created as a copy+edit from osdTime.h (which has the 2002 copyright header) and added to this branch. I don't know how this diff was created. If I run

       `git diff 7.0..HEAD`

the output looks very different (new file).

Nevertheless, thanks to your comment I found a typo in a comment in this very file, thanks ;-)

modules/libcom/src/osi/epicsEvent.h
```diff
@@ -200,6 +201,20 @@ LIBCOM_API void epicsEventMustWait(epicsEventId id);
...
+ * \param id The event identifier.
+ * \param timeout The absolute timeout. If the timeout has already
+ * expired then this is equivalent to calling epicsEventTryWait().
+ * Note that the time is measured by the underlying system (realtime)
+ * clock (and not any epics time provider). To emphasize this fact the
+ * argument is a struct timespec.
+ *
+ * \return Status indicator.
+ **/
+LIBCOM_API epicsEventStatus epicsEventWaitWithAbsTimeout(
```
anjohnson: How about naming this `epicsEventWaitUntil(id, timeStamp)` to be slightly less Germanic about it?

modules/database/src/ioc/db/dbScan.c
```diff
@@ -782,12 +811,13 @@ static void periodicTask(void *arg)
...
     double over_max = 0.0;
-    const double penalty = (ppsl->period >= 2) ? 1 : (ppsl->period / 2);
+    const epicsPeriod penalty = periodFromDouble((ppsl->period >= 2) ? 1 : (ppsl->period / 2));
+    const epicsPeriod period  = periodFromDouble( ppsl->period );
 
     taskwdInsert(0, NULL, NULL);
     epicsEventSignal(startStopEvent);
 
-    epicsTimeGetMonotonic(&next);
+    epicsTimeGetCurrent(&next);
```
anjohnson, line 824: @mdavidsaver what was the problem we were solving when we switched to `epicsTimeGetMonotonic()`? Would doing this reintroduce that issue?

modules/database/src/ioc/db/dbScan.c
```diff
@@ -59,6 +59,35 @@ static epicsEventId startStopEvent;
 
 static volatile enum ctl scanCtl;
 
+/* epicsTime does not support addition/subtraction of epicsTimeStamp
+ * but we want to do integer arithmetic...
+ */
+typedef struct epicsPeriod {
+    epicsTimeStamp ts;
+} epicsPeriod;
```
anjohnson, line 68: It seems a bit weird for a period to have a sub-component called `secPastEpoch`, why not just embed members called `sec` and `nsec` directly into `struct epicsPeriod`?

anjohnson (commented):
Some suggestions/ideas.

modules/libcom/src/osi/epicsEvent.h
till-s: I have no preferences either way. epicsEventWaitUntil() sounds good to me.

modules/database/src/ioc/db/dbScan.c
till-s: I don't know. However, deadline sleep on an monotonic clock has many more portability issues than I could address in a few days. Might not be supported on RTOSes like vxWorks-6. Windows would also be more difficult.

modules/database/src/ioc/db/dbScan.c
till-s: Agreed - my goal was to keep it consistent with whatever data type(s) epicsTimeStamp uses so that if the latter ever goes 64bit epicsPeriod would automatically follow and there would be no warnings about signedness and/or truncation. I introduced a new name/type only because it seems from the declarations that epicsTime is always relative to the epoch. epicsPeriod should represent a 'relative' time (while using the same types as epicsTimeStamp's fields). If you think that none of this matters I'd be happy to either use a plain epicsTimeStamp or redefine the period as an independent struct.
Given that none of us will live the day (maybe Elon has a few ideas?) when a 64-bit period expires I can see that it is reasonable to just use uint32 for the period's sec/nsec.

modules/database/src/ioc/db/dbScan.c
mdavidsaver: If the OS doesn't really provide monotonic, then `epicsTimeGetMonotonic()` will actually be returning wall clock time.  (I can't conjure monotonic time from nowhere)

mdavidsaver: The problem being solved was a system clock time jump leading to an overly long or short delay (eg. hours if +-time zone happens).

modules/database/src/ioc/db/dbScan.c
anjohnson: `nsec` is an `epicsUInt32` since that never needs to hold more than 1e9 (or ±2e9 after doing an addition/subtraction).

We might want to allow periods to be negative, so I see the main question as whether we need to be able to represent more than ±68 years (2**31 seconds) as a period. I'm going to say no, so use an `epicsInt32` for `secs` as well.

till-s: OK, done.

modules/libcom/src/osi/epicsEvent.h
till-s: Done.

modules/database/src/ioc/db/dbScan.c
till-s: This problem is going to come back (as I said from the beginning) but has also never really gone away. The traditional approach (dbScan uses `epicsMonotonicGet()`, computes relative timeouts and `epicsEventWaitWithTimeout()`) for the most important OS classes:
- vxworks -> OK, epicsEventWaitWithTimeout() uses relative sleep
- RTEMS -> OK, uses relative sleep
- WIN -> OK, uses relative sleep
- POSIX -> NOT OK - uses absolute sleep on `CLOCK_REALTIME` under the hood

The new approach (`epicsEventWaitUntil()`) makes basically all systems behave like POSIX - except for OSes which under the hood use relative sleep (RTEMS-4, maybe vxworks [<7 ?] - I have come across emails where someone claimed having peeked at the source code and seen a conversion to relative sleep).

Another approach I could envision is
 - `epicsEventWaitUntil()` only uses a true deadline sleep on systems that support
     - sleeping on a monotonic clock *and* 
     - `epicsMonotonicGet()` reads that same clock.
 - on other systems `epicsEventWaitUntil()` reverts to relative sleep
 - this leaves POSIX (only) systems that do not support `pthread_condattr_setclock(CLOCK_MONOTONIC)` *and* `clock_gettime(CLOCK_MONOTONIC)` in the rain; these will still have to use absolute sleep on `CLOCK_REALTIME`.
 
Finally: one *advantage* you get from using `CLOCK_REALTIME` is that on most systems this synced to NTP or similar, i.e, the scan task has much better long-term stability than if it used `CLOCK_MONOTONIC`. So it comes down to a decision: do you prefer long-term stability or no glitches when you step the clock...

anjohnson:
Core Group 2022-10-28: We appreciate the work you put into this implementation, thank-you for investigating the possibility as described for the Codeathon.

The inability of most targets OSs to use the monotonic time source for deadline scheduling is a major concern to us though. A jump in the wall-clock time on network time synchronization does happen, and we switched the periodic scan threads to using the monotonic source precisely to avoid that. We don't see the long-term precision of the scan periods as being more important than that, so unfortunately we don't think this is something we want to merge as given. If/when our non-Posix targets at the time provide equivalent facilities we could revisit this.

PR #254, opened 2022-05-12
automatic COMMANDLINE_LIBRARY w/ newer compilers

mdavidsaver:
clang, recent GCC, and even more recent MSVC support [`__has_include()`](https://en.cppreference.com/w/cpp/preprocessor/include) which allows the preprocessor to test for the existence of a header file in the configured search path (including `-I`).  This becomes standard w/ c++17.

### DIFF
--- a/configure/CONFIG_COMMON
+++ b/configure/CONFIG_COMMON
@@ -288,7 +288,6 @@ TARGET_LIB_LDFLAGS = $(if $(findstring $*, $(LOADABLE_LIBRARY)), \
 
 #--------------------------------------------------
 # Command-line input support default
-COMMANDLINE_LIBRARY = EPICS
 OP_SYS_LDLIBS += $(LDLIBS_$(COMMANDLINE_LIBRARY))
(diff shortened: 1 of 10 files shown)


anjohnson:
Needs an RTEMS check for libtecla, but this is a better approach. Need to remove the existing checks from the build system.

mdavidsaver:
I've made the requested changes.  Also, it turns out that the mingw-w64 install in the GHA windows-2019 environment includes readline.  I was able to make the necessary change for this to be used (not overwrite `OP_SYS_LDLIBS`, and omit `-lcurses`).

anjohnson:
Core group 7/6: Approved, short Release Notes entry please.

mdavidsaver:
Merged as of a0a7a949447940f446a62a0afe9065a27b08e254

mdavidsaver:
Amended by 52cc68433fab94852955c6d7f5c84e9572ab6658 to restore the fallback to `fallback to $(wildcard $(GNU_DIR) ...`.

PR #257, opened 2022-05-12
Added Doxygen annotations to modules/libcom/src/iocsh/iocsh.h

roehrig:


modules/libcom/src/iocsh/iocsh.h
```diff
@@ -10,6 +10,27 @@
...
+ * 
+ * @par Command functions:
+ *     int iocsh (const char *pathname)@n 
+ *     int iocshLoad (const char *pathname, const char *macros)@n 
+ *     int iocshCmd (const char *cmd)@n 
+ *     int iocshRun (const char *cmd, const char *macros)
+ * 
+ * @par Registration functions:
+ *     void iocshRegister (const iocshFuncDef * piocshFuncDef, iocshCallFunc func)@n 
+ *     void epicsStdCall iocshRegisterVariable (const iocshVarDef *piocshVarDef)
```
brendanchandler, line 31: I found Doxygen also supports markdown syntax, so you could write these lists with markdown list syntax and avoid the @n at the end.  No need to do that for this pull request, just FYI in case you're writing doxygen annotations in the future.

```suggestion
 * @par Command functions:
 *     - int iocsh (const char *pathname)
 *     - int iocshLoad (const char *pathname, const char *macros)
 *     - int iocshCmd (const char *cmd)
 *     - int iocshRun (const char *cmd, const char *macros)
 * 
 * @par Registration functions:
 *     void iocshRegister (const iocshFuncDef * piocshFuncDef, iocshCallFunc func)@n 
 *     void epicsStdCall iocshRegisterVariable (const iocshVarDef *piocshVarDef)
```

anjohnson:
Rebased and merged at 83a685867c0e84b44d08c14f1767186303e35a16

Thanks!

PR #262, opened 2022-05-13
Modify genVersionHeader.pl to add commit time (in epoch).

hanlet:
Tested for git.
Not tested: darcs, mercurial, or svn
Skipping: bzr

### DIFF
--- a/src/tools/genVersionHeader.pl
+++ b/src/tools/genVersionHeader.pl
@@ -29,8 +29,7 @@ our $opt_t = '.';
 our $opt_N = 'VCSVERSION';
 our $opt_V = $now;
 
-our $opt_T = 'VCSREVISION';
-our $opt_C = $now;
+our $RevDate = $now;
 
 my $vcs;
 my $cv;
@@ -57,7 +56,7 @@ if (!$vcs && -d "$opt_t/_darcs") { # Darcs
         my $hasmod = `darcs whatsnew --repodir="$opt_t" -l`;
         $opt_V .= '-dirty' unless $?;
     }
-    $cv = `darcs optimize --reorder; darcs changes --context | sort | md5sum`; # this is untested
+    $cv = ""; # ToDo
 }
 if (!$vcs && -d "$opt_t/.hg") { # Mercurial
     print "== Found <top>/.hg directory\n" if $opt_v;
@@ -74,7 +73,7 @@ if (!$vcs && -d "$opt_t/.hg") { # Mercurial
         chomp $hasmod;
         $opt_V .= '-dirty' if $hasmod ne '';
     }
-    $cv = `hg log -l1 --template '{date(date, "%s")}'` # this is untested
+    $cv = `hg log -l1 --template '{date|isodate}'` # this is untested
 }
 if (!$vcs && -d "$opt_t/.git") { # Git
     print "== Found <top>/.git directory\n" if $opt_v;
@@ -104,7 +103,7 @@ if (!$vcs && -d "$opt_t/.svn") { # Subversion
         chomp $hasmod;
         $opt_V .= '-dirty' if $hasmod ne '';
     }
-    $cv = `svn info --show-item revision`; # this is untested
+    $cv = `svn info --show-item last-changed-date`; # this is untested
 }
 if (!$vcs && -d "$opt_t/.bzr") { # Bazaar
     print "== Found <top>/.bzr directory\n" if $opt_v;
@@ -116,7 +115,7 @@ if (!$vcs && -d "$opt_t/.bzr") { # Bazaar
         $opt_V = $result;
         $vcs = 'Bazaar';
     }
-    $cv = `bzr version-info -q -timestamp`;  # this is untested
+    $cv = `bzr version-info -q --custom --template='{date}'`;
 }
 if (!$vcs) {
     print "== No VCS directories\n" if $opt_v;
@@ -130,7 +129,7 @@ if (!$vcs) {
 }
 
 chomp $cv;
-$opt_C=$vcs . ': ' . $cv;
+$RevDate=$vcs . ': ' . $cv;
 
 my $output = << "__END";
 /* Generated file, do not edit! */
@@ -140,8 +139,8 @@ my $output = << "__END";
 #ifndef $opt_N
   #define $opt_N \"$opt_V\"
 #endif
-#ifndef $opt_T
-  #define ${opt_N}_DATE \"$opt_C\"
+#ifndef ${opt_N}_DATE
+  #define ${opt_N}_DATE \"$RevDate\"
 #endif
 __END
 
--- a/src/tools/genVersionHeader.pl
+++ b/src/tools/genVersionHeader.pl
@@ -57,8 +57,7 @@ if (!$vcs && -d "$opt_t/_darcs") { # Darcs
         my $hasmod = `darcs whatsnew --repodir="$opt_t" -l`;
         $opt_V .= '-dirty' unless $?;
     }
-    # this is untested
-    $cv = `darcs optimize --reorder; darcs changes --context | sort | md5sum`;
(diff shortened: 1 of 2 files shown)


mdavidsaver:
A good first usage would be to modify `coreRelease()` to add the commit date.

https://github.com/epics-base/epics-base/blob/9ebd84173803e4f83434dcee7a8e08457d03e07c/modules/database/src/ioc/misc/epicsRelease.c#L19-L26

src/tools/genVersionHeader.pl
```diff
@@ -29,7 +29,11 @@
 our $opt_N = 'VCSVERSION';
 our $opt_V = $now;
 
+our $opt_T = 'VCSREVISION';
+our $opt_C = $now;
```
mdavidsaver: Can `$opt_T` and `$opt_C` be removed or renamed?  `$opt_` suggests that these are CLI arguments, but the call to `getopts()` isn't being modified.

src/tools/genVersionHeader.pl
```diff
@@ -128,12 +140,16 @@
 #ifndef $opt_N
   #define $opt_N \"$opt_V\"
 #endif
+#ifndef $opt_T
+  #define ${opt_N}_DATE \"$opt_C\"
+#endif
```
mdavidsaver, line 144: The `#ifndef` condition should match the variable being defined.  eg.

```c
#ifndef ${opt_N}_DATE
#  define ${opt_N}_DATE \"$opt_C\"
#endif
```

src/tools/genVersionHeader.pl
```diff
@@ -108,6 +116,7 @@
         $opt_V = $result;
         $vcs = 'Bazaar';
     }
+    $cv = `bzr version-info -q -timestamp`;  # this is untested
```
mdavidsaver: `-timestamp` isn't valid, at least for the "bzr" which ships with Debian 11 (`Breezy (brz) 3.1.0`).

```
bzr version-info -q --custom --template='{date}'
```

prints something like `2012-01-02 10:57:46 +0100`.

src/tools/genVersionHeader.pl
```diff
@@ -97,6 +104,7 @@
         chomp $hasmod;
         $opt_V .= '-dirty' if $hasmod ne '';
     }
+    $cv = `svn info --show-item revision`; # this is untested
```
mdavidsaver: Digging up an old SVN repository...

```
svn info --show-item last-changed-date
```

Prints something like `2016-09-29T19:54:09.131366Z`.

src/tools/genVersionHeader.pl
```diff
@@ -69,6 +74,7 @@
         chomp $hasmod;
         $opt_V .= '-dirty' if $hasmod ne '';
     }
+    $cv = `hg log -l1 --template '{date(date, "%s")}'` # this is untested
```
mdavidsaver:
```
hg log -l1 --template '{date|isodate}'
```
Prints something like `2015-07-08 19:54 -0400`.

src/tools/genVersionHeader.pl
```diff
@@ -53,6 +57,7 @@
         my $hasmod = `darcs whatsnew --repodir="$opt_t" -l`;
         $opt_V .= '-dirty' unless $?;
     }
+    $cv = `darcs optimize --reorder; darcs changes --context | sort | md5sum`; # this is untested
```
mdavidsaver: [`man darcs`](https://linux.die.net/man/1/darcs) for `optimize` states that it "... modifies the current repository ..." which doesn't seem like a good idea here.

I've no idea what the correct incantation would be, or even if there is one.  Darcs has a very difference conception of history.  Rather then guessing, how about just leaving this TODO?

```perl
    $cv = ""; # TODO
```

src/tools/genVersionHeader.pl
```diff
@@ -82,6 +88,7 @@
         $opt_V = $result;
         $vcs = 'Git';
     }
+    $cv = `git show -s --format=%ci HEAD`;
```
mdavidsaver, line 90: For completeness, this prints something like `2020-05-05 16:59:56 -0500` circa Git 2.30.2.

mdavidsaver (changes requested):

src/tools/genVersionHeader.pl
```diff
@@ -53,6 +56,7 @@
         my $hasmod = `darcs whatsnew --repodir="$opt_t" -l`;
         $opt_V .= '-dirty' unless $?;
     }
+    $cv = ""; # ToDo
```
anjohnson, line 59: AI on RL to check and advise on darcs.

src/tools/genVersionHeader.pl
```diff
@@ -108,6 +115,7 @@
         $opt_V = $result;
         $vcs = 'Bazaar';
     }
+    $cv = `bzr version-info -q --custom --template='{date}'`;
```
anjohnson, line 118: AI on ANJ: Check/test bzr.

anjohnson (commented):
Final checks & darcs code needed.

src/tools/genVersionHeader.pl
ralphlange: Darcs code:
`darcs log --last 1 | grep Date: | sed -s 's/Date:   //'`

This takes the date of the last commit, while darcs may reorder commits, so commits further down may have newer dates. (But in principle this also applies to git, where commits could have been moved.)

ralphlange: This prints something like `Wed Jun  8 19:01:42 CEST 2022`.

src/tools/genVersionHeader.pl
anjohnson: This gives `2017-03-03 18:18:57 -0600` (the last commit on the Base-3.14 BZR branch).

src/tools/genVersionHeader.pl
anjohnson: In case this isn't obvious, Ralph's `| grep ... | sed ...` part isn't portable to Windows, so would need to be converted into Perl.

anjohnson (commented):
This is the iocInit output from a softIoc built using this PR branch:
```
Starting iocInit
## EPICS R7.0.6.2-DEV
## Rev. R7.0.6.1-157-ga291e9d94c2f866c9bcb-dirty
## Rev. Date Git: 2022-07-19 19:58:54 -0500
epics>
```

anjohnson:
Core Group 7/20: AI on RL to post output from darcs without filter. AI on AJ to finish off and merge.

src/tools/genVersionHeader.pl
ralphlange: The first part of the command prints something like...
```
ralph@debian-testing:~/TTT/TT$ darcs log --last 1
patch 73f74629a41ec69dc09d3230489c335cbcd8b0ab
Author: ralph.lange@gmx.de
Date:   Wed Jun  8 19:01:42 CEST 2022
  * another change
ralph@debian-testing:~/TTT/TT$
```

ralphlange:
Output from Darcs without filter:
```
ralph@debian-testing:~/TTT/TT$ darcs log --last 1
patch 73f74629a41ec69dc09d3230489c335cbcd8b0ab
Author: ralph.lange@gmx.de
Date:   Wed Jun  8 19:01:42 CEST 2022
  * another change
ralph@debian-testing:~/TTT/TT$
```

src/tools/genVersionHeader.pl
anjohnson: Committing the Perl version as:
```perl
   $cv = `darcs log --last 1`;
   $cv =~ s/ \A .* Date: \s+ (\N+) .* \z /$1/sx;
```

anjohnson:
Merging with some updates for portability (removing bash-isms).

PR #263, opened 2022-05-13
Fix undef ts on first camonitor update of NORD

bhill-slac:
Moves db_post_events() calls for NORD to process() function
after it updates the timestamp.

DRAFT
Needs a test case to show bug in 7.0 branch and verify the fix.

### DIFF
--- a/modules/database/src/std/rec/waveformRecord.c
+++ b/modules/database/src/std/rec/waveformRecord.c
@@ -124,6 +124,7 @@ static long process(struct dbCommon *pcommon)
     struct waveformRecord *prec = (struct waveformRecord *)pcommon;
     wfdset *pdset = (wfdset *)(prec->dset);
     unsigned char pact=prec->pact;
+	epicsUInt32 nord = prec->nord;
     long status;
 
     if ((pdset==NULL) || (pdset->read_wf==NULL)) {
@@ -143,6 +144,8 @@ static long process(struct dbCommon *pcommon)
     prec->udf = FALSE;
     recGblGetTimeStampSimm(prec, prec->simm, &prec->siol);
 
+	if (nord != prec->nord)
+		db_post_events(prec, &prec->nord, DBE_VALUE | DBE_LOG);
     monitor(prec);
 
     /* process the forward scan link record */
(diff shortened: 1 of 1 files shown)


modules/database/src/std/rec/waveformRecord.c
```diff
@@ -206,7 +209,10 @@ static long put_array_info(DBADDR *paddr, long nNew)
         prec->nord = prec->nelm;
 
     if (nord != prec->nord)
+	{
+		printf( "put_array_info: db_post_events(DBE_VALUE\|DBE_LOG)\n" );
```
mdavidsaver: A debug `printf()` to remove?

modules/database/src/std/rec/waveformRecord.c
```diff
@@ -124,6 +124,7 @@ static long process(struct dbCommon *pcommon)
     struct waveformRecord *prec = (struct waveformRecord *)pcommon;
     wfdset *pdset = (wfdset *)(prec->dset);
     unsigned char pact=prec->pact;
+	epicsUInt32 nord = prec->nord;
```
anjohnson: Please use 4 spaces for indentation instead of tab char's. Applies to the other changes below as well.

mdavidsaver:
This PR looks like the right approach.

> Needs a test case to show bug in 7.0 branch and verify the fix.

This would certainly be an improvement, although I won't insist on this.  The testing code change seems likely to be considerably larger than the fix because none of the existing `dbUnitTest.h` helpers (`testMonitor*()`) will actually be helpful in testing this behavior.  What would be helpful is a variant of `testMonitorWait()` which calls `dbGet()` from the dbEvent callback (`testmonupdate()`).

bhill-slac:
I've fixed the tabs and extra printf.    I also looked into adding a test case but I would have had to make some big additions/changes to the test framework.    Not clear that it's worth it for this simple bug which is easy to verify by hand.

anjohnson:
Core Group 7/6: Approved. Core group would like to see fixes for the other record types too, particularly aai and aao (can be in a separate PR).

mdavidsaver:
Applied as 5ba8080f6ed698da477cc33b54ee939f372fa031.  Opened #280 to track a fix for aai and aao.

PR #265, opened 2022-05-13
Doxygen support for ipAddrToAsciiAsynchronous.h

brendanchandler:


modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
```diff
@@ -8,6 +8,55 @@
...
+ *
+ * // Do other work and wait for results to finish.
+ * // Note that typically you would use an epicsEvent or similar to coordinate
+ * // making sure the result is ready from your callback.  For this example,
+ * // we'll ignore that complexity and use a naive sleep call instead.
+ * sleep(2);
+ *
+ *
+ * trans.release();
+ * engine.release();
```
mdavidsaver, line 66: The example shown here differs from the in-tree usages (`msgForMultiplyDefinedPV` and `hostNameCache`).  These hold the `ipAddrToAsciiTransaction &` as a member of the `ipAddrToAsciiCallBack` sub-class, and call `ipAddrToAsciiTransaction::release()` from the destructor.  This gives some added exception safety.

imo. The need for manual reference counting make this a tricky API to use wrt. walking the fine line between memory leaks and use-after-free.  Sticking close to the in-tree usages is probably a good idea.

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
```diff
@@ -19,28 +68,79 @@
...
+     * \param ipAddrToAsciiCallBack Reference to the user supplied callbacks to call when the result is available
+     */
     virtual void ipAddrToAscii ( const osiSockAddr &, ipAddrToAsciiCallBack & ) = 0;
+
+    /** \brief Get the conversion address currently set
+     * \return Get the last (or current) address converted to ascii
+     */
     virtual osiSockAddr address () const  = 0;
+
+    /* \brief Prints the converted IP address
```
mdavidsaver: Should be `/**` here, and below.

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
```diff
@@ -19,28 +68,79 @@
...
 public:
+    /// Destroy this transaction object and remove from the parent engine object
     virtual void release () = 0;
+
+    /** \brief Convert an IP address to ascii, asynchronously
+     *
+     * \param osiSockAddr Reference to the address to convert
+     * \param ipAddrToAsciiCallBack Reference to the user supplied callbacks to call when the result is available
+     */
     virtual void ipAddrToAscii ( const osiSockAddr &, ipAddrToAsciiCallBack & ) = 0;
```
mdavidsaver: Add something like:
```
\note The ipAddrToAsciiCallBack referenced must remain valid until release() is called on this transaction.
```

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
```diff
@@ -19,28 +68,79 @@
...
+     * Prints to stdout
+     *
+     * \param level 0 for basic information, 1 for extra info
+     */
     virtual void show ( unsigned level ) const = 0;
+
+    /** \brief Creates a new ipAddrToAsciiEngine to convert IP addresses
+     *
+     * \return Newly created engine object
+     */
```
mdavidsaver, line 160:
```
\note Caller must release() this engine.
```

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
```diff
@@ -19,28 +68,79 @@
...
 
+/// Class which manages creating transactions for converting ipAddr's to ASCII
 class LIBCOM_API ipAddrToAsciiEngine {
 public:
+    /// Cancel any pending transactions and destroy this ipAddrToAsciiEngine object.
     virtual void release () = 0;
+
+    /** \brief Create a new transaction object used to do IP address conversions
+     * \return The newly created transaction object
+     */
```
mdavidsaver, line 144:
```
\note Caller must release() the returned transaction
```

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
brendanchandler: Updated the example so the callback class manages the transaction object's memory too.  Also added the epicsEvent synchronization rather than the sleep() call.

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
brendanchandler: fixed. nice catch.

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
brendanchandler: Added note, good suggestion.

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
brendanchandler: added note

modules/libcom/src/misc/ipAddrToAsciiAsynchronous.h
brendanchandler: added note

anjohnson:
Rebased and merged at 35f4d3f8dd6780af48c8670ac9d498582bb88b73

Thanks!

PR #267, opened 2022-05-13
Fix lp #1858467

anjohnson:
Fix [this launchpad bug](https://bugs.launchpad.net/epics-base/+bug/1858467) by collecting the record names separately in a list. Sorting them before output is optional, `-s` suppresses the sort.


modules/database/src/tools/DBD/Output.pm
```diff
@@ -42,12 +42,13 @@ sub OutputDBD {
...
     my ($out, $dbd) = @_;
-    OutputRecords($out, $dbd->records);
+    OutputRecords($out, $dbd);
 }
 
 sub OutputMenus {
     my ($out, $menus) = @_;
-    while (my ($name, $menu) = each %{$menus}) {
+    foreach my $name (sort keys %{$menus}) {
+        my $menu = $menus->{$name};
```
mdavidsaver, line 51: As you know, my perl is sketchy at best.  Is this iterating a hash after sorting?

Does this mean that `-s` effects only record instances?  (as well as COMMENT and POD)

anjohnson: Yes — extract a list of the keys, sort that and then iterate over it. It would be possible to create an array of object references sorted by their `->name` member, but that's a little bit more complicated and probably not worth it given that we have only 30 menus in total (34 recordtypes is the largest collection in Base). The `-s` flag to the `dbExpand.pl` script now disables sorting, so we sort the expanded list of records by default. Note that `dbdExpand.pl` is a different program for expanding DBD files, and now those objects are always in sorted order.

anjohnson: DBD files are always sorted, DB files it's optional.

Approved.

mdavidsaver:
Merged as of abe819b51d078aba02589a043b6a02d688f36cad

PR #269, opened 2022-05-23
Set `RSRV_SERVER_PORT`

mdavidsaver:
Publish the actual TCP port used as `$RSRV_SERVER_PORT`.

Note that RSRV is, and was, only capable of making use of a single TCP port number.  With multiple interfaces this same port must be bound on all.

https://github.com/epics-base/epics-base/blob/3fadf4a26cfe33dcb9eb9e4620634e1d3a7b9763/modules/database/src/ioc/rsrv/caservertask.c#L143-L150

@shroffk fyi.

### DIFF
--- a/modules/database/src/ioc/rsrv/caservertask.c
+++ b/modules/database/src/ioc/rsrv/caservertask.c
@@ -572,6 +572,7 @@ void rsrv_init (void)
 
     {
         unsigned short sport = ca_server_port;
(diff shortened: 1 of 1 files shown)


anjohnson:
Release notes entry please.

mdavidsaver:
Merged as 9c8aaa93cde6ff7c3ec03cccf51e79e482680712

@shroffk @kasemir fyi.

PR #270, opened 2022-05-24
Use SetThreadDescription(), if available, to set thread name

FreddieAkeroyd:
Check at runtime if new function is available, original exception based mechanism is still also called as the new mechanism also requires a recent visual studio to see thread names in debugger. 

The new function allows thread names to appear in crash dumps and also doesn't require  a debugger to be attached prior to thread creation to see named threads.


anjohnson:
Core meeting July-6 Approved; we hope it merges up to 7.0 cleanly.

PR #272, opened 2022-06-14
db: move db_field_log dtor out of union

mdavidsaver:
wrt. [an issue](https://epics.anl.gov/tech-talk/2022/msg00831.php) reported by @MarkRivers.  The "bug" turns out to be in the MMU configuration of a vxWorks BSP.  Still, I take away the idea that the current definition of `db_field_log`, where the included union field overlays a float/int with a function pointer, is a bad idea as it allows speculative execution to cross a "boundary" between data and code.  This escalated a mis-configuration in a way which was very difficult to troubleshoot, and could have been triggered remotely.

This PR moves the function pointer out of the union.  This will add 4/8 bytes to `sizeof(db_field_log)`.  More significantly, it would break any out-of-tree filter code which is accessing the `dtor` field.

Would this change be worth the trouble?

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -731,7 +731,7 @@ static db_field_log* db_create_field_log (struct dbChannel *chan, int use_val)
             /* don't make a copy yet, just reference the field value */
             pLog->u.r.field = dbChannelField(chan);
             /* indicate field value still owned by record */
-            pLog->u.r.dtor = NULL;
+            pLog->dtor = NULL;
             /* no private data yet, may be set by a filter */
             pLog->u.r.pvt = NULL;
         }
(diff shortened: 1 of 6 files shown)


mdavidsaver:
In addition to powerpc.  From some wider reading, I think this sort of issue could also turn up on ARM7 where it is [said](https://developer.arm.com/documentation/den0024/a/Memory-Ordering/Memory-types/Device-memory) that:

> Trying to execute code from a region marked as Device, is generally unpredictable.
> The implementation might either handle the instruction fetch as if it were to a memory
> location with the Normal non-cacheable attribute, or it might take a permission fault.

I don't think it would be a problem with x86.  At least based on reading the Intel docs.

anjohnson:
Core group 7/6: Approved, with Release Notes entry.

mdavidsaver:
Applied as 9048aa34386bc3d7c9b4e42acfdf5b5ace951782

PR #273, opened 2022-06-16
Add space to the message printed by errMessage()

exzombie:
The way `errMessage()` is used throughout epics-base, a space and a newline should be added around the message provided by the caller. Here is an example from `dbStaticLib.c`:

    errMessage(-1, "dbCreateAlias: Add to PVD failed");

When called this way, the message sticks to the line number on the left and to the next log entry to the right. It's a sensible way to use this macro, so instead of fixing the call sites, I propose fixing the macro instead.

### DIFF
--- a/modules/database/src/ioc/as/asCa.c
+++ b/modules/database/src/ioc/as/asCa.c
@@ -241,7 +241,7 @@ void asCaStart(void)
                                      epicsThreadGetStackSize(epicsThreadStackBig),
                                      (EPICSTHREADFUNC)asCaTask,0);
         if(threadid==0) {
-            errMessage(0,"asCaStart: taskSpawn Failure\n");
+            errMessage(0,"asCaStart: taskSpawn Failure");
         }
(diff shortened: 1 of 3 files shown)


anjohnson:
I agree that this will make for better log output. Could you also adjust the description of the message parameter in the Doxygen annotation immediately above the macro to say that it should not end with a newline.

Less important if you don't have time to fix them as well, there are 2 calls to `errMessage()` in Base that have what will become an extraneous newline at the end:
```
modules/database/src/ioc/as/asCa.c:            errMessage(0,"asCaStart: taskSpawn Failure\n");
modules/libcom/src/osi/os/posix/osdMonotonic.c:    errMessage(errlogMinor, "Warning: failed to setup monotonic time source\n");
```

Thanks!

mdavidsaver:
Applied as e6b8ecd55e8c93a37d1ef63e95abcff17186b0c3

PR #274, opened 2022-06-20
iocsh tab completion w/ readline

mdavidsaver:
Adds tab completion for command names, record names, "pdbbase", and filesystem paths to iocsh.  Only when built against libreadline.

iocsh-complete

(animation captured with [`apt-get install ttyrec`](https://tracker.debian.org/pkg/ttyrec))

The only externally visible changes are the addition of `iocshArgStringRecord` and `iocshArgStringPath`, which are functionally identical to `iocshArgString`.  The difference is the hint provided to tab completion logic.

TODO: Completion currently only works for shell-like syntax (space separated), and not for C-like (`(,)` separated).  This is a consequence of my re-packaging, but not re-writing, the argument splitting code.


mdavidsaver:
@anjohnson Suggests completion for iocshArgPdbbase since this is trivial.

mdavidsaver:
Updated to include completion for iocshArgPdbbase

mdavidsaver:
@anjohnson What is the name of the GNU readline alternative which Apple ships?

```
... use of undeclared identifier 'rl_basic_quote_characters'; ...
```

anjohnson:
[libedit(3)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/editline.3.html) has a readline-compatibility mode. There are sources [here](https://opensource.apple.com/source/libedit/libedit-39/), see the src/editline directory for the readline.h header which has the same version number as all 6 copies shipped by Apple in the SDKs installed on my laptop (and are all identical).

mdavidsaver:
That's friendly...  I see that [a version](http://cvsweb.netbsd.org/bsdweb.cgi/src/lib/libedit/readline/readline.h?annotate=1.53&sortby=date) adds `rl_basic_quote_characters`, but still doesn't change `RL_READLINE_VERSION`.

While GNU readline 4.2 has `rl_basic_quote_characters`, it is [ancient](https://ftp.gnu.org/gnu/readline/).  So I'll assume that `RL_READLINE_VERSION <= 0x402` is really libedit.

mdavidsaver:
Now that R7.0.7 is out, I'd like to see this PR considered.  In my own usage, I'm getting addicted to not typing or copy+pasting long PV names in the shell.

anjohnson:
Core Group 2022-10-28: Approved.

mdavidsaver:
Applied as of d9ca8a70f05bc61a6e4dfbe1518da8ce4f515651

anjohnson:
Got this to work on macOS by adjusting the build to point to Homebrew's install of GNU readline. See 1f75813a4 which enables that automatically if installed.

mdavidsaver:
cf. https://bugs.launchpad.net/epics-base/+bug/2004463

The regression in https://github.com/epics-base/epics-base/pull/274 is the change from

> int quote, inword, backslash;

to

> bool inword = false;
> bool backslash = false;
> char quote = EOF;

Which in retrospect is an issue because the `EOF` macro is meant to be stored/compared with `int` (eg. with `getc()` and friends).

PR #276, opened 2022-07-01
Fix potentially buggy strncpy calls and add epicsStrnCpy

JJL772:
In a few places, strings were not explicitly null terminated following calls to `strncpy`, which doesn't always null terminate the destination buffer.

libcom's string handling library is a bit lacking, so I also added `epicsStrnCpy`. It's identical to `strncpy` except that it always null terminates the destination buffer for a bit of added safety.

For C++11 and higher, `epicsStrnCpy` is also provided as a template function that can deduce the input buffer size if working with C arrays. Example:
```
char buffer[64];
epicsStrnCpy(buffer, "my cool string");
```

### DIFF
--- a/modules/database/src/ioc/db/dbDbLink.c
+++ b/modules/database/src/ioc/db/dbDbLink.c
@@ -337,6 +337,7 @@ static long dbDbGetUnits(const struct link *plink, char *units, int unitsSize)
         return status;
 
     strncpy(units, buffer.units, unitsSize);
+    units[unitsSize-1] = 0;
     return 0;
 }
 
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -442,3 +442,11 @@ done:
     free(dist1);
     return ret;
 }
+
+/* Identical to strncpy but ensures dest is terminated */
+char* epicsStrnCpy(char* dest, const char* src, size_t destSize)
+{
+    char* ret = strncpy(dest, src, destSize);
+    dest[destSize-1] = 0;
+    return ret;
+}
--- a/modules/libcom/src/misc/epicsString.h
+++ b/modules/libcom/src/misc/epicsString.h
@@ -37,6 +37,9 @@ LIBCOM_API int epicsStrPrintEscaped(FILE *fp, const char *s, size_t n);
 #define epicsStrSnPrintEscaped epicsStrnEscapedFromRaw
 LIBCOM_API size_t epicsStrnLen(const char *s, size_t maxlen);
 
+/* Behaves identically to strncpy except that it always NULL terminates dest */
+LIBCOM_API char* epicsStrnCpy(char* dest, const char* src, size_t destSize);
+
 /** Matches a string against a pattern.
  *
  * Checks if str matches the glob style pattern, which may contain ? or * wildcards.
@@ -80,6 +83,15 @@ LIBCOM_API int dbTranslateEscape(char *s, const char *ct);
 
 #ifdef __cplusplus
 }
+#endif
+
+#if __cplusplus >= 201103L
+
+template<size_t N>
+static inline char* epicsStrnCpy(char(&dest)[N], const char* src) {
+    return epicsStrnCpy(dest, src, N);
+}
+
 #endif
 
 #endif /* INC_epicsString_H */
--- a/modules/database/src/ioc/db/dbConvert.c
+++ b/modules/database/src/ioc/db/dbConvert.c
@@ -881,6 +881,7 @@ static long getMenuString(const dbAddr *paddr,
         return(S_db_badChoice);
     }
     strncpy(pdst, pchoice, MAX_STRING_SIZE);
+    pdst[MAX_STRING_SIZE-1] = 0;
     return 0;
 }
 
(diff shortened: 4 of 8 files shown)


modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -337,6 +337,7 @@ static long dbDbGetUnits(const struct link *plink, char *units, int unitsSize)
         return status;
 
     strncpy(units, buffer.units, unitsSize);
+    units[unitsSize-1] = 0;
```
mdavidsaver, line 340: Here, and elsewhere, I think the size==0 case needs to be handled when the size is not a compile time constant.

modules/libcom/src/misc/epicsString.c
```diff
@@ -442,3 +442,11 @@ double epicsStrSimilarity(const char *A, const char *B)
     free(dist1);
     return ret;
 }
+
+/* Identical to strncpy but ensures dest is terminated */
+char* epicsStrnCpy(char* dest, const char* src, size_t destSize)
+{
+    char* ret = strncpy(dest, src, destSize);
+    dest[destSize-1] = 0;
+    return ret;
```
mdavidsaver, line 451: With an added check for `destSize==0`, your `epicsStrnCpy()` is equivalent to [`strlcpy()`](https://linux.die.net/man/3/strlcpy), which has been [discussed](https://github.com/epics-base/epics-base/pull/124) previously.

I'm somewhat reluctant to go with defining our own str*() functions. This would quiet GCC's static analysis warnings by circumventing whatever hooks GCC is using (eg. the [`nonnull`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes) attribute).  This seems undesirable as the long term trend is for the SNR of these warnings to improve.

I would add to this the (probably slight) loss in potential for optimization.  For example, an out of line definition prevents usage of a compiler [builtin](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) in cases where `destSize` is a constant.

Also, if this function is to be added to libCom, it needs to be used in Base.

modules/libcom/src/misc/epicsString.c
```diff
@@ -442,3 +442,11 @@ double epicsStrSimilarity(const char *A, const char *B)
...
+}
```
mdavidsaver, line 452: Please keep the trailing newline.

modules/libcom/src/misc/epicsString.h
```diff
@@ -82,4 +85,13 @@ LIBCOM_API int dbTranslateEscape(char *s, const char *ct);
 }
 #endif
 
+#if __cplusplus >= 201103L
+
+template<size_t N>
+static inline char* epicsStrnCpy(char(&dest)[N], const char* src) {
+    return epicsStrnCpy(dest, src, N);
+}
```
mdavidsaver, line 93: This seems like it might incentivize bad code (C string manipulation in C++).

Also, it should be used somewhere in Base.  This is especially important for a C++ template to prove that it can actually be applied.

mdavidsaver (changes requested):

anjohnson:
I haven't worked out exactly where, but I have concerns that some of these changes could result in writing past the end of an allocated string buffer. In particular there is a known problem with the IOC's APIs for converting between data types (dbConvert.c and dbFastLinkConv.c), some of those conversion cases pass a buffer which can be shorter than `MAX_STRING_SIZE` through a generic API which has no way to can pass the buffer length. That code mustn't assume all string buffers are at least `MAX_STRING_SIZE` bytes long, but as long as the source string is shorter than the buffer size the code works Okay. We wouldn't design an API like that nowadays, but this is pretty ancient code (mid to late 1990s, before I was involved at least). Changing the API is difficult because if can affect out-of-tree code.

There is at least [one bug report](https://bugs.launchpad.net/epics-base/+bug/1918762) about that kind of issue, although it isn't necessarily the same one as here.

modules/libcom/src/misc/epicsString.h
JJL772: Agreed that it's a good idea to use it in base to ensure it works (if we're going to keep this function that is!)

I'm not sure it incentivizes bad code though. `std::string` allocates for sufficiently large strings and you might want to use traditional C strings (or `std::array`) to avoid that.

modules/libcom/src/misc/epicsString.c
JJL772: You're right, this does look almost equivalent to `strlcpy` (aside from return value), so I'll remove this. Should I define a `epicsStrlCpy` in this branch or let that other PR handle it?

> I'm somewhat reluctant to go with defining our own str*() functions.

My plan here was to only provide a slightly safer alternative `strncpy` because of its odd, occasionally unsafe behavior. The rest of the C string library is a bit less error-prone, so defining `epicsStr*` wouldn't really be necessary.

anjohnson:
Core group 20-July: If you want to continue with this we suggest running the tests under valgrind, which might help find the sources of the failures. We obviously can't merge this until the tests pass, and aren't sure whether it's worth the effort.

anjohnson:
[LWN Article](https://lwn.net/SubscriberLink/905777/b315fa3a18205bf7/) about the
```C
ssize_t strscpy(char *dest, const char *src, size_t count);
```
function which is likely to be replacing many calls to `strlcpy()` in the Linux Kernel. The `strlcpy()` function has to scan the whole of the source string to calculate its `strlen(src)` return value, even when the destination buffer is shorter.

`strscpy()` always Nil-terminates, and returns the number of characters it copied (excluding the Nil), or `-E2BIG` if the source won't all fit, hence the `ssize_t` return type.

anjohnson:
Core Group 2022-10-28: No progress since last review, marking as draft.

PR #277, opened 2022-07-12
Adding new epicsThreadCreateJoinable function

jwlodek:
Adding a simple utility function for shorthand creation of joinable threads. Just as with `epicsThreadCreate` I think it's easier to remember usage for this rather than initializing the opts manually for `epicsThreadCtreateOpt`

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -16,6 +16,10 @@ should also be read to understand what has changed since earlier releases.
 
 <!-- Insert new items immediately below here ... -->
 
+### Add `epicsThreadCreateJoinable` function
+
+Adds a function to the libcom API for shorthand creation of joinable threads, without the use of `epicsThreadCreateOpt`.
+
 ### Perl CA support for empty long strings
 
 The Perl CA bindings have been fixed to handle zero-length long string data
--- a/modules/database/src/ioc/db/dbCa.c
+++ b/modules/database/src/ioc/db/dbCa.c
@@ -319,11 +319,6 @@ void dbCaShutdown(void)
 
 static void dbCaLinkInitImpl(int isolate)
 {
-    epicsThreadOpts opts = EPICS_THREAD_OPTS_INIT;
-
-    opts.stackSize = epicsThreadGetStackSize(epicsThreadStackBig);
-    opts.priority = epicsThreadPriorityMedium;
-    opts.joinable = 1;
 
     dbServiceIsolate = isolate;
     dbServiceIOInit();
@@ -337,7 +332,9 @@ static void dbCaLinkInitImpl(int isolate)
         startStopEvent = epicsEventMustCreate(epicsEventEmpty);
     dbCaCtl = ctlPause;
 
-    dbCaWorker = epicsThreadCreateOpt("dbCaLink", dbCaTask, NULL, &opts);
+    dbCaWorker = epicsThreadCreateJoinable("dbCaLink", epicsThreadPriorityMedium,
+                        epicsThreadStackBig, dbCaTask, NULL);
+    
     /* wait for worker to startup and initialize dbCaClientContext */
     epicsEventMustWait(startStopEvent);
 }
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -1131,11 +1131,6 @@ int db_start_events (
     void *init_func_arg, unsigned osiPriority )
 {
      struct event_user * const evUser = (struct event_user *) ctx;
-     epicsThreadOpts opts = EPICS_THREAD_OPTS_INIT;
-
-     opts.stackSize = epicsThreadGetStackSize(epicsThreadStackMedium);
-     opts.priority = osiPriority;
-     opts.joinable = 1;
 
      epicsMutexMustLock ( evUser->lock );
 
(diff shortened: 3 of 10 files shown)


jwlodek:
I'm not sure why the one CI test failed, it doesn't seem like it should be related to the PR

anjohnson:
There is a known [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug) in `testpvalink.tap` so you don't need to worry about that failure. The Core group will review and comment on this PR soon, either individually or at one of our regular every-other-Wednesday-morning meetings.

I would suggest adding or modifying a test to use it in the `modules/libcom/test/epicsThreadTest.cpp` program — it may seem almost too trivial to test, but more automated tests help us catch issues faster. Are there any other places in Base where using this could reduce how much code we actually ship? Don't bother to look at the PVA modules though, those still have to build against Base-3.15 releases which wouldn't include this new API. We will also want an entry about it in the Release Notes.

Thanks!

jwlodek:
Thanks for the suggestions! I added an extra test to the `epicsThreadTest` suite, and simplified instances of uses of `epicsThreadCreateOpt` to the new `epicsThreadCreateJoinable` where applicable. Also added a note to the `RELEASE_NOTES`. All the tests seem to be executing the same for me as for a fresh compile of the `7.0` branch, though in both cases I am seeing:

```
osiSockTest.t ................. 19/24 
not ok 20 - Successes 0
osiSockTest.t ................. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/24 subtests 
```

mdavidsaver:
@jwlodek Please rebase instead of merging.  Merge commits complicate review and acceptance.

mdavidsaver:
My [motivation](https://code.launchpad.net/~epics-core/epics-base/+git/Com/+merge/361379) for the more verbose `epicsThreadCtreateOpt()` is to allow for new (abstracted) options to be introduced in the future without a proliferation of `epicsThreadCreate*` variants.  (eg. CPU affinity)  However, since 2019 no more have been proposed.

So I guess the questions are whether the concern about proliferation still remains?  Also, are there other ways to simplify usage?  eg. make `EPICS_THREAD_OPTS_INIT` equivalent to zero initialization.

mdavidsaver:
> ... Don't bother to look at the PVA modules ...

At this point ~all usage in the PVA modules is through the c++ `class epicsThread`, which joins automatically.

jwlodek:
While I can see the issue with adding `epicsThreadCreate*` functions as other options are added - my main logic here is that at least in my experience I only ever used `epicsThreadCreate`, or `epicsThreadCreateOpts` when I needed a joinable thread, and I more frequently use joinable threads since I often want to guarantee DAQ threads relinquish control over a device before it is deleted, so a shorthand is just a nice QOL feature. Most of the time, defaults for all options is good enough, and I only really want to specify whether the thread should be joinable or not.

In that case I think if we leave `epicsThreadCreateOpt` as it is, and have `EPICS_THREAD_OPTS_INIT` have sensible defaults for any options added down the line (for when the shorthands are called), I think it should be OK.

mdavidsaver:
> eg. make `EPICS_THREAD_OPTS_INIT` equivalent to zero initialization.

... to enable:

```c
epicsThreadOpts opts {.joinable = 1};
```

It looks like even [ancient](https://gcc.gnu.org/onlinedocs/gcc-2.95.3/gcc_4.html#SEC80) GCC supported this syntax, which has become standard with [C99](https://en.cppreference.com/w/c/language/struct_initialization) and [C++20](https://en.cppreference.com/w/cpp/language/aggregate_initialization).  So it looks like the only obstacle to our using this syntax in Base is MSVC.  I can't find a [statement](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance) on when (or if) the MS _C_ compiler started accepting this syntax.

anjohnson:
Core group July-20: @FreddieAkeroyd can you comment on Michael's question about MSVC support for struct initialization ("designated initializer" perhaps?) in [this comment above](https://github.com/epics-base/epics-base/pull/277#issuecomment-1185772295) please?

mdavidsaver:
> ```c
> epicsThreadOpts opts {.joinable = 1};
> ```

I realize that I am confusing this C++ constructor syntax with the similar C syntax, which has an additional `=`.

```c
epicsThreadOpts opts = {.joinable = 1};
```

This C syntax is accepted by default in C code (but not c++ code) by the latest version of all supported compilers, and historically with gcc/clang.  Older MSVC is still an open question.

So we probably can't use this syntax in Base without dropping support for older MSVC (which imo. will happen at some point).

details...

Remembering [godbolt](https://godbolt.org/) I find that the C designated initializer syntax is accepted by default with the latest clang, gcc, and msvc.  Although in the case of MSVC, only in C code.  With vs2019 in c++ code this gives:

```
error C7555: use of designated initializers requires at least '/std:c++20
```

Further, G++ < 8 also rejects this with:

```
... sorry, unimplemented: non-trivial designated initializers not supported
```

anjohnson:
Core Group 2022-10-28: We don't really see this as enough of a benefit to the code-base. The `epicsThreadOpts` structure was only added a few years ago with the idea of allowing additional parameters to be added in future without requiring API changes, and this function contradicts that aim.

Thanks for making the suggestion, we'd be happy to see other ideas in the future.

PR #278, opened 2022-07-26
JSON error messages

dirk-zimoch:
Fix for https://bugs.launchpad.net/epics-base/+bug/1982847

### DIFF
--- a/modules/database/src/ioc/db/dbConvertJSON.c
+++ b/modules/database/src/ioc/db/dbConvertJSON.c
@@ -30,10 +30,12 @@ typedef struct parseContext {
 } parseContext;
 
 static int dbcj_null(void *ctx) {
+    errlogPrintf("dbConvertJSON: Null objects not supported\n");
     return 0;    /* Illegal */
 }
 
 static int dbcj_boolean(void *ctx, int val) {
+    errlogPrintf("dbConvertJSON: Boolean not supported\n");
     return 0;    /* Illegal */
 }
 
@@ -51,6 +53,7 @@ static int dbcj_integer(void *ctx, long long num) {
 }
 
 static int dblsj_integer(void *ctx, long long num) {
+    errlogPrintf("dbLSConvertJSON: Numeric value %lli provided, string expected\n", num);
     return 0;    /* Illegal */
 }
 
@@ -67,6 +70,7 @@ static int dbcj_double(void *ctx, double num) {
 }
 
 static int dblsj_double(void *ctx, double num) {
+    errlogPrintf("dbLSConvertJSON: Numeric value %g provided, string expected\n", num);
     return 0;    /* Illegal */
 }
 
@@ -78,7 +82,8 @@ static int dbcj_string(void *ctx, const unsigned char *val, size_t len) {
      * metadata about the field than we have available at the moment.
      */
     if (parser->dbrType != DBF_STRING) {
-        errlogPrintf("dbConvertJSON: String provided, numeric value(s) expected\n");
+        errlogPrintf("dbConvertJSON: String \"%.*s\" provided, numeric value expected\n",
+            (int)len, val);
         return 0; /* Illegal */
     }
 
@@ -159,8 +164,10 @@ long dbPutConvertJSON(const char *json, short dbrType,
     size_t jlen = strlen(json);
     long status;
 
-    if(INVALID_DB_REQ(dbrType))
+    if (INVALID_DB_REQ(dbrType)) {
+        errlogPrintf("dbConvertJSON: Invalid dbrType %d\n", dbrType);
         return S_db_badDbrtype;
+    }
 
     if (!jlen) {
         *pnRequest = 0;
@@ -175,8 +182,10 @@ long dbPutConvertJSON(const char *json, short dbrType,
 
     yajl_set_default_alloc_funcs(&dbcj_alloc);
     yh = yajl_alloc(&dbcj_callbacks, &dbcj_alloc, parser);
-    if (!yh)
+    if (!yh) {
+        errlogPrintf("dbConvertJSON: out of memory\n");
         return S_db_noMemory;
+    }
 
     ys = yajl_parse(yh, (const unsigned char *) json, jlen);
     if (ys == yajl_status_ok)
@@ -188,15 +197,13 @@ long dbPutConvertJSON(const char *json, short dbrType,
         status = 0;
         break;
 
-    case yajl_status_error: {
-        unsigned char *err = yajl_get_error(yh, 1,
-            (const unsigned char *) json, jlen);
-        errlogPrintf("dbConvertJSON: %s\n", err);
-        yajl_free_error(yh, err);
+    default: {
+            unsigned char *err = yajl_get_error(yh, 1,
+                (const unsigned char *) json, jlen);
+            errlogPrintf("dbConvertJSON: %s", err);
+            yajl_free_error(yh, err);
+            status = S_db_badField;
         }
-        /* fall through */
-    default:
-        status = S_db_badField;
     }
 
     yajl_free(yh);
@@ -233,8 +240,10 @@ long dbLSConvertJSON(const char *json, char *pdest, epicsUInt32 size,
 
     yajl_set_default_alloc_funcs(&dbcj_alloc);
     yh = yajl_alloc(&dblsj_callbacks, &dbcj_alloc, parser);
-    if (!yh)
+    if (!yh) {
+        errlogPrintf("dbLSConvertJSON: out of memory\n");
         return S_db_noMemory;
+    }
 
     ys = yajl_parse(yh, (const unsigned char *) json, jlen);
 
@@ -244,15 +253,13 @@ long dbLSConvertJSON(const char *json, char *pdest, epicsUInt32 size,
         status = 0;
         break;
 
-    case yajl_status_error: {
-        unsigned char *err = yajl_get_error(yh, 1,
-            (const unsigned char *) json, jlen);
-        errlogPrintf("dbLoadLS_JSON: %s\n", err);
-        yajl_free_error(yh, err);
+    default: {
+            unsigned char *err = yajl_get_error(yh, 1,
+                (const unsigned char *) json, jlen);
+            errlogPrintf("dbLSConvertJSON: %s", err);
+            yajl_free_error(yh, err);
+            status = S_db_badField;
         }
-        /* fall through */
-    default:
-        status = S_db_badField;
     }
 
     yajl_free(yh);
--- a/modules/database/src/ioc/db/dbConvertJSON.c
+++ b/modules/database/src/ioc/db/dbConvertJSON.c
@@ -162,6 +162,11 @@ long dbPutConvertJSON(const char *json, short dbrType,
     if(INVALID_DB_REQ(dbrType))
         return S_db_badDbrtype;
 
+    if (!jlen) {
+        *pnRequest = 0;
+        return 0;
+    }
+
     parser->depth = 0;
     parser->dbrType = dbrType;
     parser->dbrSize = dbValueSize(dbrType);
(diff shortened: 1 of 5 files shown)


modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -183,73 +157,31 @@ long dbPutConvertJSON(const char *json, short dbrType,
...
-        yajl_free_error(yh, err);
-        }
-        /* fall through */
-    default:
-        status = S_db_badField;
-    }
-
-    yajl_free(yh);
+    status = dbPutConvertJSON(json, DBF_CHAR, pdest, &nRequest);
+    *plen = nRequest;
```
anjohnson: The original version only set `*plen` when there was no conversion error.

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -118,14 +100,6 @@ static int dbcj_start_map(void *ctx) {
...
 }
 
-static int dbcj_map_key(void *ctx, const unsigned char *key, size_t len) {
-    return 0;    /* Illegal */
-}
-
-static int dbcj_end_map(void *ctx) {
-    return 0;    /* Illegal */
-}
-
```
dirk-zimoch, line 128: We never get here, because start_map already fails.

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -135,43 +109,42 @@ static int dbcj_start_array(void *ctx) {
     return (parser->depth == 1);
 }
 
-static int dbcj_end_array(void *ctx) {
-    parseContext *parser = (parseContext *) ctx;
-
-    parser->depth--;
-    return (parser->depth == 0);
-}
-
```
dirk-zimoch, line 144: Nothing to do because we do not support nested arrays.

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -66,39 +64,23 @@ static int dbcj_double(void *ctx, double num) {
...
-    return 0;    /* Illegal */
-}
-
 static int dbcj_string(void *ctx, const unsigned char *val, size_t len) {
     parseContext *parser = (parseContext *) ctx;
     char *pdest = parser->pdest;
 
-    /* Not attempting to handle char-array fields here, they need more
-     * metadata about the field than we have available at the moment.
-     */
```
dirk-zimoch, line 79: We do have all the metadata needed: parser->dbrType and parser->elems

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -183,73 +156,32 @@ long dbPutConvertJSON(const char *json, short dbrType,
...
-static yajl_callbacks dblsj_callbacks = {
-    dbcj_null, dbcj_boolean, dblsj_integer, dblsj_double, NULL, dblsj_string,
-    dbcj_start_map, dbcj_map_key, dbcj_end_map,
-    dbcj_start_array, dbcj_end_array
-};
-
 long dbLSConvertJSON(const char *json, char *pdest, epicsUInt32 size,
     epicsUInt32 *plen)
 {
-    parseContext context, *parser = &context;
```
dirk-zimoch, line 211: Large parts of code were shared with dbPutConvertJSON() anyway. Differences are minor.

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -135,43 +109,42 @@ static int dbcj_start_array(void *ctx) {
...
     parser->depth = 0;
     parser->dbrType = dbrType;
     parser->dbrSize = dbValueSize(dbrType);
     parser->pdest = pdest;
     parser->elems = *pnRequest;
 
-    yajl_set_default_alloc_funcs(&dbcj_alloc);
-    yh = yajl_alloc(&dbcj_callbacks, &dbcj_alloc, parser);
-    if (!yh)
+    yh = yajl_alloc(&dbcj_callbacks, NULL, parser);
```
dirk-zimoch, line 157: Passing NULL for alloc_funcs will call yajl_set_default_alloc_funcs() inside yajl_alloc() anyway.

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -183,73 +156,32 @@ long dbPutConvertJSON(const char *json, short dbrType,
...
+    default: {
+            unsigned char *err = yajl_get_error(yh, 1,
+                (const unsigned char *) json, jlen);
+            errlogPrintf("dbConvertJSON: %s", err);
+            yajl_free_error(yh, err);
+            status = S_db_badField;
         }
-        /* fall through */
-    default:
-        status = S_db_badField;
```
dirk-zimoch, line 194: Print message on any error.

modules/database/src/ioc/db/dbConvertJSON.c
dirk-zimoch: Array elements will be concatenated (optionally mixed with numeric single bytes) into one long string. This allows to split a long constant string  into multiple lines in the template file.

modules/database/test/std/rec/linkInitTest.c
```diff
@@ -65,7 +65,7 @@ static void testLongStringInit()
         testdbGetFieldEqual("longstr3.VAL", DBR_STRING, "!--------------------------------------");
     }
 
-    testdbGetFieldEqual("longstr4.VAL", DBR_STRING, "One");
+    testdbGetFieldEqual("longstr4.VAL", DBR_STRING, "OneTwoThreeFour");
```
dirk-zimoch: Using the new feature of concatenating string elements in array.

modules/database/src/ioc/db/dbConvertJSON.c
anjohnson: Concatenating a list of strings seems like a reasonable idea, although the YAJL parser already removes a backslash-newline from its input strings, that was added by the JSON-5 parsing enhancements. There is a problem with concatenating though, load and look at the `VAL` fields of these two records:
```
record(stringin, "not-joined") {
  field(INP, ["One", "Two"])
}
record(waveform, "reindeer") {
  field(FTVL, "STRING")
  field(NELM, "20")
  field(INP, ["Dasher", "Dancer", "Prancer", "Vixen",
              "Comet", "Cupid", "Donner", "Blitzen", "Rudolph"])
}
```
If you fix the code so the first one works as you might expect, you're likely to break the second one (which still works with your code at the moment, initializing 9 string elements in the waveform).

I also don't like it accepting numbers as raw byte values, some users might think those numbers should be Unicode code points instead, which I could see as an equally valid possibility. Given that ambiguity I don't really want to allow numbers in the array at all (especially not those with decimal points, they should be rejected). Users can include escaped characters with the usual `\x` or `\u` notations (although `\u` doesn't work in a `dbpf` command because iocsh preprocesses the command line before it calls the `dbpf` command and whatever it uses doesn't accept `\u` escapes).

modules/database/src/ioc/db/dbConvertJSON.c
```diff
@@ -64,19 +64,21 @@ static int dbcj_double(void *ctx, double num) {
...
-
 static int dbcj_string(void *ctx, const unsigned char *val, size_t len) {
     parseContext *parser = (parseContext *) ctx;
     char *pdest = parser->pdest;
 
-    /* Not attempting to handle char-array fields here, they need more
-     * metadata about the field than we have available at the moment.
-     */
+    if (parser->dbrType == DBF_CHAR || parser->dbrType == DBF_UCHAR) {
+        /* Treating char array as long string */
```
anjohnson: Can you drop the `DBF_UCHAR` from the above test, I want to keep it so that only `DBF_CHAR` arrays are handled as long strings — a `DBF_UCHAR` array should be treated by the IOC as an array of 8-bit unsigned integers. CA makes the distinction fairly invisible since all long-string requests from CA will use `DBR_CHAR` even when the field itself is unsigned, but inside the IOC I would like to aim for this where we can. I came to that conclusion long after adding the original long-string support though, so it isn't universal in the existing code and I might not have actually told anyone else about it yet either.

modules/database/test/std/rec/linkInitTest.c
anjohnson: I don't seem to have written any test code for using a constant initializer for an array of `DBF_STRING` values, although I have checked it by hand before and I can confirm that it still works with this PR. If you find a fix for the concatenation problem I described above, please add something like the record below as a unit test to ensure that we don't break the ability to initialize a waveform like this:
```
record(waveform, "reindeer") {
  field(FTVL, "STRING")
  field(NELM, "20")
  field(INP, ["Dasher", "Dancer", "Prancer", "Vixen",
              "Comet", "Cupid", "Donner", "Blitzen", "Rudolph"])
  # Should initialize the record with 9 elements
}
```

documentation/RELEASE_NOTES.md
```diff
@@ -16,6 +16,14 @@ should also be read to understand what has changed since earlier releases.
 
 <!-- Insert new items immediately below here ... -->
 
+### JSON links for CHAR arrays (long strings)
+
+Record types that support arrays of `CHAR` or `UCHAR` like `waveform`, `aai`
+or `lsi` can now use JSON arrays in their constant input links which consist
+of multiple strings, optionally mixed with numeric values. The individual
+strings will be concatenated and the numeric values will be interpreted as
+byte codes.
```
anjohnson: As I said elsewhere I don't like allowing numbers to be included in an array of strings, if that gets removed this Release Note entry would also need adjusting.

modules/database/src/ioc/db/dbConvertJSON.c
anjohnson: True, I think I only learned that after writing this code.

anjohnson (changes requested):
If you'd stuck with just fixing the JSON error messages I would have approved this, but with the concatenation there's a problem that I'm not sure can be resolved. Details below.

modules/database/src/ioc/db/dbConvertJSON.c
dirk-zimoch: The RELEASE_NOTES message says that concatenation works on CHAR arrays, not that it would work on STRING (where due to the limited STRING size line breaks are not so pressing). I could either modify the text so that is it more explicit about what the change does NOT do or I can implement concatenation on STRING scalars, but not STRING arrays (which I find even more confusing; what about STRING[1] arrays?)

I understand your hesitation concerning raw bytes in long strings. Implementing numbers as Unicode code points would be possible as well, but with an ambiguity for code points 0x80-0xff. On the other hand, we do not have Unicode support in EPICS anywhere but in JSON strings (where is is enforced, which I don't like because historically, we still have many Latin-1 encoded strings.)

Of course I can disable accepting numbers in long strings (but not in CHAR waveforms, which may really mean 8 bit integer numbers) but I do not really see why, as it comes for free. Where does it really hurt?

Concerning floating point numbers: CHAR (or any other integer type) waveforms already now accept floating point values.

modules/database/test/std/rec/linkInitTest.c
dirk-zimoch: Sure.

modules/database/src/ioc/db/dbConvertJSON.c
dirk-zimoch: Will fix.

dirk-zimoch:
I will split the string concatenation off into a separate MR.

dirk-zimoch:
Force-pushed a scaled down merge request. All major modifications in dbConvertJSON.c removed. No concatenations.
Test for array of STRING added.

anjohnson:
Core Group 3-Aug: Looks good now, thanks!

mdavidsaver:
Merged as of c2df1c6fe5964b2ceec0d5088cf5436adad2c557

PR #281, opened 2022-08-05
Errlog fixes

dirk-zimoch:
Allow partial messages. This fixes a problem with logMsg on VxWorks when using iocLogPrefix. Because logMsg sends partial messages to errlog, each format specifier in the messages had the prefix prepended and appended.
See bug report https://bugs.launchpad.net/epics-base/+bug/1983385.
Now, the message is accumulated until a part ends in \n. The the whole message is then sent and optionally prefixed.
Also avoid usage of %z (for VxWorks), make types more consistent (for Windows).

### DIFF
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -71,6 +71,7 @@ typedef struct listenerNode{
 typedef struct {
     char *base;
     int pos;
+    int nchar;
 } buffer_t;
 
 static struct {
@@ -125,9 +126,14 @@ char* msgbufAlloc(void)
 
     errlogInit(0);
     epicsMutexMustLock(pvt.msgQueueLock); /* matched in msgbufCommit() */
-    if(pvt.bufSize - pvt.log->pos >= 1+pvt.maxMsgSize) {
+    if(pvt.bufSize - pvt.log->pos - pvt.log->nchar >= 1+pvt.maxMsgSize) {
         /* there is enough space for the worst cast */
         ret = pvt.log->base + pvt.log->pos;
+        if (pvt.log->nchar) {
+            /* append to last message */
+            return ret + 1 + pvt.log->nchar;
+        }
+        /* new message */
         ret[0] = ERL_STATE_WRITE;
         ret++;
     }
@@ -145,7 +151,7 @@ int msgbufCommit(int nchar, int localEcho)
     int isOkToBlock = epicsThreadIsOkToBlock();
     int wasEmpty = pvt.log->pos==0;
     int atExit = pvt.atExit;
-    char *start = pvt.log->base + pvt.log->pos;
+    char *start = pvt.log->base + pvt.log->pos + pvt.log->nchar;
 
     /* nchar returned by snprintf() is >= maxMsgSize when truncated */
     if(nchar >= pvt.maxMsgSize) {
@@ -164,15 +170,20 @@ int msgbufCommit(int nchar, int localEcho)
          */
         fprintf(pvt.console, "%s", start);
 
-    } else {
-        start[0] = ERL_STATE_READY | (localEcho ? ERL_LOCALECHO : 0);
+    } else if (start[nchar] != '\n') {
+        /* incomplete message, prepare to append */
+        pvt.log->nchar += nchar;
 
-        pvt.log->pos += 1 + nchar + 1;
+    } else {
+        pvt.log->base[pvt.log->pos] = ERL_STATE_READY | (localEcho ? ERL_LOCALECHO : 0);
+
+        pvt.log->pos += 1 + pvt.log->nchar + nchar + 1;
+        pvt.log->nchar = 0;
     }
 
     epicsMutexUnlock(pvt.msgQueueLock); /* matched in msgbufAlloc() */
 
-    if(wasEmpty && !atExit)
+    if(wasEmpty && !atExit && !pvt.log->nchar)
         epicsEventMustTrigger(pvt.waitForWork);
 
     if(localEcho && isOkToBlock && !atExit)
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -234,22 +234,19 @@ int isATTY(FILE* fp)
 static
 int isATTY(FILE* fp)
 {
+#ifdef ENABLE_VIRTUAL_TERMINAL_PROCESSING
     HANDLE hand = NULL;
     DWORD mode = 0;
     if(fp==stdout)
         hand = GetStdHandle(STD_OUTPUT_HANDLE);
     else if(fp==stderr)
         hand = GetStdHandle(STD_ERROR_HANDLE);
-#ifdef ENABLE_VIRTUAL_TERMINAL_PROCESSING
     if(hand && GetConsoleMode(hand, &mode)) {
         (void)SetConsoleMode(hand, mode|ENABLE_VIRTUAL_TERMINAL_PROCESSING);
         mode = 0;
         if(GetConsoleMode(hand, &mode) && (mode&ENABLE_VIRTUAL_TERMINAL_PROCESSING))
             return 1;
     }
-#else
-    (void)hand;
-    (void)mode;
 #endif
     return 0;
 }
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -70,14 +70,14 @@ typedef struct listenerNode{
 
 typedef struct {
     char *base;
-    size_t pos;
+    int pos;
 } buffer_t;
 
 static struct {
     /* const after errlogInit() */
-    size_t maxMsgSize;
+    int maxMsgSize;
     /* alloc size of both buffer_t::base */
-    size_t bufSize;
+    int bufSize;
     int    errlogInitFailed;
 
     epicsMutexId listenerLock;
@@ -98,8 +98,8 @@ static struct {
 
     /* A loop counter maintained by errlogThread. */
     epicsUInt32 flushSeq;
-    size_t nFlushers;
-    size_t nLost;
+    unsigned long nFlushers;
+    unsigned long nLost;
 
     /* 'log' and 'print' combine to form a double buffer. */
     buffer_t *log;
@@ -140,7 +140,7 @@ char* msgbufAlloc(void)
 }
 
 static
-size_t msgbufCommit(size_t nchar, int localEcho)
+int msgbufCommit(int nchar, int localEcho)
 {
     int isOkToBlock = epicsThreadIsOkToBlock();
     int wasEmpty = pvt.log->pos==0;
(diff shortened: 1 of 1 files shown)


modules/libcom/src/error/errlog.c
```diff
@@ -70,14 +70,15 @@ typedef struct listenerNode{
 
 typedef struct {
     char *base;
-    size_t pos;
+    int pos;
```
dirk-zimoch, line 73: Avoid size_t and thus %z as well as comparing size_t to int resulting in many warnings on Windows.
Message buffers are short enough for int on all supported architectures.

modules/libcom/src/error/errlog.c
```diff
@@ -70,14 +70,15 @@ typedef struct listenerNode{
...
+    int nchar;
```
dirk-zimoch, line 74: Keeping track of partial message size.

modules/libcom/src/error/errlog.c
```diff
@@ -125,9 +126,14 @@ char* msgbufAlloc(void)
...
     errlogInit(0);
     epicsMutexMustLock(pvt.msgQueueLock); /* matched in msgbufCommit() */
-    if(pvt.bufSize - pvt.log->pos >= 1+pvt.maxMsgSize) {
+    if(pvt.bufSize - pvt.log->pos - pvt.log->nchar >= 1+pvt.maxMsgSize) {
         /* there is enough space for the worst cast */
         ret = pvt.log->base + pvt.log->pos;
+        if (pvt.log->nchar) {
+            /* append to last message */
+            return ret + 1 + pvt.log->nchar;
+        }
```
dirk-zimoch, line 135: Append to existing partial message: Do not overwrite the state in ret[0]!

modules/libcom/src/error/errlog.c
```diff
@@ -140,39 +146,44 @@ char* msgbufAlloc(void)
...
     if(localEcho && isOkToBlock && atExit) {
         /* errlogThread is not running, so we print directly
          * and then abandon the buffer.
          */
         fprintf(pvt.console, "%s", start);
 
+    } else if (start[nchar] != '\n') {
+        /* incomplete message, prepare to append */
+        pvt.log->nchar += nchar;
+
```
dirk-zimoch, line 176: If message does not end in \n, store it and wait for more. Do not send it out yet.

modules/libcom/src/error/errlog.c
```diff
@@ -140,39 +146,44 @@ char* msgbufAlloc(void)
...
     int wasEmpty = pvt.log->pos==0;
     int atExit = pvt.atExit;
-    char *start = pvt.log->base + pvt.log->pos;
+    char *start = pvt.log->base + pvt.log->pos + pvt.log->nchar;
 
     /* nchar returned by snprintf() is >= maxMsgSize when truncated */
     if(nchar >= pvt.maxMsgSize) {
         const char *trunc = "<<TRUNCATED>>\n";
-        nchar = pvt.maxMsgSize - 1u;
+        nchar = pvt.maxMsgSize - 1;
```
dirk-zimoch, line 159: 0u and 1u is rather pointless. And our data types are signed now anyway.

modules/libcom/src/error/errlog.c
```diff
@@ -140,39 +146,44 @@ char* msgbufAlloc(void)
...
 
-        pvt.log->pos += 1u + nchar + 1u;
+        pvt.log->pos += 1 + pvt.log->nchar + nchar + 1;
+        pvt.log->nchar = 0;
     }
 
     epicsMutexUnlock(pvt.msgQueueLock); /* matched in msgbufAlloc() */
 
-    if(wasEmpty && !atExit)
+    if(wasEmpty && !atExit && !pvt.log->nchar)
```
dirk-zimoch, line 186: Message not complete: Do not trigger output.

modules/libcom/src/error/errlog.c
```diff
@@ -234,22 +245,19 @@ int isATTY(FILE* fp)
...
     else if(fp==stderr)
         hand = GetStdHandle(STD_ERROR_HANDLE);
-#ifdef ENABLE_VIRTUAL_TERMINAL_PROCESSING
     if(hand && GetConsoleMode(hand, &mode)) {
         (void)SetConsoleMode(hand, mode|ENABLE_VIRTUAL_TERMINAL_PROCESSING);
-        mode = 0u;
+        mode = 0;
         if(GetConsoleMode(hand, &mode) && (mode&ENABLE_VIRTUAL_TERMINAL_PROCESSING))
             return 1;
     }
```
dirk-zimoch, line 260: If ENABLE_VIRTUAL_TERMINAL_PROCESSING is not defined, calling GetStdHandle() is pointless.

modules/libcom/src/error/errlog.c
```diff
@@ -638,49 +646,49 @@ static void errlogThread(void)
...
             while(pos < print->pos) {
                 listenerNode *plistenerNode;
                 char* base = print->base + pos;
-                size_t mlen = epicsStrnLen(base+1u, pvt.bufSize - pos);
+                int mlen = (int)epicsStrnLen(base+1, pvt.bufSize - pos);
                 int stripped = 0;
 
                 if((base[0]&ERL_STATE_MASK) != ERL_STATE_READY || mlen>=pvt.bufSize - pos) {
-                    fprintf(stderr, "Logic Error: errlog buffer corruption. %02x, %zu\n",
+                    fprintf(stderr, "Logic Error: errlog buffer corruption. %02x, %u\n",
```
dirk-zimoch, line 659: VxWorks does not support %z.

mdavidsaver:
> Now, the message is accumulated until a part ends in \n

If someone has forgotten a newline, then it doesn't flush until a subsequent message with a newline?

I know that windows stdout behaves this way, and have been occasionally been confused/frustrated when output gets lost on `exit()`.  So I'm not enthused about errlog starting to do the same.  Having a timed flush (like *nix stdout) would help, although I still think there are simpler ways to achieve the same effect.  eg. in `logClientSend()` by injecting `logClientPrefix` only before the first character after a newline (and initially).

mdavidsaver:
> Also avoid usage of %z (for VxWorks), make types more consistent (for Windows).

This change is more than changing a `printf()` format  string.  What motivates the larger change?

(Seeing `int bufSize;` gives me bad Java flashbacks)

modules/libcom/src/error/errlog.c
mdavidsaver: > VxWorks does not support %z.

Could you make a make a minimal change to address just this?  eg. replace `%zu` with `%u` and cast the corresponding value to `(unsigned)`?

A small change like this could be easily reviewed and less contentious.  Potentially before the next Base release in about a month.  Mixing this up with functional changes is going to take longer.

modules/libcom/src/error/errlog.c
mdavidsaver: I'm sorry, but I just don't like replacing `size_t` with `int`.  A "size" is inherently unsigned.  Making this change to work around a vxworks limitation seems to me like the tail [wagging](https://www.collinsdictionary.com/us/dictionary/english/the-tail-is-wagging-the-dog) the dog.

modules/libcom/src/error/errlog.c
mdavidsaver: > Message not complete: Do not trigger output.

The more I think about it, the less I like this behavior.  Making multiple calls to `errlogPrintf()` and friends has always been a race, potentially resulting in corrupted output.  Delaying some output seems to me likely only to add confusion.

"fixing" this behavior would either necessitate API additions, some sort of "message builder" type.  w/ c++ code `std::ostringstream` is handy.  So perhaps this is an indicate that this is better left to dependent code.

modules/libcom/src/error/errlog.c
mdavidsaver: > comparing size_t to int resulting in many warnings on Windows

This is a warning w/ recent GCC as well, which should be addressed.  The issue in errlog.c isn't so much type width, but whether `snprintf()` can return a negative number (eg `-1` w/ `errno` set).  While no `errno` values are mentioned, [`man snprintf`](https://linux.die.net/man/3/sprintf) does say:

```
If an output error is encountered, a negative value is returned. 
```

anjohnson:
@dirk-zimoch If there's a problem with `logMsg()` sending out partial messages, maybe your "save characters until we see a newline" code belongs in the `errlogWrite()` routine in modules/libcom/src/osi/os/vxWorks/logMsgToErrlog.cpp rather than in errlog itself. That would avoid having to resolve the prefix issue too.

mdavidsaver:
As I understand `iosDrvInstall()` / `iosDevAdd()` sets up a character device file, with calls to `errlogWrite()` mapping one-to-one with calls to `write()`.  So I agree with @anjohnson , line buffering in `logMsgToErrlog.cpp` seems a better approach.  (and being vxworks specific code, I can ignore it ;) )

modules/libcom/src/error/errlog.c
dirk-zimoch: `snprintf()` has no possible output errors as the output does not go to an I/O device. It never returns negative values. The man page is for all `*printf()` flavors. Only those which use files can fail.

modules/libcom/src/error/errlog.c
dirk-zimoch: I did it here because there was already a buffering mechanism that I could (ab-)use.

modules/libcom/src/error/errlog.c
dirk-zimoch: For the complete list of `%z` usages see https://bugs.launchpad.net/epics-base/+bug/1983472

dirk-zimoch:
Just back from vacation...
> If there's a problem with logMsg() sending out partial messages, maybe your "save characters until we see a newline" code belongs in the errlogWrite() routine in modules/libcom/src/osi/os/vxWorks/logMsgToErrlog.cpp rather than in errlog itself. That would avoid having to resolve the prefix issue too.

I considered this first, but then I found that there was already a working buffering mechanism in errlog. As "normal" messages are always `\n` terminated, I did not see any conflict with existing code. Any real-life counter examples? (specially constructed ones and bugs don't count)

anjohnson:
Core Group 2023-03-06: We see the fix to [the Lunchpad bug #1983385](https://bugs.launchpad.net/epics-base/+bug/1983385) as requiring changes to the VxWorks logMsgToErrlog.cpp code, since that would also fix the prefix problem.

We've been moving towards the use `size_t` instead of `int` internally for length values that are naturally unsigned, going the other way (for generic code files like errlog at least) looks like a regression.

Closing this PR, please submit a separate PR with any fix to logMsgToErrlog.cpp.

dirk-zimoch:
I support using size_t for all things that relate to memory sizes. Can‘t wait to see all array lengths be size_t.

PR #282, opened 2022-08-05
Improve `logClientPrefix`

mdavidsaver:
@dirk-zimoch Example of what I had in mind when writing https://github.com/epics-base/epics-base/pull/281#issuecomment-1206630893.  I've only tested that epicsErrlogTest passes.  There are still cases which may need to be considered.  eg. Multiple newline `"\n\n"`, possibly split over two calls.

### DIFF
--- a/modules/libcom/src/log/logClient.c
+++ b/modules/libcom/src/log/logClient.c
@@ -53,6 +53,7 @@ typedef struct {
     unsigned            connected;
     unsigned            shutdown;
     unsigned            shutdownConfirm;
+    unsigned            atEoL;
     int                 connFailStatus;
 } logClient;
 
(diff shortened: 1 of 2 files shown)


dirk-zimoch:
I would not print a prefix after \n in the middle of one message.

mdavidsaver:
> I would not print a prefix after \n in the middle of one message.

I take it that you refer to:

```c
                     "A message with \na second prefix\n",
                     "DONE\n"
```

Which this PR renders as:

```c
                     "fac=LI21 A message with \n"
                     "fac=LI21 a second prefix\n"
                     "fac=LI21 DONE\n"
```

So you would rather have?

```c
                     "fac=LI21 A message with \n"
                     " a second prefix\n"
                     "fac=LI21 DONE\n"
```

I understand the idea of differentiating a multi-line message (by eg. a leading space).  Though this implies that:

```c
errlogPrintf("A message with \na second prefix\n");
```

would differ further from:

```c
errlogPrintf("A message with \n");
errlogPrintf("a second prefix\n");
```

I think this would be acceptable to me.  The second pair of calls has always been racing with potential concurrent loggers.

mdavidsaver:
I think this example has served its purpose.

PR #283, opened 2022-08-05
`iocShutdown()`: Always stop worker threads

mdavidsaver:
Something I'm testing following a [a tech-talk discussion](https://epics.anl.gov/tech-talk/2022/msg01081.php) involving ideas for improving IOC shutdown.

Switch `iocShutdown()` to unconditionally stop and join dbScan and callback worker threads.

The goal is to reduce the scope for race conditions in `epicsAtExit()` and `initHookAfterShutdown` callbacks.

Joining these threads opens up a possibility for a new deadlocks in user code if one of the threads being joined is eg. waiting on `epicsExit()`.  In the specific case of `epicsExit()`, the fix is switching to `epicsExitLater()` cf. 924350b362edde17f612dff575af5effaaff4886.

### DIFF
--- a/modules/database/src/ioc/db/callback.c
+++ b/modules/database/src/ioc/db/callback.c
@@ -301,13 +301,16 @@ void callbackInit(void)
             callbackQueue[i].threadsConfigured = callbackThreadsDefault;
 
         for (j = 0; j < callbackQueue[i].threadsConfigured; j++) {
+            epicsThreadOpts opts = EPICS_THREAD_OPTS_INIT;
+            opts.joinable = 0;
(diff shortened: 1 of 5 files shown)


anjohnson:
Core Group 12/7: Looks reasonable, maybe check one or more of the Python IOC tools to make sure they don't have problems, if any of them attempt IOC shutdown properly.

Un-draft, and a Release note entry please.

anjohnson:
Core group 2023-01-04: Good with Release Note entry.

mdavidsaver:
Applied as of f430389ee72f08fc1f24fd2960f69c1a14bbf9eb

PR #287, opened 2022-08-08
fix aSub with constant input links

### LINKED ISSUE #284: Constant `INP*` to aSub
cf. https://epics.anl.gov/tech-talk/2022/msg01154.php

```
record(aSub,"$(user):aSubExample")
{
   field(INAM,"myAsubInit")
   field(SNAM,"myAsubProcess")
   field(FTA,"DOUBLE")
   field(NOA,"10")
   field(INPA,"$(user):compressExample CPP")
   field(FTB,"DOUBLE")
   field(NOB,"1")
   field(INPB,"123")
}
```

Stepping through `aSubRecord.c`, it looks like `init_record()` does the right thing, and leaves `B=123` and `NEB=1`.

https://github.com/epics-base/epics-base/blob/424d678abff19887cd2fa9ff5af82e33806a3712/modules/database/src/std/rec/aSubRecord.c#L133-L135

However, later in `fetch_values()`, there is no `dbLinkIsConstant()` test, so the call to `dbGetLink()` goes to `dbConstGetValue()` where `NEB` is set to zero.

https://github.com/epics-base/epics-base/blob/424d678abff19887cd2fa9ff5af82e33806a3712/modules/database/src/std/rec/aSubRecord.c#L278-L283

https://github.com/epics-base/epics-base/blob/424d678abff19887cd2fa9ff5af82e33806a3712/modules/database/src/ioc/db/dbConstLink.c#L212-L213

mdavidsaver:
Attempt to fix #284

Open questions:

- [ ] Do other record types working with arrays need the same change? 

That is, places where `dbGetLink()` is called with `pnRequest!=NULL`.

rset and dset for aai, aao, subArray, waveform.  rset for compressRecord and printfRecord.  lnkCalc.c.

### DIFF
--- a/modules/database/src/std/rec/aSubRecord.c
+++ b/modules/database/src/std/rec/aSubRecord.c
@@ -275,8 +275,11 @@ static long fetch_values(aSubRecord *prec)
 
     /* Get the input link values */
     for (i = 0; i < NUM_ARGS; i++) {
+        DBLINK *plink = &(&prec->inpa)[i];
         long nRequest = (&prec->noa)[i];
-        status = dbGetLink(&(&prec->inpa)[i], (&prec->fta)[i], (&prec->a)[i], 0,
+        if(dbLinkIsConstant(plink))
+            continue;
+        status = dbGetLink(plink, (&prec->fta)[i], (&prec->a)[i], 0,
             &nRequest);
         if (status)
             return status;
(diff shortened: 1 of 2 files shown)


anjohnson:
I've just gone through all the modules/database/src/std calls to `dbGetLink()` and while I haven't tested anything I didn't see any other places that obviously need fixing in the same way as aSub.

I do wonder whether the aai and waveform record SIOL links should support constant simulated input values, currently neither attempt to load simulated constants from that link, but that would probably be an enhancement.

ralphlange:
I support the idea to allow constant SIOLs for array-type records.

IMHO it is an enhancement - until pretty recently it was not even possible to have constant array data in a link field.

anjohnson (approved):
Looks good to me.

mdavidsaver:
Applied as d47fa4caa404aeff8ed5a218b057b64e08499857

PR #290, opened 2022-08-18
iocinf.cpp: Hostnames may be longer than 32 bytes

tboegi:
Found here at ESS:

(all in one line)
EPICS_PVA_ADDR_LIST=averylonghostname.mylabnetwork.technicalnetwork.example.com
EPICS_PVA_AUTO_ADDR_LIST=NO
pvget somePVnam

leads to something like this:
    Warning: "Empty PV search address list"
    Source File: ../udpiiu.cpp line 403
    Current Time: Thu Jun 09 2022 10:10:47.804161447

Solution:
Increase the buf size in addAddrToChannelAccessAddressList() from 32 to 256

### DIFF
--- a/modules/ca/src/client/iocinf.cpp
+++ b/modules/ca/src/client/iocinf.cpp
@@ -79,7 +79,7 @@ extern "C" int epicsStdCall addAddrToChannelAccessAddressList
     const char *pStr;
     const char *pToken;
     struct sockaddr_in addr;
-    char buf[32u]; /* large enough to hold an IP address */
+    char buf[256u]; /* large enough to hold an IP address or hostname */
     int status, ret = -1;
 
     pStr = envGetConfigParamPtr (pEnv);


anjohnson (approved):
I suspect the real length limit for a fully-qualified hostname is likely to be larger than 256, but I'm not suggesting we worry about that. This looks fine to me.

mdavidsaver:
If @anjohnson isn't worried about the increased stack usage, then I don't see a problem with this as a quick fix.  Although being c++, `std::vector` is available.

In the longer term, our general trend has been to redesign code to remove fixed size limits like this altogether.  eg. make a temporary copy in a mutable `std::vector<char>`, and then split with `epicsStrtok_r()`.

tboegi:
> I suspect the real length limit for a fully-qualified hostname is likely to be larger than 256, but I'm not suggesting we worry about that. This looks fine to me.
@anjohnson 
I think that we are fine with 256:
https://github.com/epics-base/epics-base/commit/dca5c9ed3c038e39c1e34e3dd14ca5633ac732d4

anjohnson:
@tboegi Thanks for the RFC pointer, it's nice to see that the length limit is actually well-defined and not huge.

mdavidsaver:
Applied as a8e8d22c31aaa2bfff6511017e883b5e751a3e0a

PR #292, opened 2022-08-29
Add 'fs' link support

mdavidsaver:
Link support for file system access.

Something I played around with.  I'm not sure if I'll pursue this, either in this form, or as device supports.  I can foresee a very large number of permutations of options would be desirable.  eg. at present file are (re) opened for each `dbGet/PutLink()`, and read entirely (`IN_LINK`) or truncated (`OUT_LINK`).

My toy case for testing.

```
record(ai, "temp6") {
    field(DTYP, "Raw Soft Channel")
    field(INP, {fs:"/sys/class/thermal/thermal_zone7/temp"})
    field(SCAN, "1 second")
    field(ASLO, "0.001")
    field(EGU , "C")
    field(PREC, "3")
}

record(bo, "turbo") {
    field(OUT, {fs:{path:"/sys/devices/system/cpu/intel_pstate/no_turbo"}})
    field(ZNAM, "Enable")
    field(ONAM, "Disable")
}
```

### DIFF
--- a/modules/database/src/std/link/Makefile
+++ b/modules/database/src/std/link/Makefile
@@ -15,5 +15,6 @@ dbRecStd_SRCS += lnkConst.c
 dbRecStd_SRCS += lnkCalc.c
 dbRecStd_SRCS += lnkState.c
 dbRecStd_SRCS += lnkDebug.c
+dbRecStd_SRCS += lnkFS.c
 
 HTMLS += links.html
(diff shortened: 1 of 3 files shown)


anjohnson:
Interesting.

This is probably an annoying and pedantic comment, but the link provides access to the contents of a file, not a whole file-system.

A scanf record type might be an interesting companion to this, or a regex record type, but that might need to be in an external module for portability/dependency reasons (libpcre).

mdavidsaver:
Dropping for lack of interest.

Issue #293, opened 2022-09-01
Tests with custom .plt files do not track their dependencies


simon-ess:
There are a few tests in EPICS base that have .plt files that are meant to replace the standard .t files generated by `makeTestfile.pl`. In most cases, these are their own independent tests, but in a few cases they have source files that need to be compiled in order for the test to run:
* modules/libcom/test/epicsUnitTestTest
* modules/libcom/test/yajlTest

However, the rules from RULES_BUILD do not take this into account:
```make
# If there's a perl test script (.plt) available, use it
%.t: ../%.plt
	@$(RM) $@
	$(EXPAND_TOOL) -t $(INSTALL_LOCATION) -a $(T_A) $< $@

# Test programs (.t files) must be written in Perl.
# Generate a perl program to exec the real test binary.
%.t: %$(EXE) $(TOOLS)/makeTestfile.pl
	@$(RM) $@
	$(PERL) $(TOOLS)/makeTestfile.pl $(T_A) $(EPICS_HOST_ARCH) $@ $<
```

As a consequence, if you happen to modify or delete the source files that are used in these tests, then the tests will fail upon a repeated invocation of `make runtests`.

anjohnson:
Doesn't this just mean that we need to add something like this dependency?
```makefile
runtests: build
```
The final version won't actually be quite that, but I think it's close enough to confirm whether that does what you're looking for.

simon-ess:
Yes, you're right, and that is a much simpler fix. I will update the pull request.

anjohnson:
Fixed with 54c02e2c247def09ebfbfc501e29c1b4e25d42a8

PR #294, opened 2022-09-01
Fix missing test dependencies

simon-ess:
See https://github.com/epics-base/epics-base/issues/293. The gist is that tests with custom perl wrappers do not register their dependencies on their source files, which can result in tests failing if test are modified or removed after they are built.

### DIFF
--- a/configure/RULES_BUILD
+++ b/configure/RULES_BUILD
@@ -372,7 +372,7 @@ $(MODNAME): %$(MODEXT): %$(EXE)
 #---------------------------------------------------------------
 # Automated testing
 
-runtests: run-tap-tests
+runtests: build run-tap-tests
 run-tap-tests: $(TESTSCRIPTS.t)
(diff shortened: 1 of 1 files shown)


simon-ess:
Note that there is also a similar issue with one of the tests missing a dependency on a generated .db file; this is included in this pull request, but there wasn't an obvious generic mechanism in that case.

anjohnson:
This is fine for serial builds, I just merged 54c02e2c247def09ebfbfc501e29c1b4e25d42a8 that also works for parallel builds too. The `build` dependency needed to be added to the `run-tap-tests` target (with a couple other adjustments) so the build tasks happens *before* the tests are run, otherwise under `make -j` they could be done at the same time, which obviously wouldn't work. This recipe requires `build` to be complete before `run-tap-tests` gets executed:
```makefile
runtests: run-tap-tests
run-tap-tests: | build
```
I applied a similar change to the `%.tap: %.t` static pattern rule as well.

PR #297, opened 2022-09-15
Modify pointer type in example to match options from epicsTypes.h

mskoufis:
This example in the documentation was somewhat misleading and caused us a bit of a headache when running similar code on our real-time linux servers.  For the aSub record in this example, the pointer to input A is declared as long*, while instead it should be declared as a pointer to a type selected from epicsTypes.h.  In our case, our linuxRT C/C++ compiler assumes long to be 8 bytes, whereas input A in the aSub record was declared as an EPICS LONG, hence 4 bytes.  As a result, we would only read every other array element in the subroutine and resulted in incorrect calculations.

### DIFF
--- a/modules/database/src/std/rec/aSubRecord.dbd.pod
+++ b/modules/database/src/std/rec/aSubRecord.dbd.pod
@@ -2354,11 +2354,11 @@ The associated subroutine code that uses the A field might look like this:
 
     static long my_asub_routine(aSubRecord *prec) {
         long i;
-        epicsUInt32 *a;
+        epicsInt32 *a;
 
         double sum=0;
         ...
-        a = (epicsUInt32 *)prec->a;
+        a = (epicsInt32 *)prec->a;
         for (i=0; i<prec->noa; i++) {
             sum += a[i];
         }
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Using C `long` is certainly known to be a problem wrt. 32/64 linux.  Small point.  Strictly speaking the menu entry `LONG` maps to `epicsInt32` and `ULONG` to `epicsUInt32`.  Of course, not such a big deal in practice as both have the same `sizeof()`.

anjohnson:
@mskoufis — Sorry for the confusing documentation, and thanks for this fix. EPICS definitely shows its history in a number of places and this example code is one of them, the M68k and PowerPC target architectures that were mainly used for IOCs when the aSub was originally written had 32-bit `int` and `long` types. Some even earlier EPICS targets still had a 16-bit `int` type, which is why the Channel Access `DBF_INT` and `DBF_SHORT` field type-codes are synonyms.

BTW you should ignore the CI test failures from Appveyor, they are nothing to do with your changes.

anjohnson:
Core Group 11/23: Approved. Maybe also change `i` to be an `epicsUInt32` as well, since NOA is that type.  AI on whoever in the Core group merges this to make that change as well.

mdavidsaver:
Applied as 7ccc3ab82dd15bbec14804f93a8f66d73485ac30

Issue #286, opened 2022-08-08
UB fork() warnings from legal code


anjohnson:
```C
#include <unistd.h>

int main()
{
    return fork();
}
```

Compiling and linking the program above against libCom causes it to trigger the warning added in PR #214 on MacOS. I suspect Linux will behave the same but I haven't actually tried it. Adding calls to `epicsThreadShowAll()` demonstrate that no EPICS threads are running immediately before or after the fork, so I think the current check is too simplistic. I had assumed it would only warn if a thread is actually started before `fork()` is called.

Passing `-s` to the ca-nameserver or ca-gateway (which tells them to detach from their parent session) causes them to generate this warning incorrectly. IMHO a server process which double-forks to disconnect and become a session leader before it makes _any_ EPICS calls should not be triggering the warning.

anjohnson:
Core Group 9/28: Refer to #211 it would be nice to be able to relax this warning.

mdavidsaver:
My concern with relaxing the warning stems from concern about what could happen in global constructors running before `main()` (or on `dlopen()`).

If we care to assert that any such constructors are not, and will not, do anything `fork()` unsafe then I'm ok with relaxing the warning.  eg. not starting worker threads, or maybe even not locking a mutex.

With pvxs I make a stronger statement that no global constructors will appear in libraries at all.  In large part, this is because I know how to verify this [automatically](https://github.com/mdavidsaver/pvxs/blob/master/.ci-local/cdt-check.sh).  For GCC w/o `-flto` at least.  I don't know about [clang](https://llvm.org/docs/LangRef.html#the-llvm-global-ctors-global-variable).  (windows has no `fork()`, so I think isn't directly relevant)

From a quick check, it doesn't seem like the 4 Base libraries have any global constructors at present.  At least for my Linux build.

```
nm lib/linux-x86_64/*.a|grep __static_initialization_and_destruction
```

For comparison:

```c++
void bar();
struct Foo {
   Foo() {bar();}
} foo;
```

```
$ g++ -o ctor.o -c ctor.cpp && nm ctor.o | grep __static_initialization_and_destruction
0000000000000000 t _Z41__static_initialization_and_destruction_0ii
```

anjohnson:
[...]

PR #298, opened 2022-09-30
Remove global constructors from libCom

mdavidsaver:
C++ global constructors have well known pitfalls with respect to initialization order.  With Com they also trigger #286.

Status as of 1d19ba4cc2e8c9ab17288b684022a0682e398bde :

- [X] `epicsSingleton` constructors
  -  Fixed by 8f1243da406aebf5a8cebda65ea84e5dbeeff286 when building with `-std=c++11` or later.
- [ ] `fileDescriptorManager` singleton
  - https://github.com/epics-base/epics-base/pull/298 has an example of making the singleton lazy, now out of date.
- [X] `localRegister()` in iocsh.cpp
  - Fixed by 7830345e5927ca1f262dda17a1f819286c5b5821
- [ ] `initMainThread()`
  - Removed by #451
- [ ] `timeRegister()` really two separate instances
  - [ ] in `posix/osdTime.cpp`
  - [ ] in `Darwin/osdTime.cpp`

The remaining `timeRegister()` in the various `osdTime.cpp` seem less trivial.  Some of this work might be moved to `generalTime_Init()`.  Which might require changes around the `useOsdGetCurrent` optimization.  Something different would be needed for `osdMonotonicInit()`.

### DIFF
--- a/modules/libcom/src/osi/epicsThread.cpp
+++ b/modules/libcom/src/osi/epicsThread.cpp
@@ -382,12 +382,3 @@ extern "C" {
         return id;
     }
 } // extern "C"
-
-static epicsThreadId initMainThread(void) {
-    epicsThreadId main = epicsThreadGetIdSelf();
-    epicsThreadSetOkToBlock(1);
-    return main;
-}
-
-// Ensure the main thread gets a unique ID and allows blocking I/O
-epicsThreadId epicsThreadMainId = initMainThread();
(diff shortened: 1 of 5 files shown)


modules/libcom/src/osi/epicsThread.cpp
```diff
@@ -382,12 +382,3 @@ extern "C" {
         return id;
     }
 } // extern "C"
-
-static epicsThreadId initMainThread(void) {
-    epicsThreadId main = epicsThreadGetIdSelf();
-    epicsThreadSetOkToBlock(1);
```
mdavidsaver, line 388: This is not a no-op.  Threads default to not ok to block, except for `main()`.

mdavidsaver:
From the windows test failures, I think one of the global ctors is calling `osiSockAttach()`.

PR #299, opened 2022-09-30
Rtems mvme2100 fixes 7.0

goetzpf:
Here are three patches for EPICS Base 7.0:

* Fix generation of mvme2100 bootable files 29e70530bff6a4eb0bf6bfeba76229bb413aae7b
  The current version of EPICS Base 7.0 doesn't create bootable files for RTEMS for the mvme2100
  correctly. This patches fixes this. I have tested this with my mvme2100 CPU and RTEMS 5 in my office.
* Fix mvme2100 memory problem 03c4f093066d0abb82fe81643c0a0f354b614712
  I found a similar problem in an EPICS Mailing list and was able to apply the same patch for
  my mvme2100 CPU. This CPU has only 32MBytes of RAM, so some data structure in EPICS Base has
  to be made smaller in order to not consume all available memory.  
* Allow tftp and nfs for fetching the startup script 38ad82776e1c540ecc8c9b6c6567e3ee6376bdbe
  This patch isn't as large as it seems. Changes the code's indentation only makes it to seem a bit large.
  It enables to get the startup script via TFTP just by letting the file path start with "/TFTP/BOOTP_HOST".
  I needed this for my tests but to me this seems a useful addition to the Base.
  
  The patches above enabled me to successfully boot an example application 
  created with makeBaseApp.pl with template "example" with EPICS Base 7 and RTEMS 5 
  on my mvme2100 CPU. I verified that the database was loaded and the records were processed, so
  I could see changes of their values with a "camonitor" command.


anjohnson:
Core Group 12/7: Will this handle the case where the TFTP and BOOTP hosts are different?

@hjunkes Would you be able to check the TFTP changes here on another BSP?

hjunkes:
Unfortunately, this is not how it works.
At the point where after the string is checked:
posix/rtems_init.c" line 333
    if (strstr(rtems_bsdnet_bootp_cmdline, "/TFTP/BOOTP_HOST:")==rtems_bsdnet_bootp_cmdline) 

rtems_bsdnet_boot_cmdline contains only the FilePath.

In setBootConfigFromNVRAM.c the string "epics-string" is already split (splitRtemsBsdnetBootpCmdline(void)).

The following query worked for me (! without ":" !) :
    if (strstr(env_nfsServer, "/TFTP/BOOTP_HOST")==env_nfsServer)
    
Environment set on CPU:
epics-script=/TFTP/BOOTP_HOST:/Volumes/Epics:MVME3100/IOCDyCrVme/iocBoot/iocdycr/st.cmd

NFS- way: 
epics-script=10.0.0.1:/Volumes/Epics:MVME3100/IOCDyCrVme/iocBoot/iocdycr/st.cmd

anjohnson:
@hjunkes to re-tests and comment here.

Issue #219, opened 2021-12-17
caget record.SSCN


anjohnson:
I would like an improved response to this caget from a default-valued SSCN field:
```
woz$ caget bi.SSCN
CA.Client.Exception...............................................
    Warning: "Channel read request failed"
    Context: "op=0, channel=bi.SSCN, type=DBR_TIME_STRING, count=1, ctx="read failed""
    Source File: ../getCopy.cpp line 91
    Current Time: Thu Dec 16 2021 15:59:56.272797000
..................................................................
Read operation timed out: some PV data was not read.
bi.SSCN                 
```

On the IOC console output shown below, the first line comes from the above caget:
```
Illegal choice filename="(null)" line number=0PV: bi.SSCN error detected in routine: dbFastLinkConv(cvt_menu_st)
epics> dbgf bi.SSCN
Illegal choice filename="(null)" line number=0PV: bi.SSCN error detected in routine: dbFastLinkConv(cvt_menu_st)
DBF_STRING:         failed.   
epics> dbDumpField pdbbase bi SSCN
recordtype(bi) 
    SSCN
	         prompt: Sim. Mode Scan
	          extra: 
	      indRecordType: 66
	        special: 0 
	     field_type: DBF_MENU
	process_passive: 0
	       property: 0
	           base: 0
	    promptgroup: 90 - Simulate
	       interest: 1
	       as_level: 1
	        initial: 65535
		  menu: menuScan
	           size: 2
	         offset: 1022
Attribute: name RTYP value bi
Attribute: name VERS value none specified
```

There are several things in the above output that we could handle better (in addition to removing the filename and line number from those logged messages). The fundamental "why" of the error is that SSCN is a menu field with a default value that doesn't match any of the menuScan choices — a deliberate decision that I recommended to @ralphlange when he was working on the simulation enhancements and needed a "none of the above" choice value.

The IOC ought to behave better when it encounters that situation. Inside the IOC any out-of-range numeric puts to a menu field can cause this for almost any menu field. The behavior of a `DBF_ENUM` can be different from a `DBF_MENU` though since the IOC has separate converters for those two types.

anjohnson:
10/25/2022 In discussion with RL we agreed to convert the menu value to a string.

PR #301, opened 2022-11-01
Fix for GH issue #219, menu fields with non-choice values

anjohnson:
Menu fields may hold a value that does not correspond to one of the menu choices; the default value of the SSCN fields is the most common example (65535). Change the type conversion routines to return a numeric string instead of giving an error.

DBF_DEVICE was fixed in dbFastLinkConv.c before the 7.0.7 release.

### DIFF
--- a/modules/database/src/ioc/db/dbFastLinkConv.c
+++ b/modules/database/src/ioc/db/dbFastLinkConv.c
@@ -14,7 +14,6 @@
  */
 
 #include <stddef.h>
-#include <stdio.h>
 #include <string.h>
 #include <math.h>
 #include <float.h>
@@ -24,6 +23,7 @@
 #include "dbDefs.h"
 #include "epicsConvert.h"
 #include "epicsStdlib.h"
+#include "epicsStdio.h"
 #include "errlog.h"
 #include "errMdef.h"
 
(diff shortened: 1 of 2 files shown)


modules/database/src/ioc/db/dbFastLinkConv.c
```diff
@@ -1335,24 +1335,26 @@ static long cvt_menu_st(
...
-   char                 *pchoice;
+{
+    dbFldDes *pdbFldDes;
+    dbMenu *pdbMenu;
 
-    if(! paddr
-    || !(pdbFldDes = paddr->pfldDes)
-    || !(pdbMenu = (dbMenu *)pdbFldDes->ftPvt)
-    || *from>=pdbMenu->nChoice
-    || !(papChoiceValue = pdbMenu->papChoiceValue)
```
anjohnson, line 1348: This (papChoiceValue not NULL) + next check are not replicated in the new code — should never happen but...

anjohnson: Those checks were not being done in the original `dbGetString()` code in the other file either.

In dbLexRoutines.c `dbMenuBody()` uses `dbCalloc()` (a macro expanding out to `callocMustSucceed()`) to allocate the `papChoiceValue` array, and `dbMenuChoice()` similarly uses that for allocating the individual value strings. The only place in Base that inserts an item into `pdbbase->menuList` is `dbMenuBody()`, so the code in Base shouldn't need to check the contents separately.

anjohnson:
Core Group 11/23: Approved.

mdavidsaver:
Applied as b460c2659e8907f2aa7de9db44b1acf9ddb5a3d6

PR #303, opened 2022-11-15
Fix ringPointerTest priority

mdavidsaver:
Correctly avoid call to `epicsThreadSetPriority()` from non-epics thread `main()` (`prio==0` does not imply non-epics thread).

@hjunkes fyi.  What we looked at during the recent BNL meeting.

### DIFF
--- a/modules/libcom/test/ringPointerTest.c
+++ b/modules/libcom/test/ringPointerTest.c
@@ -240,17 +240,32 @@ static void testPair(int locked)
     epicsRingPointerDelete(ring);
 }
 
+static
+void testInThread(void *raw)
+{
+    epicsEventId stop = raw;
+    testPair(0);
+    testPair(1);
+    epicsEventMustTrigger(stop);
+}
+
 MAIN(ringPointerTest)
 {
-    int prio = epicsThreadGetPrioritySelf();
+    epicsThreadId tid;
+    epicsThreadOpts opts = EPICS_THREAD_OPTS_INIT;
+    epicsEventId stop = epicsEventMustCreate(epicsEventEmpty);
 
     testPlan(42);
     testSingle();
-    if (prio)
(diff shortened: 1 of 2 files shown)


modules/libcom/test/ringPointerTest.c
```diff
@@ -240,17 +240,32 @@ static void testPair(int locked)
...
+    /* testPair() needs to run with a priority > 0.
+     * Start a new thread since main() is a "non-epics"
+     * thread, for which we can/should not change the priority
+     */
+    opts.joinable = 1;
+    opts.priority = epicsThreadPriorityScanLow;
+    tid = epicsThreadCreateOpt("RING", testInThread, stop, &opts);
+    epicsEventMustWait(stop);
+    epicsThreadMustJoin(tid);
+    epicsEventDestroy(stop);
```
mdavidsaver, line 269: @anjohnson Maybe worth testing on vxWorks to make sure this `epicsEventDestroy()` doesn't race when `epicsThreadMustJoin()` is a no-op.

anjohnson: Forwarding to @dirk-zimoch — Dirk, this would affect pre-6.9 VxWorks systems, but since the change is only to test code it might not bother you. I don't have time to do the check that Michael is suggesting, and won't hold up merging this PR to look for a problem since we won't run EPICS 7 on older systems anyway.

If there is a problem one way to resolve it would be to convert this test program to C++ and use an `epicsThread` class, which handles joins itself.

dirk-zimoch: Fine for me.

anjohnson:
Core group 11/23: Looks good.

mdavidsaver:
Applied as 166267a32fef93aa15ffdbbd84ae2a35a1d797c4

PR #304, opened 2022-11-15
errlog fixes

mdavidsaver:
- errlog: worker exit when buffer is empty

Ensure buffered messages are printed when errlog shuts down.

This is likely a regression introduced by #113.

- errlogRemoveListeners() handle self-removal

Make `errlogRemoveListeners()` safe for self-removal from a listener callback.  Current code corrupts the `listenerList`.

This was an issue prior to #113 as well.


anjohnson:
Core group 11/23: Looks good.

mdavidsaver:
Applied as of 49fddaa13e10b3c524fa1d294f6d31392aa288af

PR #308, opened 2022-12-06
Add support for CA tools timeout from environment variable

drm-slac:
It would be useful to have EPICS Channel Access tools use an environment variable to specify a connection timeout to override the default of 1.0 seconds.  The '-w' flag will override the env variable (EPICS_CLI_TIMEOUT) for caput, caget, camonitor, and cainfo.

### DIFF
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -436,11 +436,6 @@ int main (int argc, char *argv[])
             enumAsNr = 1;
             break;
         case 'w':               /* Set CA timeout value */
-            /*
-             * epicsScanDouble is a macro defined as epicsParseDouble,
-             * (found in modules/libcom/src/misc) which will only
-             * change caTimeout here if it finds an acceptable value.
-             */
             if(epicsScanDouble(optarg, &caTimeout) != 1)
             {
                 fprintf(stderr, "'%s' is not a valid timeout value "
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -24,8 +24,6 @@
  *     Added field separators
  *  2009/04/01 Ralph Lange (HZB/BESSY)
  *     Added support for long strings (array of char) and quoting of nonprintable characters
- *  2022/12/06 Doug Murray (SLAC)
- *     Added CA timeout environment variable
  *
  */
 
@@ -391,20 +389,10 @@ int main (int argc, char *argv[])
 
     int nPvs;                   /* Number of PVs */
     pv* pvs;                    /* Array of PV structures */
-    const char* tmo;            /* timeout from environment var */
 
     LINE_BUFFER(stdout);        /* Configure stdout buffering */
 
-    if ((tmo = getenv(DEFAULT_TIMEOUT_ENV)) != NULL)
-    {
-    if(epicsScanDouble(tmo, &caTimeout) != 1)
-            {
-                fprintf(stderr, "'%s' is not a valid timeout value "
-                        "(from '%s' in the environment) - ignored. ('caget -h' for help.)\n", tmo, DEFAULT_TIMEOUT_ENV);
-                caTimeout = DEFAULT_TIMEOUT;
-            }
-
-    }
+    use_ca_timeout_env ( &caTimeout);
 
     while ((opt = getopt(argc, argv, ":taicnhsSVe:f:g:l:#:d:0:w:p:F:")) != -1) {
         switch (opt) {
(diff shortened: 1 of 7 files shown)


modules/ca/src/tools/tool_lib.h
```diff
@@ -49,6 +51,7 @@
 
 #define DEFAULT_CA_PRIORITY 0  /* Default CA priority */
 #define DEFAULT_TIMEOUT 1.0     /* Default CA timeout */
+#define DEFAULT_TIMEOUT_ENV "EPICS_CLI_TIMEOUT" /* CA timeout environment var */
```
anjohnson: Rename macro to `EPICS_CLI_TIMEOUT`

modules/ca/src/tools/caget.c
```diff
@@ -389,9 +391,21 @@ int main (int argc, char *argv[])
...
+    {
+    if(epicsScanDouble(tmo, &caTimeout) != 1)
+            {
+                fprintf(stderr, "'%s' is not a valid timeout value "
+                        "(from '%s' in the environment) - ignored. ('caget -h' for help.)\n", tmo, DEFAULT_TIMEOUT_ENV);
+                caTimeout = DEFAULT_TIMEOUT;
+            }
+
+    }
+
```
anjohnson: Yank this into a function in tool_lib.c

anjohnson: But be careful about the tool name in the error message...

anjohnson (changes requested):
Core Group 12/7: See suggestions.
If you want to add tests maybe create a unit test for the tool_lib.c.
Testing the timeout might have issues on Appveyor, we're okay with not testing the actual executables if that proves difficult.
Ask core-talk for advice if you need it.

modules/ca/src/tools/tool_lib.h
drm-slac: The macro was removed because the code common to the tool apps that referenced that macro was moved to this file. The environment variable now only appears in one place.

modules/ca/src/tools/caget.c
drm-slac: code was moved as suggested.

modules/ca/src/tools/caget.c
```diff
@@ -434,11 +436,15 @@ int main (int argc, char *argv[])
...
         case 'w':               /* Set CA timeout value */
-            if(epicsScanDouble(optarg, &caTimeout) != 1)
             {
-                fprintf(stderr, "'%s' is not a valid timeout value "
-                        "- ignored. ('caget -h' for help.)\n", optarg);
-                caTimeout = DEFAULT_TIMEOUT;
+                double prevTimeout = caTimeout;
+                if(epicsScanDouble(optarg, &caTimeout) != 1)
+                {
+                    caTimeout = prevTimeout;
```
mdavidsaver: `epicsScanDouble()` (aka. `!epicsParseDouble()`) will only write to the output pointer on success.  So I don't think there it is necessary to explicitly save/restore the default timeout.

modules/ca/src/tools/tool_lib.c
```diff
@@ -639,3 +640,32 @@ int connect_pvs (pv* pvs, int nPvs)
...
+/*+**************************************************************************
+ *
+ * Function:    use_ca_timeout_env
+ *
+ * Description: Set the variable at the given address to the
+ *              value of the CA timeout environment variable
+ *
+ * Arg(s) In:   timeout - Pointer to double
+ *
+ **************************************************************************-*/
```
mdavidsaver: Can this comment be condensed and made less redundant?

eg. the function name and argument are syntactically unambiguous, and need not be repeated.

Also, why go through the extra indirection of an argument which will always point to the `caTimeout` global?

(I understand and agree with a general aversion to globals, though in this can that ship has already sailed)

anjohnson: We don't need to maintain these decorated function header comments, feel free to change this to more like a single-line comment...

modules/ca/src/tools/caget.c
```diff
@@ -434,11 +436,16 @@ int main (int argc, char *argv[])
             enumAsNr = 1;
             break;
         case 'w':               /* Set CA timeout value */
+            /*
+             * epicsScanDouble is a macro defined as epicsParseDouble,
+             * (found in modules/libcom/src/misc) which will only
+             * change caTimeout here if it finds an acceptable value.
+             */
```
anjohnson: Move the "only change the value if it converts successfully" into the epicsStdlib.h header. Ditto below.

anjohnson (commented):
Core Group 12/21: Looks good subject to the suggested comment changes here.

anjohnson:
Core Group: Ping @drm-slac

mdavidsaver (approved):
I think this looks ok now.

anjohnson (approved):
Okay by me.

mdavidsaver:
Applied as 1d056c6fe42bc645f50c823d8b5861faa2f1b554

Issue #309, opened 2022-12-09
epicsStrPrintEscaped does not always communicate errors to the caller


smarsching:
`epicsStrPrintEscaped` (defined in `modules/libcom/src/misc/epicsString.c`) makes many calls to `fprintf`. Instead of checking the return value of `fprintf`, it simply sums up all the return values.

This will work correctly when all calls to `fprintf` are successful. However, if some of the calls fail, `fprintf` will return a negative value, so this value will be substracted from the sum. In that case, `epicsStrPrintEscaped` will return an incorrect number of characters and the caller will not even have a chance to detect this error, because the overall return value might still be positive.

I suggest fixing this by checking the return value of each call to `fprintf`, immediately returning and forwarding the return value when it is negative.

I am going to provide a PR that implements this bugfix.

PR #310, opened 2022-12-09
Fix epicsStrPrintEscaped to detect error in fprintf and return

smarsching:
This PR fixes issue #309.

`fprintf` returns a negative value in order to signal an error. We have to detect this situation in `epicsStrPrintEscaped` and return a negative when `fprintf` returns a negative value in order to give the calling code a chance to detect this situation.

The old implementation (of simply accumulating the return values of `fprintf`) was wrong anyway, because it would not only cause an error in `fprintf` to be lost but would also cause the returned number to be too small (not representing the actual number of bytes written) in such a case.

The only case where the old implementation would work correctly was when all calls to `fprintf` succeeded or all these calls failed.

### DIFF
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -233,24 +233,30 @@ int epicsStrPrintEscaped(FILE *fp, const char *s, size_t len)
 
    while (len--) {
        char c = *s++;
+       int rc = 0;
 
        switch (c) {
-       case '\a':  nout += fprintf(fp, "\\a");  break;
-       case '\b':  nout += fprintf(fp, "\\b");  break;
-       case '\f':  nout += fprintf(fp, "\\f");  break;
-       case '\n':  nout += fprintf(fp, "\\n");  break;
-       case '\r':  nout += fprintf(fp, "\\r");  break;
-       case '\t':  nout += fprintf(fp, "\\t");  break;
-       case '\v':  nout += fprintf(fp, "\\v");  break;
-       case '\\':  nout += fprintf(fp, "\\\\"); break;
-       case '\'':  nout += fprintf(fp, "\\'");  break;
-       case '\"':  nout += fprintf(fp, "\\\"");  break;
+       case '\a':  rc = fprintf(fp, "\\a");  break;
+       case '\b':  rc = fprintf(fp, "\\b");  break;
+       case '\f':  rc = fprintf(fp, "\\f");  break;
+       case '\n':  rc = fprintf(fp, "\\n");  break;
+       case '\r':  rc = fprintf(fp, "\\r");  break;
+       case '\t':  rc = fprintf(fp, "\\t");  break;
+       case '\v':  rc = fprintf(fp, "\\v");  break;
+       case '\\':  rc = fprintf(fp, "\\\\"); break;
+       case '\'':  rc = fprintf(fp, "\\'");  break;
+       case '\"':  rc = fprintf(fp, "\\\"");  break;
        default:
            if (isprint(0xff & (int)c))
-               nout += fprintf(fp, "%c", c);
+               rc = fprintf(fp, "%c", c);
            else
-               nout += fprintf(fp, "\\x%02x", (unsigned char)c);
+               rc = fprintf(fp, "\\x%02x", (unsigned char)c);
            break;
+       }
+       if (rc < 0) {
+           return rc;
+       } else {
+           nout += rc;
        }
    }
    return nout;
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -231,6 +231,12 @@ int epicsStrPrintEscaped(FILE *fp, const char *s, size_t len)
 {
    int nout = 0;
 
+   if (fp == NULL)
+       return -1;
+
+   if (s == NULL || strlen(s) == 0 || len == 0)
+       return 0; // No work to do
+
    while (len--) {
        char c = *s++;
        int rc = 0;
--- a/modules/libcom/test/epicsStringTest.c
+++ b/modules/libcom/test/epicsStringTest.c
@@ -199,24 +199,24 @@ static
 void testEpicsStrPrintEscaped(void)
 {   
     const char *filename = "testEpicsStrPrintEscaped";
-    // Create a file then re-open it read only
-    FILE *readOnly = fopen(filename, "a");
-    fclose(readOnly);
-    readOnly = fopen(filename, "r");
+    // Avoid printing to stdout by redirecting everything to a file
+    FILE *testFile = fopen(filename, "a");
+    FILE *readOnly = fopen(filename, "r");
 
     testDiag("testEpicsStrPrintEscaped()");
 
     // Passing cases
-    testOk1(epicsStrPrintEscaped(stdout, "1234", 4) == 4);
-    testOk1(epicsStrPrintEscaped(stdout, "\a\b\f\n\r\t\v\\\'\"", 10) == 20);
+    testOk1(epicsStrPrintEscaped(testFile, "1234", 4) == 4);
+    testOk1(epicsStrPrintEscaped(testFile, "\a\b\f\n\r\t\v\\\'\"", 10) == 20);
 
     //Failing cases
     testOk1(epicsStrPrintEscaped(readOnly, "1234", 4) == -1);
     testOk1(epicsStrPrintEscaped(NULL, "1234", 4) == -1);
-    testOk1(epicsStrPrintEscaped(stdout, NULL, 4) == 0);
-    testOk1(epicsStrPrintEscaped(stdout, "", 4) == 0);
-    testOk1(epicsStrPrintEscaped(stdout, "1234", 0) == 0);
+    testOk1(epicsStrPrintEscaped(testFile, NULL, 4) == 0);
+    testOk1(epicsStrPrintEscaped(testFile, "", 4) == 0);
+    testOk1(epicsStrPrintEscaped(testFile, "1234", 0) == 0);
 
+    fclose(testFile);
     fclose(readOnly);
     remove(filename);
 }
--- a/modules/libcom/test/epicsStringTest.c
+++ b/modules/libcom/test/epicsStringTest.c
@@ -195,6 +195,32 @@ void testStrTok(void)
     testTok(NULL, " \t", "bbb ", "bbb", "");
 }
 
+static
+void testEpicsStrPrintEscaped(void)
+{   
+    const char *filename = "testEpicsStrPrintEscaped";
+    // Create a file then re-open it read only
+    FILE *readOnly = fopen(filename, "a");
+    fclose(readOnly);
+    readOnly = fopen(filename, "r");
+
+    testDiag("testEpicsStrPrintEscaped()");
+
+    // Passing cases
+    testOk1(epicsStrPrintEscaped(stdout, "1234", 4) == 4);
+    testOk1(epicsStrPrintEscaped(stdout, "\a\b\f\n\r\t\v\\\'\"", 10) == 20);
+
+    //Failing cases
+    testOk1(epicsStrPrintEscaped(readOnly, "1234", 4) == -1);
+    testOk1(epicsStrPrintEscaped(NULL, "1234", 4) == -1);
+    testOk1(epicsStrPrintEscaped(stdout, NULL, 4) == 0);
+    testOk1(epicsStrPrintEscaped(stdout, "", 4) == 0);
+    testOk1(epicsStrPrintEscaped(stdout, "1234", 0) == 0);
+
+    fclose(readOnly);
+    remove(filename);
+}
+
 MAIN(epicsStringTest)
 {
     const char * const empty = "";
@@ -209,7 +235,7 @@ MAIN(epicsStringTest)
     char *s;
     int status;
 
-    testPlan(439);
+    testPlan(446);
 
     testChars();
 
@@ -407,6 +433,7 @@ MAIN(epicsStringTest)
 
     testDistance();
     testStrTok();
+    testEpicsStrPrintEscaped();
 
     return testDone();
 }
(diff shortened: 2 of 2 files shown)


mdavidsaver (approved):
Looks reasonable to me.

anjohnson:
Core Group 12/21: Approved. We will make test coverage for this a Codeathon project.

mdavidsaver:
A note following up on a digression during yesterday's meeting about possible future work (w/ a different PR).  GCC optimization may replace simple/static `fprintf()` with `putc()` or `fwrite()`.  GCC 10 replaces both of the following.  So does clang.  MSVC always calls `fprintf()`.  (tested the following with https://godbolt.org/)

```c
#include <stdio.h>

int main()
{
    fprintf(stdout, "%s", "hello");
    fprintf(stdout, " world\n");
    return 0;
}
```

AlexanderWells-diamond:
I've written a test and made a PR for it here: https://github.com/smarsching/epics-base/pull/1

I couldn't find a way to do it as a PR on this repository without also including this PR's commits inside it, which I assume is unwanted. However doing it this way may mean that @smarsching has to approve the PR on his fork first?

smarsching:
Thank you for writing the test. I just merged your changes into my branch.
Unfortunately, I don not know any other good way to make calls to `fprintf` fail in a reproducible way.
I guess one could use FUSE to simulate a file-system that causes I/O errors, but this would be a lot of work, so I don’t think it is justified. The other possible solution would be replacing `fprintf` by a mock, but if there is no existing infrastructure for link seams, this is not possible, and creating such an infrastructure in the testing framework means a lot of work as well.

So, I think that writing to a read-only file descriptor is the best solution that can be implemented with reasonable effort.

modules/libcom/test/epicsStringTest.c
```diff
@@ -195,6 +195,32 @@ void testStrTok(void)
...
+    // Create a file then re-open it read only
+    FILE *readOnly = fopen(filename, "a");
+    fclose(readOnly);
+    readOnly = fopen(filename, "r");
+
+    testDiag("testEpicsStrPrintEscaped()");
+
+    // Passing cases
+    testOk1(epicsStrPrintEscaped(stdout, "1234", 4) == 4);
+    testOk1(epicsStrPrintEscaped(stdout, "\a\b\f\n\r\t\v\\\'\"", 10) == 20);
```
mdavidsaver: I didn't notice that CI had not run.  There is an issue in the test output with the special characters confusing things.

```
...
ok 439 - inp: "(nil)" delim: "  " tok: "bbb"=="bbb" save: ""==""
# testEpicsStrPrintEscaped()
1234ok 440 - epicsStrPrintEscaped(stdout, "1234", 4) == 4
\a\b\f\n\r\t\v\\\'\"ok 441 - epicsStrPrintEscaped(stdout, "\a\b\f\n\r\t\v\\\'\"", 10) == 20
ok 442 - epicsStrPrintEscaped(readOnly, "1234", 4) == -1
...
```

results in

```
  Test Summary Report
  -------------------
  epicsStringTest.tap           (Wstat: 0 Tests: 444 Failed: 0)
    Parse errors: Tests out of sequence.  Found (442) but expected (440)
                  Tests out of sequence.  Found (443) but expected (441)
                  Tests out of sequence.  Found (444) but expected (442)
                  Tests out of sequence.  Found (445) but expected (443)
                  Tests out of sequence.  Found (446) but expected (444)
  Displayed the first 5 of 6 TAP syntax errors.
  Re-run prove with the -p option to see them all.
  Files=49, Tests=4376,  0 wallclock secs ( 0.17 usr  0.05 sys +  0.05 cusr  0.01 csys =  0.28 CPU)
  Result: FAIL
````

**Comment by AlexanderWells-diamond** [comment]
@smarsching Thanks for approving my request. Unfortunately as @mdavidsaver pointed out there are issues with the tests. Would it be possible to get push access to your repo, so I can push directly to this PR? Otherwise I'll have to make another PR for your repo and you'll have to approve it. Thank you.

EDIT: Given that the fix seemed pretty simple I've made it a new PR here: https://github.com/smarsching/epics-base/pull/2
Hopefully this is the last one!

**Comment by mdavidsaver** [comment]
@AlexanderWells-diamond In the interest of handling this likely simple fix today, just push the update to a branch in your fork and let me know which one.

**Comment by AlexanderWells-diamond** [comment]
> @AlexanderWells-diamond In the interest of handling this likely simple fix today, just push the update to a branch in your fork and let me know which one.

Thanks, it's here: https://github.com/AlexanderWells-diamond/epics-base/tree/add_test
There's also a PR here: https://github.com/smarsching/epics-base/pull/2

**Comment by smarsching** [comment]
@AlexanderWells-diamond I just merged your second PR.

**Comment by AlexanderWells-diamond** [comment]
There's one failing test. It looks like read-only filehandles and `fprintf` on Windows behaves differently to Linux. I'll try and find a Windows machine to test it.

**Comment by AlexanderWells-diamond** [comment]
Unfortunately my Windows machine (Windows Server 2012 with Visual Studio 2015) cannot recreate the issue - I always see the expected "-1" return from `fprintf` when writing to a read-only file handle. I don't have a mingw setup anywhere to test that particular compiler.

The only thing I can think to do is to modify the test to print the actual value returned, in the hopes it sheds some light on it, but as the CI takes 12+ hours and the issue that I cannot directly commit to this PR means I'm loathe to do that.

**Comment by ralphlange** [comment]
Hm.
I do have _some_ compilers on my Windows box. What do I need to do?

**Comment by AlexanderWells-diamond** [comment]
Compile and run this test program should prove the behaviour:
 
```
// ConsoleApplication1.cpp : Defines the entry point for the console application.
//

#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <ctype.h>

int epicsStrPrintEscaped(FILE *fp, const char *s, size_t len)
{
	int nout = 0;

	if (fp == NULL)
		return -1;

	if (s == NULL || strlen(s) == 0 || len == 0)
		return 0; // No work to do

	while (len--) {
		char c = *s++;
		int rc = 0;

		switch (c) {
		case '\a':  rc = fprintf(fp, "\\a");  break;
		case '\b':  rc = fprintf(fp, "\\b");  break;
		case '\f':  rc = fprintf(fp, "\\f");  break;
		case '\n':  rc = fprintf(fp, "\\n");  break;
		case '\r':  rc = fprintf(fp, "\\r");  break;
		case '\t':  rc = fprintf(fp, "\\t");  break;
		case '\v':  rc = fprintf(fp, "\\v");  break;
		case '\\':  rc = fprintf(fp, "\\\\"); break;
		case '\'':  rc = fprintf(fp, "\\'");  break;
		case '\"':  rc = fprintf(fp, "\\\"");  break;
		default:
			if (isprint(0xff & (int)c))
				rc = fprintf(fp, "%c", c);
			else
				rc = fprintf(fp, "\\x%02x", (unsigned char)c);
			break;
		}
		if (rc < 0) {
			return rc;
		}
		else {
			nout += rc;
		}
	}
	return nout;
}

int main()
{
	const char *filename = "testEpicsStrPrintEscaped";
 
	FILE *testFile = fopen(filename, "a");
	FILE *readOnly = fopen(filename, "r");

	printf("%d\n", epicsStrPrintEscaped(testFile, "12345", 5));
	printf("%d\n", epicsStrPrintEscaped(testFile, "\a\b\f\n\r\t\v\\\'\"", 10));

	printf("%d\n", epicsStrPrintEscaped(readOnly, "12345", 5));

    return 0;
}
```

The expected output should be:
 ```
5
20
-1
```

The failing test returns something that isn't `-1`.

ralphlange:
Nice! On my Windows10 box I get

MSYS2 MinGW gcc 12.1.0 (64bit):
> 5
> 20
> 5

Cygwin g++ 11.3.0 (64bit):
> 5
> 20
> -1

MSVC 14.0 (64bit):
> 5
> 20
> -1

MSVC 16.1 (64bit):
> 5
> 20
> -1

AlexanderWells-diamond:
Right, so it looks like that `MSYS2 MinGW gcc 12.1.0 (64bit):` build is not standard compliant. Out of interest, what is the contents of the "testEpicsStrPrintEscaped" file for that platform? It should be `1234\a\b\f\n\r\t\v\\\'\"`.

Is there a standard thing done in EPICS base to handle (apparently) non-conforming implementations of certain functions?

ralphlange:
The testEpicsStrPrintEscaped file contains
```
12345\a\b\f\n\r\t\v\\\'\"
```

AlexanderWells-diamond:
> The testEpicsStrPrintEscaped file contains
> 
> ```
> 12345\a\b\f\n\r\t\v\\\'\"
> ```

So it appears the MinGW fprintf is not printing the data given to `fprintf`, but is returning a value as though it is. Definitely looks like a bug to me.

ralphlange:
Would that be a bug in the compiler or in the runtime environment?
With these Gnu-on-Windows things, I'm never sure where things are happening at run-time.

smarsching:
It seems like `fprintf` is implemented by a wrapper around the Windows C library that is provided by the MinGW-w64 project. There seem to be two versions and I haven’t figured out yet which of the two is used by default:

1. https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-crt/stdio/mingw_fprintf.c
2. https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-crt/stdio/ucrt_fprintf.c

So, I think it is a bug in MinGW, not GCC.

Should we maybe work around this bug by first testing that `fprintf` actually returns -1, and only running the failing part of the test when this condition holds? After all, we cannot expect epicsStrPrintEscaped to work correctly when the standard library does not work correctly…

smarsching:
A small addition to my last comment: If the first version is the one being used, I think I found the cause of the bug: `fputc` is called (by `__pformat_putc`) without ever checking the return value. So, the function cannot know whether `fputc` succeeded or not. This most definitely is a bug in MinGW-w64.

@ralphlange Would you like to submit a bug report with the MinGW-w64 project?

ralphlange:
@smarsching: Yep, will do. Thanks for the detective work.

smarsching:
I just pushed a commit that excludes the test with the read-only file descriptor when running on a platform where `fprintf` is buggy and does not return -1 on error. Let’s see whether this now makes all tests succeed.

AlexanderWells-diamond:
> I just pushed a commit that excludes the test with the read-only file descriptor when running on a platform where `fprintf` is buggy and does not return -1 on error. Let’s see whether this now makes all tests succeed.

Thank you for doing the legwork. I'm interested to see the bug report to MinGW.

One question: How does conditionally not running one test affect the declaration of the number of tests in the file? i.e. `testPlan(446)` - in some cases (i.e. on MinGW platforms) this will actually only be `445`. Is this an issue? (I must admit I don't know why we even need to declare the number of tests).

smarsching:
@AlexanderWells-diamond Thank you for pointing out the issue with `testPlan(…)`. I just amended my commit to include a dummy call to `testOk1` when the test is not included in order to keep the number of tests constant.

To answer your question why the total number of tests needs to be declared: It seems like it does not have to. `testPlan` can also be called with a parameter of zero, and in this case `testDone` will simply report the number of tests run. I guess that declaring the number of tests ensures that no tests were skipped accidentally (which could happen due to a bug and then hide an actual problem).

Other testing frameworks (e.g. Boost.Test) do this in a more implicit way by using special preprocessor macros for test declaration, but I have to admit that I actually like how simple and straight-forward EPICS’s testing framework looks.

ralphlange:
> @AlexanderWells-diamond Thank you for pointing out the issue with testPlan(…). I just amended my commit to include a dummy call to testOk1 when the test is not included in order to keep the number of tests constant.

_Skipping_ a test is the intended way to do that, as skipped tests are counted at execution when comparing declared and run number of tests. From the Doxygen documentation in `epicsUnitTest.h`:
```
 * If some program condition or failure makes it impossible to run some tests,
 * the testSkip() routine can be used to indicate how many tests are being omitted
 * from the run, thus keeping the test counts correct; the constant string why is
 * displayed as an explanation to the user (this string is not printf-like).
```

smarsching:
@ralphlange Thanks for letting me know. This makes much more sense. I just changed the code to call `testSkip(1, "fprintf is broken on this system")` instead of `testOk1`.

AlexanderWells-diamond:
The test failure appears to be unrelated to the changes on this branch - it's in `regressTest` which is not a file we've modified. Do I recall hearing that the tests are sometimes flaky and subject to random failures? Is there a way to re-run the CI?

ralphlange:
Just started "rebuild failed jobs".

ralphlange:
MinGW-w64 bug created at https://sourceforge.net/p/mingw-w64/bugs/967/

mdavidsaver:
This change looks ok to me, and the CI jobs are now passing.  @ralphlange Do you know of any outstanding issues?

anjohnson:
Core Group: Approved, this PR adds a fix that resolves broken `fprintf()` behaviour in MinGW (and potentially others).

AlexanderWells-diamond:
Out of curiosity, is there a reason this isn't merged? Not sure if I'm missing a process step or somesuch. Thanks.

anjohnson:
Sorry, no significant reason. We have approved the PR so it will get merged when one of us has the time to do that, but $WORK projects have been taking up our time recently, and nobody has asked us for a new EPICS release recently, which also tends to bump up the priority a bit.

mdavidsaver:
Applied as of f488765631998c9d0b27a14c13aa412ec81a5fe9

modules/libcom/src/misc/epicsString.c
```diff
@@ -231,27 +231,39 @@ int epicsStrPrintEscaped(FILE *fp, const char *s, size_t len)
 {
    int nout = 0;
 
+   if (fp == NULL)
+       return -1;
+
+   if (s == NULL || strlen(s) == 0 || len == 0)
```
mdavidsaver, line 237: While not explicitly stated in documentation, having arguments `const char* s` and `size_t len` implies a contract to operate on an unterminated string.  I have belatedly realized that this added `strlen(s)` violates that implied contract.

smarsching: Good catch. Instead of using `strnlen`, `strnlen` should be used. Do you want to open a PR for this or should I?

PR #311, opened 2022-12-20
Return an error if subrecord processing fails due to bad INP links

brendanchandler:
If a sub record has an invalid INPx link, it was silently failing (and not running the proc function).  This change plumbs in the error, so the put fails and the user knows something went wrong.

### DIFF
--- a/modules/database/src/std/rec/subRecord.c
+++ b/modules/database/src/std/rec/subRecord.c
@@ -162,7 +162,7 @@ static long process(struct dbCommon *pcommon)
     recGblFwdLink(prec);
     prec->pact = FALSE;
 
-    return 0;
+    return status;
 }
 
 static long special(DBADDR *paddr, int after)
--- a/modules/database/test/std/rec/subproctest.c
+++ b/modules/database/test/std/rec/subproctest.c
@@ -44,7 +44,7 @@ MAIN(subproctest)
     // Since the put to PROC above fails, subproc() never runs
     // and the value of PROC will not be set by subproc().  However,
     // the testdbPutField call above goes through, so we get a partial
-    // result of the PROC file being left as 1.
+    // result of the PROC field being left as 1.
     testdbGetFieldEqual("InvalidINPARec.PROC", DBF_LONG, 1);
     
     testIocShutdownOk();
--- a/modules/database/test/std/rec/subproctest.c
+++ b/modules/database/test/std/rec/subproctest.c
@@ -1,0 +1,55 @@
+/*************************************************************************\
+* Copyright (c) 2022 UChicago Argonne LLC, as Operator of Argonne
+*     National Laboratory.
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+/* This test covers tests related to invoking subrecords
+ */
+
+#include <testMain.h>
+#include <dbUnitTest.h>
+#include <dbAccess.h>
+#include <iocsh.h>
+#include "registryFunction.h"
+#include <subRecord.h>
+
+static
+long subproc(subRecord *prec)
+{
+    prec->proc = 77;
+    return 0;
+}
+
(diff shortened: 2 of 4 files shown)


modules/database/test/std/rec/subproctest.c
```diff
@@ -0,0 +1,55 @@
...
+
+    testIocInitOk();
+    testDiag("===== Test that invalid link in INPA field fails a put request ======");
+
+    testdbPutFieldFail(-1, "InvalidINPARec.PROC", DBF_LONG, 1);
+
+    // Since the put to PROC above fails, subproc() never runs
+    // and the value of PROC will not be set by subproc().  However,
+    // the testdbPutField call above goes through, so we get a partial
+    // result of the PROC file being left as 1.
```
anjohnson: Typo in comment, `s/file/field/`

modules/database/src/std/rec/subRecord.c
```diff
@@ -162,7 +162,7 @@ static long process(struct dbCommon *pcommon)
     recGblFwdLink(prec);
     prec->pact = FALSE;
 
-    return 0;
+    return status;
```
anjohnson, line 165: This may be the most controversial change in this PR, although I think it's likely that we'll be fine with it. To document what I'm seeing: Some record types (e.g. fanoutRecord.c) have always returned `0` from their process() routine; some (e.g. aoRecord.c) have always returned `status`; some (e.g. calcRecord.c) had their `return status` replaced by `return 0` (2b6886385ba0546be3c3eade8079effbf2986489 in 1990), and others (e.g. waveformRecord.c much more recently in commit 23d917677221fc9d24421aa02dc63d2767c563e0) went the other way.

The rules have never been properly written down, the example of `process()` given [in the AppDevGuide](https://epics.anl.gov/base/R3-16/2-docs/AppDevGuide/RecordSupport.html#x12-49000011.4.3) (and I suspect most I/O record types) returns `0` from the first pass of an asynchronous device processing but `status` from the second or from a synchronous device. I don't see anything explaining how the return value is used, IIRC it's only fairly recently that it was actually propagated out of the parent `dbProcess()` routine.

anjohnson (approved):
Looks good to me personally, thanks!

The Core Group will take a look at this in due course.

modules/database/test/std/rec/subproctest.c
```diff
@@ -0,0 +1,55 @@
...
+
+    testIocInitOk();
+    testDiag("===== Test that invalid link in INPA field fails a put request ======");
+
+    testdbPutFieldFail(-1, "InvalidINPARec.PROC", DBF_LONG, 1);
+
+    // Since the put to PROC above fails, subproc() never runs
+    // and the value of PROC will not be set by subproc().  However,
+    // the testdbPutField call above goes through, so we get a partial
+    // result of the PROC field being left as 1.
```
anjohnson: C++-style comments in a .c file, please switch to C-style.

modules/database/test/std/rec/subproctest.c
```diff
@@ -0,0 +1,55 @@
...
+
+    testdbReadDatabase("subproctest.dbd", NULL, NULL);
+    subproctest_registerRecordDeviceDriver(pdbbase);
+    registryFunctionAdd("subproc", (REGISTRYFUNCTION) subproc);
+    testdbReadDatabase("subproctest.db", NULL, "TPRO=0");
+
+    testIocInitOk();
+    testDiag("===== Test that invalid link in INPA field fails a put request ======");
+
+    testdbPutFieldFail(-1, "InvalidINPARec.PROC", DBF_LONG, 1);
```
anjohnson, line 42: Could also check for a STAT=LINK alarm, which is also sub-record specific behavior.

anjohnson (commented):
Core Group 12/21: Approve subject to comment changes, others optional but would be nice.

Maybe change `subRecord::fetch_values()` to return the failure status, not just -1.

mdavidsaver:
Applied as 832abbd3b1fcbf5d15f1beb7da1a4bf3d8a4f0b5

PR #320, opened 2023-01-04
Update shareLib API docs, directing the reader to makeAPIheader.pl

exzombie:
To my eyes, the documentation in `epicsExport.h` indicates that including that header is preferable to manually defining `epicsExportSharedSymbols` prior to importing `shareLib.h`. Thus, I propose to change the wording in the latter.

### DIFF
--- a/modules/libcom/src/misc/epicsExport.h
+++ b/modules/libcom/src/misc/epicsExport.h
@@ -14,16 +14,22 @@
 /** \file epicsExport.h
  *  \brief Exporting IOC objects.
  *
- * This header is unique, as it defines epicsExportSharedSymbols and thus
- * triggers a transition between importing declarations from other libraries,
- * to exporting symbols from our own library. The comments in shareLib.h
- * provide more information.
+ * This header defines macros that allow registering IOC shell commands,
+ * subroutines, device support, etc.
  *
- * This header should be included with a trailing comment to make it stand
- * out from other includes, something like this:
- \code
-    #include <epicsExport.h>  // defines epicsExportSharedSymbols
- \endcode
+ * Because this header defines the `epicsExportSharedSymbols` macro, it triggers
+ * a transition between importing declarations from other libraries, to
+ * exporting symbols from our own library. For this reason, it must be included
+ * *last*, after all other EPICS-related includes. The comments in shareLib.h
+ * provide more information. It is recommended to mark this with a comment, e.g.
+ * \code
+ *    #include <epicsExport.h>  // defines epicsExportSharedSymbols, do not move
+ * \endcode
+ *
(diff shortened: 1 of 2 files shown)


anjohnson:
The disadvantages of using epicsExport.h are that it also provides one typedef and several macro definitions, and when placed within a group of `#include` lines it's often hard to recognize that it is special and requires these special rules. New or infrequent maintainers can very easily add or move `#include` lines above and/or below it without noticing it, and breaking the placement rules only breaks builds on Windows, which most people don't build for anyway.

The epicsExport.h file was intended for people writing record types, subRecord subroutines, device support and device drivers. For them it usually provides exactly the semantics they need, and in most cases their code symbols require no explicit decoration at all to build and run on Windows.

For other developers we have adopted a new approach to decorations that doesn't require strict segregation of the header files like shareLib.h does. There are build rules provided with Base and a Perl script called [makeAPIheader.pl](https://epics.anl.gov/base/R7-0/7-docs/doxygen/makeAPIheader.html) that can generate a header file containing decoration macros for a specific library. The above link provides documentation, which you can also see by running the script with a `-h` flag or by looking in the base/html directory.

I'm sure that our documentation of the new approach could be improved, but having the shareLib.h comments refer users to epicsExport.h doesn't feel right to me. I don't know how we could adapt epicsExport.h to the new approach though, and I suspect we may have to live with what we have now given how wide-spread the shareLib.h usage is.

My €0.02, others may have different ideas though...

exzombie:
A, so there's a *third* way to do this. Looking at it, I rather like it (ordering requirement always seemed icky), and as always, I wish I knew about this sooner :)

> I'm sure that our documentation of the new approach could be improved, but having the shareLib.h comments refer users to epicsExport.h doesn't feel right to me. I don't know how we could adapt epicsExport.h to the new approach though, and I suspect we may have to live with what we have now given how wide-spread the shareLib.h usage is.

I see nothing wrong with keeping the existing approach. I seems to me it would be a significant improvement if the doxygen of `shareLib.h` and `epicsExport.h` simply pointed the reader to `makeAPIHeader.pl`; if it did, I would have found it easily. I will think about how to word this.

ralphlange:
> A, so there's a third way to do this. Looking at it, I rather like it (ordering requirement always seemed icky), and as always, I wish I knew about this sooner :)

I think there's a pattern of _always_ being a third way that is _always_ the least documented one.

> It seems to me it would be a significant improvement if the doxygen of shareLib.h and epicsExport.h simply pointed the reader to makeAPIHeader.pl.

+1

anjohnson:
Core Group: Looks good, thanks.

mdavidsaver:
Applied as bdaca51d9696249bdac10fdcb75086324e4a3d53

PR #325, opened 2023-01-18
iocsh expose `dbCreateAlias` 

mdavidsaver:
Allow record aliases to be added via iocsh.  This is fully valid and safe prior to `iocInit`.  After `iocInit` it is not thread safe due to a lack of locking for `struct dbBase` and friends.  This PR still allows it as it will usually work as iteration of `dbRecordType::recList` isn't so frequent (although `recsync` does so on (re)connect).

My motivation comes from managing mrfioc2 driver installations at nsls2 where user friendly aliases for connected trigger channels are preferred to numbers, but IOC restarts were inconvenient.  Restarts to add aliases turned out to be the leading cause of downtime in the timing system during nsls2 installation.

Live additions to `recList` could be made safer by the use of atomic operations.  So far I'm not sure this would be worth the effort.

Also, cleanup error handling in `dbCreateAlias()`.  Currently this is not an issue as there are many other ways in which .db parsing errors are not recoverable.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
@@ -9,6 +9,7 @@
 \*************************************************************************/
 
 #include "iocsh.h"
+#include "errSymTbl.h"
 
 #include "dbStaticIocRegister.h"
 #include "dbStaticLib.h"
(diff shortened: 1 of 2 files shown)


anjohnson:
Core Group Feb 15: Needs Release Notes, with example(s). Concept approved.

mdavidsaver:
Release note...

### Expose `dbCreateAlias` in IOC shell

Add a new IOC shell command `dbCreateAlias` allow record aliases to be added.  Intended for use before `iocInit`.  eg. to add an alias "bar" for a record "foo".

```
dbLoadRecords("some.db") # includes: record(ai, "foo") { ...
dbCreateAlias("foo", "bar")
iocInit()
```

mdavidsaver:
Applied as of 3500a020343c4771cad94ed4a83615bc283718a4

PR #326, opened 2023-01-18
dbEvent: correct eventsRemaining

### LINKED ISSUE #324: dbEvent `eventsRemaining` missed on cancel
A non-zero `eventsRemaining` count passed to one `dbEventSubscription` callback promises that another callback is imminent.  However, a subsequent `db_cancel_event()` for any `dbEventSubscription` sharing the same `event_que` can break this promise.

RSRV uses `eventsRemaining!=0` to defer `send()` to avoid excessive packet fragmentation (in conjunction with `TCP_NODELAY`).  So if an callback "lost" by a `db_cancel_event()` can result in subscription updates being "stuck" in the user space send buffer until some other action triggers a flush.

@ralphlange provides a [test case](https://github.com/epics-base/epics-base/files/10448845/test.cpp.txt) which reliably triggers a "stuck" update.

mdavidsaver:
Only pass eventsRemaining!=0 if no pending events have been canceled.  Attempts a partial fix #324 without excess message fragmentation.

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -76,6 +76,7 @@ struct event_que {
     unsigned short          quota;          /* the number of assigned entries*/
     unsigned short          nDuplicates;    /* N events duplicated on this q */
     unsigned short          nCanceled;      /* the number of canceled entries */
+    unsigned                possibleStall;
 };
 
 struct event_user {
@@ -934,6 +935,7 @@ void db_post_single_event (dbEventSubscription event)
 static int event_read ( struct event_que *ev_que )
 {
     db_field_log *pfl;
+    int notifiedRemaining = 0;
     void ( *user_sub ) ( void *user_arg, struct dbChannel *chan,
             int eventsRemaining, db_field_log *pfl );
 
@@ -1012,6 +1014,7 @@ static int event_read ( struct event_que *ev_que )
                 /* Issue user callback */
                 ( *user_sub ) ( pevent->user_arg, pevent->chan,
                                 eventsRemaining, pfl );
+                notifiedRemaining = eventsRemaining;
             }
             LOCKEVQUE (ev_que);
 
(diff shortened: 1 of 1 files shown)


modules/database/src/ioc/db/dbEvent.c
```diff
@@ -1009,7 +1011,7 @@ static int event_read ( struct event_que *ev_que )
             if (pfl) {
                 /* Issue user callback */
                 ( *user_sub ) ( pevent->user_arg, pevent->chan,
-                                ev_que->evque[ev_que->getix] != EVENTQEMPTY, pfl );
+                                eventsRemaining, pfl );
```
mdavidsaver, line 1016: Looking again at the previous code.  `ev_que->evque[ev_que->getix]` is being evaluated without locking, so a race condition as well.

ralphlange:
Thanks!
I will prepare a test library with this fix for my colleague Walter tomorrow to have him run the original tests that were showing the issue.

mdavidsaver:
I'm starting to worry about adverse performance impacts.  This change would likely result in `eventsRemaining==0` more often, leading to more message fragmentation.  This could be mitigated by not setting `TCP_NODELAY`, which imo. would be reasonable for targets other than RTEMS and vxWorks (with coarse timer granularity).

mdavidsaver:
I think I have found a way avoid most of the excess message fragmentation by not passing `eventsRemaining==1` if any event on a queue has been canceled.  This may result in more missed opportunities for RSRV to combine messages than is strictly necessary.  The issue is only if that last event has been canceled.  On the other hand, assuming that cancelling subscriptions is relatively rare, the performance impact should be minimal.

I have also added a second commit which tries to warn in all cases where a `eventsRemaining==0` update is lost.  Including as described in https://github.com/epics-base/epics-base/issues/324#issuecomment-1399611177 .  As written `WARNING dbEvent possible queue stall` isn't very informative, but I'm not sure what information can usefully be printed.

ralphlange:
This fix works in the simplified and the original tests. (Run ~100 times with no failure.)

anjohnson:
Core Group Feb 15: Approved, needs a Release note entry to describe what changed.

mdavidsaver:
I've seen this new warning `WARNING dbEvent possible queue stall` appear once during IOC shutdown.  Probably spurious.  Need to investigate how the dbEvent worker exits...

mdavidsaver:
Changed to limit new warning message to once per dbEvent queue (aka. once per client w/ RSRV).

mdavidsaver:
Applied as b6626e4f60697d577097e335ce79e1ecbce3fee6

PR #329, opened 2023-02-02
Add support for loongarch64 architecture.

kira-96:
### Add `loongarch64` architecture support.

Now base can be compiled successfully on Loongson 3A5000(loongarch64) CPU.

### DIFF
--- a/configure/os/CONFIG_SITE.Common.linux-loong64
+++ b/configure/os/CONFIG_SITE.Common.linux-loong64
@@ -1,0 +1,48 @@
+# CONFIG_SITE.Common.linux-loong64
+#
+# Site Specific definitions for all linux-loong64 targets
+#-------------------------------------------------------
+
+# NOTE for SHARED_LIBRARIES: In most cases if this is set to YES the
+# shared libraries will be found automatically.  However if the .so
+# files are installed at a different path to their compile-time path
+# then in order to be found at runtime do one of these:
+# a) LD_LIBRARY_PATH must include the full absolute pathname to
+#    $(INSTALL_LOCATION)/lib/$(EPICS_HOST_ARCH) when invoking base
+#    executables.
+# b) Add the runtime path to SHRLIB_DEPLIB_DIRS and PROD_DEPLIB_DIRS, which 
+#    will add the named directory to the list contained in the executables.
+# c) Add the runtime path to /etc/ld.so.conf and run ldconfig
+#    to inform the system of the shared library location.
+
+# Depending on your version of Linux you'll want one of the following
+# lines to enable command-line editing and history in iocsh.  If you're
+# not sure which, start with the top one and work downwards until the
+# build doesn't fail to link the readline library.  If none of them work,
+# comment them all out to build without readline support.
+
+# No other libraries needed (recent Fedora, Ubuntu etc.):
+#COMMANDLINE_LIBRARY = READLINE
+
+# Needs -lncurses (RHEL 5 etc.):
+#COMMANDLINE_LIBRARY = READLINE_NCURSES
+
+# Needs -lcurses (older versions)
+#COMMANDLINE_LIBRARY = READLINE_CURSES
+
+# Readline is broken or you don't want use it:
+#COMMANDLINE_LIBRARY = EPICS
+
+
+# WARNING: Variables that are set in $(CONFIG)/CONFIG.gnuCommon cannot be
+# overridden in this file for native builds, e.g. variables such as
+#    OPT_CFLAGS_YES, WARN_CFLAGS, SHRLIB_LDFLAGS
+# They must be set in CONFIG_SITE.linux-loong64.linux-loong64 or for
+# cross-builds in CONFIG_SITE.<host-arch>.linux-loong64 instead.
+
+# Tune GNU compiler output for a specific cpu-type
+# (e.g. loongarch64, la264, la364, la464 etc.)
+#GNU_ARCH_CFLAGS = -march=loongarch64
+#GNU_TUNE_CFLAGS = -mtune=loongarch64
+# Enable soft-float feature (e.g. none, 32, 64)
+#GNU_DEP_CFLAGS = -mfpu=none
--- a/configure/CONFIG_SITE
+++ b/configure/CONFIG_SITE
@@ -25,6 +25,7 @@
 #	linux-arm              (GNU compiler used for host builds)
 #	linux-ppc              (GNU compiler used for host builds)
 #	linux-ppc64            (GNU compiler used for host builds)
+#	linux-loong64          (GNU compiler used for host builds)
 #	linux-x86              (GNU compiler used for host builds)
 #	linux-x86_64           (GNU compiler used for host builds)
 #	solaris-sparc          (Sun compiler used for host builds)
@@ -63,6 +64,7 @@
 #       ios-386                 (darwin-x86 host)
 #       linux-arm               (linux-x86 or -x86_64 host)
 #       linux-aarch64           (linux-x86_64 host)
+#       linux-loong64           (linux-x86_64 host)
 #       linux-microblaze
 #       linux-xscale_be
 #       vxWorks-486
--- a/configure/os/CONFIG.Common.linux-loong64
+++ b/configure/os/CONFIG.Common.linux-loong64
@@ -1,0 +1,13 @@
+# CONFIG.Common.linux-loong64
+#
+# Definitions for linux-loong64 target builds
+# Override these settings in CONFIG_SITE.Common.linux-loong64
+#-------------------------------------------------------
+
+# Include definitions common to all Linux targets
+include $(CONFIG)/os/CONFIG.Common.linuxCommon
+
+ARCH_CLASS = loongarch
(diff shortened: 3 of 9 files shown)


anjohnson:
Core Group: We can't test builds on this architecture.

Does Ubuntu provide a cross-compiler for this target yet? We would like to add GitHub Actions builds for new targets, but that does require at least an Ubuntu PPA with such a compiler to be available.

kira-96:
I cannot provide an Ubuntu PPA with this cross-compiler. Currently, the official provides only a download link for the cross-compiler. Is this approach feasible?
[loongson gnu toolchain](http://www.loongnix.cn/zh/toolchain/GNU/)
[loongson gnu toolchain x86_64-loongarch64-linux-gnu](http://ftp.loongnix.cn/toolchain/gcc/release/loongarch/gcc8/loongson-gnu-toolchain-8.3.novec-x86_64-loongarch64-linux-gnu-rc1.1.tar.xz)

``` sh
CC_PREFIX=/opt/loongson-gnu-toolchain-8.3.novec-x86_64-loongarch64-linux-gnu-rc1.1
export PATH=$CC_PREFIX/bin:$PATH
export LD_LIBRARY_PATH=$CC_PREFIX/lib:$LD_LIBRARY_PATH
export LD_LIBRARY_PATH=$CC_PREFIX/loongarch64-linux-gnu/lib64:$LD_LIBRARY_PATH
```

anjohnson:
Core Group: We are reluctant to add build files for a new target architecture that is not yet in wide use, [we read](https://wiki.debian.org/Ports/loong64) that currently loongarch CPUs are not available to us. We are happy to leave this PR open until the architecture becomes popular and other EPICS users ask for it, but we don't want to be responsible for maintaining a new architecture until then.

Thank-you for contributing to EPICS.

mdavidsaver:
As an example.  From 2007 (e96f5bc1adfbe9eebd40d3774ce995b42d9236b1)  until 2021 (6d96ea06d06ed51e635ad57bc616e1b0253501f3) we included build configuration files for Linux on [CRIS](https://en.wikipedia.org/wiki/ETRAX_CRIS), another single vendor architecture.  It isn't clear if/when Base was actually used, or even usable, on that platform.

kira-96:
Thank you for following this PR. I am not sure if anyone else has attempted to run `Base` on this architecture besides us, so I will maintain a repository for this architecture to ensure that `Base` can run on it, which is not difficult.

I hope that in the future, `Base` can officially support this architecture. Thanks.

configure/os/CONFIG.Common.linux-la64
```diff
@@ -0,0 +1,13 @@
+# CONFIG.Common.linux-la64
+#
+# Definitions for linux-loongarch64 target builds
```
anjohnson: Could you also replace `linux-loongarch64` with `linux-la64` in the comments as well please. Other files below also have similar comments.

anjohnson (changes requested):
The EPICS Core Developers Group will still not want to merge this PR until our CI system has the ability to at least compile for this target architecture (either native or cross-compiling) and I haven't heard that this situation has changed yet.

I do prefer the new shorter target name though, and since you are updating the PR I want to point out a few more minor changes that I'd want to see before we would merge it.

Thank-you.

kira-96:
@anjohnson I have modified the code based on your suggestions.

Regarding the CI system:

I tried to modify the build script (commit [8316f14](https://github.com/kira-96/epics-ci-scripts/commit/8316f1471f0345c89538fccc35b5574e8d335318)), and now it is possible to cross compile the `base` on an x64 computer. see [job/23253686937](https://github.com/kira-96/epics-base/actions/runs/8486777397/job/23253686937)

But I don't think this is a good method because I can't install the cross compiler via `apt`.

configure/os/CONFIG.Common.linux-la64
kira-96: > Could you also replace `linux-loongarch64` with `linux-la64` in the comments as well please. Other files below also have similar comments.

Today, I saw an article from the Longson community [如何称呼龙架构？](https://areweloongyet.com/docs/loong-or-loongarch/), and realized that it is necessary to keep the naming conventions of the architecture consistent with those of other open source communities.
Some communities choose to use shorter names like `loong` or `loong64` to refer to the `LoongArch` after the discussion.
Why not call it `la64`?
Because there used to be an architecture called `ia64`, although it has declined, it still exists in the memory of many people. People generally avoid situations where two names have only one difference in `i/l/1` and `O/0` when using Latin letters, which can easily lead to confusion.

|Distro|ARCH identifier|
|------|----------|
|AOSC OS|`loongarch64`|
|Debian|<ul><li>old world：`loongarch64`</li><li>new world：`loong64`</li></ul>|
|Gentoo|`loong`|
|Loong Arch Linux|`loong64`|
|RPM|`loongarch64`|
|Slackware|`loong64`|

|Language|ARCH identifier|
|----|----|
|C#|[`LoongArch64`](https://learn.microsoft.com/zh-cn/dotnet/api/system.runtime.interopservices.architecture?view=net-8.0)|
|Go|`GOARCH=loong64`|
|Rust|`loongarch64-unknown-linux-gnu`|


So I chose to use `loong64` as the name to refer to the LoongArch.

configure/os/CONFIG_SITE.Common.linux-loong64
```diff
@@ -0,0 +1,48 @@
...
+#    OPT_CFLAGS_YES, WARN_CFLAGS, SHRLIB_LDFLAGS
+# They must be set in CONFIG_SITE.linux-loong64.linux-loong64 or for
+# cross-builds in CONFIG_SITE.<host-arch>.linux-loong64 instead.
+
+# Tune GNU compiler output for a specific cpu-type
+# (e.g. loongarch64, la264, la364, la464 etc.)
+GNU_ARCH_CFLAGS = -march=loongarch64
+GNU_TUNE_CFLAGS = -mtune=loongarch64
+# Enable soft-float feature (e.g. none, 32, 64)
+GNU_DEP_CFLAGS = -mfpu=none
```
wuruilong01: According to the manual when march is loongarch64, mfpu needs to be 64 to match, otherwise a compile error will occur. The reference link is as follows: [https://loongson.github.io/LoongArch-Documentation/LoongArch-toolchain-conventions-EN.html](url)

anjohnson:
Core Group: Reviewed, keeping this PR open until it's possible to compile this architecture in our CI builds. Apparently a cross-compiler may become available in Ubuntu 25.04, which GitHub Actions doesn't have yet.

mdavidsaver:
> Apparently a cross-compiler may become available in Ubuntu 25.04 ...

I was mistaken this morning.  [`g++-loongarch64-linux-gnu`](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=g%2B%2B-loong&searchon=names) appears to be available beginning with Ubuntu 24.04, which is now available for GHA jobs as as `ubuntu-latest`.  So it may now be possible to add a job to cross build for `loongarch64-linux-gnu` following the pattern of ARM32.

https://github.com/epics-base/epics-base/blob/f911f8ca3eb94e0a7be3e3ed26166cc63ad8dac4/.github/workflows/ci-scripts-build.yml#L186-L190

mdavidsaver:
oops.  I did not read closely enough and am doubly mistaken.  The `g++-loongarch64-linux-gnu` cross compiler is added in 24.10, which is __not__  `ubuntu-latest`.

However, some loongarch packages are available in 24.04.  https://packages.ubuntu.com/search?keywords=loong  The current version of our ci-scripts tooling won't automatically handle these though.

So things progress, slowly...

kira-96:
Actually, you are right. Ubuntu 24.04 already has the loongarch64 cross compiler, which is called [g++-13-loongarch64-linux-gnu](https://packages.ubuntu.com/search?suite=noble&section=all&arch=any&keywords=g%2B%2B-13-loong&searchon=names) or [g++-14-loongarch64-linux-gnu](https://packages.ubuntu.com/search?suite=noble&section=all&arch=any&keywords=g%2B%2B-14-loong&searchon=names). But this is not convenient to use, when you use it, you need to call `loongarch64-linux-gnu-g++-14`. It may be necessary to modify [ci-scripts](https://github.com/epics-base/ci-scripts) in order for CI to run correctly.

mdavidsaver:
Looking again at compiler availability in github actions [runners](https://github.com/actions/runner-images), I think a loongarch64 compiler will become available for easy use once Ubuntu 26.04 LTS is released, and available.  Maybe some time this fall?

Debian 13 includes [`g++-loongarch64-linux-gnu`](https://packages.debian.org/search?keywords=g%2B%2B-loongarch64-linux-gnu).  I have not checked for availability with Fedora.  A container build could provide a more immediate route to adding an automatic build for loongarch64.  See also the `docker:` jobs in `.github/workflows/ci-scripts-build.yml`.

anjohnson:
Core: @mdavidsaver or someone else to check whether this can cross-build on Ubuntu-26.04 release. We won't require QEMU to run tests to merge this, but eventually that will be desirable when available.

PR #335, opened 2023-03-06
For links, treat "" the same as unset.

mdavidsaver:
Reverse an apparent behavior change some time since the 3.14 series where a link which is never set (implicitly a const link with NULL) behaves differently to an empty string (const link with empty string).

cf. https://epics.anl.gov/tech-talk/2023/msg00279.php

```
record(longout, "lo1") {
    field(DOL, "")
    field(VAL, "42")
}
record(longout, "lo2") {
    field(VAL, "42")
}
record(longout, "lo3") {
    field(DOL, "42")
}
```

All should result in `VAL` set to `42`.  However at present, `lo1.VAL` is set to `0`.


mdavidsaver:
Applied as 3b484f58d3f12fa0fb9d15fc09f0f654569f97c8

Issue #312, opened 2022-12-21
dbLoadRecords() should warn or error if an alias is specified with a field part.


anjohnson:
See https://epics.anl.gov/tech-talk/2022/msg01938.php

Note that this issue is _not_ about supporting aliases that can point to fields, that would be a much larger and non-trivial change.

EmilioPeJu:
I would like to do this one

EmilioPeJu:
Fixed by PR 336

anjohnson:
This was fixed by #336

PR #336, opened 2023-03-07
Validate target record name when creating an alias

EmilioPeJu:
This fixes issue https://github.com/epics-base/epics-base/issues/312 by printing an error when a field is specified

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1257,7 +1257,7 @@ static void dbAlias(char *name, char *alias)
     DBENTRY dbEntry;
     DBENTRY *pdbEntry = &dbEntry;
 
-    if(dbRecordNameValidate(alias))
+    if(dbRecordNameValidate(alias) || dbRecordNameValidate(name))
         return;
 
     dbInitEntry(pdbbase, pdbEntry);
--- a/modules/database/test/ioc/db/dbStaticTest.c
+++ b/modules/database/test/ioc/db/dbStaticTest.c
@@ -291,6 +291,18 @@ static void testDbVerify(const char *record)
     dbFinishEntry(&entry);
 }
 
+static void testWrongAliasRecord(const char *filename)
+{
+    FILE *fp = NULL;
+    dbPath(pdbbase,"." OSI_PATH_LIST_SEPARATOR "..");
+    dbOpenFile(pdbbase, filename, &fp);
+    if(!fp) {
+        testAbort("Unable to read %s", filename);
+    }
+    testOk(dbReadDatabaseFP(&pdbbase, fp, NULL, NULL) != 0,
+           "Wrong alias record in %s is expected to fail", filename);
+}
+
 void dbTestIoc_registerRecordDeviceDriver(struct dbBase *);
 
 MAIN(dbStaticTest)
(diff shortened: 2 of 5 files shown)


anjohnson:
Please add a unit test (within one of the existing test programs) to show that an alias that specifies a field name will be rejected and the alias is not created in that case.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1257,7 +1257,7 @@ static void dbAlias(char *name, char *alias)
     DBENTRY dbEntry;
     DBENTRY *pdbEntry = &dbEntry;
 
-    if(dbRecordNameValidate(alias))
+    if(dbRecordNameValidate(alias) || dbRecordNameValidate(name))
```
mdavidsaver, line 1260: I think this would work.  I had been thinking more along the lines of checking that `pdbEntry->pflddes==NULL` after `dbFindRecord()`, but I think this would accomplish the same.

EmilioPeJu:
Thanks Andrew and Michael for reviewing this PR, I've amended the commit to add unit tests for the case of wrong record name as well as the case of wrong alias name (which was already handled before).

modules/database/src/ioc/dbStatic/dbLexRoutines.c
EmilioPeJu: Thanks Michael, yes, I assumed a parsing error was the right way to handle this

EmilioPeJu:
I have a couple of comments that are worth mentioning:
- the way `dbOpenFile` was checked for success was wrong (given the current `dbOpenFile` implementation, there are cases in which the function returns 0 and it was read successfully), the right way seems to be checking the file pointer
- every time `dbReadDatabase*` is called, `dbFreePath` is called (deleting all the paths in `dbbase->pathPvt`) and if you want to read one more database using the same search path, you need to call dbPath again

anjohnson:
Currently your version of the dbStaticTest program is failing on GitHub because you have added 2 tests to the program without adjusting the number of tests in the `testPlan()` call that appears towards the top of the `MAIN()` routine:

```
  dbStaticTest.tap ............. 
  ok 311 - dbVerify('testrec.SCAN', 'zzzz') => 'Not a valid menu choice'
  ok 312 - dbVerify('testrec.FLNK', 'Anything works here!') => 'OK'
  All 310 subtests passed 
  
  Test Summary Report
  -------------------
  dbStaticTest.tap           (Wstat: 0 Tests: 312 Failed: 2)
    Failed tests:  311-312
    Parse errors: Bad plan.  You planned 310 tests but ran 312.
```

EmilioPeJu:
Thanks for spotting that, I've amended the commit one more time

modules/database/test/ioc/db/dbStaticTest.c
```diff
@@ -291,27 +291,43 @@ static void testDbVerify(const char *record)
...
+    if(!fp) {
         testAbort("Unable to read dbStaticTest.db");
     }
     if(dbReadDatabaseFP(&pdbbase, fp, NULL, NULL)) {
         testAbort("Unable to load %s%sdbStaticTest.db",
                   ldir, OSI_PATH_LIST_SEPARATOR);
     }
 
+    testWrongAliasRecord("dbStaticTestAlias1.db");
+    testWrongAliasRecord("dbStaticTestAlias2.db");
```
mdavidsaver, line 329: I thought I remembered an issue with testing .db parser failure triggering the `dbReadCOM: Parser stack dirty w/o error...` and failure to load subsequent correct .db files.  However, I'm not able to reproduce this now.  I tested moving the `testWrongAliasRecord()` calls before the load of `dbStaticTest.db`, which still succeeds.  Which means my memory was incorrect.

mdavidsaver (approved):
Looks ok to me.

modules/database/test/ioc/db/dbStaticTest.c
```diff
@@ -291,27 +291,43 @@ static void testDbVerify(const char *record)
...
 
+static void testWrongAliasRecord(const char *filename)
+{
+    FILE *fp = NULL;
+    dbPath(pdbbase,"." OSI_PATH_LIST_SEPARATOR "..");
+    dbOpenFile(pdbbase, filename, &fp);
+    if(!fp) {
+        testAbort("Unable to read %s", filename);
+    }
+    testOk(dbReadDatabaseFP(&pdbbase, fp, NULL, NULL),
```
simon-ess: I find this a bit confusing. I assume that `testOk` should pass if the resulting argument passes; but in this case we are using the fact that `dbReadDatabaseFP` returns a non-zero status if it _fails_. I think something like
```c
testOk(dbReadDatabaseFP(&pdbbase, fp, NULL, NULL)!=0
```
or something similar.

EmilioPeJu: Thanks for reviewing, I've amended the commit as requested

mdavidsaver:
Applied as 500a57738b149cd1124f2d1cee01ed60f637c53b

PR #337, opened 2023-03-07
fix missing newline at end of file

dirk-zimoch:
Some compilers complain about files not ending in newline. freeList.h ended in a space.

### DIFF
--- a/modules/libcom/src/freeList/freeList.h
+++ b/modules/libcom/src/freeList/freeList.h
@@ -41,4 +41,3 @@ LIBCOM_API size_t epicsStdCall freeListItemsAvail(void *pvt);
 #endif
 
 #endif /*INCfreeListh*/
- 


anjohnson:
The Single Unix Specification actually specifies that a file _must_ end in a newline to be considered a text file. Missing newlines often come from Windows editors which don't care about such rules. In this case `git blame` points to a merge from a previous Codeathon participant.

minijackson:
For a lot of commonly used editors / IDEs, this kinds of issues might be solved by adding an [EditorConfig](https://editorconfig.org/) file at the root folder of epics-base. Tabs vs spaces issues aside, it might be nice to have a configuration like this:

```dosini
root = true

# Unix-style newlines with a newline ending every file
[*]
end_of_line = lf
insert_final_newline = true
```

anjohnson:
Hi @minijackson, EPICS can be checked out on Windows and it builds fine when using the Windows convention of CR+LF line endings. Thus I don't think we need/want to enforce LF line-endings since Git is good at converting them for us, but if `insert_final_newline = true` in a `.editorconfig` file would help I'd be happy with that, please submit a PR that includes that URL in the first comment as in the website example.

Thanks!

mdavidsaver:
Applied as bcdeeed20699176c0ef0850a192f7cbaae6a75de

PR #338, opened 2023-03-07
avoid accessing dbr_text[type] when type is out of range

dirk-zimoch:
No need to call the rest of ca_dump_dbr when type is out of range. Compiler complains that dbr_text[type] may be out of range.


mdavidsaver:
Looks ok.

mdavidsaver:
Applied as 172597e0e6348308ae8c24c7a1c200bdd0e0226c

PR #339, opened 2023-03-07
fix warning "if clause does not guard..."

dirk-zimoch:
The compiler sees an `if` followed by two statements on the same line. This is often a mistake as only the first statement depends on the `if` and the author (or the reader of the code) may be misled to think both are covered. In this case, it is intended. It is just all in one line to keep the macro compact. Use explicit braces to tell the compiler the intention.


mdavidsaver:
Ok.

mdavidsaver:
Applied as e862f0e95fe79b8ccd6b692176d6e5b790a60abf

PR #340, opened 2023-03-07
Unit tests for standard record types

karlvestin:
New unit test as requested for codeathon 2023


modules/database/test/std/rec/Makefile
```diff
@@ -93,6 +93,13 @@ testHarness_SRCS += longoutTest.c
 TESTFILES += ../longoutTest.db
 TESTS += longoutTest
 
+TESTPROD_HOST += boTest
+boTest_SRCS += boTest.c
+boTest_SRCS += recTestIoc_registerRecordDeviceDriver.cpp
+testHarness_SRCS += boTest.c
```
mdavidsaver, line 104: Adding to the testHarness executable also requires a change to `modules/database/test/std/rec/epicsRunRecordTests.c`.  eg. the corresponding change to hook in `MAIN(analogMonitorTest)`.

https://github.com/epics-base/epics-base/blob/fa00572780a00bea2f57dbff69653b14bd792be0/modules/database/test/std/rec/epicsRunRecordTests.c#L16

https://github.com/epics-base/epics-base/blob/fa00572780a00bea2f57dbff69653b14bd792be0/modules/database/test/std/rec/epicsRunRecordTests.c#L32

modules/database/test/std/rec/boTest.c
```diff
@@ -0,0 +1,143 @@
...
+    epicsTimeGetCurrent(&startTime);
+
+    /* set VAL to process record */
+    testdbPutFieldOk("test_bo_rec.VAL", DBF_SHORT, TRUE);
+
+    /* wait and verfiy time */
+    testMonitorWait(test_mon);
+    epicsTimeGetCurrent(&endTime);
+
+    /* EPICS timers have a tendency to trip slightly early, hence the test tolerance is added here to avoid false positives in testing */
```
mdavidsaver, line 60: fyi. https://github.com/epics-base/epics-base/issues/106

mdavidsaver:
Looks good so far.

modules/database/test/std/rec/Makefile
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/6e8955f93ef14dbeda8596a82b87c876de6c8587

modules/database/test/std/rec/boTest.c
karlvestin: Good to know, thanks.

modules/database/test/std/rec/biTest.c
```diff
@@ -0,0 +1,115 @@
...
+
+static void test_operator_display(void){
+    /* set operator display parameters */
+    testdbPutFieldOk("test_bi_rec.ZNAM", DBF_STRING, "ZNAM_TEST");
+    testdbPutFieldOk("test_bi_rec.ONAM", DBF_STRING, "ONAM_TEST");
+    testdbPutFieldOk("test_bi_rec.DESC", DBF_STRING, "DESC_TEST");
+
+    /* verify operator display parameters */
+    testdbGetFieldEqual("test_bi_rec.ZNAM", DBF_STRING, "ZNAM_TEST");   
+    testdbGetFieldEqual("test_bi_rec.ONAM", DBF_STRING, "ONAM_TEST");
```
henrique-silva, line 55: Maybe you could set the value of the record to 0 or 1 and test that the retrieved VAL as string matches `ZNAM` and `ONAM`, this way the test includes the conversion as well, not only the field setting.

modules/database/test/std/rec/biTest.c
```diff
@@ -0,0 +1,115 @@
...
+    /* Verify alarm status is NO_ALARM*/
+    testdbGetFieldEqual("test_bi_rec.SEVR", DBF_SHORT, menuAlarmSevrNO_ALARM);
+
+    /* Set ZSV to MAJOR and verify that SEVR is now MAJOR */
+    testdbPutFieldOk("test_bi_rec.ZSV", DBF_SHORT, menuAlarmSevrMAJOR);
+    testdbGetFieldEqual("test_bi_rec.SEVR", DBF_SHORT, menuAlarmSevrMAJOR);
+
+    /* Set VAL to 1 on lined record and verify that COSV now sets the SEVR to INVALID */
+    testdbPutFieldOk("test_bi_link_rec.VAL", DBF_SHORT, TRUE);
+    testdbGetFieldEqual("test_bi_rec.SEVR", DBF_SHORT, menuAlarmSevrINVALID);
```
henrique-silva, line 91: You can also include here a test to check that `LALM` field holds the last value that caused the alarm, in this case it should be `1`

modules/database/test/std/rec/biTest.c
```diff
@@ -0,0 +1,115 @@
...
+    testdbPutFieldOk("test_bi_rec.COSV", DBF_SHORT, menuAlarmSevrINVALID);
+    
+    /* Verify alarm status is NO_ALARM*/
+    testdbGetFieldEqual("test_bi_rec.SEVR", DBF_SHORT, menuAlarmSevrNO_ALARM);
+
+    /* Set ZSV to MAJOR and verify that SEVR is now MAJOR */
+    testdbPutFieldOk("test_bi_rec.ZSV", DBF_SHORT, menuAlarmSevrMAJOR);
+    testdbGetFieldEqual("test_bi_rec.SEVR", DBF_SHORT, menuAlarmSevrMAJOR);
+
+    /* Set VAL to 1 on lined record and verify that COSV now sets the SEVR to INVALID */
```
henrique-silva: Small typo

modules/database/test/std/rec/boTest.c
```diff
@@ -0,0 +1,143 @@
...
+    diffTime = epicsTimeDiffInSeconds(&endTime, &startTime) + diffTimeTolerance;
+    testOk(diffTime >= high_time, "HIGH time %lf", diffTime);
+
+    /* verfiy that both records are set back to 0 */
```
henrique-silva: Small typo

modules/database/test/std/rec/boTest.c
```diff
@@ -0,0 +1,143 @@
...
+
+static void test_operator_display(void){
+    /* set operator display parameters */
+    testdbPutFieldOk("test_bo_rec.ZNAM", DBF_STRING, "ZNAM_TEST");
+    testdbPutFieldOk("test_bo_rec.ONAM", DBF_STRING, "ONAM_TEST");
+    testdbPutFieldOk("test_bo_rec.DESC", DBF_STRING, "DESC_TEST");
+
+    /* verify operator display parameters */
+    testdbGetFieldEqual("test_bo_rec.ZNAM", DBF_STRING, "ZNAM_TEST");   
+    testdbGetFieldEqual("test_bo_rec.ONAM", DBF_STRING, "ONAM_TEST");
```
henrique-silva, line 80: Same as comments for `bi` record. Could implicitly test the enum conversion here.

modules/database/test/std/rec/boTest.c
```diff
@@ -0,0 +1,143 @@
...
+
+    /* Set start VAL to FALSE*/
+    testdbPutFieldOk("test_bo_rec.VAL", DBF_SHORT, FALSE);
+
+    /* set alarm parameters */
+    testdbPutFieldOk("test_bo_rec.ZSV", DBF_SHORT, menuAlarmSevrNO_ALARM);
+    testdbPutFieldOk("test_bo_rec.OSV", DBF_SHORT, menuAlarmSevrMINOR);
+    testdbPutFieldOk("test_bo_rec.COSV", DBF_SHORT, menuAlarmSevrINVALID);
+    testdbPutFieldOk("test_bo_rec.IVOA", DBF_SHORT, menuIvoaSet_output_to_IVOV);
+    testdbPutFieldOk("test_bo_rec.IVOV", DBF_SHORT, FALSE);
```
henrique-silva, line 106: Should we also test the other `IVOA` actions?

modules/database/test/std/rec/biTest.c
```diff
@@ -0,0 +1,115 @@
...
+    testdbGetFieldEqual("test_bi_rec.RVAL", DBF_SHORT, TRUE);
+
+    // number of tests = 6
+}
+
+static void test_operator_display(void){
+    /* set operator display parameters */
+    testdbPutFieldOk("test_bi_rec.ZNAM", DBF_STRING, "ZNAM_TEST");
+    testdbPutFieldOk("test_bi_rec.ONAM", DBF_STRING, "ONAM_TEST");
+    testdbPutFieldOk("test_bi_rec.DESC", DBF_STRING, "DESC_TEST");
```
simon-ess, line 51: Is there any real need to test the `.DESC` field? Does the `bi` record have any particularly special functionality there?

modules/database/test/std/rec/biTest.c
```diff
@@ -0,0 +1,115 @@
...
+    testdbGetFieldEqual("test_bi_rec.ONAM", DBF_STRING, "ONAM_TEST"); 
+    testdbGetFieldEqual("test_bi_rec.NAME", DBF_STRING, "test_bi_rec");
+    testdbGetFieldEqual("test_bi_rec.DESC", DBF_STRING, "DESC_TEST");
+
+    // number of tests = 7
+}
+
+static void test_alarm(void){
+    /* set soft channel */
+    testdbPutFieldOk("test_bi_rec.DTYP", DBF_STRING, "Soft Channel");
```
simon-ess, line 70: If every test is a `Soft Channel`, maybe this should just be included in the `biTest.db` file.

modules/database/test/std/rec/biTest.c
```diff
@@ -0,0 +1,115 @@
...
+    test_soft_input();
+    test_raw_soft_input();
+    test_operator_display();
+    test_alarm();
+
+    testIocShutdownOk();
+    testdbCleanup();
+
+    return testDone();
+}
```
simon-ess: End of file newlines are missing!

modules/database/test/std/rec/biTest.c
henrique-silva: But they are not, there's a specific test for `Raw Soft Channel`. Arguably, the `DTYP` could be overriden just in that test, but I don't think it's any better for clarity.

modules/database/test/std/rec/boTest.c
simon-ess: I agree with @henrique-silva about both of these.

modules/database/test/std/rec/printfTest.c
```diff
@@ -0,0 +1,551 @@
...
+static void test_hash_flag(void){
+
+    const char format_string[] = "Format test string %#o";
+    const char result_string[] = "Format test string 014";
+
+    /* set format string */
+    testdbPutFieldOk("test_printf_rec.FMT", DBF_STRING, format_string);
+
+    /* set value on inp0 */
+    testdbPutFieldOk("test_printf_inp0_rec.VAL", DBF_SHORT, 12);
```
simon-ess: Picky, but this might be a clearer test if (as above) you specify this as an octal value `014`

modules/database/test/std/rec/biTest.c
simon-ess: Ah, I missed that. I think you're right about clarity in that case.

modules/database/test/std/rec/biTest.c
karlvestin: !

modules/database/test/std/rec/biTest.c
karlvestin: I actually prefer to keep the test code in the C file for readability, even if it does get a bit repetitive. Also more flexible for adding tests that may want a raw soft channel for instance.

modules/database/test/std/rec/biTest.c
karlvestin: Not strictly needed I guess, but it is mentioned in the record reference for bi, hence I tested it.

modules/database/test/std/rec/biTest.c
karlvestin: Fixed in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/biTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/biTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/boTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/boTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/boTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/biTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

modules/database/test/std/rec/printfTest.c
karlvestin: Resolved in https://github.com/epics-base/epics-base/pull/340/commits/28ecf0af3c3f2585f1571215c8b16a15ccbcab37

ralphlange:
%F does work fine on my laptop:
Windows 10 using `gcc.exe (Rev1, Built by MSYS2 project) 12.1.0`

The failing GitHub Actions build uses `gcc (x86_64-posix-seh, Built by strawberryperl.com project) 8.3.0`

karlvestin:
@ralphlange I am open to any non-supernatural explanation as to why the %F fails to format on this and only this platform. %f works fine on the same platform.

karlvestin:
Made https://github.com/epics-base/epics-base/issues/368 for the strange behaviour of %F on Win2019 mingw

modules/database/test/std/rec/printfTest.c
```diff
@@ -0,0 +1,568 @@
...
+    const char result_string[] = "Format test string 6729982.999000";
+
+    /* set format string */
+    testdbPutFieldOk("test_printf_rec.FMT", DBF_STRING, format_string);
+
+    /* set value on inp0 */
+    testdbPutFieldOk("test_printf_inp0_rec.VAL", DBF_DOUBLE, 6729982.999);
+
+    /* verify that string is formatted as expected */
+    #ifdef _WIN32
```
FreddieAkeroyd, line 205: Is this possibly a gcc rather than Win2019/mingw specific issue? @ralphlange mentioned this platform used gcc 8.* and I see the ubuntu tests use gcc 9 and pass, so maybe using
```
#if defined(__GNUC__)  &&  __GNUC__ < 9
```
would be sufficient

FreddieAkeroyd: I see the Appveyor Win2019 mingw build uses gcc 12.2.0 so if this issues is gcc version specific then that build should pass

modules/database/test/std/rec/aiTest.c
```diff
@@ -0,0 +1,429 @@
...
+
+    /* set soft channel */
+    testdbPutFieldOk("test_ai_rec.DTYP", DBF_STRING, "Raw Soft Channel");
+    testdbPutFieldOk("test_ai_rec.INP", DBF_STRING, "test_ai_link_rec.VAL");
+    testdbPutFieldOk("test_ai_link_rec.FLNK", DBF_STRING, "test_ai_rec");
+
+    /* set unit conversion parameters */
+    testdbPutFieldOk("test_ai_rec.ROFF", DBF_SHORT, roff);
+    testdbPutFieldOk("test_ai_rec.ASLO", DBF_SHORT, aslo);
+    testdbPutFieldOk("test_ai_rec.AOFF", DBF_SHORT, aoff);
```
henrique-silva, line 97: I know the default value of `LINR` is already `NO CONVERSION`, but it would be good to explicitly set it here, for the sake of clarity of the test routine.

modules/database/test/std/rec/aiTest.c
```diff
@@ -0,0 +1,429 @@
...
+
+    /* set unit conversion parameters */
+    testdbPutFieldOk("test_ai_rec.ROFF", DBF_LONG, roff);
+    testdbPutFieldOk("test_ai_rec.ASLO", DBF_LONG, aslo);
+    testdbPutFieldOk("test_ai_rec.AOFF", DBF_LONG, aoff);
+    testdbPutFieldOk("test_ai_rec.ESLO", DBF_LONG, eslo);
+    testdbPutFieldOk("test_ai_rec.EOFF", DBF_LONG, eoff);
+    /* Since our raw soft input does not support EGUL and EGUF conversion we set them to lagre values here just to check they dont break anything */
+    testdbPutFieldOk("test_ai_rec.EGUL", DBF_LONG, egul);
+    testdbPutFieldOk("test_ai_rec.EGUF", DBF_LONG, eguf);
```
henrique-silva: A suggestion here could be to "mock" a device support calculation by using calc records to perform the EGUL/EGUF calculation.

modules/database/test/std/rec/aiTest.c
karlvestin: Ok. Correcting this.

modules/database/test/std/rec/aiTest.c
karlvestin: I am not convinced this would add any value to the test code. The calc record would just set the same conversion fields I am already testing.

modules/database/test/std/rec/printfTest.c
karlvestin: I suspect this may not have anything to do with the %F specifier and everything to do with the fact that the number I am trying to convert does not fit into a float without truncation. Testing a bit to see if I can get rid of this ugly todoTest.

karlvestin: I was wrong. A float with fewer digits does not fare any better.

karlvestin: Appveyor shows that several windows based platforms fails, not only on %F but also on %lx. I have skipped these two tests on all windows platforms for now and updated issue #368 accordingly. Resolving this conversation in favor of the issue for future discussion.

FreddieAkeroyd:
looking at https://github.com/karlvestin/epics-base/commits/bo-test and examining the results of the github actions i see:

* d64a262 "mingw win2019" failed 
* 3432c4d "mingw win2019" worked

so checking for gcc < 9 would appear to resolve the issue there and Win2019 Visual Studio 2019 builds work too in github actions

So why does appveyor fail? Looking at https://ci.appveyor.com/project/epics-base-7/epics-base/builds/46510746 I see that 

* Win2019 VS2019 x64 work (this is also tested by github actions)
* Win2019 VS2019 x86 fails (this is not tested by github actions)
* VS2013 and earlier fail
* mingw fails

The x86 failure is
```
not ok 73 - dbGetField("test_printf_rec.VAL", 0) -> "Format test string a0a1c0c1" == "Format test string ffffffff"
```
which looks to be from `%llx` test with an `epicsInt64` . The old visual studio and mingw are
```
not ok 36 - dbGetField("test_printf_rec.VAL", 0) -> "Format test string 6729982.999000" == "Format test string "
```
The mingw on appveyor is
```
gcc (Rev10, Built by MSYS2 project) 12.2.0
```
which is more recent than the github action ubuntu which is gcc 9, so i would have thought gcc 12 would be ok. However i don't know if mingw handles printf itself or passes it to an underlying microsoft library, so gcc version may be irrelevant in the mingw case.

@ralphlange  you say above Windows 10 using `gcc.exe (Rev1, Built by MSYS2 project) 12.1.0` worked for you, so it is surprising  that `gcc (Rev10, Built by MSYS2 project) 12.2.0` on appveyor does not work. Maybe there is an issue with the appveyor installation, or maybe the epics appveyor build is linking to an old/wrong mingw library?

modules/database/test/std/rec/printfTest.c
```diff
@@ -0,0 +1,573 @@
...
+
+    const char format_string[] = "Format test string %lx";
+    const char result_string[] = "Format test string a0a1c0c1";
+
+    /* set format string */
+    testdbPutFieldOk("test_printf_rec.FMT", DBF_STRING, format_string);
+
+    /* set value on inp0 */
+    testdbPutFieldOk("test_printf_inp0_rec.VAL", DBF_LONG, 0xba0a1c0c1);
+
```
FreddieAkeroyd: Should that start `0xa0a1` and not `0xba0a1` ?

FreddieAkeroyd: I think that may be the cause of the x86 failure above with `%lx`-  ignore my post about `%llx` as i was looking at the wrong test

FreddieAkeroyd: After above change, i suspect that the following could work to protect just %F test
```
// visual studio less than 2015 does not support %F
// mingw/gcc also fails, suspect this may be gcc version
// related and checking __GNUC__ could resolve but 
// initial attempts didn't work so excluding mingw entirely for now
#ifdef _WIN32
#if (defined(_MSC_VER) && _MSC_VER < 1900) || defined(_MINGW)
testTodoBegin("Fails on windows with old visual studio versions and mingw");
#endif
#endif
```

karlvestin:
All comments are addressed and all test and checks now pass on this PR. The AppVeyor tests that failed seems to be random intermittent failures on a test that is not impacted by this PR. Request merge while I remember what I was doing.

FreddieAkeroyd:
Hi, i think the `testdbPutFieldOk("test_printf_inp0_rec.VAL", DBF_LONG, 0xba0a1c0c1);` line mention above is not correct, sorry if my later comment about lx and llx made you think i had changed my mind.  I feel if the `b`  was removed it would pass on win32. With the `b` the value passed is > 32bits so is probably causing an issue in some circumstances.

karlvestin:
> Hi, i think the `testdbPutFieldOk("test_printf_inp0_rec.VAL", DBF_LONG, 0xba0a1c0c1);` line mention above is not correct, sorry if my later comment about lx and llx made you think i had changed my mind. I feel if the `b` was removed it would pass on win32. With the `b` the value passed is > 32bits so is probably causing an issue in some circumstances.

Sadly it does not. In appveyour I get the following:

not ok 73 - dbGetField("test_printf_rec.VAL", 0) -> "Format test string a0a1c0c1" == "Format test string ffffffff"

Will revert last commit, since I feel it is not in scope for this test to trouble shoot weird behavior in specific windows versions.

simon-ess:
If the test only fails on win32, can you just disable that test for that architecture?

FreddieAkeroyd:
It only fails on 32bit windows, 64bit windows passes. I thought that was due to  `0xba0a1c0c1` being bigger than 32bits and leading to undefined behaviour, but as `0xa0a1c0c1` still results in `0xffffffff` my only guess would be it could be a signed/unsigned issue somewhere in the intervening logic. EPICS doesn't do unsigned types natively, so at some point  `0xa0a1c0c1` may get converted/cast to `long` and that behaves differently on 32bit and 64bit windows. That may be testable by making the test data `0x20a1c0c1` for example so it fits into both signed and unsigned data types, but it could of course be that printf is broken on 32bit windows.

However it does raise the question about the logic of the test - that we are testing  `DBF_LONG`  against `0xa0a1c0c1`, if `0xa0a1c0c1` was taken to be a positive number that we needed to store then it overflows the signed long data type of DBF_LONG, so are we making an assumption that the compiler will interpret this hex number as a negative number rather than overflow?

karlvestin:
> It only fails on 32bit windows, 64bit windows passes. I thought that was due to `0xba0a1c0c1` being bigger than 32bits and leading to undefined behaviour, but as `0xa0a1c0c1` still results in `0xffffffff` my only guess would be it could be a signed/unsigned issue somewhere in the intervening logic. EPICS doesn't do unsigned types natively, so at some point `0xa0a1c0c1` may get converted/cast to `long` and that behaves differently on 32bit and 64bit windows. That may be testable by making the test data `0x20a1c0c1` for example so it fits into both signed and unsigned data types, but it could of course be that printf is broken on 32bit windows.
> 
> However it does raise the question about the logic of the test - that we are testing `DBF_LONG` against `0xa0a1c0c1`, if `0xa0a1c0c1` was taken to be a positive number that we needed to store then it overflows the signed long data type of DBF_LONG, so are we making an assumption that the compiler will interpret this hex number as a negative number rather than overflow?

My thinking behind the test case was to send a number larger-than-long with DBF_LONG and check that it was correctly truncated and formatted by printf. I could retest with 0x90a1c0c1, so its a big but valid and positive long instead if you think that makes more sense?

FreddieAkeroyd:
I think `0x90a1c0c1` may give a similar issue, as a decimal integer it takes more than 31 bits to represent which could result in an overflow rather than negative number. You would need to use a number less than `0x80000000` so `0x70a1c0c1` for example to be safe. Is it defined in any standard what the correct behaviour should be? I'm guessing 0x is treated as an unsigned number so if you assign a 32bit unsigned number to a 32bit signed type and the unsigned number is too big to store with the correct sign should it (a) just copy/cast the bit pattern resulting in a negative signed number or (b) record some other number to indicate it can't store so high a positive number.

FreddieAkeroyd:
It used to be that unsigned integer overflow had a defined behaviour by the standard, but signed integer overflow did not. Maybe that has changed and x86 builds have not caught up?

karlvestin:
After implementing change proposed by @FreddieAkeroyd tests now runs on all architectures. 

Requesting merge.

modules/database/test/std/rec/aiTest.c
```diff
@@ -0,0 +1,430 @@
...
+    test_raw_soft_input();
+    test_operator_display();
+    test_no_linr_unit_conversion();
+    test_slope_linr_unit_conversion();
+    test_linear_linr_unit_conversion();
+    test_bpt_conversion();
+    test_smoothing_filter();
+    test_udf();
+    test_alarm();
+    /* TODO: Investigate and make the AFTC test work correctly
```
simon-ess: If these tests aren't working, it might be a good idea to document better what the issue is.

simon-ess:
It would be good to do a rebase of these commits to clean up the history a little bit.

anjohnson:
Core Group note: Squash commits when merging this. Merge approved.

mdavidsaver:
Applied as of 39b6fa26dad299596deb5dd4d07b874f7b0e2208

PR #341, opened 2023-03-07
ci: add cross-compilation tests for aarch64, arm soft and hard float

minijackson:
Part of Codeathon 2023.

Pinging @mdavidsaver as that was his proposal, and @ralphlange as this involves CI.

Right now, only tests with gcc and using shared libraries are run. Let me know if you want me to add more.

Linking implementation PR: epics-base/ci-scripts#69

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -151,6 +151,26 @@ jobs:
             configuration: default
             name: "Win2019 mingw"
 
+          # Cross builds
+
+          - os: ubuntu-latest
+            cmp: gcc
+            configuration: default
+            name: "Cross linux-aarch64"
+            cross: linux-aarch64
+
+          - os: ubuntu-latest
+            cmp: gcc
+            configuration: default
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Rather than adding another copy of the `steps:`, would you be willing to look at switching the RTEMS jobs to `CI_CROSS_TARGETS`?  eg. an example of what this would look like from [PVXS](https://github.com/mdavidsaver/pvxs/blob/9c546f0ed448ec9f17d9b676119fa98d66217f6a/.github/workflows/ci-scripts-build.yml#L77).

minijackson:
For future reference, @mdavidsaver did the RTEMS refactoring in d9052f71056ab38217344e0f6d395e6937f15965. I just rebased, so if CI passes it should be good to go.

mdavidsaver:
The added jobs passed, so I think this is ok.

mdavidsaver:
Applied as 9f97f25669d9fd05b0fa9f21f5dc2487906eada5

PR #342, opened 2023-03-07
editorconfig: add initial version forcing final newlines

minijackson:
As per comment https://github.com/epics-base/epics-base/pull/337#issuecomment-1458388011

CC @anjohnson 

This file may be extended later if we want to enforce tabs / space length by filetype.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# Documentation for this file: https://EditorConfig.org
+
+root = true
+
+# Unix-style newlines ending every file,
+# as some compilers complain about files not ending in newline
+[*]
+insert_final_newline = true
--- a/.github/workflows/check-editorconfig.yml
+++ b/.github/workflows/check-editorconfig.yml
@@ -1,0 +1,13 @@
+name: Check EditorConfig
+
+on:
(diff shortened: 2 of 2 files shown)


minijackson:
If I remember correctly, there is the possibility of adding a GitHub action that checks that the EditorConfig file is upheld in the codebase, I'll add it in this PR in a bit.

minijackson:
There we go! It should work now.

anjohnson:
This looks great to me (and Ralph)!

mdavidsaver:
Just to note, this PR adds a 3rd party github action.  At the moment, the repository policy allows any action.  I don't see an issue since we don't currently have any secrets associated with this repo.

anjohnson:
I did see a note on another repo recently where the project's maintainer explicitly specified the SHA of the 3rd party action script instead of its version number. The owner of that action could move the version tag at a later date but couldn't (easily) change the commit with that SHA to add anything malicious. Is it worth our adopting that convention?

mdavidsaver:
Applied as of c7a769e5da9ce7253e58a321e2fd7e0da1ba8cf3

PR #344, opened 2023-03-07
make buffer large enough for any argument value

dirk-zimoch:
The compiler complains that a huge `int` may overrun the buffer `char rform[10]` when printing the argument with `sprintf( rform, "\\%.3o", c )` . In reality, only chars are passed as `c`, but the function takes an int argument and I hesitate to change that. Increasing the buffer size a bit is the least invasive solution.


mdavidsaver:
ok.

mdavidsaver:
Applied as 0b01fb20db0102d045a0daba24453ea6f37c8a96

PR #345, opened 2023-03-07
Fix generalTimeReport interest level argument type

henrique-silva:
Currently the `interest_level` argument is being parsed as `iocshArgArgv`, which makes it always be tested as true in the `generalTimeReport` function.

Setting to iocshArgInt actually makes the function verbosity to be able to be controlled by the value


mdavidsaver:
Good catch!

mdavidsaver:
Applied as 3e51491628096642b26db7ca34999673463de44f

PR #346, opened 2023-03-07
fix compiler warning on 32 bit Linux

dirk-zimoch:
When compiling for 32 bit Linux, the compiler complains because a 64 bit integer is converted to a (32 bit) pointer. The cause is in the macro `ARR` which expands to something like `condition ? a_32_bit_integer : a_64_bit_integer`. This promotes the result to a 64 bit integer, even when the 32 bit integer is chosen. The intermediate cast to `uintptr_t` tells the compiler that we know what we are doing.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdElfFindAddr.c
+++ b/modules/libcom/src/osi/os/posix/osdElfFindAddr.c
@@ -639,7 +639,7 @@ epicsFindAddr(void *addr, epicsSymbol *sym_p)
 
     if ( nearest.raw && ( (idx = ARR(c,nearest,0,st_name)) < es->strMap->max ) ) {
         sym_p->s_nam = strtab + idx;
-        sym_p->s_val = (char*) ARR(c, nearest, 0, st_value) + es->addr;
+        sym_p->s_val = (char*)(uintptr_t) ARR(c, nearest, 0, st_value) + es->addr;
     }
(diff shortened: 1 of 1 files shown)


mdavidsaver:
I think this is ok.

@dirk-zimoch For these issues/PRs, can you include the verbatim text of the warning?

dirk-zimoch:
The warning was:
```
In file included from ../osi/os/Linux/osdFindAddr.c:11:
../osi/os/posix/osdElfFindAddr.c: In function `epicsFindAddr':
../osi/os/posix/osdElfFindAddr.c:642: warning: cast to pointer from integer of different size
```

mdavidsaver (approved):
wrt. `uintptr_t`, I find [references](https://manpages.ubuntu.com/manpages/lunar/man3/intptr_t.3type.html) that this is "C99 and later; POSIX.1-2001 and later.".  I think this won't be a problem in `os/posix` code with currently supported gcc/clang.

dirk-zimoch:
I could use `size_t` instead but found `uintptr_t` cleaner.
No problem at least on my oldest Linux system, which is MontaVista 4.0 with gcc 3.4.3.

mdavidsaver:
Applied as f41f11c7f6e4c27a6e25eec36e3cbe3cf6a79688

Issue #155, opened 2021-04-06
Apply some GCC attributes


mdavidsaver:
GCC (and usually clang) has a number of `__attribute__(())`s which it might be useful to (conditionally) apply to some Base functions to facilitate static analysis/optimization by the compiler.

https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

eg. `alloc_size(...)` to provide size hints, or `malloc(deallocator, ptr-index)` to match alloc and dealloc functions, could be added to various memory allocation functions.  eg.

```c
__attribute__((malloc(freeListFree,2)))
void* freeListMalloc(void *pvt)

__attribute__((alloc_size(1,2),malloc(free,1),returns_nonnull))
void* callocMustSucceed(size_t count, size_t size,  const char *errorMessage);
```

Some of the simpler, or more common, might be abstracted through `compilerDependencies.h` (eg. `deprecated/error/warning` or `warn_unused_result`).  Though the more specialized/complicated probably have to be special cases.

Complications include identifying supported GCC versions ranges.  (although GCC usually treats unknown attributes as a warning).

FreddieAkeroyd:
Visual Studio has some similar notation  https://docs.microsoft.com/en-us/cpp/code-quality/using-sal-annotations-to-reduce-c-cpp-code-defects?view=msvc-160 but I think it is just for static analysis. I'm not sure if it is possible to combine the two notations in a common way

mdavidsaver:
Are the MSVC annotations meant to be used on declarations, definitions, or both?  The examples I've seen so far are on definitions.  The GCC annotations are exclusively for declarations, and it would actually be nice if the MSVC were definitions only as this could avoid some extra clutter in headers.

Also, the MSVC notation examples look more preprocessor friendly (no extra commas).  So it might be easier to add than the GCC notation, which might need `#ifdef __GNUC__` in more cases.

FreddieAkeroyd:
The examples on https://docs.microsoft.com/en-us/cpp/code-quality/best-practices-and-examples-sal?view=msvc-160 seem to have a mixture of use on declarations and definitions. I'm afraid i've never actually used this annotations, it may be some parts  only make sense on the definition, some may work well on the declaration and allow static analysis to check both caller usage and implementation. Maybe you can annotate both declaration and definition separately and it combines the relevant bits during static analysis

dirk-zimoch:
[...]

PR #347, opened 2023-03-07
Adding gcc attributes

rosesyrett:
added some gcc attributes to things, plus a few simple test cases that need to be removed

Note; the arrayFunction in the errlog.c will be removed. This was just for my benefit in understanding pointers/ how allocation and attributes work.

### DIFF
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -719,3 +719,16 @@ static void errlogThread(void)
     if(wakeFlusher)
         epicsEventMustTrigger(pvt.waitForSeq);
 }
+
+int * arrayFunction(int size, size_t size_t) {
+    static int * array;
+
+    //array = calloc(size, sizeof(int));
+    //array = malloc(size * sizeof(int));
+    array = malloc(size * size_t);
+
+    for (int i=0; i< size;i++){
+        array[i] = i;
+    }
+    return array;
+}
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -719,16 +719,3 @@ static void errlogThread(void)
     if(wakeFlusher)
         epicsEventMustTrigger(pvt.waitForSeq);
 }
-
-int * arrayFunction(int size, size_t size_t) {
-    static int * array;
-
-    //array = calloc(size, sizeof(int));
-    //array = malloc(size * sizeof(int));
-    array = malloc(size * size_t);
-
-    for (int i=0; i< size;i++){
-        array[i] = i;
-    }
-    return array;
-}
--- a/modules/libcom/test/epicsErrlogTest.c
+++ b/modules/libcom/test/epicsErrlogTest.c
@@ -88,6 +88,7 @@ typedef struct {
 } clientPvt;
 
 static void testLogPrefix(void);
+static void testArrayFunction(void); //some comment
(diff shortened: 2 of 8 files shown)


rosesyrett:
resolves https://github.com/epics-base/epics-base/issues/155

modules/libcom/src/error/errlog.c
```diff
@@ -719,3 +719,16 @@ static void errlogThread(void)
...
+
+    //array = calloc(size, sizeof(int));
+    //array = malloc(size * sizeof(int));
+    array = malloc(size * size_t);
+
+    for (int i=0; i< size;i++){
+        array[i] = i;
+    }
+    return array;
+}
```
minijackson: github reports "no newline at end of tile"

modules/libcom/test/epicsErrlogTest.c
```diff
@@ -552,3 +555,18 @@ static void readFromClient(void *pParam)
...
+
+    if (myArray != NULL) {
+        testOk(1, "array initialised, elems are %d, %d, %d", myArray[0], myArray[1], myArray[2]);
+    }
+    else {
+        testAbort("test failed, array is Null.");
+    }
+
+    free(myArray);
+}
```
minijackson: github reports "no newline at end of tile"

minijackson (commented):
Are you using a text editor on windows? GitHub reports that some newlines are missing at some end of files, which can trigger compiler warnings

mdavidsaver:
I'm marking this as a "draft" pending some cleanup.  (eg. removal of `arrayFunction()`)

mdavidsaver:
I have enabled github actions builds for this PR, which is failing with

```
  ../error/errlog.h:250:1: error: wrong number of arguments specified for ‘malloc’ attribute
```

I'm disappointed that this is an error.  I had hoped that GCC would only warn about malformed attributes.

This build is running with

```
$ gcc --version
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
```

modules/libcom/src/error/errlog.c
rosesyrett: I am using vscode on a linux machine, but I will look into this now

minijackson: You should be able to force vscode to insert newlines by adding this file at the top of epics-base: https://github.com/epics-base/epics-base/blob/01b6ea9c1d35f2cba80fc7a3ef30da5cccd4ff6a/.editorconfig

rosesyrett:
@mdavidsaver that is unfortunate, I will have a look at how to force the compiler to ignore attribute syntax for specific versions of the compiler...

modules/libcom/src/error/errlog.c
rosesyrett: Thanks for pointing this out; I have changed my IDE settings so it automatically inserts a newline at the end of any edited files. I have since deleted this function, which I really just used to aid my understanding, but hopefully this should not be a problem again.

rosesyrett:
Ok, I have discovered why the failure above was happening. It was to do with the redundant `arrayFunction` I had left behind in the `errlog.h` file... this is now removed.

Have discussed with Michael any other gcc attributes that are desired, and I think we've concluded for now the changes I've made are fine. So if all the tests pass I think it's good to merge (if it looks good to everyone else).

PR #350, opened 2023-03-08
replace deprecated decrementing volatile with atomic decrement

dirk-zimoch:
The compiler complains that using the `--` operator on a `volatile` variable is deprecated (likely because `volatile` has no defined meaning in this context). Maybe the author meant to decrement atomically. Commit f8647ac5c6 which introduced the `volatile` qualifier has the message "Guard against overzealous optimizers." No idea what the "overzealous optimizer" actually did wrong.

### DIFF
--- a/modules/libcom/test/epicsTimerTest.cpp
+++ b/modules/libcom/test/epicsTimerTest.cpp
@@ -21,6 +21,7 @@
 #include "epicsEvent.h"
 #include "epicsAssert.h"
 #include "epicsGuard.h"
+#include "epicsAtomic.h"
 #include "tsFreeList.h"
 #include "epicsUnitTest.h"
 #include "testMain.h"
@@ -84,7 +85,7 @@ private:
     delayVerify & operator = ( const delayVerify & );
 };
 
-static volatile unsigned expireCount;
+static int expireCount;
 static epicsEvent expireEvent;
 
 delayVerify::delayVerify ( double expectedDelayIn, epicsTimerQueue &queueIn ) :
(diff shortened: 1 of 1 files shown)


dirk-zimoch:
PS: I also could not find a problem with the test when simply removing the `volatile` keword. But I don't know on which architecture a problem occured.

mdavidsaver:
I've realized that the name `expireCount` is overloaded within `epicsTimerTest.cpp` in both the global scope, and as a class static data member `cancelVerify::expireCount`.  Confusing...

> PS: I also could not find a problem with the test when simply removing the `volatile` keword. But I don't know on which architecture a problem occured.

A compiler can't recognize that the timer callbacks are invoked from another thread.  Dynamic analysis like valgrind could.  So strictly speaking, both of the `expireCount` globals should use atomic access.

Also, `epicsAtomic.h` has C++ wrappers which can save a few characters as they are overloaded by type.  eg. `epics::atomic::set(size_t&, size_t)` and `epics::atomic::decrement(size_t&)`.

dirk-zimoch:
New approach using `epics::atomic::set` and `epics::atomic::decrement`. I also decided that using an `int` instead of `size_t` is better because `size_t` is for memory sizes, not for large unsigned integers (that would be `uintmax_t`) and I consider it bad style to abuse it for things like counters. Also this counter counts down from 25 only.

mdavidsaver:
> I also decided that using an `int` instead of `size_t` is better because `size_t` is for memory sizes, not for large unsigned integers (that would be `uintmax_t`) ...

> ... Also this counter counts down from 25 only.

I do not share this opinion, and would not accept a change like this in other circumstances.  However, I don't think it is worth quibbling over in this instance (test code).

imo. epicsTimerTest would benefit from some future work.  eg. the confusion with `cancelVerify::expireCount`, which appears to have the same concurrency issues as the global `expireCount`.

mdavidsaver:
Applied as 912a82c0b521d8e894b0930d25582e98fcfe5686

PR #353, opened 2023-03-08
silence "does not compute number of elements" warning

dirk-zimoch:
Recent gcc (from version 11 on) warns when dividing the size of an array by the size of something with a type different from the type of the array elements. This is because "usually" such a division is used to compute the number of elements in the array and dividing by something else gives a "wrong" result. However, in this case the mismatch between array type and divider is intentional. The compiler tells how to silence the warning in this case:
```
../db/dbTest.c:514:57: warning: expression does not compute the number of elements in this array; element type is 'long int', not 'epicsInt8' {aka 'signed char'} [-Wsizeof-array-div]
  514 |     no_elements = MIN(addr.no_elements,((sizeof(buffer))/sizeof(epicsInt8)));
      |                                                         ^
  472 |     long       buffer[400];
      |                ^~~~~~
../db/dbTest.c:514:57: note: add parentheses around the second 'sizeof' to silence this warning
```

The extra parentheses around the first `sizeof` are not required.

### DIFF
--- a/modules/database/src/ioc/db/dbTest.c
+++ b/modules/database/src/ioc/db/dbTest.c
@@ -506,62 +506,62 @@ long dbtgf(const char *pname)
     ret_options=0;
 
     dbr_type = DBR_STRING;
-    no_elements = MIN(addr.no_elements,((sizeof(buffer))/MAX_STRING_SIZE));
+    no_elements = MIN(addr.no_elements,(sizeof(buffer)/MAX_STRING_SIZE));
     status = dbGetField(&addr,dbr_type,pbuffer,&ret_options,&no_elements,NULL);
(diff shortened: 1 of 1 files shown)


mdavidsaver:
fyi. As I understand it, this is a more conformant way to express an aligned byte buffer (`buf.b`).

```c
union {
    double alignme;
    char b[400];
} buf;
```

dirk-zimoch:
I tried to be minimally invasive with my fix.

anjohnson:
Core Group: Approved.

mdavidsaver:
Applied as e1a51e2839b408cbcc73cbf1be40e8a3077ce96c

PR #355, opened 2023-03-08
Added the new annotation EPICS_PRINTF_STYLE_ANNOTATION

evvaaaa:
Added a preprocessor definition to the `printf` wrapper which will make them use [`_Printf_format_string_`](https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values) when compiled with mcvs.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -162,13 +162,15 @@ jobs:
 
           - os: windows-2019
             cmp: vs2019
-            configuration: default
+            configuration: debug
             name: "Win2019 MSC-19"
+            extra: "CMD_CXXFLAGS=-analysis"
 
           - os: windows-2019
             cmp: vs2019
-            configuration: static
+            configuration: static-debug
             name: "Win2019 MSC-19, static"
+            extra: "CMD_CXXFLAGS=-analysis"
 
           - os: windows-2019
             cmp: vs2019
--- a/modules/database/src/ioc/db/recGbl.h
+++ b/modules/database/src/ioc/db/recGbl.h
@@ -75,7 +75,7 @@ DBCORE_API void recGblInheritSevr(int msMode, void *precord, epicsEnum16 stat,
     epicsEnum16 sevr);
 DBCORE_API int recGblSetSevrMsg(void *precord, epicsEnum16 new_stat,
                                 epicsEnum16 new_sevr,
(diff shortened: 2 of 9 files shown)


mdavidsaver:
This looks ok to me.

@anjohnson Any thoughts about the `EPICS_PRINTF_STYLE_ANNOTATION` macro name?

anjohnson:
@mdavidsaver asked so…

`EPICS_PRINTF_STYLE_ANNOTATION` is rather long, would `EPICS_PRINTF_FMT` or `EPICS_PRINTF_FMT_ARG` be understandable as a shorter name? Example:

```C
void dbMsgPrint(DBENTRY *pdbentry, EPICS_PRINTF_FMT_ARG const char *fmt, ...)
    EPICS_PRINTF_STYLE(2,3);
```

Taking the argument as a macro argument might be even better:
```C
void dbMsgPrint(DBENTRY *pdbentry, EPICS_PRINTF_FMT(const char *fmt), ...)
    EPICS_PRINTF_STYLE(2,3);
```
That way if some other compiler wanted an annotation _after_ the argument instead we could still support it, although it does require the non-windows implementation to not be empty.

I would also ask that you line-wrap these so the lines aren't wider than about 75 char's. At this point I'm getting into dangerously personal and religious areas, but I like to indent the continuation line(s) by 4 characters — indenting up to the open parenthesis doesn't really make much sense to me.

evvaaaa:
@anjohnson It seems those files were already indented up to the parenthesis, e.g:
``` C
DBCORE_API int recGblSetSevrMsg(void *precord, epicsEnum16 new_stat,
                                epicsEnum16 new_sevr,
                                const char *msg, ...) EPICS_PRINTF_STYLE(4,5);
DBCORE_API int recGblSetSevrVMsg(void *precord, epicsEnum16 new_stat,
                                 epicsEnum16 new_sevr,
                                 const char *msg, va_list args);
```

Just to make sure I'm understanding you, should I go through the files changed and change the untouched indents as well?

anjohnson:
Hi @evalott100 I didn't mean to suggest that you change any lines that you aren't already modifying. This looks fine to me now, thanks!

mdavidsaver:
Looks ok to me.  Combined with a successful appveyor run, I think this PR is ready to go.

mdavidsaver:
Applied as b878295d06de383125693ca3fd533cf5c11d5a01

Issue #268, opened 2022-05-17
errSymbolAdd() only works if called before errlogInit()


anjohnson:
Support modules can (and do, e.g. see [here](https://github.com/epics-modules/vme/blob/071bcea9f7e4273d08e9729469523b3e3ce5f61e/vmeApp/src/drvIK320ErrStat.c#L8)) add error symbols to the error symbol table by calling `errSymbolAdd()`, but must do so before `errlogInitPvt()` has called `errSymBld()`. If called too late the symbols are not added to the hash table so don't get found when they are looked up for printing (if I've understood the code properly).

The implementation needs correcting so error symbols can be added at any time and get inserted into the hash table as well, but there are currently no locks protecting the data structures in errSymLib.c so they need adding as well, preferably in a way that doesn't require a separate lock/unlock for each symbol that gets added by `errSymBld()` though.

EmilioPeJu:
I'm interested in doing this one

EmilioPeJu:
Fixed by PR 357

EmilioPeJu:
Just to be clear, the name of the PR doesn't match this issue, but it fixes it, the link you mentioned is a support module trying to add error symbols without the expected effect

PR #357, opened 2023-03-08
Allow adding error symbols in support modules

EmilioPeJu:
This was acomplished by making errSymbolAdd add the error symbol directly into the global hash table and removing errnumlist which is not needed anymore.

Unit tests were added for checking the following cases:
- Adding a valid symbol and checking that it exists (fixed by this change)
- Getting an existing error symbol
- Getting a non existing error symbol
- Adding an invalid error symbol (fixed by this change)
- Adding an error symbol with a code that already exists (fixed by this change)

Therefore, issue #268 was fixed

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -102,6 +102,10 @@ and string formats, some of which support full nanosecond precision.
 More information is included in the filters documentation, which can be found in
 the `html/filters.html` document that is generated during the build
 
+### Allow adding new error symbols at any time
+
+`errSymbolAdd` can now be called after early initialization.
+
 ### Add conditional output (OOPT) to the longout record
 
 The longout record can now be configured using its new OOPT and OOCH fields
--- a/modules/libcom/src/error/errSymLib.c
+++ b/modules/libcom/src/error/errSymLib.c
@@ -80,7 +80,7 @@ static epicsUInt16 errhash(long errNum)
 
     modnum = (unsigned short) (errNum >> 16);
     errnum = (unsigned short) (errNum & 0xffff);
-    return (((modnum - 500) * 20) + errnum) % NHASH;
+    return ((modnum - (MIN_MODULE_NUM - 1)) * 20 + errnum) % NHASH;
 }
 
 /****************************************************************
@@ -94,7 +94,7 @@ int errSymbolAdd(long errNum, const char *name)
     ERRNUMNODE *pNew = NULL;
     int modnum = (epicsUInt16) (errNum >> 16);
 
-    if (modnum < 501)
+    if (modnum < MIN_MODULE_NUM)
         return S_err_invCode;
 
     epicsUInt16 hashInd = errhash(errNum);
@@ -154,7 +154,7 @@ const char* errSymLookupInternal(long status)
     modNum = (unsigned) status;
     modNum >>= 16;
     modNum &= 0xffff;
-    if (modNum <= 500) {
+    if (modNum < MIN_MODULE_NUM) {
         const char * pStr = strerror ((int) status);
         if (pStr) {
             return pStr;
@@ -244,9 +244,10 @@ void errSymTestPrint(long errNum)
     message[0] = '\0';
     modnum = (epicsUInt16) (errNum >> 16);
     errnum = (epicsUInt16) (errNum & 0xffff);
-    if (modnum < 501) {
+    if (modnum < MIN_MODULE_NUM) {
         fprintf(stderr, "Usage:  errSymTestPrint(long errNum) \n");
-        fprintf(stderr, "errSymTestPrint: module number < 501 \n");
+        fprintf(stderr,
+                "errSymTestPrint: module number < %d\n", MIN_MODULE_NUM);
         return;
     }
     errSymLookup(errNum, message, sizeof(message));
@@ -266,7 +267,7 @@ void errSymTest(epicsUInt16 modnum, epicsUInt16 begErrNum,
     epicsUInt16  errnum;
 
     if (!initialized) errSymBld();
-    if (modnum < 501)
+    if (modnum < MIN_MODULE_NUM)
         return;
 
     /* print range of error messages */
--- a/modules/libcom/src/error/errSymLib.c
+++ b/modules/libcom/src/error/errSymLib.c
@@ -33,14 +33,12 @@
 static epicsUInt16 errhash(long errNum);
 
 typedef struct errnumnode {
-    ELLNODE            node;
     long               errNum;
     struct errnumnode *hashnode;
     const char        *message;
     long               pad;
 } ERRNUMNODE;
 
-static ELLLIST errnumlist = ELLLIST_INIT;
 static ERRNUMNODE **hashtable;
 static int initialized = 0;
 extern ERRSYMTAB_ID errSymTbl;
@@ -56,44 +54,20 @@ extern ERRSYMTAB_ID errSymTbl;
 int errSymBld(void)
 {
     ERRSYMBOL      *errArray = errSymTbl->symbols;
-    ERRNUMNODE     *perrNumNode = NULL;
-    ERRNUMNODE     *pNextNode = NULL;
-    ERRNUMNODE    **phashnode = NULL;
     int             i;
-    int             modnum;
 
     if (initialized)
         return(0);
 
     hashtable = (ERRNUMNODE**)callocMustSucceed
         (NHASH, sizeof(ERRNUMNODE*),"errSymBld");
-    for (i = 0; i < errSymTbl->nsymbols; i++, errArray++) {
-        modnum = errArray->errNum >> 16;
-        if (modnum < 501) {
-            fprintf(stderr, "errSymBld: ERROR - Module number in errSymTbl < 501 was Module=%lx Name=%s\n",
-                errArray->errNum, errArray->name);
-            continue;
-        }
-        if ((errSymbolAdd(errArray->errNum, errArray->name)) < 0) {
-            fprintf(stderr, "errSymBld: ERROR - errSymbolAdd() failed \n");
-            continue;
-        }
-    }
-    perrNumNode = (ERRNUMNODE *) ellFirst(&errnumlist);
-    while (perrNumNode) {
-        /* hash each perrNumNode->errNum */
-        epicsUInt16 hashInd = errhash(perrNumNode->errNum);
 
-        phashnode = (ERRNUMNODE**)&hashtable[hashInd];
-        pNextNode = (ERRNUMNODE*) *phashnode;
-        /* search for last node (NULL) of hashnode linked list */
-        while (pNextNode) {
-            phashnode = &pNextNode->hashnode;
-            pNextNode = *phashnode;
+    for (i = 0; i < errSymTbl->nsymbols; i++, errArray++) {
+        if (errSymbolAdd(errArray->errNum, errArray->name)) {
+            fprintf(stderr, "errSymBld: ERROR - errSymbolAdd() failed \n");
         }
-        *phashnode = perrNumNode;
-        perrNumNode = (ERRNUMNODE *) ellNext((ELLNODE *) perrNumNode);
     }
+
     initialized = 1;
     return(0);
 }
@@ -114,16 +88,38 @@ static epicsUInt16 errhash(long errNum)
 
 /****************************************************************
  * ERRSYMBOLADD
- * adds symbols to the master errnumlist as compiled from errSymTbl.c
+ * adds symbols to the global error symbol hash table
  ***************************************************************/
 int errSymbolAdd(long errNum, const char *name)
 {
-    ERRNUMNODE *pNew = (ERRNUMNODE*) callocMustSucceed(1,
-        sizeof(ERRNUMNODE), "errSymbolAdd");
+    ERRNUMNODE *pNextNode = NULL;
+    ERRNUMNODE **phashnode = NULL;
+    ERRNUMNODE *pNew = NULL;
+    int modnum = (epicsUInt16) (errNum >> 16);
 
+    if (modnum < 501)
+        return S_err_invCode;
+
+    epicsUInt16 hashInd = errhash(errNum);
+
+    phashnode = (ERRNUMNODE**)&hashtable[hashInd];
+    pNextNode = (ERRNUMNODE*) *phashnode;
+
+    /* search for last node (NULL) of hashnode linked list */
+    while (pNextNode) {
+        if (pNextNode->errNum == errNum) {
+            return S_err_codeExists;
+        }
+        phashnode = &pNextNode->hashnode;
+        pNextNode = *phashnode;
+    }
+
+    pNew = (ERRNUMNODE*) callocMustSucceed(
+        1, sizeof(ERRNUMNODE), "errSymbolAdd");
     pNew->errNum = errNum;
     pNew->message = name;
-    ellAdd(&errnumlist, (ELLNODE*)pNew);
+    *phashnode = pNew;
+
     return 0;
 }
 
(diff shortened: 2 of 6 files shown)


EmilioPeJu:
should we pull the constant `501` to a define?  it appears a few times

mdavidsaver:
@anjohnson If we are going to encourage custom error codes in user code, should we take some steps to prevent collisions?

EmilioPeJu:
BTW, I also did the change for not allowing double adding the same error code (not sure if that's what you meant).

One thing to consider is how to make it easier to add error symbols in a support modules (an easier way other than having to call `errSymbolAdd` in some driver init function), it could be something similar to what epics base does, extracting the symbols from from some specified headers files

EmilioPeJu:
I realised that the size of the error symbol hash table is known at compile time, so, I changed it to be allocated statically (having the extra advantage of allowing adding error symbols much earlier). Please, note that because it's in the bss section, it will be zero which is the proper initial state for the hash table

anjohnson:
Could you modify the check for duplicate symbols being added so it doesn't return an error if both the number and the name string are identical to an already-registered entry? This reduces unnecessary errors in circumstances like testing where the symbol registration code might be run more than once — this should be allowed.

#265 says:
> there are currently no locks protecting the data structures in errSymLib.c so they need adding as well, preferably in a way that doesn't require a separate lock/unlock for each symbol that gets added.

I'm not seeing any added `epicsMutex` objects in your changes to protect the hashtable from being modified while it is being searched, was this something you looked at?

The other thing I would like to see is an entry in the `documentation/RELEASE_NOTES.md` file explaining how external software can add its private error symbols to the table.

EmilioPeJu:
Done, instead of using an `epicsMutex`, the add operation is done atomically with `epicsAtomicSetPtrT`, each position of the error symbol hash table keeps a single linked list, therefore, adding just requires to set a pointer.

mdavidsaver:
Atomic container algorithms are tricky to get correct...

fyi. I think that you would need to use a loop with `epicsAtomicCmpAndSwapPtrT()` instead of `epicsAtomicSetPtrT()` to prevent concurrent calls (however unlikely) from leaking memory.  Also, it is generally advisable to use atomic operations for all access to a variable.  `epicsAtomicGetPtrT()` instead of simple dereference.

modules/libcom/src/error/errSymLib.c
```diff
@@ -23,25 +23,25 @@
 #include "cantProceed.h"
 #include "epicsAssert.h"
 #include "epicsStdio.h"
+#include "epicsAtomic.h"
 #include "errMdef.h"
 #include "errSymTbl.h"
 #include "ellLib.h"
 #include "errlog.h"
 
+#define MIN_MODULE_NUM 501
```
exzombie: I would put this into `errMdef.h`, next to the existing comment explaining the reasoning behind the number.

EmilioPeJu:
I'm not protecting the case of simultaneous adding symbols on purpose because I don't think the extra complexity is worth it. The error symbols will be added in some init function (for example driver init), which are called in the main thread and therefore, there is not such risk.

As I said earlier, the focus was in the case of simultaneously adding and fetching a symbol, which is something that could happen because any error logging  could fetch a symbol (which is something that could easily happen in a thread parallel to the main thread adding)

modules/libcom/src/error/errSymLib.c
```diff
@@ -23,25 +23,25 @@
...
-    ELLNODE            node;
     long               errNum;
     struct errnumnode *hashnode;
     const char        *message;
     long               pad;
 } ERRNUMNODE;
 
-static ELLLIST errnumlist = ELLLIST_INIT;
-static ERRNUMNODE **hashtable;
+static ERRNUMNODE *hashtable[NHASH * sizeof(ERRNUMNODE *)];
```
exzombie: Inappropriate use of `sizeof`: this a static array, not `malloc()`, so its size is given as number of elements, not bytes.

modules/libcom/src/error/errSymLib.c
```diff
@@ -23,25 +23,25 @@
...
 
 static epicsUInt16 errhash(long errNum);
 
 typedef struct errnumnode {
-    ELLNODE            node;
     long               errNum;
     struct errnumnode *hashnode;
     const char        *message;
     long               pad;
 } ERRNUMNODE;
```
exzombie, line 43: As I understand it up to this point, the removed `node` field and the `hashnode` field have overlapping roles: both make this struct an element of a linked list. Personally, I don't mind a "hand-rolled" linked list, but in that case, the `hashnode` field should be renamed to make it obvious what it represents. On the other hand, it would be preferable to use a consistent API for linked lists everywhere, which means removing the `hashnode` field and restoring `node` instead.

Edit: ok, I get it now. This thing used to be a part of two linked lists, and if you keep `hashnode` instead of `node`, you get to keep existing code for iterating over the hashmap list. I would use this opportunity to change the code, but ultimately, it's up to you. It's just style, after all.

modules/libcom/src/error/errSymLib.c
```diff
@@ -56,44 +56,17 @@ extern ERRSYMTAB_ID errSymTbl;
 int errSymBld(void)
```
exzombie: The description of the function can be simplified, just as the function itself has been.

exzombie:
> As I said earlier, the focus was in the case of simultaneously adding and fetching a symbol, which is something that could happen because any error logging could fetch a symbol (which is something that could easily happen in a thread parallel to the main thread adding)

If you choose to limit the support to that use case, that's sufficient reasoning for using atomic-set instead of compare-and-swap, but you should explain it in a comment somewhere so that it's clear for future readers that writing from multiple threads is not supported. However, Michael's point that you need to use atomic-get to read the values stands: if you don't, there is no guarantee that long-lived thread would *ever* get an updated value.

Edit: thinking about it again, it's probably easier to just do compare-and-swap than try to explain the reasoning in a comment :)

modules/libcom/src/error/errSymLib.c
```diff
@@ -109,21 +82,47 @@ static epicsUInt16 errhash(long errNum)
 
     modnum = (unsigned short) (errNum >> 16);
     errnum = (unsigned short) (errNum & 0xffff);
```
exzombie, line 94: This is a very common pattern here, so I would put it in a function, even if it is just two lines of code.

mdavidsaver:
> ... I'm not protecting the case of simultaneous adding symbols on purpose because I don't think the extra complexity is worth it. ...

I'm don't think intentionally leaky code is acceptable.  The simple way would be to use a mutex and live with a likely minor performance hit when looking up error messages (hopefully a relatively rare occurrence).

modules/libcom/src/error/errSymLib.c
EmilioPeJu: Please note that we are allocating pointers, so, what we have is the same as `sizeof(void *)`

modules/libcom/src/error/errSymLib.c
EmilioPeJu: Will do, thanks for the suggestion

modules/libcom/src/error/errSymLib.c
exzombie: Let me put this another way. The following three statements allocate an object of the same size:
```
void *dynmem = malloc(num_elements * sizeof(void*));
void *zeromem = calloc(num_elements, sizeof(void*));
void *staticmem[num_elements];
```

EmilioPeJu: Ohh, now I understand, thanks for spotting that, I'll fix it

modules/libcom/src/error/errSymLib.c
EmilioPeJu: Done

modules/libcom/src/error/errSymLib.c
EmilioPeJu: The linked list we had before was used only as an intermediate list that is iterated and added to the hash table, so I optimized that and now the `errSymbolAdd` function adds directly to the hash table. The linked list inside the hash table doesn't require a linking to the previous node, therefore the current approach is simpler and with less overhead than using `ELLNODE`

modules/libcom/src/error/errSymLib.c
EmilioPeJu: what do you think about having 2 macros at the top of the file: `ERROR_NUMBER(status)` and `ERROR_MODULE(status)`?

EmilioPeJu:
Thanks @exzombie and @mdavidsaver for reviewing, I have added the epics mutex as you suggested, I hope I got it right this time

documentation/RELEASE_NOTES.md
```diff
@@ -17,6 +17,14 @@ should also be read to understand what has changed since earlier releases.
...
 <!-- Insert new items immediately below here ... -->
 
+### Support modules can now add error symbols even before calling `errlogInit`
+
+A support module can add an error symbol by calling `errSymbolAdd`, for
+example: `errSymbolAdd(S_newError, "New error message")`
+
+This used to work in older EPICS versions but got broken at some point and is
+now restored.
+
```
mdavidsaver: I would describe this change as "Allow errSymbolAdd() to be called after early initialization".  I would also like to see some guidance to users about how avoid error number collisions.  Also, "got broken at some point" is not helpful to users.

For that matter, I think this sort of information should be added as a doxygen comment for `errSymbolAdd()`.  eg.

```c
/** Add custom error message to be returned by subsequent `errSym*` calls.
 * \since 7.0.2 Calls after errlogInit() have no effect.
 * \since UNRELEASED Can be called at any time.
 */
```

The `7.0.2` is a guess that 0fdb995a0bda9a7d9591b6c2b1bde5b2aab43421 contained the "breakage".

modules/libcom/src/error/errSymLib.c
```diff
@@ -33,69 +35,55 @@
...
 extern ERRSYMTAB_ID errSymTbl;
 
+static void populateEpicsErrorSymbols()
+{
+    ERRSYMBOL *errArray = errSymTbl->symbols;
+    int i;
+
+    for (i = 0; i < errSymTbl->nsymbols; i++, errArray++) {
+        if (errSymbolAdd(errArray->errNum, errArray->name)) {
+            fprintf(stderr, "errSymBld: ERROR - errSymbolAdd() failed \n");
```
mdavidsaver: This new `populateEpicsErrorSymbols()` is only called from `errSymBld()`, which does nothing but call `populateEpicsErrorSymbols()`.  Why introduce another function name?

EmilioPeJu: it was just for having a clearer code (easier to read without needing to go deeper), the new function has a name that describes what it does... however, I'm happy to move it to `errSymBld()` if that is preferred.

EmilioPeJu: OK, I've moved it, given that the big comment at the top says what it does, it was redundant

documentation/RELEASE_NOTES.md
EmilioPeJu: I simplified the release note using the text you suggest, unfortunately I can't build your guess commit to verify it, not sure if because build was broken or because there is something nasty in my environment

anjohnson:
Core Group: Code now looks good, but needs test results from Appveyor before pushing the merged version.

mdavidsaver:
I have rebased this PR on top of a change to fix the appveyor builds.  I have also added an `#include <epicsTypes.h>` which will hopefully fix a build failure with vs2012.

I am removing Approval until an appveyor run passes.

modules/libcom/src/error/errSymLib.c
```diff
@@ -109,21 +92,48 @@ static epicsUInt16 errhash(long errNum)
...
+    ERRNUMNODE **phashnode = NULL;
+    ERRNUMNODE *pNew = NULL;
+    int modnum = (epicsUInt16) (errNum >> 16);
+
+    if (modnum < MIN_MODULE_NUM)
+        return S_err_invCode;
+
+    initErrorHashTable();
+
+    epicsUInt16 hashInd = errhash(errNum);
```
mdavidsaver: The appveyor vs2012 build is still failing on this line.  My attempt to include `epicsTypes.h` above did not help, which confuses me.  I am not certain why this does not include the typedef for `epicsUInt16`?

```
cl               -nologo -FC -D__STDC__=0 -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DBUILDING_libCom_API   -Ox -GL -Oy-   -W3        -MD -DEPICS_BUILD_DLL -DEPICS_CALL_DLL    -I. -I../O.Common -I. -I../osi/compiler/msvc -I../osi/compiler/default -I. -I../osi/os/WIN32 -I../osi/os/default -I.. -I../as -I../bucketLib -I../calc -I../cvtFast -I../cppStd -I../cxxTemplates -I../dbmf -I../ellLib -I../env -I../error -I../fdmgr -I../flex -I../freeList -I../gpHash -I../iocsh -I../log -I../macLib -I../misc -I../osi -I../pool -I../ring -I../taskwd -I../timer -I../yacc -I../yacc -I../yajl -I../../../../include/compiler/msvc -I../../../../include/os/WIN32 -I../../../../include         -c ../error/errSymLib.c
errSymLib.c
c:\projects\epics-base\modules\libcom\src\error\errsymlib.c(114) : error C2275: 'epicsUInt16' : illegal use of this type as an expression
        c:\projects\epics-base\modules\libcom\src\misc\epicstypes.h(47) : see declaration of 'epicsUInt16'
c:\projects\epics-base\modules\libcom\src\error\errsymlib.c(114) : error C2146: syntax error : missing ';' before identifier 'hashInd'
c:\projects\epics-base\modules\libcom\src\error\errsymlib.c(114) : error C2065: 'hashInd' : undeclared identifier
c:\projects\epics-base\modules\libcom\src\error\errsymlib.c(117) : error C2065: 'hashInd' : undeclared identifier
make[4]: *** [../../../../configure/RULES_BUILD:258: errSymLib.obj] Error 2
make[4]: Leaving directory 'C:/projects/epics-base/modules/libcom/src/O.win32-x86'
make[4]: Entering directory 'C:/projects/epics-base/modules/libcom/src/O.win32-x86'
cl               -nologo -FC -D__STDC__=0 -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DBUILDING_libCom_API   -Ox -GL -Oy-   -W3        -MD -DEPICS_BUILD_DLL -DEPICS_CALL_DLL    -I. -I../O.Common -I. -I../osi/compiler/msvc -I../osi/compiler/default -I. -I../osi/os/WIN32 -I../osi/os/default -I.. -I../as -I../bucketLib -I../calc -I../cvtFast -I../cppStd -I../cxxTemplates -I../dbmf -I../ellLib -I../env -I../error -I../fdmgr -I../flex -I../freeList -I../gpHash -I../iocsh -I../log -I../macLib -I../misc -I../osi -I../pool -I../ring -I../taskwd -I../timer -I../yacc -I../yacc -I../yajl -I../../../../include/compiler/msvc -I../../../../include/os/WIN32 -I../../../../include         -c errSymTbl.c
errSymTbl.c
```

https://ci.appveyor.com/project/epics-base-7/epics-base/builds/47895623/job/6a76h0mefu03jsr7#L1429

ralphlange:
Maybe it's

> When you name your source files *.c, MSVC assumes it's compiling C, which means C89. All block-local variables need to be declared at the beginning of the block.

From: https://stackoverflow.com/questions/9903582/error-c2275-illegal-use-of-this-type-as-an-expression

mdavidsaver:
Ah, yes you are right of course.  I took the error message too literally.

I have tried to push up another change.  Hopefully github.com will pick this up whenever their various issues are resolved.

mdavidsaver:
Ok.  that did the trick.

mdavidsaver:
Merged as of 88ea1507f46daabb0520d773193b271ac6c9b56d

PR #358, opened 2023-03-08
EpicsHostArch: add -g option to convert a GNU arch tuplet

minijackson:
For GNU to EPICS architecture conversion. This might be useful for interfacing with external build / packaging programs, like Nix / Debian tooling / etc.

Pinging @anjohnson as this involves the Perl tooling.

### DIFF
--- a/src/tools/EpicsHostArch.pl
+++ b/src/tools/EpicsHostArch.pl
@@ -7,30 +7,103 @@
 # in file LICENSE that is included with this distribution.
 #*************************************************************************
 
-# Returns an architecture name for EPICS_HOST_ARCH that should be
-# appropriate for the CPU that this version of Perl was built for.
-# Any arguments to the program will be appended with separator '-'
-# to allow flags like -gnu -debug and/or -static to be added.
-
-# Before Base has been built, use a command like this:
-#   bash$ export EPICS_HOST_ARCH=`perl src/tools/EpicsHostArch.pl`
-#
-# If Base is already built, use
-#   tcsh% setenv EPICS_HOST_ARCH `perl base/lib/perl/EpicsHostArch.pl`
-
-# If your architecture is not recognized by this script, please send
-# the output from running 'perl --version' to the EPICS tech-talk
-# mailing list to have it added.
-
 use strict;
+use warnings;
+
+use Getopt::Std;
+$Getopt::Std::STANDARD_HELP_VERSION = 1;
 
 use Config;
 use POSIX;
 
-print join('-', HostArch(), @ARGV), "\n";
+use FindBin qw($Bin);
+use lib ("$Bin/../../lib/perl", $Bin);
 
-sub HostArch {
-    my $arch = $Config{archname};
+use Pod::Usage;
+
+=head1 NAME
+
+EpicsHostArch.pl - Prints the current host architecture
+
+=head1 SYNOPSIS
+
+B<EpicsHostArch.pl> [extension]
+
+B<EpicsHostArch.pl> -g <GNU-like arch> [extension]
+
+B<EpicsHostArch.pl> -h
+
+=head1 DESCRIPTION
+
+Returns an architecture name for EPICS_HOST_ARCH that should be
+appropriate for the CPU that this version of Perl was built for.
+Any arguments to the program will be appended with separator '-'
+to allow flags like -gnu -debug and/or -static to be added.
+
+Before Base has been built, use a command like this:
+
+C<bash$ export EPICS_HOST_ARCH=`perl src/tools/EpicsHostArch.pl`>
+
+If Base is already built, use:
+
+C<tcsh% setenv EPICS_HOST_ARCH `perl base/lib/perl/EpicsHostArch.pl`>
+
+If your architecture is not recognized by this script, please send
+the output from running C<perl --version> to the EPICS tech-talk
+mailing list to have it added.
+
+If the C<-g> option is provided with an argument, print the EPICS
+architecture corresponding to the given GNU architecture tuplet.
+
+=head1 OPTIONS
+
+B<EpicsHostArch.pl> understands the following options:
+
+=over 4
+
+=item B<-h>
+
+Help, display this document as text.
+
+=item B<-g> <GNU arch tuplet>
+
+If specified, convert the given GNU architecture tuplet instead.
+
+=back
+
+=head1 EXAMPLES
+
+C<EpicsHostArch.pl>
+
+C<EpicsHostArch.pl static>
+
+C<EpicsHostArch.pl -g powerpc64-linux>
+
+C<EpicsHostArch.pl -g MSWin32-x64 static>
+
+=cut
+
+our ($opt_h);
+our ($opt_g);
+$opt_h = 0;
+
+sub HELP_MESSAGE {
+    pod2usage(-exitval => 2, -verbose => $opt_h * 3);
+}
+
+HELP_MESSAGE() if !getopts('hg:') || $opt_h;
+
+# Convert GNU-like architecture tuples (<cpu>-<system>)
+# to EPICS terminology (<system>-<cpu>)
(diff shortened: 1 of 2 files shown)


minijackson:
Updated as discussed.

Pinging @mdavidsaver as he was a part of the discussion.

minijackson:
I'm also proposing to install `EpicsHostArch` into the bin folder, so that it is more easily available in `PATH` (see 7ce3ea7)

src/tools/EpicsHostArch.pl
```diff
@@ -1,36 +1,109 @@
 #!/usr/bin/env perl
 #*************************************************************************
-# Copyright (c) 2018 UChicago Argonne LLC, as Operator of Argonne
+# Copyright (c) 2023 UChicago Argonne LLC, as Operator of Argonne
```
anjohnson: In general you shouldn't modify other people's copyright statements, here the Argonne copyright term started in 2018 when I last modified this file, and the fact that _you_ have made changes to it doesn't affect someone else's start-date. You or your employer may want your own separate copyright line to be added, and that's fine if you do.

src/tools/EpicsHostArch.pl
```diff
@@ -1,36 +1,109 @@
...
+# to EPICS terminology (<system>-<cpu>)
+#
+# Documentation for GNU-like terminology:
+# - https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/System-Type.html#System-Type
+# - https://git.savannah.gnu.org/cgit/config.git/tree/
+#
+# Some examples from the Debian project:
+# - https://wiki.debian.org/Multiarch/Tuples
+sub toEpicsArch {
+	my $arch = shift;
```
anjohnson: If the argument to this subroutine starts with `darwin` the routine will call `uname` on line 120 and attempt to use that host-specific data to determine the macOS CPU type. Thus running `EpicsHostArch.pl -g darwin-ppc` on a `linux-x86_64` box will probably return `darwin-x86_64` which is obviously wrong. On this macOS (`darwin-aarch64`) laptop, the Perl `$Config{archname}` configuration variable is `darwin-2level` which is why using `uname` is necessary.

anjohnson (changes requested):
See comments inline.

src/tools/EpicsHostArch.pl
minijackson: Good catch, I completely missed it

src/tools/EpicsHostArch.pl
```diff
@@ -7,30 +7,103 @@
...
+
+use Pod::Usage;
+
+=head1 NAME
+
+EpicsHostArch.pl - Prints the current host architecture
+
+=head1 SYNOPSIS
+
+B<toEpicsArch.pl> [extension]
```
mdavidsaver: Is `toEpicsArch.pl` the script name?

minijackson: Oops, that's from a previous version, good catch

src/tools/EpicsHostArch.pl
```diff
@@ -43,16 +116,28 @@ sub HostArch {
...
-        my ($kernel, $hostname, $release, $version, $cpu) = uname;
-        if (m/^darwin/) {
-            for ($cpu) {
-                return 'darwin-x86'     if m/^x86_64/;
-                return 'darwin-aarch64' if m/^arm64/;
-            }
-            die "$0: macOS CPU type '$cpu' not recognized\n";
-        }
+        return 'darwin-x86'     if m/^x86_64-darwin/;
+        return 'darwin-x86'     if m/^x86(_64)?-darwin/;
```
anjohnson: Did you mean to remove one of these two?

minijackson: yes, sorry about that

anjohnson (approved):
I don't see any other issues now, thanks!

mdavidsaver:
@anjohnson There have been changes since your approval, can you have another look?

minijackson:
Updated to fix the merge conflict

src/tools/Makefile
```diff
@@ -22,9 +22,10 @@ PERL_MODULES += EPICS/PodXHtml.pm
 PERL_MODULES += Pod/Markdown.pm
 PERL_MODULES += URI/Escape.pm
 
-# This goes into lib/perl, not bin/<host>
+# This also goes into lib/perl
 PERL_MODULES += EpicsHostArch.pl
 
+PERL_SCRIPTS += EpicsHostArch.pl
```
anjohnson, line 28: Adding `EpicsHostArch.pl` to `PERL_SCRIPTS` installs it into the `bin/$(EPICS_HOST_ARCH)` directory. There's a bootstrapping problem there, if you know which `bin` subdirectory it's in, why do you need to run the script at all?

src/tools/EpicsHostArch.pl
```diff
@@ -45,15 +118,9 @@ sub HostArch {
...
+        return 'darwin-aarch64' if m/^(arm64|aarch64)-darwin/;
 
-        my ($kernel, $hostname, $release, $version, $cpu) = uname;
-        if (m/^darwin/) {
-            for ($cpu) {
-                return 'darwin-x86'     if m/^x86_64/;
-                return 'darwin-aarch64' if m/^arm64/;
-            }
-            die "$0: macOS CPU type '$cpu' not recognized\n";
-        }
```
anjohnson, line 56: This is failing on MacOS, because Perl's `$Config{archname}` doesn't contain CPU information at all on Apple builds:
```
Cox$ perl -V:archname
archname='darwin-2level';
```
The original script only matches the start of the archname on MacOS, then uses `uname` to get the CPU type. You deleted that additional step, breaking the script.

I don't know how this affects the new `-g` option processing.

anjohnson (changes requested):
Can you describe the use-case for the new `-g` option, where do you need it? EPICS requires C & C++ compilers to be provided for the host architecture as it runs the `e_flex` and `antelope` binaries that it builds later in the build process, so spoofing a different architecture isn't going to work.

src/tools/EpicsHostArch.pl
anjohnson: Not sure if I was looking at an old build here, I see there is code to handle the darwin Perl version. The last CI build didn't fail for the reason I mentioned above, but it did fail on MacOS for a different reason. Can you rebase this to trigger another CI build please? The Core Group tell me the use-case for this is good.

minijackson: Done!

src/tools/Makefile
minijackson: For example for cross-compiling a software that depends on epics-base (e.g. `x86_64-linux` to `aarch64-linux`), the `EpicsHostArch.pl` would be installed on both `bin/linux-x86_64` and `bin/linux-aarch64`, and only the `bin/linux-x86_64` would be in the `PATH`.

The build system then calls `EpicsHostArch.pl -g aarch64-linux` to figure out:

- where to install its own executables and libraries, if it wants to follow the EPICS `bin/${ARCH}` structure
- where to find the target libraries
  - either by adding manually `lib/linux-aarch64` to its ld path
  - or by using the `pkg-config` file `epics-base-linux-aarch64.pc`

I have an example usage here: https://github.com/minijackson/pvxs/blob/b9ed77f942ac42abcec5dcc63ebaaf6b3b589b4e/meson.build

In my mind, putting `epics-base/bin/linux-x86_64` in the `PATH` would be the job of whatever programs sets up the build environment, but the build system still needs the other info.

anjohnson (approved):
Thanks for adding the usage text! Should this be added to the `HTMLS` list to publish that?

It's a pity that RTD doesn't work very well with the Perl-style Usage output (see the comment at the bottom of the Makefile). It would be nice to include these in the website, but that's not an issue for this PR to worry about.

Issue #280, opened 2022-07-28
Timestamp when monitoring non-VAL


mdavidsaver:
#263 Fixes an issue where `process()` for waveformRecord called `db_post_events()` before `TIME` was updated by `recGblGetTimeStampSimm()`.  So the posted update had the wrong timestamp.

At least `aai` and `aao` have the same code.

anjohnson:
Core Group 9/28: @bhill-slac 's fix #263 only fixed the waveform record, we need to apply the same change to aai and aao, maybe others too.

henrique-silva:
I'll work on applying this fix to the other records as well

henrique-silva:
So, while investigating this issue I noticed that currently the `NORD` field is updated twice every time the record processes. It keeps the same value, but gets a new timestamp.
This is due to  `db_post_events` being called both on the record support and device support layers.

As agreed with @mdavidsaver and @anjohnson , I'll work on removing the duplicate call on the device support layer, so we keep it only on the record support

anjohnson:
Core Group: PR #359 needs final cleanup and merging.

PR #359, opened 2023-03-08
Fix undefined timestamp for NORD field

henrique-silva:
This PR addresses the issues mentioned in #280.

- Adds posting event when NORD is updated only after getting a timestamp.
- Removes posting this same event in the device support layer

Affects aai, aao and waveform records

### DIFF
--- a/documentation/new-notes/PR-359.md
+++ b/documentation/new-notes/PR-359.md
@@ -1,0 +1,11 @@
+### Fix undefined timestamp for NORD field
+
+When monitoring non-val fields such as `NORD` for several record types updates
+are not sent out correctly resulting in incorrect timestamps. This fixes this
+for the record types `aai`, `aao`, `subArray`.
+
+Note that support modules that already implement this in the device support
+layer will need to be updated to avoid doubled monitor updates. If those modules
+are to be built against older versions of EPICS base then they will need to use
+the defined macros `AAO_POSTS_NORD`, `AAI_POSTS_NORD`, `SA_POSTS_NORD`, or
+`WF_POSTS_NORD` as appropriate to detect this change.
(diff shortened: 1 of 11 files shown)


simon-ess:
If I understand this correctly, it also affects the sub-array record: https://github.com/epics-base/epics-base/blob/7.0/modules/database/src/std/dev/devSASoft.c#L123

mdavidsaver (approved):
I think this is acceptable with the addition of a release note entry with a note for any external device support which might be posting `NORD` that this is no longer necessary, and will result in duplicate updates.

anjohnson (changes requested):
A release note entry will definitely be needed, and should include a description and example of how out-of-tree device support (such as Asyn) should be modified so it still works when compiled against older versions of Base. The best way to handle that is shown below, the first part you'll have to do for each of the record types that you've modified. I'm using the aaiRecord as an example, you can probably copy/paste the code in the second part as your example for the release notes.

Thanks!

----

Add the line
```
%#define AAI_POSTS_NORD
```
to the aaiRecord.dbd.pod file — find where the similar named `HAS_…dset` is defined in the file and add this near to that. The generated aaiRecord.h file will then define that macro.

----

Each device support which already #includes the aaiRecord.h file can use something like this in its `read_aai()` method:
```C
#ifndef AAI_POSTS_NORD
    epicsUInt32 nord = prec->nord;
#endif

…

#ifndef AAI_POSTS_NORD
        if (nord != prec->nord)
            db_post_events(prec, &prec->nord, DBE_VALUE | DBE_LOG);
#endif
```

anjohnson:
Core Group: AI on @simon-ess to rebase and add release notes.

simon-ess:
Rebased, will add release notes shortly.

simon-ess:
Release note added. @anjohnson

mdavidsaver (approved):
This still looks ok to me.  With the added release notes, I think this PR is ready to go.

PR #360, opened 2023-03-08
Add missing AFTC field documentation

### LINKED ISSUE #313: Some record types are missing AFTC documentation
The AFTC field was added to 5 record types, but only 2 of them have it documented in their .dbd.pod files. Using those 2 as examples, a develop text for the other 3, adjusting the wording as appropriate.

henrique-silva:
Fixes #313 

Also added a missing paragraph describing the HYST field in longin

### DIFF
--- a/modules/database/src/std/rec/calcRecord.dbd.pod
+++ b/modules/database/src/std/rec/calcRecord.dbd.pod
@@ -460,12 +460,17 @@ conditions.
 
 The HYST field defines an alarm deadband for each limit.
 
+The AFTC field sets the time constant on a low-pass filter that delays the
+reporting of limit alarms until the signal has been within the alarm range for
+that number of seconds (the default AFTC value of zero retains the previous
+behavior).
+
 See L<Alarm Specification|https://docs.epics-controls.org/en/latest/guides/EPICS_Process_Database_Concepts.html#alarm-specification>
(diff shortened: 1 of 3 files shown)


simon-ess:
LGTM

karlvestin:
I had a conversation with Andrew today about the AFTC field. The actual function of this field is somewhat weird, and not particularly well described by the documentation. I think it would make sense to add a statement in the documentation that this is primarily an internal value to implement HYST, and not of general user relevance.

henrique-silva:
But as far as I can see from the code, it has nothing to do with the `HYST` field itself:
https://github.com/epics-base/epics-base/blob/d9052f71056ab38217344e0f6d395e6937f15965/modules/database/src/std/rec/aiRecord.c#L358-L400

It is a further calculation after the `HYST` check, but it is only used in conjunction with the very much undocumented field `AFVL`

anjohnson:
IIRC our conversation was about the `LALM` field, I think Karl used the wrong field name in his comment above about it being an internal implementation detail.

This looks good to me, thanks!

anjohnson (approved):
Doc changes only.

mdavidsaver:
Applied as of 5c9903115719dde7f265d5201d4f15d1a00938de

PR #363, opened 2023-03-09
Reformat help

AlexanderWells-diamond:
Rework the help command output to improve formatting and readability, and include a visual marker (a line of underlines) between different help commands.

Unfortunately there's no way to accurately copy the formatting in Markdown. You can see examples of it in modules/libcom/test/iocshTestHelpFunctions and iocshTestHelpFunction1, or by running `help *` in `softIoc`.


modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -888,15 +888,21 @@ static void helpCallFunc(const iocshArgBuf *args)
...
                 piocshFuncDef = pcmd->def.pFuncDef;
                 if (epicsStrGlobMatch(piocshFuncDef->name, argv[iarg]) != 0) {
-                    if(piocshFuncDef->usage) {
-                        fputs("\nUsage: ", epicsGetStdout());
+
+                    if (printSeparator) {
+                        fprintf(epicsGetStdout(), 
+                            ANSI_UNDERLINE("                                                            \n"));
+                    } else {
+                        printSeparator = true;
```
simon-ess: This might be overly pedantic, but I find this logic a bit confusing. It ties the printing of a separator to whether or not we already have printed one. But the real reason we print a separator is whether or not we are printing multiple commands.

I would almost prefer that `printSeparator = true` is simply added unconditionally as the last command in the outer `for` loop. To me, this fits better with the reason that we add the separator, possibly with a separate name e.g.
```c
bool first_function = true;
for (arg in iarg) {
  for (pcmd in pcmds) {
    if (not first_function) {
      print_separator();
    }
  }
  first_function = false;
}
```
or something like that.

minijackson (approved):
Tested, on my machine, looks great!

Unrelated to this PR, `help` completion doesn't work for the second argument, it completes local files.

20230310_10h12m55s_grim
20230310_10h14m36s_grim

simon-ess:
@minijackson - you should probably raise an issue for that.

mdavidsaver:
> ... completion doesn't work for the second argument ...

I did not think of this...  May be as simple as changing `arg==1` to `arg>=1` in iocsh.cpp

```c
            } else if(arg==1 && strcmp(def->pFuncDef->name, "help")==0) {
                return rl_completion_matches(word, iocsh_complete_command);
```

modules/libcom/src/iocsh/iocsh.cpp
AlexanderWells-diamond: Thank you for the feedback. I agree. Implemented in 8e49ba6ebb2388987deea9dbe1f12ea9e4a73b0b

AlexanderWells-diamond:
I don't really understand why the tests have re-run, nor what the problem is - each failing platform appears to show a completely different problem. I suspect it's nothing to do with my changes, given that there are compile failures on some platforms?

mdavidsaver:
There is a compile error hiding in the GHA log.

```
   ../iocshTest.cpp: In function 'bool {anonymous}::compareFiles(const string&, const string&)':
  ../iocshTest.cpp:108:64: error: no matching function for call to 'std::basic_ifstream<char>::basic_ifstream(const string&, std::_Ios_Openmode)'
     std::ifstream f1(p1, std::ifstream::binary|std::ifstream::ate);
```

In short, until c++11 the `ifstream` constructor only accepted filename as `const char*`, not `const std::string&`.  Adding a `.c_str()` should be enough to maintain compatibility with older compilers.

cf. https://en.cppreference.com/w/cpp/io/basic_ifstream/basic_ifstream

AlexanderWells-diamond:
Thank you @mdavidsaver, I've made that fix in https://github.com/epics-base/epics-base/pull/363/commits/baa97518e1a6725edf42a992bf16fc26b1fd6409

AlexanderWells-diamond:
The  two failing tests appear to be `regressTest` on one platform, and `testChannelAccess` on the other. Neither of these are tests or code areas that I've touched, so I assume they are flaky tests?

modules/libcom/test/iocshTest.cpp
```diff
@@ -83,14 +84,85 @@ void assertCallFunc(const iocshArgBuf *args)
...
+const iocshFuncDef testHelpFunction3Def = {"testHelpFunction3",0,0,
+                                    "Usage message of testHelpFunction3\n"};
+void doNothing(const iocshArgBuf *args)
+{
+    return;
+}
+
+std::string readFile(std::string filename)
+{
+    std::ifstream t(filename.c_str());
```
mdavidsaver, line 101: Should check for `fail()` or `!is_open()` and throw an exception for a cleaner error if a test file is missing?

mdavidsaver (commented):
The changes to formatting look ok to me.

modules/libcom/test/iocshTest.cpp
AlexanderWells-diamond: Added in 561235d987dda3f42bf2481f2fd899e83b38ae24.

I note that raising the exception causes all subsequent tests to be skipped - I assume this is what you intended?

AlexanderWells-diamond:
> x [Build EPICS Base 7 base-7.0-907 failed](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/46871040) (commit [86615476cb](https://github.com/epics-base/epics-base/commit/86615476cb) by @AlexanderWells-diamond)

This is an old run, there's another run pending for the newer commit

AlexanderWells-diamond:
I believe the test failures here are false - they're for areas I haven't touched. Is it possible to have them re-run please?

modules/libcom/test/iocshTest.cpp
anjohnson: I haven't checked the code, is that exception just skipping the tests which depend on that file, or any/the tests beyond that as well? The correct behaviour should be that being unable to open a file will fail a test for that file opening, and subsequent tests that depend on that file should be skipped and something like `testSkip("File open failed", 5);` called instead (where 5 is the number of tests that can't be run).

BTW you can ignore the test failures from Appveyor in unrelated test programs, we are aware that some of our tests do fail there. Again, my apologies that we haven't finished reviewing and merging all of the Codeathon PRs yet.

anjohnson:
Core: This probably needs a rebase to trigger the GHA builds.

modules/libcom/test/iocshTest.cpp
mdavidsaver: > I note that raising the exception causes all subsequent tests to be skipped - I assume this is what you intended?

What I would like to avoid is having the following tests using this input silently given an empty string.  Whether this lack of silence is an exception, returning (and acting on) an error code, or just a `testFail()`, is less important to me.

mdavidsaver:
I have done a rebase, and the GHA and appveyor jobs are running...

mdavidsaver:
The RTEMS jobs fail with:

```
  iocshTest.tap ................ 
  not ok 24 - compareFiles(filename, "iocshTestHelpFunction1") == true
  not ok 25 - compareFiles(filename, "iocshTestHelpFunctions") == true
```

It is not immediately obvious to me which condition in `compareFiles()` does not match.  Maybe turning those comments into `testDiag()` would clarify?

AlexanderWells-diamond:
I've added additional `testDiag()` messages in `compareFiles`, hopefully that'll shed some light on the RTEMS issue

AlexanderWells-diamond:
I have added `testDiag` calls to all the failure paths now, however the output doesn't seem to show any of them:
```
iocshTest.tap ................ 
  not ok 24 - compareFiles(filename, "iocshTestHelpFunction1") == true
  not ok 25 - compareFiles(filename, "iocshTestHelpFunctions") == true
  Failed 2/25 subtests 
```

Have I misunderstood how `testDiag` works, or is there some other mechanism to print data? Thank you.

mdavidsaver:
> Have I misunderstood how `testDiag` works, or is there some other mechanism to print data? Thank you.

`testDiag()` and `testOk()` both print to `stdout`.  The "trick" is the the `prove` tool which parses stdout only prints `ok` and `not ok` lines (at least as base current invokes it...).

The full output may be found in the tapfile archives which appear with each completed GHA run.  I think you should be to download these files for builds associated with this PR.  Also for runs for your personal fork repository.

From [the last run](https://github.com/epics-base/epics-base/actions/runs/6340803577?pr=363).

```
...
ok 23 - contents.find("testHelpFunction3") != std::string::npos
# One or more files failed to open
# f1.fail(): 0 f2.fail(): 1
not ok 24 - compareFiles(filename, "iocshTestHelpFunction1") == true
# One or more files failed to open
# f1.fail(): 0 f2.fail(): 1
not ok 25 - compareFiles(filename, "iocshTestHelpFunctions") == true
...
```

mdavidsaver:
Early in the RTEMS boot..

```
...
***** Using compiled in file data *****
-> /iocshTestBadArgIndirect.cmd - ok
-> /iocshTestBadArg.cmd - ok
-> /iocshTestSuccessIndirect.cmd - ok
-> /iocshTestSuccess.cmd - ok
***** Initializing NFS *****
...
```

So it looks like your new files are not captured along with the others for lack of `.cmd`.

```make
TESTFILES += $(wildcard ../iocshTest*.cmd)
```

https://github.com/epics-base/epics-base/blob/e3b1ef9defa76c7654ec571e5010afe42670a5cb/modules/libcom/test/Makefile#L271C43-L271C43

mdavidsaver:
> So it looks like your new files are not captured along with the others for lack of `.cmd`.

If you don't think `.cmd` is appropriate, then add them to `TESTFILES` in some other way...

AlexanderWells-diamond:
Thank you @mdavidsaver . I think I've fixed it this time.

AlexanderWells-diamond:
I believe the AppVeyor failure is due to a flaky Windows test:
```
Tests failed in:
    modules/database/test/std/rec/O.windows-x64-debug
        regressTest.tap
```
I have not touched this area at all.

ralphlange:
Re-running the failed test right now...

mdavidsaver:
Core Meeting:  Tests passing now, so ready to go.

mdavidsaver:
Applied as 6dec68554c30daab64823ef57e7757002cf4f644

Issue #322, opened 2023-01-17
Enhancement to tab completion — fields


anjohnson:
The GNU readline-based tab-completion that was added to the IOC shell by #274 (which is not included in any released versions of Base yet) doesn't currently support PVs that include a field name, it only offers suggestions as far as the record name. It would be even more useful if it could suggest a list of fields that match the current PV name beyond the record. This is what it looks like at the moment:
```
epics> dbpf CA<tab>
epics> dbpf CA_TESTS:<tab>
CA_TESTS:DONE  CA_TESTS:LONG  CA_TESTS:SEQ   
epics> dbpf CA_TESTS:S<tab>
epics> dbpf CA_TESTS:SEQ.<tab>
CA_TESTS:  
epics> dbpf CA_TESTS:SEQ.D<tab>
CA_TESTS:  
```

This project may be fairly advanced and require an understanding of dbStaticLib and how the GNU Readline API works, but the initial implementation by @mdavidsaver in #274 should give some decent pointers where the enhancement would need to be added.

rosesyrett:
I'm going to take a crack at this... perhaps foolishly

SourabhHalli:
Will try my luck on this!

PR #365, opened 2023-03-10
WIP: Add autocompletion for Fields in the ioc shell

rosesyrett:
first working version of autocompletion for PVs

At the moment this was only tested for the dbpf command.


rosesyrett:
relates to https://github.com/epics-base/epics-base/issues/322

rosesyrett:
have just noticed that, if I try to autocomplete a non-existing field entry, I get a corruption error.

my test record:
```
record(ai, "something") {
    field(DESC, "some test thing")
}
```

running the ioc shell:
```
epics> dbpf something.G<tab>
epics> dbpf somethree.Gdouble free or corruption (out)
```

This is the call stack:
```
libc.so.6!__pthread_kill_implementation(int no_tid, int signo, pthread_t threadid) (pthread_kill.c:44)
libc.so.6!__pthread_kill_internal(int signo, pthread_t threadid) (pthread_kill.c:78)
libc.so.6!__GI___pthread_kill(pthread_t threadid, int signo) (pthread_kill.c:89)
libc.so.6!__GI_raise(int sig) (raise.c:26)
libc.so.6!__GI_abort() (abort.c:79)
libc.so.6!__libc_message(enum __libc_message_action action, const char * fmt) (libc_fatal.c:155)
libc.so.6!malloc_printerr(const char * str) (malloc.c:5664)
libc.so.6!_int_free(mstate av, mchunkptr p, int have_lock) (malloc.c:4588)
libc.so.6!__GI___libc_free(void * mem) (malloc.c:3391)
libCom.so.3.22.1!__gnu_cxx::new_allocator<char const*>::deallocate(__gnu_cxx::new_allocator<char const*> * const this, const char ** __p, __gnu_cxx::new_allocator<char const*>::size_type __t) (/usr/include/c++/11/ext/new_allocator.h:145)
libCom.so.3.22.1!std::allocator_traits<std::allocator<char const*> >::deallocate(std::allocator_traits<std::allocator<char const*> >::allocator_type & __a, std::allocator_traits<std::allocator<char const*> >::pointer __p, std::allocator_traits<std::allocator<char const*> >::size_type __n) (/usr/include/c++/11/bits/alloc_traits.h:496)
libCom.so.3.22.1!std::_Vector_base<char const*, std::allocator<char const*> >::_M_deallocate(std::_Vector_base<char const*, std::allocator<char const*> > * const this, std::_Vector_base<char const*, std::allocator<char const*> >::pointer __p, std::size_t __n) (/usr/include/c++/11/bits/stl_vector.h:354)
libCom.so.3.22.1!std::_Vector_base<char const*, std::allocator<char const*> >::~_Vector_base(std::_Vector_base<char const*, std::allocator<char const*> > * const this) (/usr/include/c++/11/bits/stl_vector.h:335)
libCom.so.3.22.1!std::vector<char const*, std::allocator<char const*> >::~vector(std::vector<char const*, std::allocator<char const*> > * const this) (/usr/include/c++/11/bits/stl_vector.h:683)
libCom.so.3.22.1!(anonymous namespace)::Tokenize::~Tokenize((anonymous namespace)::Tokenize * const this) (/home/rose/Documents/projects/epics-base/modules/libcom/src/iocsh/iocsh.cpp:245)
libCom.so.3.22.1!(anonymous namespace)::iocsh_attempt_completion(const char * word, int start, int end) (/home/rose/Documents/projects/epics-base/modules/libcom/src/iocsh/iocsh.cpp:619)
libreadline.so.8![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libreadline.so.8!rl_complete_internal (Unknown Source:0)
libreadline.so.8!_rl_dispatch_subseq (Unknown Source:0)
libreadline.so.8!readline_internal_char (Unknown Source:0)
```

So, that obviously needs to be fixed first. Putting this as a WIP PR as a result.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -41,6 +41,9 @@
 
 #include "epicsReadlinePvt.h"
 
+//DO NOT COMMIT THIS BIT...
+#define EPICS_COMMANDLINE_LIBRARY EPICS_COMMANDLINE_LIBRARY_READLINE
+
```
henrique-silva: I suspect this causes the following compiler warning:
```
../iocsh/iocsh.cpp:45: warning: "EPICS_COMMANDLINE_LIBRARY" redefined
   45 | #define EPICS_COMMANDLINE_LIBRARY EPICS_COMMANDLINE_LIBRARY_READLINE
      | 
```

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -571,8 +576,25 @@ char** iocsh_attempt_completion(const char* word, int start, int end)
...
+            if (dot != std::string::npos) {
+                pv = str.substr(0, dot);
+                init_field = str.substr(dot+1u, str.length());
+            }
+
+            if(!init_field.empty()) {
+                return iocshCompleteField(pv.c_str(), init_field.c_str());
+            }
+
+            //Q: where is iocshCompleteRecord instantiated?
```
henrique-silva: Apparently it comes from here: https://github.com/epics-base/epics-base/blob/5eff3803a83fbc3ea855b63841f5359c09830ade/modules/database/src/ioc/db/dbIocRegister.c#L506

https://github.com/epics-base/epics-base/blob/5eff3803a83fbc3ea855b63841f5359c09830ade/modules/database/src/ioc/dbStatic/dbCompleteRecord.cpp#L90

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -571,8 +576,25 @@ char** iocsh_attempt_completion(const char* word, int start, int end)
...
             if(argdef->type==iocshArgStringRecord && iocshCompleteRecord) {
-                return (*iocshCompleteRecord)(word);
+                return iocshCompleteRecord(word);
```
henrique-silva, line 592: Does this actually work? Because `iocshCompleteRecord` is a pointer to a function, so you would need to dereference it first to call the actual function.  
See: https://github.com/epics-base/epics-base/blob/5eff3803a83fbc3ea855b63841f5359c09830ade/modules/database/src/ioc/db/dbIocRegister.c#L506

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -526,10 +531,10 @@ char** iocsh_attempt_completion(const char* word, int start, int end)
         size_t linelen = strlen(line);
         std::vector<char> lbuf(linelen+1u);
         memcpy(&lbuf[0], line, linelen);
-        lbuf[linelen] = '\0';
+        lbuf[linelen] = '\0';
```
henrique-silva: Extra space at end of line

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -507,6 +511,7 @@ char* iocsh_complete_variable(const char* word, int notfirst)
 char** iocsh_attempt_completion(const char* word, int start, int end)
 {
     const char *line = rl_line_buffer;
+    // printf("\n%d\n", isspace(*line));
```
henrique-silva: Debug printf?

modules/libcom/src/iocsh/iocsh.cpp
rosesyrett: Yes, I can't quite remember who advised me on this, but effectively in c++ you can call function pointers in the same way you call regular functions. I've tested commands in the shell and it seems to work just fine.

modules/libcom/src/iocsh/iocsh.cpp
rosesyrett: yes, I have removed this in the latest commit. This was just so my IDE understood what was going on

rosesyrett:
In the latest commit, I have changed how the field completion works such that if something like the following is typed in:

```
epics> dbpf MY-IOC:MY-PV.g<tab>
```
(where g above is obviously not the start of any field names), then, all possible suggestions of MY-PV's fields are returned (if MY-PV is itself valid).


However, I seem to be getting a segfault... and I think it is to do with the readline part of this. 

```
if (total_suggestions == 0) {
            result= getAllFields(&ent, pv);
            char * fullInput = (char *) malloc(pvlen + strlen(init_field) + 2u);
            snprintf(fullInput, pvlen + strlen(init_field) + 2u, "%s.", pv);
            strncat(fullInput, init_field, strlen(init_field));
            result[0] = fullInput;
            return result;
        } 
```
The above gets executed when a non-valid field name is passed through to it - there are no "suggestions" so it goes into the 'if' statement. If I replace 'return result' with a 'return NULL' then the code executes fine. This bit of code is called within a function called by iocsh_attempt_completion in modules/libcom/src/iocsh/iocsh.cpp, and I'm not sure exactly where this gets called but my debugger suggests it is by modules/libcom/src/osi/os/default/gnuReadline.c in line 70. 

The result array ends up having some 95 elements in my case. Could it be that the segfault comes from readline not correctly freeing the memory in this array? I'm not sure why else it'd be producing a segfault.

modules/libcom/src/iocsh/iocsh.cpp
rosesyrett: good point, I'll remove this now.

rosesyrett:
I think to progress further with this I would need a bit of help on what readline does with the `char**` 's it recieves from the `iocsh_attempt_completion` function. I'm not sure where to start as I'm not even sure how or where to find the caller of this function. Thanks.

mdavidsaver:
> ... I would need a bit of help on what readline does with the `char**` 's it recieves from the `iocsh_attempt_completion` function ...

You aren't alone.  Test best "documentation" I've found is the comments and code of `rl_completion_matches()` in the readline source.  From this I infer that for this array of strings `char **ret`, the first string `ret[0]` is the longest common prefix of all possible matches.  Readline will complete with this string.  The remaining strings are a NULL terminated list of possible completions which are shown by a double tab.  eg. two possibilities:

```
ret[0] = "prefix:"
ret[1] = "prefix:one"
ret[2] = "prefix:two"
ret[3] = NULL
```

would complete up to `prefix:` with a single tab, and a double tab will show `prefix:one` and `prefix:two`.

Additionally, it seems to be the case that when only one suggestion is given, whitespace is also added to begin the next argument.  eg.

```
ret[0] = "prefix:one"
ret[1] = "prefix:one"
ret[2] = NULL
```

would complete with `prefix:one ` (with a trailing space).

This is what currently happens when `dbCompleteRecord()` is called with a full record name.  To facilitate field name completion, I think it would be convenient to instead give two suggestions `prefix:one` and `prefix:one.` (with a trailing dot) to avoid needing to hit backspace each time.

Further, instead of adding a separate `dbCompleteField()` and file, I think it would be better to extend `dbCompleteRecord()`.  Maybe splitting out some bits into `static` helper functions.  I think the skeleton of this logic might look something like the following pseudo-code:

```
char** dbCompleteRecord(const char *cword) {
    const char *orig = cword;
    if(0==dbFindRecordPart(&cword)) { // have entire record name
        // cword updated to skip past record name.  eg. to '.' or '\0' or first char of field name
        if(cword[0]=='.') {
            if(0==dbFindField(&cword[1]) { // have complete field name.  all done
                return {orig, orig, NULL};
            } else { // partial, or no, field name
                // .. like current logic, replacing eg. dbFirstRecord() with dbFirstField()
            }
        } else { // just the record name.  suggest a trailing '.'
            return {orig, orig, orig+".", NULL};
        }
    } else { // partial, or no, record name
        // .. current logic
    }
}
```

modules/libcom/src/iocsh/iocsh.cpp
anjohnson: It works in C code too, there are implicit conversions defined by the language and we use this elsewhere in the IOC code. The Pointers to functions section of [this page](https://en.cppreference.com/w/c/language/pointer) is a good place to start if you want to read all about it.

mdavidsaver: The current code is my personal style to make it clear that there is a pointer de-reference here.  Otherwise someone reading later (eg. me) can miss that this is not a regular function call.  Not a big deal in this case as the NULL check on the previous line should be a clue (surely I wouldn't miss that!).

mdavidsaver:
With the "WIP" in the title, I'm going to switch this to a draft.  Feel free to switch it back when you think this becomes when ready.

huyong1979:
I have interest of continuing rosesyrett's work on this feature during EPICS & Phoebus Developers Meeting 2024 at Brookhaven National Lab (BNL).

Issue #362, opened 2023-03-09
iocsh cannot handle escaped quotes in quotes


dirk-zimoch:
Commands like `echo "x\"y"` falsely trigger an "unbalanced quote" error.
The reason is that within the quoted string, backslash is not recognizes and thus the escaped inner quote is interpreted as the closing quote of the string.

daykin:
The solution seems trivial enough - when you are checking for 'quote=c' e.g. in iocsh.cpp:337, check if preceding character is a backslash as well. But, do we want to handle [literal backslashes, which themselves are escaped?](https://xkcd.com/1638)

dirk-zimoch:
Simply checking for `\"` is not enough. What about `echo "\\"` or `echo "\\\\\\\\\\\""`?

dirk-zimoch:
But the soulution of course goes into this direction: Properly handle the `backslash` variable in the code.

daykin:
OK, how about this:
1. During scanning, a backslash is seen - point to it with `firstBackslash`. 
2. Continue scanning (icin++), until a backslash is not seen. 
3. `backslash` is set to `(icin - firstBackslash)%2`
4. even numbers of backslashes are thus paired off as literal backslashes (the C way), and odd numbers will escape the character immediately following.

daykin:
Ah, you've beaten me to the punch 👍 
I agree that this whole thing should probably be simplified.

PR #367, opened 2023-03-10
fix escaped quotes in quoted strings

dirk-zimoch:
Fix for issue #362 
Needs tests.

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -299,7 +299,7 @@ struct Tokenize {
                 }
             }
             if (inword) {
-                if (c == quote) {
+                if (c == quote && !backslash) {
                     quote = 0;
                 }
                 else {
@@ -319,6 +319,10 @@ struct Tokenize {
                     }
                     else {
                         line[icout++] = c;
+                            if (c == '\\' && !backslash) {
+                                backslash = true;
+                                continue;
+                            }
                     }
                 }
             }
--- a/modules/libcom/test/iocshTest.cpp
+++ b/modules/libcom/test/iocshTest.cpp
@@ -83,14 +83,23 @@ void assertCallFunc(const iocshArgBuf *args)
     iocshSetError(args[0].ival);
 }
 
+const iocshArg printArg0 = {"string", iocshArgString};
+const iocshArg * const printArgs[1] = { &printArg0 };
+const iocshFuncDef printFuncDef = {"print",1,printArgs};
+void printCallFunc(const iocshArgBuf *args)
+{
+    reached.insert(args[0].sval);
+}
+
 } // namespace
 
 MAIN(iocshTest)
 {
-    testPlan(19);
+    testPlan(39);
     libComRegister();
     iocshRegister(&positionFuncDef, &positionCallFunc);
     iocshRegister(&assertFuncDef, &assertCallFunc);
+    iocshRegister(&printFuncDef, &printCallFunc);
     findTestData();
 
     testFile("iocshTestSuccess.cmd");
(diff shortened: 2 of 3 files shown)


dirk-zimoch:
The code is so convoluted that I think I will rewrite it.

mdavidsaver:
@dirk-zimoch Could you add a regression test for quoting?

I think this should be as simple as adding another `testCmd()` to `iocshTest.cpp`.

https://github.com/epics-base/epics-base/blob/216359974c97b26ed9e8b4e5931baa3af2f7a2a6/modules/libcom/test/iocshTest.cpp#L109-L110

dirk-zimoch:
I had started with it but found no time the last few weeks.

dirk-zimoch:
Tests added and rebased onto current 7.0

mdavidsaver:
This test coverage seems reasonable to me.  Thanks Dirk.

dirk-zimoch:
For information: The lines that were failing without the fix are:
`position "Test \"escaped double quotes\" in double quotes"` and
`position 'Test \'escaped single quotes\' in single quotes'`.

anjohnson:
Core Group: Add tests that use environment variables, both setting them and using ones set outside that have escaped chars in them. Release notes explaining how we think escaping should now work would be good.

dirk-zimoch:
Macros that expand to quotes are tricky:
```
epics> epicsEnvSet SQ0 '
Unbalanced quote.
epics> epicsEnvSet DQ0 "
Unbalanced quote.
epics> epicsEnvSet SQ1 \'
epics> epicsEnvShow SQ1
SQ1='
epics> epicsEnvSet DQ1 \"
epics> epicsEnvShow DQ1
DQ1="
epics> echo $(SQ1)

epics> echo $(DQ1)

epics> echo "$(SQ1)"

epics> echo "$(DQ1)"

# bash would print ' and " here.

epics> epicsEnvSet SQ2 \\'
Unbalanced quote.
epics> epicsEnvSet DQ2 \\'
Unbalanced quote.
epics> epicsEnvSet SQ3 \\\'
epics> epicsEnvShow SQ3
SQ3=\'
epics> epicsEnvSet DQ3 \\\"
epics> epicsEnvShow DQ3
DQ3=\"
epics> echo $(SQ3)
Unbalanced quote.
epics> echo "$(SQ3)"
'
epics> echo $(DQ3)
Unbalanced quote.
epics> echo "$(DQ3)"
Unbalanced quote.
# bash would print \'  and \" here.

epics> epicsEnvSet SQ5 \\\\\'
epics> epicsEnvShow SQ5
SQ5=\\'
epics> epicsEnvSet DQ5 \\\\\"
epics> epicsEnvShow DQ5
DQ5=\\"
epics> echo $(SQ5)
Trailing backslash.
epics> echo $(DQ5)
Trailing backslash.
epics> echo "$(SQ5)"
Unbalanced quote.
epics> echo "$(DQ5)"
Unbalanced quote.
# bash would print \\' and \\" here.

epics> epicsEnvSet DQ7 \\\\\\\"
epics> epicsEnvShow SQ7
SQ7=\\\'
epics> epicsEnvSet DQ7 \\\\\\\"
epics> epicsEnvShow DQ7
DQ7=\\\"
epics> echo $(SQ7)
'
epics> echo "$(SQ7)"
'
epics> echo $(DQ7)
"
epics> echo "$(DQ7)"
"
# bash would print \\\' and \\\" here.
```

This is very different from similar macros in bash. But the strangeness is from macro expansion rather than iocsh reading the command. bash escapes quotes in expanded macros properly, in context of surrounding quotes. Our macro expansion is totally ignorant of its quotation context.

dirk-zimoch:
See also:
```
epics> echo "\x40"
@
epics> epicsEnvSet at "\x40"
epics> epicsEnvShow at
at=\x40
epics> echo $(at)
x40
epics> echo "$(at)"
x40
```

Again in bash, with `at="\x40"` `echo -e $at` and  `echo -e "$at"` both print `@`.

mdavidsaver:
Core Group: We acknowledge that current macro expansion behavior is terrible.  Please add tests to document the current situation.

dirk-zimoch:
Not yet quite as I would like to have it, i.e like in bash: In double quotes, double quotes can be escaped with backslash and the backslash is removed. Otherwise backslashes are kept except when escaping a backslash which would otherwise escape a double quote. In single quotes, backslashes are always kept and it is not possible to escape single quotes.

irierr:
Is this still something that will be implemented? 
And is there any workaround to include double quotes in an environment variable string for now?

dirk-zimoch:
I had no time to fix the macro quote expansion. And I would not be sure not to break existing expected behavior.

PR #371, opened 2023-04-05
fix problem with commands returning multiple trailing newlines

dirk-zimoch:
The command `git show -s --format=%ci HEAD` may output more than one trailing newline under some circumstances (git version 1.8.3.1 if the commit is a merge). This leads to invalid C-code in epicsVCS.h if not all trailing newlines are removed. The default for `chomp` is to remove only one newline, unless one sets `$/='';` first. https://perldoc.perl.org/functions/chomp.

### DIFF
--- a/src/tools/genVersionHeader.pl
+++ b/src/tools/genVersionHeader.pl
@@ -55,7 +55,7 @@ if (-d '_darcs') { # Darcs
     # v1-4-dirty
     # is tag 'v1' plus 4 patches
     # with uncommited modifications
-    my @tags = `darcs show tags`;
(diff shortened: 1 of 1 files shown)


dirk-zimoch:
See https://bugs.launchpad.net/epics-base/+bug/2015234

anjohnson:
Core Group: @ralphlange Were you going to test this on Darcs, just to show that the generalized setting change hasn't broken that VCS? If not we're happy with this.

mdavidsaver:
@ralphlange ping.

ralphlange:
With that change - fine for rebase-and-merge.

PR #373, opened 2023-04-11
Change compiler for FreeBSD 13: Use clang

tboegi:
FreeBSD 13 uses clang, not gcc.

### DIFF
--- a/configure/os/CONFIG.freebsd-x86.freebsd-x86
+++ b/configure/os/CONFIG.freebsd-x86.freebsd-x86
@@ -2,6 +2,7 @@
 # Definitions for freebsd-x86 host - freebsd-x86 target builds
 # Sites may override these definitions in CONFIG_SITE.freebsd-x86.freebsd-x86
 #-------------------------------------------------------
+GNU_DIR=/usr/local
 
 # Include common gnu compiler definitions
 include $(CONFIG)/CONFIG.gnuCommon
--- a/configure/os/CONFIG.Common.freebsdCommon
+++ b/configure/os/CONFIG.Common.freebsdCommon
@@ -7,6 +7,10 @@
 
 # Include definitions common to all Unix targets
 include $(CONFIG)/os/CONFIG.Common.UnixCommon
+GNU         = NO
+CMPLR_CLASS = clang
+CC          = clang
+CCC         = clang++
 
 OS_CLASS = freebsd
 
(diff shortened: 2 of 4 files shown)


configure/os/CONFIG.freebsd-x86.freebsd-x86
```diff
@@ -4,7 +4,75 @@
 #-------------------------------------------------------
 
 # Include common gnu compiler definitions
-include $(CONFIG)/CONFIG.gnuCommon
+#include $(CONFIG)/CONFIG.gnuCommon
```
mdavidsaver: What necessitates (almost?) duplicating the contents of `CONFIG.gnuCommon` twice?

tboegi: @mdavidsaver :  Thanks for the rewiew. The code duplication was removed now.

mdavidsaver: Yes, much shorter now!

anjohnson:
Looks good to me now, thanks!

mdavidsaver:
Whenever freebsd support comes up, I do some searching about the practicality of our adding CI coverage.

I find an [acknowledgement](https://wiki.freebsd.org/BuildingOnNonFreeBSD) that cross-compiling from Linux is useful for CI, but find no concrete recipe.

An older, but more interesting, [recipe](https://smallhacks.wordpress.com/2018/01/08/c-cross-compilation-on-linux-for-the-freebsd-target-using-gcc-or-clang/) which claims that cross-compiling from Linux does not need a special (re)build of `clang`.  [referenced example](https://github.com/smartmontools/docker-build/blob/master/Dockerfile#L72)

Apparently travis-ci.com has some support for [freebsd](https://docs.travis-ci.com/user/reference/freebsd/).  Of course, I stopped paying attention with the demise of travis-ci.org, so  I don't know how usable this is to us now.

mdavidsaver:
Also worth noting that qemu has [some](https://wiki.freebsd.org/QemuUserModeToDo) [support](https://www.qemu.org/docs/master/user/main.html#bsd-user-space-emulator) for user mode emulation of freebsd.

I know from past experience that user mode emulation doesn't support some of the more esoteric `ioctl()` used in CA code, so it wasn't possible to run a full IOC that way.  However, most of our unittests can run.

mdavidsaver:
Applied as 17ad04505e52f1d712c9073c66cfb6ae77ff47ac

PR #374, opened 2023-04-11
rtems: Close NTP socket

kiwichris:
This patch fixes a bug where the socket is not closed. This code is called once during start up.

### DIFF
--- a/modules/libcom/RTEMS/epicsNtp.c
+++ b/modules/libcom/RTEMS/epicsNtp.c
@@ -51,6 +51,7 @@ int epicsNtpGetTime(char *ntpIp, struct timespec *now)
 
   // Call up the server using its IP address and port number.
   if ( connect( sockfd, ( struct sockaddr * ) &serv_addr, sizeof( serv_addr) ) < 0 ) {
+    close( sockfd );
     perror( "epicsNtpGetTime" );
     return -1;
   }
@@ -58,6 +59,7 @@ int epicsNtpGetTime(char *ntpIp, struct timespec *now)
   // Send it the NTP packet it wants. If n == -1, it failed.
   n = write( sockfd, ( char* ) &packet, sizeof( ntp_packet ) );
   if ( n < 0 ) {
+    close( sockfd );
     perror( "epicsNtpGetTime" );
     return -1;
   }
@@ -65,9 +67,12 @@ int epicsNtpGetTime(char *ntpIp, struct timespec *now)
 // Wait and receive the packet back from the server. If n == -1, it failed.
   n = read( sockfd, ( char* ) &packet, sizeof( ntp_packet ) );
   if ( n < 0 ) {
+    close( sockfd );
     perror( "epicsNtpGetTime" );
     return -1;
   }
+
+  close( sockfd );
 
   // These two fields contain the time-stamp seconds as the packet left the NTP server.
   // The number of seconds correspond to the seconds passed since 1900.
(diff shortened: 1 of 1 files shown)


mdavidsaver:
@hjunkes fyi.

kiwichris:
I see there is a CI fail for a platform I do not think is RTEMS and I do not think is related to the code touched in this merge request. Is there a problem?

mdavidsaver:
> ... Is there a problem?

Yes, but not one which concerns this PR.  cf. https://github.com/epics-base/epics-base/discussions/162#discussioncomment-5599472

kiwichris:
> > ... Is there a problem?
> 
> Yes, but not one which concerns this PR. cf. [#162 (reply in thread)](https://github.com/epics-base/epics-base/discussions/162#discussioncomment-5599472)

Thanks. I will ignore this result for other pull requests I create.

modules/libcom/RTEMS/epicsNtp.c
```diff
@@ -51,24 +51,29 @@ int epicsNtpGetTime(char *ntpIp, struct timespec *now)
...
     perror( "epicsNtpGetTime" );
     return -1;
   }
 
 // Wait and receive the packet back from the server. If n == -1, it failed.
   n = read( sockfd, ( char* ) &packet, sizeof( ntp_packet ) );
   if ( n < 0 ) {
+    close( sockfd );
     perror( "epicsNtpGetTime" );
     return -1;
```
mdavidsaver, line 72: `close()` may clobber the `errno` which `perror()` sees.  Simplest to `close()` after printing the error.  Also, it would be nice if the three calls to `perror()` printed different messages so that they could be distinguished.

kiwichris: Thanks for the review. I can make those changes.

A couple of things to note:

1. I am planning to make further changes to this module so it can be used with LibBSD and the Legacy networking stack. The NTP provider for RTEMS in EPICS uses code that is only part of the Legacy stack and that code will not be moved to LibBSD. The code in EPICS calls [rtems_bsdnet_get_ntp()](https://git.rtems.org/rtems-net-legacy/tree/lib/rtems_bsdnet_ntp.c#n154) and I think this code is best moved into this file and the support be self contained. This change could be merged into that work?
2. The change does nothing at the moment because the code is disabled. The patch [ntp-provider-with-rtems-posix](https://github.com/epics-base/epics-base/compare/7.0...kiwichris:epics-base:ntp-provider-with-rtems-posix) enables it but this change will not build with LibBSD.

anjohnson: Core Group: Can we resolve the issue with losing the error status here since this code is in EPICS and we'll merge that; future changes could come in later PRs.

kiwichris: An update from me on this PR. 

We have moved to using `ntpd` in RTEMS for Gemini and not this code or this option.

hjunkes:
We unfortunately neglected Dierk's remark that here 'perror' belongs before 'close( sockfd )'.
Another point. 
Our RTEMS efforts are diverging and we need to synchronize those. I'll write about this on Core Talk in a moment.

anjohnson:
I have just pushed a commit to this PR branch that fixes the issue of `close()` clobbering `errno` before `perror()` displays the error. Gemini and @kiwichris are now using ntpd instead of this routine to fetch NTP time, but other sites may still be relying on this, so this PR is still valuable as it fixes the resource leak for them.

mdavidsaver (approved):
Looks fine now.

kiwichris:
@anjohnson thank you for following this up and providing a patch.

mdavidsaver:
Applied as 3ea29f581b07a8baea8988c7a37a316e4ffab34a

Issue #300, opened 2022-11-01
`select()` portability and FD limits on RTEMS


mdavidsaver:
From https://docs.rtems.org/branches/master/user/migration/v4_11-to-v5.html

> In RTEMS 5.1, the list of free file descriptors has a LIFO ordering in contrast to previous versions where it was a FIFO. This means if an application regularly opens and closes files (or sockets) it sees the whole range of file descriptors. The reason for this change was to increase the time before file descriptors are reused to more likely catch a file descriptor use after close.
> 
> This change may surface application issues. If the configured file descriptor maximum (CONFIGURE_MAXIMUM_FILE_DESCRIPTORS) is greater than the FD_SETSIZE defined by Newlib to 64, then calls to select() are undefined behaviour and may corrupt the thread stack. In particular, FD_SET() may result in an out of bounds access. It is possible to define a custom FD_SETSIZE. The application must ensure that the custom FD_SETSIZE is defined before <sys/select.h> is included in all modules used by the application, for example via a global compiler command line define. This applies also to all third-party libraries used by the application.

This is the reason that the configured maximum number of file descriptors has been reduced from 150 (RTEMS <= 4.x) to 64 (RTEMS >= 5.x)

https://github.com/epics-base/epics-base/blob/0f8ea3aa3671cb0433b4350020430a7740d6fdc6/modules/libcom/RTEMS/score/rtems_config.c#L36

https://github.com/epics-base/epics-base/blob/0f8ea3aa3671cb0433b4350020430a7740d6fdc6/modules/libcom/RTEMS/posix/rtems_config.c#L64

To be clear this reduction is made out of an abundance of caution.  `FD_SETSIZE` effects mainly (only?) the `select()` function.  Applications/sites which don't call `select()` may safely raise this limit.

At present, IOC core and the PVA modules do not use `select()`.

The only usage of `select()` in Base is in the `fdManager` code, which is mainly encountered in the PCAS and ca-gateway code.

mdavidsaver:
The only usage of `select()` in Base is the fdManager utility[1].  I'm not aware of any usage in IOCs.

https://github.com/epics-base/epics-base/blob/0f8ea3aa3671cb0433b4350020430a7740d6fdc6/modules/libcom/src/fdmgr/fdManager.cpp#L121

[...]

PR #375, opened 2023-04-13
RTEMS 6 Updates for Legacy Networking, Makefile.inc and MVME2700

kiwichris:
These patches update the RTEMS 6 in EPICS 7.0 to:

1. Add support for the Legacy Networking stack as a separate package. RTEMS 6 moved the legacy network stack from the `rtems.git` repo to the `rtems-net-legacy.git` repo for long term maintenance support. The legacy stack can be built as a 3rd party library like LibBSD and this change lets EPICS 7 support it. The change is backwards compatible with RTEMS 5.3.0 with the networking enabled. It automatically detects the install stack or an error is raised.
2. Fix `CONFIG.Common.RTEMS` to use the `Makefile.inc` provided `LDFLAGS`. A patch for RTEMS 6's `Makefile.inc` support fixes the build issue related to `DBLINK` errors and the `link.h` header file. The RTEMS 6 patch is needed for RTEMS 6 and EPICS 7 to build for the MVME2700 BSP.
3. RTEMS 6 now supports the MVME2700 BSP. The patch supports MVME2700 on RTEMS 6 while supporting `mvme2307` on RTEMS 5.
4. Clean up of the bootp server address declaration in `rtems_config.c`.
5. Fix the shell command declarations for use with the legacy stack as a 3rd party library. I also added `top` and `rtems` while making the change.

### DIFF
--- a/configure/os/CONFIG.Common.RTEMS
+++ b/configure/os/CONFIG.Common.RTEMS
@@ -65,7 +65,7 @@ CXXFLAGS = $($(BUILD_CLASS)_CXXFLAGS) $(POSIX_CXXFLAGS) $(OPT_CXXFLAGS)\
  $(USR_CXXFLAGS) $(CMD_CXXFLAGS) $(ARCH_DEP_CXXFLAGS) $(CODE_CXXFLAGS)\
  $(STATIC_CXXFLAGS) $(OP_SYS_CXXFLAGS) $(LIBRARY_SRC_CFLAGS)
 
-LDFLAGS = $(OPT_LDFLAGS) $(TARGET_LDFLAGS) $(USR_LDFLAGS) $(CMD_LDFLAGS)\
+LDFLAGS += $(OPT_LDFLAGS) $(TARGET_LDFLAGS) $(USR_LDFLAGS) $(CMD_LDFLAGS)\
  $(POSIX_LDFLAGS) $(ARCH_DEP_LDFLAGS) $(DEBUG_LDFLAGS) $(OP_SYS_LDFLAGS)\
  $($(BUILD_CLASS)_LDFLAGS) $(RUNTIME_LDFLAGS) $(CODE_LDFLAGS)
 
@@ -102,15 +102,47 @@ OS_CLASS = RTEMS
 # Operating system compile & link flags
 OP_SYS_CFLAGS += -D__LINUX_ERRNO_EXTENSIONS__
 
-OP_SYS_CFLAGS_NET_yes = -DRTEMS_LEGACY_STACK
-OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_HAS_NETWORKING))
-
-ifeq ($(RTEMS_HAS_POSIX_API),yes)
-POSIX_CPPFLAGS = -D_GNU_SOURCE -D_DEFAULT_SOURCE
+# Has RTEMS been built with the internal legacy stack?
+ifeq ($(RTEMS_LEGACY_NETWORKING_INTERNAL),yes)
+RTEMS_HAS_NETWORKING = yes
+RTEMS_NETWORKING = legacy_internal
 endif
 
-OP_SYS_LDLIBS_posix_NET_yes = -ltftpfs -lnfs -lz -ltelnetd
-OP_SYS_LDLIBS_posix_NET_no = -ltftpfs -lbsd -lz
+# Has RTEMS been built with the legacy stack as a separate package?
+ifeq ($(RTEMS_LEGACY_NETWORKING),yes)
+RTEMS_HAS_NETWORKING = yes
+RTEMS_NETWORKING = legacy
+endif
+
+# Has RTEMS been built with the libbsd stack as a separate package?
+ifeq ($(RTEMS_BSD_NETWORKING),yes)
+RTEMS_HAS_NETWORKING = yes
+RTEMS_NETWORKING = bsd
+endif
+
+RTEMS_LEGACY_NET_LIB_no=
+
+# Legacy network with RTEMS 5 and earlier
+RTEMS_NET_LIB_legacy_internal=-lnfs
+OP_SYS_CFLAGS_NET_legacy_internal = -DRTEMS_LEGACY_STACK
+
+# Legacy network with RTEMS 6 is a separate package and library
+RTEMS_NET_LIB_legacy=-lnfs -lnetworking -lnfs
+OP_SYS_CFLAGS_NET_legacy = -DRTEMS_LEGACY_STACK
+
+# LibBSD network with RTEMS 5 and 6 is a separate package and library
+RTEMS_NET_LIB_bsd=-lbsd
+OP_SYS_CFLAGS_NET_bsd = -DRTEMS_LIBBSD_STACK
+
+# Set the networking flags
+OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_NETWORKING))
+
+POSIX_CPPFLAGS_posix = -D_GNU_SOURCE -D_DEFAULT_SOURCE
+POSIX_CPPFLAGS = $(POSIX_CPPFLAGS_$(OS_API))
+
+OP_SYS_LDLIBS_posix_NET_yes = -ltftpfs -lz -ltelnetd
+OP_SYS_LDLIBS_posix_NET_yes += $(RTEMS_NET_LIB_$(RTEMS_NETWORKING))
+OP_SYS_LDLIBS_posix_NET_no = -ltftpfs -lz
 OP_SYS_LDLIBS_score_NET_yes = -lnfs
 OP_SYS_LDLIBS_score_NET_no = -lnfs
 OP_SYS_LDLIBS += -lrtemsCom -lCom
--- a/configure/os/CONFIG.Common.RTEMS-mvme2700
+++ b/configure/os/CONFIG.Common.RTEMS-mvme2700
@@ -23,7 +23,7 @@ endef
 
 include $(CONFIG)/os/CONFIG.Common.RTEMS
 
-ifeq ($(RTEMS_VERSION), $(word 1, $(sort 5 $(RTEMS_VERSION))))
+ifeq ($(shell test $(RTEMS_VERSION) -ge 5; echo $$?),0)
 RTEMS_BSP = mvme2700
 else
 RTEMS_BSP = mvme2307
--- a/configure/os/CONFIG.Common.RTEMS-mvme2700
+++ b/configure/os/CONFIG.Common.RTEMS-mvme2700
@@ -1,7 +1,6 @@
 #
 # Author: Matt Rippa
 #
-RTEMS_BSP = mvme2307
 RTEMS_TARGET_CPU = powerpc
 ARCH_DEP_CFLAGS += -DMY_DO_BOOTP=NULL
 ARCH_DEP_CFLAGS += -DHAVE_PPCBUG
@@ -23,3 +22,9 @@ define MUNCH_CMD
 endef
 
 include $(CONFIG)/os/CONFIG.Common.RTEMS
+
+ifeq ($(RTEMS_VERSION), $(word 1, $(sort 5 $(RTEMS_VERSION))))
+RTEMS_BSP = mvme2700
+else
+RTEMS_BSP = mvme2307
+endif
--- a/configure/toolchain.c
+++ b/configure/toolchain.c
@@ -30,9 +30,26 @@ MSVC_VER = _MSC_VER
 #ifdef __rtems__
 #include <rtems/score/cpuopts.h>
 #  if __RTEMS_MAJOR__>=5
-OS_API = posix
+     OS_API = posix
 #  else
-OS_API = score
+     OS_API = score
+#  endif
+#  if defined(RTEMS_NETWORKING)
+     /* legacy stack circa RTEMS <= 5 and networking internal to RTEMS */
+     RTEMS_LEGACY_NETWORKING_INTERNAL = yes
+#  else
+#    if !defined(__has_include)
+       /* assume old GCC implies RTEMS < 5 with mis-configured BSP */
+#      error rebuild BSP with --enable-network
+#    elif __has_include(<machine/rtems-net-legacy.h>)
+       /* legacy stack circa RTEMS > 5 */
+       RTEMS_LEGACY_NETWORKING = yes
+#    elif __has_include(<machine/rtems-bsd-version.h>)
+       /* libbsd stack */
+       RTEMS_BSD_NETWORKING = yes
+#    else
+#      error Cannot determine RTEMS network configuration
+#    endif
 #  endif
 #endif
 
@@ -51,4 +68,3 @@ COMMANDLINE_LIBRARY ?= $(strip $(if $(wildcard $(if $(GNU_DIR),$(GNU_DIR)/includ
 #if defined(_FORTIFY_SOURCE) && _FORTIFY_SOURCE>2
 OP_SYS_CPPFLAGS += -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
 #endif
-
--- a/modules/libcom/RTEMS/posix/rtems_config.c
+++ b/modules/libcom/RTEMS/posix/rtems_config.c
@@ -2,7 +2,7 @@
 * Copyright (c) 2002 The University of Saskatchewan
 * EPICS BASE Versions 3.13.7
 * and higher are distributed subject to a Software License Agreement found
-* in file LICENSE that is included with this distribution. 
+* in file LICENSE that is included with this distribution.
 \*************************************************************************/
 /*
  * RTEMS configuration for EPICS
@@ -128,16 +128,17 @@ extern void *POSIX_Init(void *argument);
   &rtems_shell_SYSCTL_Command
 #else // LEGACY_STACK:
 #define CONFIGURE_SHELL_USER_COMMANDS \
-  &bsp_interrupt_shell_command, \
-  &rtems_shell_PING_Command, \
-  &rtems_shell_ROUTE_Command, \
-  &rtems_shell_IFCONFIG_Command
-#endif
+  &bsp_interrupt_shell_command
+#endif // not LEGACY_STACK
 
 #define CONFIGURE_SHELL_COMMAND_CPUUSE
 #define CONFIGURE_SHELL_COMMAND_PERIODUSE
 #define CONFIGURE_SHELL_COMMAND_STACKUSE
 #define CONFIGURE_SHELL_COMMAND_PROFREPORT
+#define CONFIGURE_SHELL_COMMAND_TOP
+#define CONFIGURE_SHELL_COMMAND_RTEMS
+
+#define CONFIGURE_SHELL_COMMANDS_ALL_NETWORKING
 
 #define CONFIGURE_SHELL_COMMAND_CP
 #define CONFIGURE_SHELL_COMMAND_PWD
@@ -155,14 +156,15 @@ extern void *POSIX_Init(void *argument);
 #define CONFIGURE_SHELL_COMMAND_SHUTDOWN
 
 #include <rtems/shellconfig.h>
+
 #define RTEMS_BSD_CONFIG_BSP_CONFIG
 #define RTEMS_BSD_CONFIG_SERVICE_TELNETD
 #define RTEMS_BSD_CONFIG_TELNETD_STACK_SIZE (16 * 1024)
 #define RTEMS_BSD_CONFIG_SERVICE_FTPD
 #define RTEMS_BSD_CONFIG_FIREWALL_PF
-#else
+#else // __RTEMS_MAJOR__ > 4
 #include <rtems/shellconfig.h>
-#endif // not LEGACY_STACK
+#endif // __RTEMS_MAJOR__ > 4
 
 #if __RTEMS_MAJOR__ < 5 // still needed in Version 4?
 #define CONFIGURE_MAXIMUM_TASKS             rtems_resource_unlimited(30)
--- a/configure/os/CONFIG.Common.RTEMS-pc386
+++ b/configure/os/CONFIG.Common.RTEMS-pc386
@@ -25,7 +25,7 @@ include $(CONFIG)/os/CONFIG.Common.RTEMS
 OP_SYS_LDFLAGS += -Wl,-Ttext,0x100000
 
 # This check must appear after the above include
-ifeq ($(RTEMS_VERSION),5)
+ifeq ($(shell test $(RTEMS_VERSION) -ge 5; echo $$?),0)
   $(info *** This target is not compatible with the configured RTEMS version.)
   $(info *** Build the RTEMS-pc686 (-qemu) target for RTEMS 5.x)
   $(error Can't continue)
--- a/configure/os/CONFIG.Common.RTEMS-pc686
+++ b/configure/os/CONFIG.Common.RTEMS-pc686
@@ -32,7 +32,7 @@ OP_SYS_LDFLAGS += -Wl,-Ttext,0x100000
 
 
 # This check must appear after the above include
-ifneq ($(RTEMS_VERSION),5)
+ifeq ($(shell test $(RTEMS_VERSION) -lt 5; echo $$?),0)
   $(info *** This target is not compatible with the configured RTEMS version.)
   $(info *** Build the RTEMS-pc386 (-qemu) target for RTEMS 4.x)
   $(error Can't continue)
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -102,9 +102,9 @@ char bootp_server_name_init[128] = "1001.1001@10.0.5.1:/epics";
 char bootp_boot_file_name_init[128] = "/epics/myExample/bin/RTEMS-beatnik/myExample.boot";
 char bootp_cmdline_init[128] = "/epics/myExample/iocBoot/iocmyExample/st.cmd";
 
-struct in_addr rtems_bsdnet_bootp_server_address;
 /* TODO check rtems_bsdnet_bootp_cmdline */
 #ifndef RTEMS_LEGACY_STACK
+struct in_addr rtems_bsdnet_bootp_server_address;
 char *rtems_bsdnet_bootp_server_name = bootp_server_name_init;
 char *rtems_bsdnet_bootp_boot_file_name = bootp_boot_file_name_init;
 char *rtems_bsdnet_bootp_cmdline = bootp_cmdline_init;
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c
@@ -1,0 +1,52 @@
+/*************************************************************************\
+* Copyright (c) 2002 The University of Chicago, as Operator of Argonne
+*     National Laboratory.
+* Copyright (c) 2002 The Regents of the University of California, as
+*     Operator of Los Alamos National Laboratory.
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+/* Author:  Marty Kraimer Date:    18JAN2000 */
+/* Author:  Chris Johns   Date:    21JUL2023 */
+
+/* This is part of the posix implementation of epicsThread */
+
+#include "epicsStdio.h"
+#include "ellLib.h"
+#include "epicsEvent.h"
+#include "epicsThread.h"
+
+#include <pthread.h>
+
+void epicsThreadShowInfo(epicsThreadOSD *pthreadInfo, unsigned int level)
+{
+    if(!pthreadInfo) {
+        fprintf(epicsGetStdout(),"            NAME       EPICS ID   "
+            "PTHREAD ID   OSIPRI  OSSPRI  STATE\n");
+    } else {
+        struct sched_param param;
+        int policy;
+        int priority = 0;
+
+        if(pthreadInfo->tid) {
+            int status;
+            status = pthread_getschedparam(pthreadInfo->tid,&policy,&param);
+            if(!status) priority = param.sched_priority;
+        }
+        fprintf(epicsGetStdout(),"%16.16s %14p %12lu    %3d%8d %8.8s\n",
+             pthreadInfo->name,(void *)
+             pthreadInfo,(unsigned long)pthreadInfo->tid,
+             pthreadInfo->osiPriority,priority,
+             pthreadInfo->isSuspended?"SUSPEND":"OK");
+    }
+}
+
+static void thread_hook(epicsThreadId pthreadInfo)
+{
+    pthread_setname_np(pthreadInfo->tid, pthreadInfo->name);
+}
+
+EPICS_THREAD_HOOK_ROUTINE epicsThreadHookDefault = thread_hook;
+EPICS_THREAD_HOOK_ROUTINE epicsThreadHookMain = thread_hook;


mdavidsaver:
> 1. A user adds `RTEMS_LEGACY_NETWORKING = yes` to their `CONFIG_SITE.local`

Is there a way in which this situation could be automatically detected to avoid burdening users with an extra manual configuration step?

configure/os/CONFIG.Common.RTEMS-mvme2700
```diff
@@ -1,7 +1,11 @@
 #
 # Author: Matt Rippa
 #
+ifeq ($(shell test $(RTEMS_VERSION) -gt 5; echo $$?),0)
+RTEMS_BSP = mvme2700
+else
 RTEMS_BSP = mvme2307
+endif
```
mdavidsaver: I'm confused by this condition.  And for that matter, why ce876d6f114fa1e293c1b278bc11f05bba4cb886 changed mvme2700 to mvme2307 in the first place.  (@anjohnson do you remember what motivated this?)

mdavidsaver:
@kiwichris What is the status of RTEMS 6 wrt. rtems-source-builder?  I don't yet see a `6` branch in git://git.rtems.org/rtems-source-builder.  This is what I look for when updating https://github.com/mdavidsaver/rsb/releases which builds toolchains for our CI jobs.  (at 2 years, I'm past due for an update...)

kiwichris:
> > 1. A user adds `RTEMS_LEGACY_NETWORKING = yes` to their `CONFIG_SITE.local`
> 
> Is there a way in which this situation could be automatically detected to avoid burdening users with an extra manual configuration step?

Yes it could if there is a way to detect libraries installed with the BSP? Doing that was beyond my knowledge of the EPICS build system. If there is an example I can follow I can look at adding that type of support.

Previously the setting was handled by the `RTEMS_NETWORKING` define however that is always `no` or empty with RTEMS 6 builds and that cannot be changed. I am also reluctant to modify a file installed by an RTEMS build when installing a 3rd party package like `rtems-net-legacy`. It can effect the configuration control of a user's deployment.

I am not sure if LibBSD and the Legacy stack can be installed at the same time. If they can the default would be LibBSD however something like this may be useful as an override?

mdavidsaver:
> ... when installing a 3rd party package like `rtems-net-legacy` ...

Could this package define some identification and/or version macros in one its headers (eg. `sys/socket.h`)?  Something like `__GLIBC__` from glibc, or our own `EPICS_VERSION` and friends.  Maybe `#define RTEMS_NETWORK_LEGACY_VERSION 1`?

> Yes it could if there is a way to detect libraries installed with the BSP?

We can't currently do test linking.  However, we have hooks to do pre-processor tests to set Make variables.  eg. to distinguish RTEMS `>= 5`.

https://github.com/epics-base/epics-base/blob/5eff3803a83fbc3ea855b63841f5359c09830ade/configure/toolchain.c#L30-L37

kiwichris:
> @kiwichris What is the status of RTEMS 6 wrt. rtems-source-builder? I don't yet see a `6` branch in git://git.rtems.org/rtems-source-builder. >

RTEMS 6 is the `master` branch of that repo. We are wanting to branch and create a release however until then it is the `master` branch. RTEMS 6 is stable and my efforts with EPICS is to try and smooth the compatibility with previous versions of RTEMS and EPICS. I have the `HEAD` of the EPICS base 7.0 branch running on a MVME2700 with the legacy stack.

> This is what I look for when updating https://github.com/mdavidsaver/rsb/releases which builds toolchains for our CI jobs. (at 2 years, I'm past due for an update...)

I have created the [rtems-deployment.git](https://git.rtems.org/chrisj/rtems-deployment.git/) to provide a way to build RTEMS and related packages like LibBSD in a controlled manner without users needs to work through the dependencies and complexities. It is being used in Gemini and their gitlab CI to build RPMs and then containers. If you are interested in this please reach out to me offline and I am happy to help.

kiwichris:
> > ... when installing a 3rd party package like `rtems-net-legacy` ...
> 
> Could this package define some identification and/or version macros in one its headers (eg. `sys/socket.h`)? Something like `__GLIBC__` from glibc, or our own `EPICS_VERSION` and friends. Maybe `#define RTEMS_NETWORK_LEGACY_VERSION 1`?

That is a good idea. I will chat to Vijay and Gedare about this. I think a header in the includes under `rtems` is what we need.

> > Yes it could if there is a way to detect libraries installed with the BSP?
> 
> We can't currently do test linking. However, we have hooks to do pre-processor tests to set Make variables. eg. to distinguish RTEMS `>= 5`.

If there is a header file a preprocess test would work. I think the `CONFIG_SITE.local` variable is worth keeping.

mdavidsaver:
> That is a good idea. I will chat to Vijay and Gedare about this. I think a header in the includes under `rtems` is what we need.

Ideally, I think such a version ID header should only define macros.  This allows it to be included from many contexts.  eg. assembly code, or in our case from what is actually a Makefile.  If this is burdensome, we could fall back to testing for a specific header file name.  eg. `__has_include(<rtems_net_legacy.h>)`.

kiwichris:
> > That is a good idea. I will chat to Vijay and Gedare about this. I think a header in the includes under `rtems` is what we need.
> 
> Ideally, I think such a version ID header should only define macros.

Agreed

> This allows it to be included from many contexts. eg. assembly code, or in our case from what is actually a Makefile. If this is burdensome, we could fall back to testing for a specific header file name. eg. `__has_include(<rtems_net_legacy.h>)`.

I will leave the network programming in assembler to you 😃

configure/os/CONFIG.Common.RTEMS-mvme2700
anjohnson: I'm not aware of anyone in the EPICS world using an actual MVME2307 board (Dirk might have some though); Gemini and APS both use the MVME2700 which was a later derivative with I believe similar or identical onboard I/O devices. Thus I wanted EPICS to use that model number in our target name. I'm guessing that RTEMS 6 now provides an mvme2700 BSP, but RTEMS 5 used the older model number in the BSP name.

kiwichris: I recently added the `powerpc/mvme2700` BSP to RTEMS 6. I have no immediate plans to update the RTEMS 5 branch to have a `mvme2700`. If that was added would that help? I am not sure how far back compatibility with RTEMS versions EPCIS 7 extends?

If we plan a change to RTEMS 5 can this change be included until RTEMS is released with an `mvme2700` BSP? Note, this assumes it is easy to add the BSP to RTEMS 5. I have not looked into it.

mdavidsaver: I'm not sure an RTEMS change is warranted.  I think we (EPICS devs.) have to own the consequences of making the EPICS target `RTEMS-mvme2700` actually use the `mvme2307` RTEMS BSP.

Is there any benefit to trying to change this name mapping going forward?

mrippa: I've long since requested to properly identify the MVME2700 bsp by the physical board name 2700, not 2307. When Eric Norum helped me test the MVME2700 back around 2009, the change was made in EPICS and RTEMS 4.9. But when RTEMS 4.10 was released this reverted back to MVM2307.  

This is all just semantic, but sound bookkeeping would favor using the 2700 model number and updating EPICS configs to call that name in RTEMS too. I support your best judgement on the issue.

kiwichris: If I make RTEMS 5 support the MVME2700 is that enough for EPICS? Does EPICS 7 support 4.10?

mrippa: EPiCS would still need to update ce876d6 to support the change in RTEMS.

mrippa: @kiwichris  I didn't see the full diff in the conversation page. What I see here looks like it would work for us.

mdavidsaver:
@kiwichris My need to distinguish between the two RTEMS network stacks is becoming more pressing.  Based on a comparison of the install tree of the RTEMS 5 BSPs pc686 (w/ libbsd) vs. mvme3100 (w/ classic/kernel) I see several possibilities.

* Test for the presence of `<sys/jail.h>` which I think is a clearly BSD header name.
* Test for the presence of `<rtems/bsd/bsd.h>` (hopefully) not to be confused with `<rtems/bsd.h>` which exists back to 4.9
* Test for the absence of `<rtems/rtems_bsdnet_internal.h>`
* Test and then include `<machine/rtems-bsd-version.h>` which has `#define __FreeBSD_version 1200018` which looks like almost exactly what I want.  I notice that [1200018](https://docs.freebsd.org/en/books/porters-handbook/versions/) is ~7 years old.  Is this number being updated in libbsd?

Can you provide guidance on whether any of these would be appropriate tests for external code to make going forward?

kiwichris:
> @kiwichris My need to distinguish between the two RTEMS network stacks is becoming more pressing. 

I understand.

> Based on a comparison of the install tree of the RTEMS 5 BSPs pc686 (w/ libbsd) vs. mvme3100 (w/ classic/kernel) I see several possibilities.
> 
>     * Test for the presence of `<sys/jail.h>` which I think is a clearly BSD header name.
> 
>     * Test for the presence of `<rtems/bsd/bsd.h>` (hopefully) not to be confused with `<rtems/bsd.h>` which exists back to 4.9
> 
>     * Test for the absence of `<rtems/rtems_bsdnet_internal.h>`
> 
>     * Test and then include `<machine/rtems-bsd-version.h>` which has `#define __FreeBSD_version 1200018` which looks like almost exactly what I want.  I notice that [1200018](https://docs.freebsd.org/en/books/porters-handbook/versions/) is ~7 years old.  Is this number being updated in libbsd?
> 
> 
> Can you provide guidance on whether any of these would be appropriate tests for external code to make going forward?

I will document detecting networking libraries and headers for RTEMS 6 and RTEMS 5. I seem to remember something about only being able to check headers but I could be wrong.

### RTEMS 6

We now support 3 different stacks and we working to align the features. The library vs stack table is for RTEMS 6 is:

- `libbsd` -> `libbsd.a`
- `legacy` -> `libnetworking.a`
- `lwip` -> `liblwip.a`

If you wish to use headers as a solution I suggest you check for `rtems-net-legacy.h`. I added this to the legacy stack after this patch was posted and I have not had time to revisit this change to update it. The patch to [add rtems-net-legacy.h](https://git.rtems.org/rtems-net-legacy/commit/?id=bef50b9ff76d3f6231e3c92bb35d47ee5bc7f0cb). If this header is present it is the legacy stack and if not you can assume the networking stack is libbsd. Note this is only for RTEMS 6 where the legacy stack is built externally to the RTEMS kernel.

Currently there is no version header for LwIP but we could add that however LwIP needs a lot more work to be usable with EPCIS, eg NFS.

### RTEMS 5

For RTEMS 5 there is only the legacy stack and libbsd and the legacy stack is built as a part of RTEMS so the `RTEMS_NETWORKING` define can be checked. The specific installed file is `rtems/score/cpuopts.h`. If the define is not defined then you can assume it is libbsd.

mdavidsaver:
> ... If the define is not defined then you can assume it is libbsd.

My past experiences with these sorts of compatibility tests lead me to prefer positive conditions over: `#else /* assume the latest and greatest */`.  Given that, does the following make sense?

```c
#include <rtems/score/cpuopts.h>
#if defined(RTEMS_NETWORKING)
// legacy stack circa RTEMS <= 5
#else
#  if !defined(__has_include)
     // assume old GCC implies RTEMS < 5 with mis-configured BSP
#    error rebuild BSP with --enable-network
#  elif __has_include(<rtems/rtems-net-legacy.h>)
     // legacy stack circa RTEMS > 5
#  else __has_include(<machine/rtems-bsd-version.h>)
     // libbsd stack
#  else
#    error Unexpected configuration
#  endif
#endif
```

kiwichris:
> > ... If the define is not defined then you can assume it is libbsd.
> 
> My past experiences with these sorts of compatibility tests lead me to prefer positive conditions over: `#else /* assume the latest and greatest */`. 

That make sense.

> Given that, does the following make sense?
> 
> ```c
> #include <rtems/score/cpuopts.h>
> #if defined(RTEMS_NETWORKING)
> // legacy stack circa RTEMS <= 5
> #else
> #  if !defined(__has_include)
>      // assume old GCC implies RTEMS < 5 with mis-configured BSP
> #    error rebuild BSP with --enable-network
> #  elif __has_include(<rtems/rtems-net-legacy.h>)
>      // legacy stack circa RTEMS > 5
> #  else __has_include(<machine/rtems-bsd-version.h>)
>      // libbsd stack
> #  else
> #    error Unexpected configuration
> #  endif
> #endif
> ```

Yes this looks good with a minor point. Now you have brought this together I am wondering if I should install the legacy header into `machine/rtems-net-legacy.h` so both stacks are in similar locations?

Will your change be merged into this patch? I welcome your input with this patch on auto detection.

mdavidsaver:
> ... I am wondering if I should install the legacy header into `machine/rtems-net-legacy.h` so both stacks are in similar locations?

Gosh Chris, I honestly couldn't care less ;)

> Will your change be merged into this patch?

I hope so (this is your PR).

My immediate usage will be for https://github.com/ChannelFinder/recsync/pull/68 to decide on `select()` vs. `poll()`.

kiwichris:
> > ... I am wondering if I should install the legacy header into `machine/rtems-net-legacy.h` so both stacks are in similar locations?
> 
> Gosh Chris, I honestly couldn't care less ;)

Sure. Let go with `machine/rtems-net-legacy.h` and I will update the legacy repo.
 
> 
> > Will your change be merged into this patch?
> 
> I hope so (this is your PR).

OK, I will update the PR next. I am still working on getting `ntpd` working and a new NTP time provider for epics-base.

> My immediate usage will be for [ChannelFinder/recsync#68](https://github.com/ChannelFinder/recsync/pull/68) to decide on `select()` vs. `poll()`.

Ah I see. We moved the networking headers into newlib so they are installed with the compiler when the tools are built. This has the unfortunate side effect of breaking configure probes for calls such as `poll` when using the legacy stack.

I also see you discovered the `select` and `fd` count issue. The `select` interface is a problem with the `fd_set` size defined in newlib and set to 64. Any more effects effects small memory targets so changing the default is not an option. If the file ops count becomes an issue maybe adding `poll` to the legacy stack is a solution?

mdavidsaver:
wrt. `select()` and the changes to the FD allocation scheme.  We have been aware of this for some time.  see #300 and #384.  I would also be interested in your thoughts about whether and when #382 would actually work.

kiwichris:
> > > Will your change be merged into this patch?
> > 
> > 
> > I hope so (this is your PR).
> 
> OK, I will update the PR next. I am still working on getting `ntpd` working and a new NTP time provider for epics-base.

@mdavidsaver is there an example in the `configure` directory I can follow that shows how I use the header check to determine the networking stack settings? I assume I handle the detection there so `RTEMS_LEGACY_NETWORKING` can be set to `yes`.

mdavidsaver:
> @mdavidsaver is there an example in the `configure` directory

I think this new test of network stack should go along with the current test of `__RTEMS_MAJOR__` in `configure/toolchain.c`, as mentioned above (https://github.com/epics-base/epics-base/pull/375#issuecomment-1506323363).

kiwichris:
> > @mdavidsaver is there an example in the `configure` directory
> 
> I think this new test of network stack should go along with the current test of `__RTEMS_MAJOR__` in `configure/toolchain.c`, as mentioned above ([#375 (comment)](https://github.com/epics-base/epics-base/pull/375#issuecomment-1506323363)).

Oh I see and I am sorry I did not understand the context before. Thanks.

kiwichris:
@mdavidsaver I have posted a new version of the patch. I have tested it with the MVME2700 BSP and the legacy stack.

mrippa (approved):
I have reviewed these changes. I don't know what needs to change to fix the build failure above.

kiwichris:
> I have reviewed these changes. I don't know what needs to change to fix the build failure above.

Thanks @mrippa. I saw the failure was in the x86 build and I assumed it was something unrelated. I have reviewed builds logs and I cannot see anything. Also the patch in question is in the RTEMS configuration and `Init` area and should be only visible in RTEMS builds?

kiwichris:
I have updated the pull request after testing with `libbsd` and the `xilinx_zynqmp_lp64_zu3eg` BSP.

thelunatic:
I added a suggested change to this PR in @kiwichris' original PR branch: https://github.com/kiwichris/epics-base/pull/1/commits/acde5b647654d13fa7dea5cf42a1120cf5760a1d

Except this change, the rest of the PR looks great and I have tested it with all EPICS tests with only one failure (`ipAddrToAsciiTest`).

kiwichris:
I have force pushed new changes to my `rtems-legacy-net-support` branch. The changes build with libbsd and the legacy stack.

configure/os/CONFIG.Common.RTEMS
```diff
@@ -102,15 +102,38 @@ OS_CLASS = RTEMS
...
+# Set the networking flags
+OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_NETWORKING))
 
 ifeq ($(RTEMS_HAS_POSIX_API),yes)
 POSIX_CPPFLAGS = -D_GNU_SOURCE -D_DEFAULT_SOURCE
 endif
 
-OP_SYS_LDLIBS_posix_NET_yes = -ltftpfs -lnfs -lz -ltelnetd
-OP_SYS_LDLIBS_posix_NET_no = -ltftpfs -lbsd -lz
+OP_SYS_LDLIBS_posix_NET_yes = -ltftpfs -lz -ltelnetd -lntp
```
thelunatic: `libntp` is provided by `rtems-net-services` repo. Does this change make `rtems-net-services` a requirement to build with any networking stack?

configure/os/CONFIG.Common.RTEMS-mvme2700
```diff
@@ -1,7 +1,11 @@
 #
 # Author: Matt Rippa
 #
+ifeq ($(shell test $(RTEMS_VERSION) -gt 5; echo $$?),0)
```
thelunatic: Is this expecting an environment variable `RTEMS_VERSION`? Currently `RTEMS_VERSION` is set by the user from `$(CONFIG)/os/CONFIG_SITE.Common.RTEMS`, which is included by `$(CONFIG)/os/CONFIG.Common.RTEMS`.

To use the `RTEMS_VERSION` from the config files instead of an environment variable, can we put this check after the config include? Moreover, the shell `test` command will probably not be compatible with Windows.

I propose a small change to this commit here: https://github.com/kiwichris/epics-base/pull/1/commits/acde5b647654d13fa7dea5cf42a1120cf5760a1d

My proposed PR uses only make functions and puts the check at the bottom to enable setting the version from make config files instead of build environment. Can we use something like this?

thelunatic (commented):
Thanks for updating the patch. I have made two comments. The rest of the patch looks great, I have tested the previous version of this patch on legacy build, I will test it again with legacy and libbsd. Thank you!

kiwichris:
> ❌ [Build EPICS Base 7 base-7.0-966 failed](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/47525983) (commit [b8b7c7f0a9](https://github.com/epics-base/epics-base/commit/b8b7c7f0a9) by @kiwichris)

The build is not found or I cannot access it. Is this a problem?

configure/os/CONFIG.Common.RTEMS-mvme2700
kiwichris: Your change is valid. I have not merged it into this patch set but I can if you are OK with this?

configure/os/CONFIG.Common.RTEMS
kiwichris: I seems to have pulled down the `libntp` library change from the NTP patch. I can rework the patch set if this is a problem? The RSB can build tools, kernel, networking and net services as a build set for suitable BSPs such as the MVME2700.

> Does this change make `rtems-net-services` a requirement to build with any networking stack?

Good question and I am not in a position to answer this. It depends on what the core team see as the base requirements for an RTEMS EPICS build. My input is to say "yes" and my reasons are...

EPICS 3 + RTEMS 4.10 provided NTP via the EPICS built in client so this functionality is expected. RTEMS 6 + EPICS 7 will support NTP via `ntpd` and not the EPICS NTP client. The `ntpd` solution provides functional backward compatibility at the NTP level with the improved performance and feature set `ntpd` has.

My analysis of the history is changes by Heinz for RTEMS on EPICS 7 commented out the EPICS NTP client support in RTEMS specific code as `libbsd` was the only networking option and it does not have the code the EPICS NTP client depends on. This situation has changed with the legacy networking stack now available however I would encourage the use of `ntpd` on RTEMS over the EPICS NTP support.

The EPICS NTP client for RTEMS depended on non-standard calls to the handle the NTP response that only exist in the legacy code. I do not see that code being moved anywhere else.

The topic of `ntpd` is not the focus of this PR and I suggest it be discussed in a new issue or a PR for that change. I have a branch in my repo that adds support. We are close to finishing our testing at Gemini and my recent changes look good.

So back to your original question. I feel building net services and having `libntpd` will happen if you want an NTP solution that matches other systems like Linux, Window, MacOS etc.

configure/os/CONFIG.Common.RTEMS-mvme2700
thelunatic: Thanks!
Yes, I am OK with it being merged into this patch set.

kiwichris:
Changes in this version:

1. Removed `libntpd` from the build. It was mistakenly added in the last version of this PR
2. Merged @thelunatic's patch to fix the MVME2700 BSP

hjunkes:
I'm a little surprised that legacy-stack is coming back into play here. I thought gemini requires NFSv4 and that can only be realised with the new libbsd-stack? (See also my comment in core-talk)
Regardless, I can't get the legacy stack installed.
With source-builder:
Checking for RTEMS_BSP_NETWORK_DRIVER_NAME    : no
**powerpc/beatnik does not provide RTEMS_BSP_NETWORK_DRIVER_NAME**
(complete log in /home/rtems/MVME6100_6_legacy_INST/rsb/rtems/build/rtems-net-legacy-ba35f73d2ddc82d2b9a7b728dc63552be2274968-x86_64-linux-gnu-1/rtems-net-legacy-ba35f73d2ddc82d2b9a7b728dc63552be2274968/bui
ld/config.log)
shell cmd failed: /bin/sh -ex  /home/rtems/MVME6100_6_legacy_INST/rsb/rtems/build/rtems-net-legacy-ba35f73d2ddc82d2b9a7b728dc63552be2274968-x86_64-linux-gnu-1/do-build
error: building rtems-net-legacy-ba35f73d2ddc82d2b9a7b728dc63552be2274968-x86_64-linux-gnu-1

In beatnik you can find 
./bsps/powerpc/beatnik/include/bsp/bsp_bsdnet_attach.h:#define RTEMS_BSP_NETWORK_DRIVER_NAME	BSP_auto_network_driver_name

thelunatic:
Hello @hjunkes, I have a patch for `rtems-net-legacy` to add the missing driver. I will post the patch soon.

thelunatic:
@hjunkes, this patch should fix the issue with mvme6100 build: https://lists.rtems.org/pipermail/devel/2023-July/075812.html

kiwichris:
> I'm a little surprised that legacy-stack is coming back into play here.

The legacy stack was removed from the main RTEMS repo (`rtems.git`) because being in that repo made it appear as the preferred networking solution for RTEMS when it is not. We always planned to make it available and to maintain it as an external package. RTEMS has a history of providing support for long life cycle projects and products. We encourage users to move to `libbsd` but we understand that is not always possible or it may take time.

Adding support for the legacy stack back into EPICS is easy and EPICS has long stable history with it.
 
> I thought gemini requires NFSv4 and that can only be realised with the new libbsd-stack? (See also my comment in core-talk)

Yes NFSv4 is only in `libbsd` and only on the `6-freebsd-12` branch. The support needs to be moved to the `master` branch at some point in time. The work to add NFSv4 took much longer because the `libbsd` FreeBSD sources had drifted away from how FreeBSD uses file descriptors and files over time and NFSv4 uses VFS and that is based around FreeBSD file descriptors and file objects. The `6-freebsd-12` branch brings the source closer to the FreeBSD base sources.

Gemini is still planing to move to NFSv4 however other projects have higher priorities, for example adding RTEMS `ntpd` support to EPICS. The MVME2700 has a `libbsd` driver for the Tulip chip however the driver, from FreeBSD, is for rev 3 or later versions of the chip while the MVME2700 has rev 1 silicon. The driver runs however I am not comfortable it is fully functional and so stable. Once Gemini has the other EPICS/RTEMS issues resolved we will be in a better position to test the driver to make sure it is fit for production.

mrippa:
We've come so far in the last year thanks to all the efforts from the EPICS and RTEMS communities. Gemini is grateful to all of your hardwork!

EPICS 7 + RTEMS 6 + ntpd is on our projects critical path. We descoped NFSv4 for this year to keep the risk in check. 
We're upgrading 22 telescope control systems and removing a dependency on our old Bancomm 635 IRIG-B time board. We aim to complete all testing at night and on-sky by Sep 2023 using the legacy network stack.
 
We're aiming to test NFSv4 + libbsd in a new project in FY2024.

kiwichris:
I have added an additional change requested by @anjohnson to ignore the RTEMS build flag `RTEMS_HAS_POSIX_API`. The commit message is:
```
    rtems: Always build RTEMS-posix and ignore the RTEMS_HAS_POSIX_API flag
    
    The RTEMS_HAS_POSIX_API is based on the POSIX build switch for RTEMS and
    that option only controls signals, the sporadic scheduler and maybe some
    other obscure POSIX features not often used. This change makes EPICS
    not depend on this RTEMS build flag.
```
This is the result of a discussion between @anjohnson and @joelsherrill on the RTEMS discord channel.

kiwichris:
I have added `pthread` naming of EPICS threads. The naming uses the non-portable POSX call `pthread_setname_np()` and that requires `_GNU_SOURCE` be set. The previous patch contains the change to add `_GNU_SOURCE` to the compiler's command line so enabling this call in `epics-base`. With this change the EPICS shell command `rt` produces:
```
10.1.2.173> rt pthread
ID       NAME                 SHED PRI STATE  MODES    EVENTS WAITINFO
------------------------------------------------------------------------------
0b010001  _main_              UPD  253 READY  P:T:nA   NONE   
0b010003  osdNTP_Runner       UPD   27 SYSEV  P:T:nA   NONE   
0b010004  taskwd              UPD  229 SEM    P:T:nA   NONE   
0b010005  osdNTP_Monitor      UPD  128 SEM    P:T:nA   NONE   
0b010006  NTPTimeSync         UPD   27 SEM    P:T:nA   NONE   
0b010007  errlog              UPD  229 SEM    P:T:nA   NONE   
0b010008  timerQueue          UPD   77 SEM    P:T:nA   NONE   
0b010009  cbLow               UPD  105 SEM    P:T:nA   NONE   
0b01000a  cbMedium            UPD   93 SEM    P:T:nA   NONE   
0b01000b  cbHigh              UPD   75 SEM    P:T:nA   NONE   
0b01000c  dbCaLink            UPD  128 SEM    P:T:nA   NONE   
0b01000d  scanOnce            UPD   85 SEM    P:T:nA   NONE   
0b01000e  scan-10             UPD  103 SEM    P:T:nA   NONE   
0b01000f  scan-5              UPD  100 SEM    P:T:nA   NONE   
0b010010  scan-2              UPD   98 SEM    P:T:nA   NONE   
0b010011  scan-1              UPD   95 SEM    P:T:nA   NONE   
0b010012  scan-0.5            UPD   93 SEM    P:T:nA   NONE   
0b010013  scan-0.2            UPD   90 SEM    P:T:nA   NONE   
0b010014  scan-0.1            UPD   88 SEM    P:T:nA   NONE   
0b010015  CAS-TCP             UPD  209 SYSEV  P:T:nA   NONE   
0b010016  CAS-UDP             UPD  214 SYSEV  P:T:nA   NONE   
0b010017  CAS-beacon          UPD  211 TIME:IS P:T:nA   NONE   Nanosleep
0b010018  CAC-event           UPD  125 SEM    P:T:nA   NONE   
```

configure/os/CONFIG.Common.RTEMS-mvme2700
```diff
@@ -23,3 +22,9 @@ define MUNCH_CMD
 endef
 
 include $(CONFIG)/os/CONFIG.Common.RTEMS
+
+ifeq ($(shell test $(RTEMS_VERSION) -gt 5; echo $$?),0)
+RTEMS_BSP = mvme2700
+else
+RTEMS_BSP = mvme2307
+endif
```
anjohnson: Vijay's PR to this branch included a most ingenious fix for this `ifeq` test that can handle `RTEMS_VERSION` values like `4.10.2` — an actual setting that we use here and other sites may also be using:
```makefile
ifeq ($(RTEMS_VERSION), $(word 1, $(sort 5 $(RTEMS_VERSION))))
```

His version should also work on Windows (where `test` isn't necessarily available) although it's unlikely (but not impossible) that anybody would actually want to cross-build EPICS for RTEMS from a Windows system.

configure/toolchain.c
```diff
@@ -34,6 +34,23 @@ OS_API = posix
 #  else
 OS_API = score
 #  endif
+#  if defined(RTEMS_NETWORKING)
+     // legacy stack circa RTEMS <= 5 and networking internal to RTEMS
```
anjohnson: I'm a little concerned about including comments like this in this particular file — see the remarks at the top. Maybe they're okay for all the RTEMS preprocessor versions, but developers on Windows might see them in the file and try to add them to other parts.

anjohnson (commented):
A couple of small change requests inline, the rest of this looks good to me now. We're going to want to add some Release Notes about all the modifications for RTEMS-6 but I won't make you write them here (unless you want to, in the documentation subdirectory).

configure/toolchain.c
kiwichris: Thanks for the review and highlighting the issue with comments on Windows. I have removed the comments.

configure/os/CONFIG.Common.RTEMS-mvme2700
kiwichris: Ah yes I moved my change and had not noticed. I have change the test to Vijay's solution.

kiwichris:
@anjohnson I have made the changes from your review. Thanks.

configure/toolchain.c
mdavidsaver: Comments in `toolchain.c` within `#ifdef __rtems__` should not be a problem.  As noted at the top of that file, GCC drops C comments from the output.  Please leave/return these comments as I think they add valuable information about the history of RTEMS networking which will, I can but hope, one day give someone confidence to remove it when this compatibility is no longer needed.

@anjohnson rather than removing comments from `toolchain.c`, wrap them with `#ifdef _COMMENT_` to hide them from MSVC.

modules/libcom/RTEMS/posix/rtems_config.c
```diff
@@ -148,14 +149,15 @@ extern void *POSIX_Init(void *argument);
...
 #include <rtems/shellconfig.h>
+
 #define RTEMS_BSD_CONFIG_BSP_CONFIG
 #define RTEMS_BSD_CONFIG_SERVICE_TELNETD
 #define RTEMS_BSD_CONFIG_TELNETD_STACK_SIZE (16 * 1024)
 #define RTEMS_BSD_CONFIG_SERVICE_FTPD
 #define RTEMS_BSD_CONFIG_FIREWALL_PF
-#else
+#else // __RTEMS_MAJOR__ > 4
 #include <rtems/shellconfig.h>
```
mdavidsaver, line 166: Does some additional header need to be included for shell commands coming from libbsd?

Any differences with RTEMS 5 vs. 6?

anjohnson:
Core Group: AI on ANJ: Poll tech-talk if anyone needs RTEMS-5 and can't easily upgrade to RTEMS-6 for the next EPICS release.
AI on @kiwichris Try to resolve the RTEMS-5 CI builds if not much work.

JJL772:
`configure/os/CONFIG.Common.RTEMS-pc686` and `configure/os/CONFIG.Common.RTEMS-pc386` also need changes for rtems 6+:

```diff
diff --git a/configure/os/CONFIG.Common.RTEMS-pc686 b/configure/os/CONFIG.Common.RTEMS-pc686
index 882a2f3cb..0bfbf1a55 100644
--- a/configure/os/CONFIG.Common.RTEMS-pc686
+++ b/configure/os/CONFIG.Common.RTEMS-pc686
@@ -32,7 +32,7 @@ OP_SYS_LDFLAGS += -Wl,-Ttext,0x100000
 
 
 # This check must appear after the above include
-ifneq ($(RTEMS_VERSION),5)
+ifneq ($(RTEMS_VERSION),$(filter $(RTEMS_VERSION),5 6 7))
   $(info *** This target is not compatible with the configured RTEMS version.)
   $(info *** Build the RTEMS-pc386 (-qemu) target for RTEMS 4.x)
   $(error Can't continue)

```

```diff
diff --git a/configure/os/CONFIG.Common.RTEMS-pc386 b/configure/os/CONFIG.Common.RTEMS-pc386
index 9fc518d6b..d7904af11 100644
--- a/configure/os/CONFIG.Common.RTEMS-pc386
+++ b/configure/os/CONFIG.Common.RTEMS-pc386
@@ -25,7 +25,7 @@ include $(CONFIG)/os/CONFIG.Common.RTEMS
 OP_SYS_LDFLAGS += -Wl,-Ttext,0x100000
 
 # This check must appear after the above include
-ifeq ($(RTEMS_VERSION),5)
+ifeq ($(RTEMS_VERSION),$(filter $(RTEMS_VERSION),5 6 7))
   $(info *** This target is not compatible with the configured RTEMS version.)
   $(info *** Build the RTEMS-pc686 (-qemu) target for RTEMS 5.x)
   $(error Can't continue)
```

krafel:
Containers with RTEMS6 (master branch) builds can be found here:
- with libbsd stack
  ```
  registry.gitlab.com/nsf-noirlab/gemini/rtsw/gem-ci/epics-base_unstable/fkraemer-issue58-libbsd:latest
  ```
- with legacy stack
  ```
  registry.gitlab.com/nsf-noirlab/gemini/rtsw/gem-ci/epics-base_unstable/fkraemer-issue58:latest
  ```
  
  There were built from the [`rtems-source-builder` commit bddb17c9d20f1c00c550b11b387dd915aa5c29de](https://git.rtems.org/rtems-source-builder/commit/?id=bddb17c9d20f1c00c550b11b387dd915aa5c29de) using @kiwichris  [`rtems-deployment` project at revision 9494f267cf8d77465fac78e2026ded25267ebec1](https://git.rtems.org/chrisj/rtems-deployment.git/commit/?id=9494f267cf8d77465fac78e2026ded25267ebec1)

kiwichris:
I have updated thsi PR with some minor formatting changes. What is the status of it?

kiwichris:
I have fixed the RTEMS 5 support and tested RTEMS 4.10, 5 (internal legacy) and 6 builds. The 5.1 p686 is suppose to fail.

mdavidsaver:
> ...  The 5.1 p686 is suppose to fail.

Please elaborate on why.  This target succeeds w/ current 7.0.

kiwichris:
> > ...  The 5.1 p686 is suppose to fail.
> 
> Please elaborate on why. This target succeeds w/ current 7.0.

Thanks, I missed the `ifneq` in the i686 when I fixed the way the versions are checked so we avoid needing to add every release. Fixed and pushed.

kiwichris:
All RTEMS CI builds pass. I have no idea what the Codacy stage is.

anjohnson:
The Codacy error says

> failed to evaluate #elif condition, division/modulo by zero

for the elif line in this snippet:
```C
       /* assume old GCC implies RTEMS < 5 with mis-configured BSP */
#      error rebuild BSP with --enable-network
#    elif __has_include(<machine/rtems-net-legacy.h>)
       /* legacy stack circa RTEMS > 5 */
       RTEMS_LEGACY_NETWORKING = yes
```

My guess that Codacy doesn't recognize/understand the `__has_include()` macro, which was only added in the C23 standard. I suspect there's more chance of it passing if the path were given in double-quotes, although [your use of angle-brackets there is correct](https://en.cppreference.com/w/c/preprocessor/include) for C23.

@ralphlange do you have any insights into that?

I personally don't think you need to fix this at all. We aren't going to hold up this PR for that failure now that all the other CI builds pass, but you're welcome to try by changing it to use `""` if you're curious.

ralphlange:
I agree and set the issue to "ignored" on Codacy.

gilesknap:
@anjohnson would it be possible to have a release made with this change please?

mrippa:
@gilesknap 
Since the RTEMS 6.1 release is now official, I will start testing in the Gemini Lab. I have 2700 hardware to start with.
This MR is specific to rtems-legacy-net-support. Building EPICS for used with libBSD will likely require some EPICS frontend changes TBD on another thread. 

https://ftp.rtems.org/pub/rtems/releases/6/6.1/

gilesknap:
Thanks @mrippa. At DLS we are using the legacy stack which is why I would like this release. I have previously tested https://github.com/kiwichris/epics-base/tree/rtems-legacy-net-support against RTEMS 6.1rc2 BSP with legacy networking on our mvme5500 boards.

I'm about to test the current 7.0 branch of epics-base with RTEMS 6.1 release in the same configuration.

hjunkes:
I miss ntp support in rtems-libbsd in the 6.1 release. In the develop-branch I had so far "config/net/ntp.bset -> rsb/rtems/config/net/ntp-4.2.6p5-1.cfg"

gilesknap:
UPDATE: RTEMS 6.1 plus legacy network + HEAD of EPICS branch 7.0 running on our mvme5500 appears to be working according to early testing. (ppc m4700)

hjunkes:
Did you use the beatnik bsp?
There is a discussion at RTEMS:
https://gitlab.rtems.org/rtems/rtos/rtems/-/issues/5216

hjunkes:
Hi Giles, I saw your comment about autosave in the RTEMS discord. I remember that we had caching problems with the ALTIVEC on the MVME6100. I have therefore introduced this on two boards:
CONFIG.Common.RTEMS-beatnik:ARCH_DEP_CFLAGS += -DRTEMS_HAS_ALTIVEC
CONFIG.Common.RTEMS-qoriq_e500:ARCH_DEP_CFLAGS += -DRTEMS_HAS_ALTIVEC

This is used in ./modules/libcom/src/calc/postfix.c to switch off the optimisation.

gilesknap:
@hjunkes, Thanks. I resolved this issue in a different way also to do with optimization flags. I found using -mstrict-align in the BSP build has fixed all the issues I've seen so far. See https://gitlab.rtems.org/rtems/rtos/rtems/-/issues/3767.
EDIT: I just realized that the discord you mention is a log of the above comment so you already knew that!

kiwichris:
> Hi Giles, I saw your comment about autosave in the RTEMS discord. I remember that we had caching problems with the ALTIVEC on the MVME6100. I have therefore introduced this on two boards: CONFIG.Common.RTEMS-beatnik:ARCH_DEP_CFLAGS += -DRTEMS_HAS_ALTIVEC CONFIG.Common.RTEMS-qoriq_e500:ARCH_DEP_CFLAGS += -DRTEMS_HAS_ALTIVEC
> 
> This is used in ./modules/libcom/src/calc/postfix.c to switch off the optimisation.

I brief check of the file does not show a reason for this being needed. The score vairant of EPICS used classic tasks and FP support was an option so a task with it off could appear to corrupt FP registers. POSIX always enables FP support.

mrippa:
I posted recently that [new changes for ntpd](https://github.com/epics-base/epics-base/pull/528) are successful on the mvme2700 target. This is based on [RTEMS6.2-rc1](https://ftp.rtems.org/pub/rtems/releases/6/rc/6.2-rc1/).

PR #377, opened 2023-04-19
Allow access to the TIME field

exzombie:
This is a followup to #126, which it was split from. Based on the discussion in that pull request, I implemented access to TIME by simply chopping away the ".TIME" part of the channel name, replacing it with a JSON fragment for the `ts` filter, and letting things happen as if the user had requested a timestamp as a `double`.

### DIFF
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -51,6 +51,7 @@ typedef struct parseContext {
 
 static void *dbChannelFreeList;
 static void *chFilterFreeList;
+static int tsFilterRegistered;
 
 void dbChannelExit(void)
 {
@@ -67,6 +68,8 @@ void dbChannelInit (void)
     freeListInitPvt(&dbChannelFreeList,  sizeof(dbChannel), 128);
     freeListInitPvt(&chFilterFreeList,  sizeof(chFilter), 64);
     db_init_event_freelists();
+
+    tsFilterRegistered = dbFindFilter("ts", 2) != NULL;
 }
 
 static void chf_value(parseContext *parser, parse_result *presult)
@@ -461,18 +464,12 @@ dbChannel * dbChannelCreate(const char *name)
        field and replace the field with a channel filter string. */
     {
         static char const tsfilter[] = ".{\"ts\":{\"num\":\"dbl\"}}";
-        static int exists_ts = -1;
 
         size_t namelen = strlen(name);
         size_t reclen = namelen - sizeof(".TIME") + 1;
         int has_time = namelen >= sizeof(".TIME") && !strcmp(name + reclen, ".TIME");
 
-        if (exists_ts == -1) {
-            /* Only look for the filter on first use. */
-            exists_ts = dbFindFilter("ts", 2) != NULL;
-        }
-
-        if (has_time && exists_ts) {
+        if (has_time && tsFilterRegistered) {
             cname = malloc(reclen + strlen(tsfilter) + 1);
             if (!cname)
                 goto finish;
@@ -800,10 +797,16 @@ void dbRegisterFilter(const char *name, const chFilterIf *fif, void *puser)
 
 const chFilterPlugin * dbFindFilter(const char *name, size_t len)
 {
-    GPHENTRY *pgph = gphFindParse(pdbbase->pgpHash, name, len,
-            &pdbbase->filterList);
+    GPHENTRY *pgph;
+
+    if (!name || !len || !*name || !pdbbase)
+        return NULL;
+
+    pgph = gphFindParse(pdbbase->pgpHash, name, len,
+                        &pdbbase->filterList);
 
     if (!pgph)
         return NULL;
+
     return (chFilterPlugin *) pgph->userPvt;
 }
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -464,7 +464,7 @@ dbChannel * dbChannelCreate(const char *name)
         static int exists_ts = -1;
 
         size_t namelen = strlen(name);
-        size_t reclen = namelen - 5; /* subtract strlen(".TIME"), MSVC won't optimize */
+        size_t reclen = namelen - sizeof(".TIME") + 1;
         int has_time = reclen > 0 && !strcmp(name + reclen, ".TIME");
 
         if (exists_ts == -1) {
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -450,12 +450,42 @@ dbChannel * dbChannelCreate(const char *name)
     const char *pname = name;
     DBENTRY dbEntry;
     dbChannel *chan = NULL;
-    char *cname;
+    char *cname = NULL;
     dbAddr *paddr;
     long status;
 
     if (!name || !*name || !pdbbase)
         return NULL;
+
+    /* If the timestamp filter is available, intercept requests for the TIME
+       field and replace the field with a channel filter string. */
+    {
+        static char const tsfilter[] = ".{\"ts\":{\"num\":\"dbl\"}}";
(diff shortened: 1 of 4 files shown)


modules/database/src/ioc/db/dbChannel.c
```diff
@@ -450,26 +450,53 @@ dbChannel * dbChannelCreate(const char *name)
...
+        static char const tsfilter[] = ".{\"ts\":{\"num\":\"dbl\"}}";
+        static int exists_ts = -1;
+
+        size_t namelen = strlen(name);
+        size_t reclen = namelen - 5; /* subtract strlen(".TIME"), MSVC won't optimize */
+        int has_time = reclen > 0 && !strcmp(name + reclen, ".TIME");
+
+        if (exists_ts == -1) {
+            /* Only look for the filter on first use. */
+            exists_ts = dbFindFilter("ts", 2) != NULL;
```
mdavidsaver: This lazy initialization should be done in `dbChannelInit()`, which runs as part of `iocInit()`.  Well after filters are registered.

modules/database/src/ioc/db/dbChannel.c
```diff
@@ -450,26 +450,53 @@ dbChannel * dbChannelCreate(const char *name)
...
 
+    /* If the timestamp filter is available, intercept requests for the TIME
+       field and replace the field with a channel filter string. */
+    {
+        static char const tsfilter[] = ".{\"ts\":{\"num\":\"dbl\"}}";
+        static int exists_ts = -1;
+
+        size_t namelen = strlen(name);
+        size_t reclen = namelen - 5; /* subtract strlen(".TIME"), MSVC won't optimize */
+        int has_time = reclen > 0 && !strcmp(name + reclen, ".TIME");
```
mdavidsaver: `reclen` is unsigned, so `reclen > 0` isn't going to work as desired for a really short name (a buffer overrun).

```c
int has_time = namelen > 5 && reclen > 0 && !strcmp(name + reclen, ".TIME");
```

mdavidsaver (commented):
wrt. `/* subtract strlen(".TIME"), MSVC won't optimize */` and  `strlen(tsfilter)`.  `sizeof(".TIME")` is a portable way to express this.  Though remember that this will include the trailing nil of the string constant.  (`sizeof(".TIME")-1==strlen(".TIME")`)

modules/database/src/ioc/db/dbChannel.c
exzombie: Good catch. I got too used to the compiler warning about these things, but there's nothing it would know to warn about here ...

modules/database/src/ioc/db/dbChannel.c
exzombie: Well, I'm not a fan of global variables that are only used in one place. But it's a matter of taste, I guess. Done.

exzombie:
Review comments have been addressed, and the only remaining failing CI tests are known to be flaky.

modules/database/src/ioc/db/dbChannel.c
mdavidsaver: Please continue to use `size_t` for `reclen` as well as `namelen`.  The `name` string can come from a remote client (CA or PVA), and imo. should be regarded as "untrusted".  I know it seems ridiculous to think about a PV name being >= 4GB long, but I don't like making such assumptions when a violation results in an out of bounds array access.  (how many "hacks" have been built on bugs like this?)

exzombie: No problem, `ssize_t` it is. Overflow that, sneaky attacker! Unless 2^31-1 is still too small and we really need that additional bit?

Sorry, couldn't help myself. I actually wrote a rant about C strings, but I decided not to post, you already know what I wanted to say. I'll just add that it you're worried about that, the name needs to be sanitized, and using `ssize_t` instead of `int` is cosmetics.

Or maybe not even that, because it doesn't work on Windows. Thank `$DEITY` for AppVeyor.

exzombie: Ugh. Revisiting this PR and re-reading my last comment, I'm sure it must have come across as more caustic than I intended. I apologize.

In any case, I believe this was solved in 1cf558f32a9432527fd2ab1fd829fb472ffab26e as the check is now performed on `namelen`, avoiding use of `reclen` unless it checks out.

BTW, do you prefer that I resolve conversations when I think the thread is done, or do you prefer that I keep them open?

mdavidsaver: > I'm sure it must have come across as more caustic than I intended. I apologize.

Don't worry about it.  I didn't experience an excessive burn ;)  In part because I understand that it is reasonable to see this sort of "secure" code nitpicking as paranoia.  It is.  I think that at this point in time, experience with C coding has shown that, if issues like this aren't being found and exploited, it is only because there was lower hanging fruit.  (so, about that .acf file...)

mdavidsaver: > BTW, do you prefer that I resolve conversations when I think the thread is done, or do you prefer that I keep them open?

Leave them open.  At some point the technical point will be resolved, then at some later point I will get tired of scrolling past it.

mdavidsaver: Core Meeting: Current logic seems safe, although the fact `reclen` can underflow looks odd.

@dirk-zimoch suggests looking at a use of `strstr()` to find `.TIME`, then checking that the result is at the end of the PV name.

exzombie: I wouldn't use the word "odd", I would call it "nasty" 😉. But to my eyes, all C string byte manipulations look nasty. To wit, here's what changes when using `strstr()`:

```diff
diff --git a/modules/database/src/ioc/db/dbChannel.c b/modules/database/src/ioc/db/dbChannel.c
index 7689c9420..a71548d28 100644
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -466,10 +466,11 @@ dbChannel * dbChannelCreate(const char *name)
         static char const tsfilter[] = ".{\"ts\":{\"num\":\"dbl\"}}";
 
         size_t namelen = strlen(name);
-        size_t reclen = namelen - sizeof(".TIME") + 1;
-        int has_time = namelen >= sizeof(".TIME") && !strcmp(name + reclen, ".TIME");
+        char *timef = strstr(name, ".TIME");
+        int has_time = timef && (namelen - (timef - name) == sizeof(".TIME") - 1);
 
         if (has_time && tsFilterRegistered) {
+            size_t reclen = timef - name;
             cname = malloc(reclen + strlen(tsfilter) + 1);
             if (!cname)
                 goto finish;
```

I'm not convinced this increases readability, while passing over the string twice. But if you find it more readable, just say and I'll add a commit.

exzombie:
I've been thinking about this PR as it seems to me it could be handled more efficiently. At this point, it's all about taste. In fact, it's been about taste since the beginning, my original proposal approached this in a completely different way. As such, it seems to me it's much less effort for you to just implement the changes in these 25-ish lines of code, rather than try and explain to me what you would like to see done.

I don't mind doing the work, but I just don't know what (and how!) you would like to see done. And I really don't mind maintainers pushing fixes to "my" branches; that's why I enabled that permission on this PR in the first place. So please, do go ahead and commit changes. Or, if you prefer that I do the changes, please provide clear direction, because, as an example, suggesting that I try `strstr()` doesn't help me decide whether you find it acceptable or not.

anjohnson:
Core Group: Since the `ts{}` filter changes haven't been released yet we'd like to pause adding this syntactic sugar until the filter has undergone some use and we get feedback about that.

anjohnson:
Core Group: We're not convinced that adding this syntactic sugar would be worth the additional confusion that it could lead to. Given than DB designers can specify the `{ts:{…}}` filter themselves this doesn't provide any extra functionality, and seeing that filter documents the value that it expects to fetch.

All readers: Please comment here if there is interest in this shortcut being implemented in the IOC.

PR #379, opened 2023-04-21
Discard search requests in CA client

### LINKED ISSUE #378: "Identical process variable names on multiple servers" with invalid address
This issue that manifest itself in camonitor (with any PV, no matter the server) as follows:

```
        Warning: "Identical process variable names on multiple servers"
        Context: "Channel: "SR-DI-DCCT-01:SIGNAL.NAME", Connecting to: cs03r-cs-gate-01.cs.diamond.ac.uk:5064, Ignored: 0.0.0.1:5"
        Source File: ../cac.cpp line 1308
```

The problem has been identified and it happens as a consequence of the following events:

1. Some IOC was mistakenly started with EPICS_CA_SERVER_PORT=5065
2.  That IOC contains CA links that will try and retry (forever) to send CA_PROTO_SEARCH request messages (using port 5065)
3. Meanwhile, in the same network, there is one or many CA clients to random PVs, the caRepeater will forward the CA_PROTO_SEARCH request message to every CA client (mistakenly thinking that it is a beacon message)
4.  The CA client mistakenly parses the CA_PROTO_SEARCH request thinking that is a CA_PROTO_SEARCH response (given that channel access doesn't have any flag to indicate if it's a request or a response)
5.  The CA client will check the last field of the header of CA_PROTO_SEARCH message which in both cases (reply and response) correspond to the Search ID... this means the message is most of the times discarded because of a non-matching search ID, but because the faulty IOC is sending CA_PROTO_SEARCH requests continuously (starting with low values of search ID), it eventually clashes with the Search ID that that client used.
6. The CA client will mistakenly parse the fields server port and IP which results in the message with invalid IP shown above

A pull request with a proposal to deal with this will be submitted shortly.

EmilioPeJu:
Search requests could accidentally reach a CA client if an IOC is configured with `EPICS_CA_SERVER_PORT=5065`.

This change filters those messages to prevent parsing a search request as a search reply and receiving several log with
"Identical process variable names on multiple servers" in which the address is invalid.

The filtering is done by checking the data count field (which in a search response, should be 0).

This fixes #378

### DIFF
--- a/modules/ca/src/client/udpiiu.cpp
+++ b/modules/ca/src/client/udpiiu.cpp
@@ -681,6 +681,13 @@ bool udpiiu :: searchRespAction (
     }
 
     /*
+     * CA requires this field to be 0 in the search response
+     */
+    if (msg.m_count != 0) {
+        return true;
+    }
+
+    /*
      * Starting with CA V4.1 the minor version number
      * is appended to the end of each UDP search reply.
      * This value is ignored by earlier clients.


anjohnson:
Core Group: Thanks for the analysis, we agree with it for RSRV and recent servers.

We would like to also check this on both the Java and PCAS servers before accepting it though, @shroffk and @ralphlange would you be willing and able to find time to do that? If so please check with @mdavidsaver for details of what to look for.

mdavidsaver:
The current CA protocol ["spec"](https://docs.epics-controls.org/en/latest/specs/ca_protocol.html#id2) does say that `count` in a search reply "Must be 0.".

In current RSRV I can convince myself that this is true:

https://github.com/epics-base/epics-base/blob/42604fc794b6f26b60db953a230b52c1118d3039/modules/database/src/ioc/rsrv/camessage.c#L2219-L2226

However, I have become cautious about assuming that what is true of RSRV now was always true, or whether this is/was true of other CA servers.  With the largest question being PCAS.

anjohnson:
Core Group: We don't see this as particularly necessary. It's a fix for a mis-configuration which probably doesn't happen very often, and we're trying to avoid making changes to the CA code that might break compatibility in subtle ways.

PR #381, opened 2023-04-25
add comment to cvtArg explaining default iocsharg behaviour

rerpha:


mdavidsaver:
Applied as 49ea46ee5e85a518fa3afd189bee06f43e1a3f24

PR #382, opened 2023-05-01
RTEMS: override default FD_SETSIZE

mdavidsaver:
Makes default `CONFIGURE_MAXIMUM_FILE_DESCRIPTORS` with RTEMS >=5 match the 4.x value while hopefully not breaking usage of `select()` in cases where it was not already broken.

User override of `FD_SETSIZE` is [documented](https://docs.rtems.org/branches/master/user/migration/v4_11-to-v5.html) as being supported provided provided that:

> ... The application must ensure that the custom FD_SETSIZE is defined before <sys/select.h> is included in all modules used by the application, for example via a global compiler command line define. This applies also to all third-party libraries used by the application. ...

I have __not tested__ this.

cf. #300


configure/os/CONFIG.Common.RTEMS
```diff
@@ -107,6 +107,7 @@ OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_HAS_NETWORKING))
 
 ifeq ($(RTEMS_HAS_POSIX_API),yes)
 POSIX_CPPFLAGS = -D_GNU_SOURCE -D_DEFAULT_SOURCE
+POSIX_CPPFLAGS += -DFD_SETSIZE=150
```
anjohnson, line 110: I'm not sure what we did on RTEMS-4, but 150 is not a multiple of 4. Doesn't `select()` map FDs to individual bits in the memory buffer? If so wouldn't it be better to make this a multiple of 8 or 32?

mdavidsaver: No need to use the past tense here...

https://github.com/epics-base/epics-base/blob/fe9995c0b50e0aacd43e7aaa7fe6fab1e866db2a/modules/libcom/RTEMS/score/rtems_config.c#L36

I agree that 150 seems arbitrary.  As were the previous values of 50 and 100.  08297f56b385a7ff1517dc854c67fb536b508418  46068e813dc7333810411f1b156bc89e4ffb83d7  deae06a694dd31c06cacca4e04b9f41100c4e548

More significant than bits in `fd_set`, `CONFIGURE_MAXIMUM_FILE_DESCRIPTORS` is sizing the `rtems_libio_iops[]` array.  I suppose we could try to tie this setting to `sizeof(rtems_libio_t)`, but I suspect it would not be worth the effort.

anjohnson:
Core Group: Question on how & where to test this, not clear if different RTEMS series can all work with a modified FD_SETSIZE and FD limit. Limit for `select()` may be different than just the max number of FDs. Probably needs a specific test program.

configure/os/CONFIG.Common.RTEMS
kiwichris: I have no used this option. If it works as documented it is good solution for applications that define the number of descriptors and can control the `FD_SETSIZE` for all `select` calls. For libraries built and used by EPICS that do not have access to the number of descriptors this does not work and the [now standard work](https://git.rtems.org/rtems/tree/cpukit/mghttpd/mongoose.c#n1525) around in the code is needed.

Note, in libbsd the [`select` kernel call](https://git.rtems.org/rtems-libbsd/tree/freebsd/sys/kern/sys_generic.c?h=6-freebsd-12#n998) locally allocates 2048 bits which should be more than enough.

anjohnson: Core Group:
@kiwichris This approach should work with newlib (implementation is messy).
ANJ: Please make the 150 a build variable that can be overridden in the BSP's CONFIG_SITE file.

kiwichris:
The work around RTEMS uses to internally handle the this socket FD issue can be seen with https://git.rtems.org/rtems/tree/cpukit/mghttpd/mongoose.c#n1537 and a different approach in `ntpd` here https://git.rtems.org/rtems-net-services/tree/bsd/freebsd/contrib/ntp/ntpd/ntp_io.c#n3694

kiwichris:
Is this PR valid since we fixed the issue on RTEMS 6 by increasing newlib to default to 256?

anjohnson:
Core Group: Not worth doing any more.

PR #383, opened 2023-05-03
dbRecordsOnceOnly allow append only with "*"

mdavidsaver:
Currently with

```
record(ai, "myrec") {}
```

`dbRecordsOnceOnly!=0` currently disallows appending fields with either form:

```
record("*", "myrec") {} # error
record(ai, "myrec") {}  # error
```

Change the meaning such that `dbRecordsOnceOnly!=0` always allows appending when explicitly intended (rtype `"*"`).

```
record("*", "myrec") {} # allowed
record(ai, "myrec") {}  # error
```

Also clearly label this parse error.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1110,14 +1110,10 @@ static void dbRecordHead(char *recordType, char *name, int visible)
     allocTemp(pdbentry);
 
     if (recordType[0] == '*' && recordType[1] == 0) {
-        if (dbRecordsOnceOnly)
-            epicsPrintf("Record-type \"*\" not valid with dbRecordsOnceOnly\n");
-        else {
-            status = dbFindRecord(pdbentry, name);
-            if (status == 0)
-                return; /* done */
-            epicsPrintf("Record \"%s\" not found\n", name);
-        }
+        status = dbFindRecord(pdbentry, name);
+        if (status == 0)
+            return; /* done */
+        epicsPrintf(ERL_ERROR ": Record \"%s\" not found\n", name);
         yyerror(NULL);
         duplicate = TRUE;
         return;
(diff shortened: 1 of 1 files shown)


ralphlange:
Are you sure the behavior you found was not the intended one?

anjohnson:
IIRC that strict `dbRecordsOnceOnly` functionality was requested by the synapps maintainers (or at least by someone in the APS BCDA group, maybe @timmmooney or Ron Sluiter) but I don't know if they are still using it (@keenanlang can you comment on that? A GitHub [search of epics-modules](https://github.com/search?q=org%3Aepics-modules+dbRecordsOnceOnly&type=commits) only shows the pcas commits which added it to Base).

The change seems like a reasonable behavior to me so I'd be okay with it, but we might want to poll a wider audience (i.e. ask tech-talk). If there are objections we could give the flag a third choice to support both the strict and relaxed meaning of "once".

Whichever we finally decide it will need Release Notes, and a change to the [database definition documentation](https://docs.epics-controls.org/en/latest/appdevguide/databaseDefinition.html?highlight=dbRecordsOnceOnly#definitions-8) (@ttkorhonen FYI).

timmmooney:
I didn't know about dbRecordsOnceOnly until just now.

Tim Mooney ***@***.***) (630)252-5417
Beamline Controls Group (www.aps.anl.gov)
Advanced Photon Source, Argonne National Lab

________________________________

mdavidsaver:
> Are you sure the behavior you found was not the intended one?

I'll let others speak to the history.  To my mind, the append syntax `"*"` is a clear expression of intent by the author of a .db file.  I was (briefly) surprised to find that the current code didn't work this way.

My use case was in loading to unrelated databases and worrying about accidental record name collision, then finding that one .db intentionally appended with `"*"`.

fyi. The record type append `"*"` was added by 3cf2d9057f78ec7693b77f41f285e1d8a21bfe52 while `dbRecordsOnceOnly` goes back to 2dc1e771741fd0579a309548c712bbfe3ff6ca92 .

> ... If there are objections we could give the flag a third choice to support both the strict and relaxed meaning of "once".  ...

I've kept a copy of the original patch I showed at the last codeathon, which did just this.

keenanlang:
> IIRC that strict `dbRecordsOnceOnly` functionality was requested by the synapps maintainers (or at least by someone in the APS BCDA group, maybe @timmmooney or Ron Sluiter) but I don't know if they are still using it (@keenanlang can you comment on that?

Yeah, nothing in synApps currently uses dbRecordsOnceOnly. I don't think anything that I have touched would have necessarily been looking for any specific behavior here and since it was added into epics almost two decades ago, I don't have any communications in-group that discuss a need for it. We would be fine with whatever changes seem reasonable to the EPICS community.

anjohnson:
Thanks Tim & Keenan. I'm okay with accepting @mdavidsaver's proposal then, should we pose the question to tech-talk or not bother? I'm fine with not doing that

mdavidsaver:
I went ahead and sent https://epics.anl.gov/tech-talk/2023/msg00698.php

Also, a draft release note:

### Allow explicit append with `dbRecordsOnceOnly!=0`

Previously setting `dbRecordsOnceOnly!=0` prevented any further changes to a record via a .db file.  eg.

```
record(ai, "myrec") {}
```

`dbRecordsOnceOnly!=0` previously disallowed appending fields with either form:

```
record("*", "myrec") {} # error
record(ai, "myrec") {}  # error
```

Beginning with this release, `dbRecordsOnceOnly!=0` allows appending when explicitly intended (when record type is `"*"`).

```
record("*", "myrec") {} # allowed
record(ai, "myrec") {}  # error
```

mdavidsaver:
Since @anjohnson and @ralphlange express agreement, and my tech-talk post drew no objections, I will consider this change accepted.

mdavidsaver:
Applied as 92cae86ff2ad7e747967808b65ec4834bf7b7d94

PR #385, opened 2023-05-11
fix: devTimestamp.c record_ai was without POSIX_TIME_AT_EPICS_EPOCH

jerzyjamroz:
This fix makes the stringin and ai record (DTYP, "Soft Timestamp") coherent otherwise there is a difference of 20 years between them.

### DIFF
--- a/modules/database/src/std/dev/devTimestamp.c
+++ b/modules/database/src/std/dev/devTimestamp.c
@@ -22,42 +22,42 @@
 #include "stringinRecord.h"
 #include "epicsExport.h"
 
-
 /* Extended device support to allow INP field changes */
 
-static long initAllow(int pass) {
-    if (pass == 0) devExtend(&devSoft_DSXT);
+static long initAllow(int pass)
+{
+    if (pass == 0)
+        devExtend(&devSoft_DSXT);
     return 0;
 }
-
 
 /* ai record */
 
 static long read_ai(aiRecord *prec)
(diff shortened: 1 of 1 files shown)


anjohnson:
The devTimestamp ai support has always provided its value as an offset from the EPICS epoch, which was defined to start 20 years after the POSIX epoch. The behavior you're seeing is thus by design and many EPICS sites are relying on it as written. The IOC code uses that epoch internally in the `struct epicsTimeStamp` and C++ `epicsTime` data types, and we provide APIs in epicsTime.h which use that epoch with them.

Thus this PR can't be merged as is, but there is a parameter that the device support can see which it doesn't currently use — the INP field accepts an `INST_IO` address string, so you could add code to examine that and allow a record to be configured to provide the POSIX time or the EPICS one. The default (an empty INP field) would need to stay the same for backwards-compatibility.

Before you implement that though, please take a quick look at the new functionality offered in the recently merged PR #126 which will appear in the next release and provides the ability to select the Unix epoch when fetching a timestamp as a number — the documentation for that is [here](https://github.com/epics-base/epics-base/pull/126/files#diff-ddd0b9179ca8d6f266382ac53afdcffdc8bdf5679749cd39651e4fde3f6d8c1b). If you do decide to add this parameter to the devTimestamp ai code please ensure that your epoch choice strings are the same as the choices for the timestamp filter.

PR #390, opened 2023-05-23
epicsThreadShow() zombies

mdavidsaver:
Flag when a thread has returned, but its tracking struct is still around.  eg. in need of joining.

This change effects all targets except vxWorks.

In response to https://github.com/epics-base/pvAccessCPP/issues/190

On Linux, output looks like:

```
epics> epicsThreadShowAll 
          _main_ 0x55e771cb0050   399465      0       0       OK
...
UDP-rx 224.0.0.1 0x55e771f1b420   399502     50       0       OK
          TCP-tx 0x7f6eb42d95a0   400851     20       0       OK ZOMBIE
          TCP-rx 0x7f6eb42d91b0   400850     20       0       OK ZOMBIE
```

### DIFF
--- a/modules/libcom/src/osi/os/Linux/osdThread.h
+++ b/modules/libcom/src/osi/os/Linux/osdThread.h
@@ -36,6 +36,7 @@ typedef struct epicsThreadOSD {
     int                isEpicsThread;
     int                isRealTimeScheduled;
     int                isOnThreadList;
+    int                isRunning;
     unsigned int       osiPriority;
     int                joinable;
     char               name[1];     /* actually larger */
--- a/modules/libcom/src/osi/os/Linux/osdThreadExtra.c
+++ b/modules/libcom/src/osi/os/Linux/osdThreadExtra.c
@@ -23,6 +23,7 @@
 #include <sys/types.h>
 #include <sys/prctl.h>
 
+#include "epicsAtomic.h"
 #include "epicsStdio.h"
 #include "ellLib.h"
 #include "epicsEvent.h"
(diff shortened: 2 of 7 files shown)


simon-ess:
This isn't working for me: when I run the test you provided the IOC crashes with
```
pthread_create error Resource temporarily unavailable
epicsThread: Unexpected C++ exception "unable to create thread" with type "N11epicsThread20unableToCreateThreadE" in thread "TCP-acceptor" at Wed Jul 12 2023 13:34:12.364443527
terminate called after throwing an instance of 'epicsThread::unableToCreateThread'
  what():  unable to create thread
```

mdavidsaver:
Which test I provided?  (Sorry, I guess I have forgotten...)

simon-ess:
The one here: https://github.com/epics-base/pvAccessCPP/issues/190#issuecomment-1559757262

When I do the exact same, the IOC simply crashes with the error above.

mdavidsaver:
> This isn't working for me: when I run the test you provided the IOC crashes with

Was this on the first attempt after starting softIocPVA?  With 100 clients?

> pthread_create error Resource temporarily unavailable

This is likely `EAGAIN`, which on Linux suggests that `RLIMIT_NPROC` was reached.  Which could conceivably happen after enough zombies have accumulated.  (61896 would be quite a horde...)

```
$ ulimit -a
...
max user processes                  (-u) 61896
```

simon-ess:
Perhaps I'm running into a different issue. When I run this now with 100 clients, there are no zombie threads; everything seems to exit properly. So I increased it to 500 monitors and I get this crash instead.

Edit: Ah, when I up it to around 300 clients I start seeing zombies but no crash.

mdavidsaver:
I think this change is ~low risk.  `epicsThreadOSD*` isn't being handled in any new places, so the sorts of bugs I can foresee would be incorrectly showing (or not) the `ZOMBIE` label.

mdavidsaver:
Applied as 45b3bce51534178a0baaf3912a17827caa1b000d

PR #391, opened 2023-05-25
clang-cl support for Windows builds

dirk-zimoch:
Allows to use clang-cl instead of MS cl to (cross) compile for Windows targets.

A new CI target using clang-cl is included.


ralphlange:
Interesting.
This is Microsoft providing clang for their VisualStudio ecosystem, correct? With Window-ish commandline args? Ha.

dirk-zimoch:
No, I think it is LLVM providing a cl compatible compiler (one that understands dllimport and stuff). Of course you need the MSVC header files and libraries to link with. But the point is that you can smoothly cross-compile on Linux (or Mac) and maybe to get the "nicer" error messages of clang.
@xiaoqiangwang can give you more details.

dirk-zimoch:
> clang-cl is a cl.exe compatible driver mode for clang
> It understands the environment, the flags, and the tools
> Integrates with Visual Studio

Hans Wennborg, Euro-LLVM 2014, https://llvm.org/devmtg/2014-04/PDFs/Talks/clang-cl.pdf

xiaoqiangwang:
I will not repeat the benefits of having clang-cl, but raise questions for discussion.
* The EPICS_HOST_ARCH suffix _-clang_. clang can emulate MinGW in the same way (with target x86_64-pc-windows-gnu). So the _-clang_ may not be sufficient.  _-clang-msvc_ and _-clang-mingw_ seem clumsy and will be more so when adding _-debug_ and _-static_ suffixes. Or on the other extreme, why not use just win32-x86/windows-x64,  since clang-cl is suppose to be drop-in replacement of msvc cl.
* The CMPLR_CLASS is set to _msvc_ instead of _clang_. The same thought as above.
* The changes done to pvDataCPP module are more than I like to. It also requires MinGW 9+, which breaks the current CI.

configure/os/CONFIG.Common.windows-x64-clang
```diff
@@ -0,0 +1,11 @@
+# CONFIG.Common.windows-x64-clang
+#
+# Definitions for windows-x64-clang target builds
+# Sites may override these definitions in CONFIG_SITE.Common.windows-x64-clang
+#-------------------------------------------------------
+
+#Include definitions common to win32-x86-clang builds
+include $(CONFIG)/os/CONFIG.Common.win32-x86-clang
```
anjohnson: Now that there's a `CONFIG.msvcCommon` file, does it still make sense for the windows-x64 builds to include the win32-x86 files instead of just using the common one? The only reason we haven't had something like your `CONFIG.msvcCommon` file before was the time and manual testing it would have taken back when we added the x64 architecture. I could see a time in the future when we might want to drop support for the win32-x86 target completely, so now might be a good time to decouple the two.

configure/CONFIG.msvcCommon
```diff
@@ -0,0 +1,280 @@
...
+
+OPT_WHOLE_PROGRAM = YES
+
+#-------------------------------------------------------
+
+WINLINK = link
+
+RCCMD = $(LLVM_PREFIX)rc$(LLVM_SUFFIX) -l 0x409 $(INCLUDES) -fo $@ $<
+
+ARCMD = $(LLVM_PREFIX)lib$(LLVM_SUFFIX) -nologo -verbose -out:$@ $(LIB_OPT_LDFLAGS) $(LIBRARY_LD_OBJS)
```
anjohnson: Could you rename `LLVM_PREFIX` and `LLVM_SUFFIX` to `MSVC_PREFIX` and `MSVC_SUFFIX` respectively, that decouples this file more from the -clang ones.

anjohnson (commented):
A couple of minor comments inline, but in general I'm happy to see the new `CONFIG.msvcCommon` file, I've wanted to add something like that for a while but it never became a priority for me. I would like to see comments on this PR from @FreddieAkeroyd but it looks reasonable to me.

I haven't looked at the changes in the ci-scripts or pvData submodules at all, those will need merging first and the .gitmodules file change undone before this PR could be merged (so I hope they don't require these changes to be made at the same time, that gets complicated).

configure/CONFIG.msvcCommon
xiaoqiangwang: Thanks @anjohnson for the review. I made the changes.

configure/os/CONFIG.Common.windows-x64-clang
anjohnson: Thanks. Could you also do a similar direct include of that in `CONFIG.windows-x64.windows-x64` which also pulls in the `CONFIG.win32-x86.win32-x86` file that you mostly moved to `CONFIG.msvcCommon`. I believe that's the only other file which could use it though.

xiaoqiangwang: Done in c68d5a8. In the same direction to cleanup/consolidate:
* _CONFIG.win32-x86-mingw.Common_ currently includes _CONFIG.UnixCommon.Common_ probably for usage inside an MSYS2 shell? But later these tool settings are overridden by equivalent Perl scripts. 
* _CONFIG.win32-x86.Common_ and _CONFIG.win32-x86-mingw.Common_ are very similar. Maybe extract the common parts to a new file _CONFIG.windowsCommon.Common_ (as _CONFIG.UnixCommon.Common_ to *NIX hosts)?

FreddieAkeroyd:
Hi, thanks, looks interesting, I will take a look later this week when I have a bit of time.

FreddieAkeroyd:
Hi,

Sorry, only just now been able to start to take a proper look. Nice work @xiaoqiangwang  

I tested with the `clang-cl` available via the visual studio installer, VS2019 supplies llvm 12.0.0 that does not work (`llvm-rc` fails to build Com.rc with "Invalid language id: 0x409") however VS2022 supplies llvm 15.0.1 which resolves this issue, however the build failed in pvdata with link issues for testSerialisation.exe which I will check tomorrow if this is an issue my side as CI seems to work though I notice that is using version 16.0.3. Overall changes and moving to the `CONFIG.msvcCommon` look good, just a few thoughts it raises: 

Compiling prints a lot of `lld-link: warning: ignoring unknown argument '-release'` however I am not sure we really need this option on either compiler as we are not writing windows drivers https://learn.microsoft.com/en-us/cpp/build/reference/release-set-the-checksum?view=msvc-170

A `windows-x64` build will no longer pull in any local `CONFIG_SITE.win32-x86.win32-x86` but this can be documented in the release notes in case anybody was relying on it.

Regarding `windows-x64` v `windows-x64-clang` v `windows-x64-clang-msvc` this may come down to what sort of drop in replacement and compatibility we can/will/want to expect. If somebody has a `windows-x64` tree built with `cl` for example and swaps to use `clang-cl` to continue within the same built tree, should this work? It would seem that object file compatibility is the aim https://clang.llvm.org/docs/MSVCCompatibility.html but it isn't clear if that would (ever) work if any of object files were created with /GL (or only work with MSVC linker but not lld-link). The /GL and /LTCG options appear not to be supported with clang-cl and its EPICS CONFIG forces `OPT_WHOLE_PROGRAM = NO`. The /GL option is a bit picky on compatibility even with MSVC compilers https://learn.microsoft.com/en-us/cpp/build/reference/gl-whole-program-optimization?view=msvc-170
and restricts binary compatibility of static libraries and object files across MSVC versions 
https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170#restrictions (for which reasons I default OPT_WHOLE_PROGRAM=NO on all our windows builds so build trees are more portable). If such object file mixing creates obvious link errors then we may wish to go with `windows-x64` for both `cl` and `clang-cl` but `windows-x64-clang` may be clearer if more cumbersome. Not a strong opinion, what do others feel about compatibility?

xiaoqiangwang:
My local build is with LLVM 9.0.0 + VS 2017.

> I tested with the `clang-cl` available via the visual studio installer, VS2019 supplies llvm 12.0.0 that does not work (`llvm-rc` fails to build Com.rc with "Invalid language id: 0x409") however VS2022 supplies llvm 15.0.1 which resolves this issue

I back traced LLVM releases and found that llvm-rc works starting with LLVM ~~9.0.0~~ 13.0.0. I used the installer from https://releases.llvm.org/download.html.

> however the build failed in pvdata with link issues for testSerialisation.exe which I will check tomorrow if this is an issue my side as CI seems to work though I notice that is using version 16.0.3. 

The previous workaround to pvData module only works with LLVM 16 and causes other problems, https://github.com/epics-base/pvDataCPP/pull/87#issuecomment-1565549643. Now I pushed a new workaround to pvData module, using C++11 constexpr feature.

FreddieAkeroyd:
Thanks - my llvm 15.0.1 installed by VS2022 now all works fine and builds everything.

I still have issues with the VS2019 supplied llvm 12.0.0, with the compile line `llvm-rc -l 0x409` i get `Invalid language id: 0x409`  if i change to decimal  `llvm-rc -l 1033` I then get a file parse error `llvm-rc: Error parsing file: expected '-', '~', integer or '(', got EPICS_VERSION`  which feels like it has failed to preprocess properly. It is the same source tree that compiles fine with llvm 15.0.1 so i'm not going to worry anymore.

xiaoqiangwang:
It seems the **-l** is not implemented until LLVM 11,  https://github.com/llvm/llvm-project/commit/f36f8b41beb9cc0353d5624934406d39b78753e0. It was simply ignore before and assumes en_US.

And this option accepts a hex number, without *0x* prefix, until LLVM 13, https://github.com/llvm/llvm-project/commit/37e964d8a623f6ae6a9e3e95bfc2d6503b9586eb.

Interestingly Microsoft [RC command manual](https://learn.microsoft.com/en-us/windows/win32/menurc/using-rc-the-rc-command-line-#options) does not include _0x_ prefix

> /l langid
Specifies the default language for compilation. For example, -l409 is equivalent to including the following statement at the top of the resource script file: LANGUAGE LANG_ENGLISH,SUBLANG_ENGLISH_US

If the one without _0x_ prefix is more standard, shall we change this line?
```
RCCMD = $(MSVC_PREFIX)rc$(MSVC_SUFFIX) -l 0x409 $(INCLUDES) -fo $@ $<
```

FreddieAkeroyd:
`-l409` works for me on llvm 12.0.0 so I agree with your idea to change to this format as it is more widely accepted

FreddieAkeroyd:
Spoke too soon, it didn't actually rebuild properly so I though it had worked. So i still think `-l409` is worth doing, but not the whole solution.  Error is the same parse error as above `llvm-rc: Error parsing file: expected '-', '~', integer or '(', got EPICS_VERSION` If I remove all `#define` related constants including `VOS__WINDOWS32` etc. then it will compile, so it seems to be some preprocessing issue. Annoyingly if it fails to compile properly it still creates a `.res` file, so a subsequent `make` does not try again unless you remember to remove this first...

FreddieAkeroyd:
Could we use `.DELETE_ON_ERROR` in our Makefiles?

xiaoqiangwang:
You are right, I was tricked by Makefile as well. Only in LLVM 13 https://github.com/llvm/llvm-project/commit/64bc44f5ddfb6da4b6a8b51ea9a03f8772b3ae95, llvm-rc added preprocessing.

FreddieAkeroyd:
We could change the rc build to be two-step and do the C preprocessing as a separate build step ourselves, do you feel it would be worth the effort compared to just specifting a minimum LLVM version of 13? We would only be inconveniencing people wishing to use the default LLVM in VS2019, but I suspect this is a very small group.

xiaoqiangwang:
For me it is fine to make LLVM 13+ as a requirement.

For native Windows build, one can download from https://releases.llvm.org. For cross-compiling from Linux, debian/ubuntu can choose from https://apt.llvm.org and centos/rhel can use from EPEL/AppStream.

PR #392, opened 2023-05-26
Fix issues detected by cppcheck / sonarqube

ralphlange:
A set of (non-functional) changes that fix a few issues found by cppcheck / sonar. Intended to make EPICS Base pass the ITER quality gates again.

- zero-initialization of local variables and structure members
- missing `free()` of allocated memory when returning an error
- memory leaks when `realloc()` returns an error

### DIFF
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -575,18 +575,10 @@ long dbChannelOpen(dbChannel *chan)
     }
 
     /* Set up type probe */
-    probe.type = dbfl_type_val;
-    probe.ctx = dbfl_context_read;
-    probe.mask = 0;
+    memset(&probe, 0, sizeof(probe));
     probe.field_type  = dbChannelExportType(chan);
     probe.no_elements = dbChannelElements(chan);
     probe.field_size  = dbChannelFieldSize(chan);
-    probe.utag = 0;
-    probe.sevr = NO_ALARM;
-    probe.stat = NO_ALARM;
-    probe.time.secPastEpoch = 0;
-    probe.time.nsec = 0;
-    probe.dtor = NULL;
 
     p = probe;
 
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -577,13 +577,16 @@ long dbChannelOpen(dbChannel *chan)
     /* Set up type probe */
     probe.type = dbfl_type_val;
     probe.ctx = dbfl_context_read;
+    probe.mask = 0;
     probe.field_type  = dbChannelExportType(chan);
     probe.no_elements = dbChannelElements(chan);
     probe.field_size  = dbChannelFieldSize(chan);
+    probe.utag = 0;
     probe.sevr = NO_ALARM;
     probe.stat = NO_ALARM;
     probe.time.secPastEpoch = 0;
     probe.time.nsec = 0;
+    probe.dtor = NULL;
 
     p = probe;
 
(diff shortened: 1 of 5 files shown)


modules/database/src/ioc/db/dbChannel.c
```diff
@@ -577,13 +577,16 @@ long dbChannelOpen(dbChannel *chan)
     /* Set up type probe */
     probe.type = dbfl_type_val;
     probe.ctx = dbfl_context_read;
+    probe.mask = 0;
```
mdavidsaver: Maybe more future proof to start with `memset(&probe, 0, sizeof(probe))`.

modules/libcom/src/log/iocLogServer.c
```diff
@@ -174,6 +176,7 @@ int main(void)
         char sockErrBuf[64];
         epicsSocketConvertErrnoToString ( sockErrBuf, sizeof ( sockErrBuf ) );
         fprintf(stderr, "iocLogServer: ioctl FIONBIO err %s\n", sockErrBuf);
+        free(pserver);
```
mdavidsaver, line 179: There are more return statements below which seem like they should also `free(pserver);`.

modules/libcom/src/osi/os/RTEMS-posix/osdMessageQueue.c
```diff
@@ -47,6 +47,7 @@ epicsMessageQueueCreate(unsigned int capacity, unsigned int maximumMessageSize)
     id->id = mq_open(id->name, O_RDWR | O_CREAT | O_EXCL, 0644, &the_attr);
     if (id->id <0) {
         fprintf (stderr, "Can't create message queue: %s\n", strerror (errno));
+        free(id);
```
mdavidsaver, line 54: As an example of the limits of static analysis...  The above `calloc()` which sets `id` does not have a `NULL` check.

modules/libcom/src/yacc/output.c
```diff
@@ -600,10 +601,16 @@ pack_vector(int vector)
 
                 newmax = maxtable;
                 do { newmax += 200; } while (newmax <= loc);
-                table = (short *) REALLOC(table, newmax*sizeof(short));
-                if (table == 0) no_space();
```
mdavidsaver: Are there enough instances of this pattern to justify a helper function?  Maybe a `static inline` in `defs.h` where `REALLOC()` is defined?

anjohnson: These `REALLOC()` calls actually look to be fine even in the face of OOM errors. The `no_space()` routine is marked `NORETURN` and calls `exit()` via `done()`. Is there no way to teach these analysis tools about `__attribute__((noreturn))`, maybe by adjusting the `#ifdef __GNUC__` around the `NORETURN` definitions?

mdavidsaver: Good point.  So the `libCom/yacc: handle realloc() failures correctly` commit is based on a false positive.

ralphlange: I see the point.

I would still insist that this use of `realloc()` is wrong and loses memory.
The fact that - in that case - the program exits makes it irrelevant, but not right. So: yes, well-justified 'won't fix', but definitely not a false positive.

ralphlange: I will reset this commit with the next push and mark the issues as 'won't fix' locally.
(Made it a separate commit and the last one for a reason...)

anjohnson:
Core Group: Approved.

mdavidsaver:
Applied as of 6636b4b9e7086643ef188ec0db4462203834036e

PR #393, opened 2023-06-09
FreeBSD: Update to use clang

kiwichris:
FreeBSD 13 does not ship `gcc` by default these days.

The resulting build has not been tested. I do not know how I run the tests. I am using it to do RTEMS cross builds.

### DIFF
--- a/configure/os/CONFIG.freebsd-x86_64.freebsd-x86_64
+++ b/configure/os/CONFIG.freebsd-x86_64.freebsd-x86_64
@@ -2,13 +2,23 @@
 # Definitions for freebsd-x86_64 host - freebsd-x86_64 target builds
 # Sites may override these definitions in CONFIG_SITE.freebsd-x86_64.freebsd-x86_64
 #-------------------------------------------------------
-GNU_DIR=/usr/local
+GNU_DIR=/usr
 
 # Include common gnu compiler definitions
 include $(CONFIG)/CONFIG.gnuCommon
+
+GNU_BIN = $(GNU_DIR)/bin
+GNU_LIB = $(GNU_DIR)/lib
+
+CMPLR_CLASS = clang
+
+CC = $(GNU_BIN)/$(CMPLR_PREFIX)cc$(CMPLR_SUFFIX)
+CCC = $(GNU_BIN)/$(CMPLR_PREFIX)c++$(CMPLR_SUFFIX)
+CPP = $(CC) -x c -E
+
+GNU_LDLIBS_YES =
 
 STATIC_LDFLAGS_YES= -Wl,-Bstatic
 STATIC_LDFLAGS_NO=
 STATIC_LDLIBS_YES= -Wl,-Bdynamic
 STATIC_LDLIBS_NO=
-
--- a/modules/libcom/src/osi/os/default/epicsMMIODef.h
+++ b/modules/libcom/src/osi/os/default/epicsMMIODef.h
@@ -139,8 +139,12 @@ bswap32(epicsUInt32 value)
 /* hton* is optimized or a builtin for most compilers
  * so use it if possible
  */
+#ifndef bswap16
 #define bswap16(v) htons(v)
+#endif
+#ifndef bswap32
 #define bswap32(v) htonl(v)
+#endif
 
 #  define be_ioread16(A)    bswap16(nat_ioread16(A))
 #  define be_ioread32(A)    bswap32(nat_ioread32(A))
--- a/src/tools/EpicsHostArch.pl
+++ b/src/tools/EpicsHostArch.pl
@@ -43,6 +43,8 @@ sub HostArch {
         return "cygwin-x86"     if m/^i[3-6]86-cygwin/;
         return 'solaris-sparc'  if m/^sun4-solaris/;
         return 'solaris-x86'    if m/^i86pc-solaris/;
+        return 'freebsd-x86_64' if m/^x86_64-freebsd/;
+        return 'freebsd-x86_64' if m/^amd64-freebsd/;
 
         my ($kernel, $hostname, $release, $version, $cpu) = uname;
         if (m/^darwin/) {


anjohnson:
Core Group: Why did you introduce a new CONFIG.clangCommon, does the existing CONFIG.gnuCommon not work here for some reason?

kiwichris:
> Core Group: Why did you introduce a new CONFIG.clangCommon, does the existing CONFIG.gnuCommon not work here for some reason?

FreeBSD does not have `gcc` or `g++` by default and the C compiler is either `cc` or `c++` or specifically `clang` and `clang++`.

The objective is to use the operating system's standard compiler. A FreeBSD release is a kernel and base system which means executables under `/usr/bin` such as `clang` are part of the release. Users can install packages or ports under `/usr/local` but they are on their own and that may include `gcc`. I do not install gcc as a package or port to make sure things I maintain or support depend only on the standard release.

anjohnson:
Hi Chris, we understand that FreeBSD now uses Clang and we completely agree with switching over to it by default on that target. The point of the Core Group's question was that our `configure/CONFIG.gnuCommon` file is already being used for other targets that have switched to Clang, e.g. `configure/os/CONFIG.darwinCommon.darwinCommon` only had to override a few of the variables that the gnuCommon file sets when macOS XCode switched, and Intel Linux users can switch by uncommenting 4 lines in the appropriate `CONFIG_SITE.linux-<cpu>.linux-<cpu>` file. We would rather avoid adding an almost-duplicated file that's only used for one target, although if it reduced duplication by being used for other Clang targets too that might be a different story. There are typo's in your value for `RANLIB` in that file BTW.

We're being careful here since we have no easy way to test the FreeBSD target, we're not aware that GitHub can run that OS at all (but we'd be happy for any help doing that).

Thanks,

- Andrew

kiwichris:
Hi Andrew,

Here is the output of FreeBSD:
```
$ uname -a
FreeBSD ruru 13.1-RELEASE-p2 FreeBSD 13.1-RELEASE-p2 GENERIC amd64
$ type gcc
gcc: not found
$ type g++
g++: not found
```
And MacOS:
```
 % uname -a
Darwin weka 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:19 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6020 arm64
% type gcc
gcc is /usr/bin/gcc
% type g++
g++ is /usr/bin/g++
```
Looking a little deeper on MacOS:
```
% gcc -v
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
Apple has provided a `gcc` wrapper for `clang`:
```
% clang -v
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
There is no `gcc` or `g++` by default on FreeBSD and FreeBSD has not provided a wrapper so you know what you are using. Apple is confusing things because users may think `gcc` or `g++` is being used but it is not while those that do know `clang` is the compile are dependent on wrapper always being available. My preference is transparency. 

In relation to the structuring of this file, yes most is shared and a common file both could include would help but is that a good idea? The other option is conditional logic in the GNU file however I prefer the names of files and the contents to match.

I selected this option because:

1. If a change is needed it is easier to edit one file than a common file and then to find the include places and update
3. Not over investing time on something I am new to in the hope reviews and guidance would come and help. The PR was quick and easy and a possible solution at the same time

You could step back a little and consider what the core team sees as important in relation to bug reports or known performance on an operating system. For example on MacOS is a bug report related to the Xcode and Apple released `clang` used via `gcc` or a Homebrew or Macports install with real `gcc` of some version? Should the default for EPICS base be `cc` and `c++`? This works on MacOS and FreeBSD and ensures the "standard" compiler is being used. In the RTEMS project we are very careful about the compiler being used because you are building a kernel so my view may be outside the accepted norms.

Thanks for the `RANLIB` review comment. If there is life in this PR I can update.

I know very ittle about github and how it works.

Chris

mdavidsaver:
> There is no `gcc` or `g++` by default on FreeBSD ...

Maybe worth quoting:

https://github.com/epics-base/epics-base/blob/42604fc794b6f26b60db953a230b52c1118d3039/configure/os/CONFIG_SITE.linux-x86_64.linux-x86_64#L13-L18

Which is currently all that is needed to switch the `linux-x86_64` target from using `gcc` to `clang`.

kiwichris:
I am not sure I follow this suggestion well enough...

Do you suggest adding a `configure/os/CONFIG_SITE.freebsd-x86_64.freebsd-x86_64` with those defines?

mdavidsaver:
> In relation to the structuring of this file, yes most is shared and a common file both could include would help but is that a good idea?

To my mind, a clear an emphatic __yes__.  imo. `configure/` already contains too much almost copy+pasted boilerplate which must be manually maintained.

Also, consider the recent #391, adding support for `clang-cl` which is clang, but with an entirely different set of CLI arguments.  Mimicking `cl.exe` instead of `gcc`.  So `CONFIG.clangCommon` would not in fact be common to both clang frontends.  As I see it, the developers of clang have made a conscious decision not to strike out on their own wrt. conventions (CLI arguments, pre-defined macros, etc.).  Our shared `configure/CONFIG.gnuCommon` seems entirely in keeping with this.

mdavidsaver:
> Do you suggest adding a `configure/os/CONFIG_SITE.freebsd-x86_64.freebsd-x86_64` with those defines?

Maybe `configure/os/CONFIG.Common.freebsdCommon` would be better?  @anjohnson is likely have an opinion on this.

kiwichris:
Thanks. I am happy to handle this as instructed.

In relation to `epics-base/configure/os/CONFIG_SITE.linux-x86_64.linux-x86_64` there is a case that could argued this should be:
```
GNU         = YES 
CMPLR_CLASS = gnu 
CC          = gcc 
CCC         = g++ 
```
 And the default be `cc` and `c++`. 😄

kiwichris:
I have updated this branch to pick up recent changes on `7.0`. The change uses the GNU compiler settings.

It also lets the `pl` tools run.

kiwichris:
The patch is not complete because `CPP` is not defined. 

If I add a `CPP` the `toolchain.c` file is correctly processed for `TOOLCHAIN.freebsd-x86_64.freebsd-x86_64`. If a `CROSS_COMPILER_TARGET_ARCHS` is set to `RTEMS-xilinx_zynqmp_lp64_zu3eg` the `CPP` used to parse `toolchain.c` to create `TOOLCHAIN.freebsd-x86_64.RTEMS-xilinx_zynqmp_lp64_zu3eg` is `/usr/bin/gcc` and not the RTEMS compiler.

I am not sure the reason for this as the logic for RTEMS looks fine.

kiwichris:
> I am not sure the reason for this as the logic for RTEMS looks fine.

There is no BSP support in `configure/os` so `toolchain.c` is preprocessed with the default which is `gcc` and I do not have `gcc` so I see an error. I wonder what the error is on a system with `gcc`?

kiwichris:
I can build the host EPICS and the `RTEMS-xilinx_zynq_a9_qemu` BSP EPICS with #375 rebased with this PR on FreeBSD. I found this PR alone failed to build the RTEMS EPICS with linker errors.

mdavidsaver:
> The patch is not complete because `CPP` is not defined.

Since `CONFIG.gnuCommon` is being included, I would expect that definition to be used.

https://github.com/epics-base/epics-base/blob/524f81b8bd147bb714c9ea7b7462b8912a134246/configure/CONFIG.gnuCommon#L23

Can you try running: `make -C configure/O.freebsd-x86_64 PRINT.CPP`?

mdavidsaver:
`make -C configure/O.freebsd-x86_64 show-makefiles` may also be helpful to understand which makefiles are being included.

If this isn't enough, you may unfortunately have to resort to `remake` to understand where the definition(s) are coming from.  :(

kiwichris:
FYI the current patch has been updated and is working. The error I was seeing from due to no RTEMS BSP file for the specified BSP. With no RTEMS BSP it seems the RTEMS compiler overrides do not happen and so `gcc` is left as the processor.

> Can you try running: `make -C configure/O.freebsd-x86_64 PRINT.CPP`?

```
ruru epics-base.git $ gmake -C configure/O.freebsd-x86_64 PRINT.CPP
gmake: Entering directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
gmake -f ../Makefile TOP=../.. T_A=freebsd-x86_64  PRINT.CPP
gmake[1]: Entering directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
../../src/tools/EpicsHostArch.pl: Architecture 'amd64-freebsd-thread-multi' not recognized
clang -x c -E  -D_POSIX_THREADS          -D_X86_64_ -DUNIX  -D_BSD_SOURCE -Dfreebsd  -D_REENTRANT  -I. -I../O.Common -I. -I. -I.. -I/opt/src/noirlab/gemini/epics-base/libbsd/include/compiler/clang -I/opt/src/noirlab/gemini/epics-base/libbsd/include/os/freebsd -I/opt/src/noirlab/gemini/epics-base/libbsd/include         ../toolchain.c > TOOLCHAIN..freebsd-x86_64
Installing CFG file /opt/src/noirlab/gemini/epics-base/libbsd/cfg/TOOLCHAIN..freebsd-x86_64
../../configure/CONFIG:67: ../../configure/os/CONFIG..Common: No such file or directory
gmake[1]: *** No rule to make target '../../configure/os/CONFIG..Common'.  Stop.
gmake[1]: Leaving directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
gmake: *** [Makefile:8: PRINT.CPP] Error 2
gmake: Leaving directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
```
The use of `clang` here is not what I expected.

kiwichris:
Sorry I was on a branch that was not rebased with this patch. Here are the results:
```
ruru epics-base.git $ gmake -C configure/O.freebsd-x86_64 PRINT.CPP
gmake: Entering directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
gmake -f ../Makefile TOP=../.. T_A=freebsd-x86_64  PRINT.CPP
gmake[1]: Entering directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
CPP = /usr/bin/cc -x c -E
gmake[1]: Leaving directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
gmake: Leaving directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
```
```
ruru epics-base.git $ gmake -C configure/O.freebsd-x86_64 show-makefiles
gmake: Entering directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
gmake -f ../Makefile TOP=../.. T_A=freebsd-x86_64  show-makefiles
gmake[1]: Entering directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'

Makefiles read:
    ../Makefile
    ../../configure/CONFIG
    ../../configure/RELEASE
    ../../configure/CONFIG_COMMON
    ../../configure/CONFIG_FILE_TYPE
    ../../configure/CONFIG_BASE
    ../../configure/CONFIG_SITE
    ../../configure/CONFIG_SITE.local
    ../../configure/CONFIG_BASE_VERSION
    ../../configure/os/CONFIG.freebsd-x86_64.Common
    ../../configure/os/CONFIG.UnixCommon.Common
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/TOOLCHAIN.freebsd-x86_64.freebsd-x86_64
    ../../configure/os/CONFIG.Common.freebsd-x86_64
    ../../configure/os/CONFIG.Common.freebsdCommon
    ../../configure/os/CONFIG.Common.UnixCommon
    ../../configure/os/CONFIG.freebsd-x86_64.freebsd-x86_64
    ../../configure/CONFIG.gnuCommon
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/CONFIG_CA_MODULE
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/CONFIG_CA_VERSION
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/CONFIG_DATABASE_MODULE
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/CONFIG_DATABASE_VERSION
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/CONFIG_LIBCOM_MODULE
    /opt/src/noirlab/gemini/epics-base/libbsd/cfg/CONFIG_LIBCOM_VERSION
    ../../configure/RULES
    ../../configure/RULES_BUILD
    ../../configure/CONFIG_ADDONS
    ../../configure/RULES_FILE_TYPE
    ../../configure/RULES.Db
    ../../configure/RULES_EXPAND
    ../../configure/RULES_COMMON
gmake[1]: Leaving directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
gmake: Leaving directory '/opt/work/chris/noirlab/epics-base.git/configure/O.freebsd-x86_64'
```

anjohnson:
So `gmake -C configure/O.freebsd-x86_64 PRINT.CPP` shows us the CPP setting for your host builds; for the RTEMS cross-targets you would need to replace the host-arch with that specific RTEMS target-arch, so please run something like `gmake -C configure/O.RTEMS-MVME2700 PRINT.CPP` for one of your cross-target.

BTW if you ever see this
```
Installing CFG file /opt/src/noirlab/gemini/epics-base/libbsd/cfg/TOOLCHAIN..freebsd-x86_64
```
where there's no host-arch given in the file-name it might help if you set `EPICS_HOST_ARCH` in your environment — that shouldn't really be necessary now, but apparently there are places such as the above where it still is. Earlier versions of EPICS did require the host architecture to be set in the environment, so this may be an area that we still need to fix but the environment variable will get you past that problem.

PR #394, opened 2023-06-09
rtems: Add AARCH64 support and xilinx_zynqmp_lp64_zu3eg BSP

kiwichris:
I have built this BSP on the TRENZ TE0802 board. This is a ZynqMP processor so the patch adds `aarch64` support for RTEMS to EPICS.

Most tests run and I see no error but some seem to need system level support such as NFS are failing.

The build is the latest as of today and is RTEMS 6 (master), libbsd (6-freebsd-12) and the [Legacy Stack](https://github.com/epics-base/epics-base/pull/375) pull request. I use FreeBSD so I also had the (FreeBSD)[https://github.com/epics-base/epics-base/pull/393] pull request included.

### DIFF
--- a/configure/os/CONFIG.Common.RTEMS-xilinx_zynqmp_lp64_zu3eg
+++ b/configure/os/CONFIG.Common.RTEMS-xilinx_zynqmp_lp64_zu3eg
@@ -1,0 +1,15 @@
+#
+# CONFIG.Common.RTEMS-xilinx_zynq_zedboard
+# Author: Chris Johns <chris@contemporary.software>
+#
+# All RTEMS targets use the same Makefile fragment
+#
+#EXE = .elf
+RTEMS_BSP = xilinx_zynqmp_lp64_zu3eg
+RTEMS_TARGET_CPU = aarch64
+GNU_TARGET = aarch64-rtems
+
+OP_SYS_LDLIBS += -Wl,--gc-sections
+ARCH_DEP_LDFLAGS = -L$(RTEMS_BASE)/$(GNU_TARGET)$(RTEMS_VERSION)/xilinx_zynqmp_lp64_zu3eg/lib/
+
+include $(CONFIG)/os/CONFIG.Common.RTEMS
--- a/modules/libcom/RTEMS/posix/rtems_config.c
+++ b/modules/libcom/RTEMS/posix/rtems_config.c
@@ -169,7 +169,8 @@ extern void *POSIX_Init(void *argument);
  * The new general time support makes including the RTC driver less important.
  */
 #if !defined(mpc604) && !defined(__mc68040__) && !defined(__mcf5200__) && \
-    !defined(mpc7455) && !defined(__arm__)  && !defined(__nios2__)
+    !defined(mpc7455) && !defined(__arm__)  && !defined(__aarch64__) && \
+    !defined(__nios2__)
     /* don't have RTC code */
 #define CONFIGURE_APPLICATION_NEEDS_RTC_DRIVER
 #endif
--- a/modules/libcom/src/osi/os/RTEMS/osdVME.h
+++ b/modules/libcom/src/osi/os/RTEMS/osdVME.h
@@ -11,12 +11,12 @@
 /*
  * OS-dependent VME support
  */
-#ifndef __i386__
-#ifndef __mc68000
-#ifndef __arm__
+/*
+ * ?? Is PowerPC the only EPICS arch to support VME?
+ */
+#if !defined(__i386__) && !defined(__mc68000) && \
+    !defined(__arm__) && !defined(__aarch64__)
 #include <bsp/VME.h>
-#endif
-#endif
 #endif
 
 #ifdef __cplusplus


anjohnson:
Core Group: We need to update our tool-chains for CI builds to support this target, which @mdavidsaver will probably have to do. We would also prefer to be able to move detection of e.g. VME from "it's a 68K or PowerPC so it has VME" to either detecting the bsp/VME.h header, or to adding flags to the target-specific CONFIG file. The header-file detection in gcc doesn't work with the RTEMS 4.9 versions of GCC though. We will probably have more discussions about these.

Which released versions of RTEMS provide this BSP? We haven't added any RTEMS-6 specific targets to Base yet since it hasn't been released and we haven't merged the changes to support it.

mdavidsaver:
> Which released versions of RTEMS provide this BSP?

From a quick look, this BSP has no yet appeared in a release.  cf. https://github.com/RTEMS/rtems/commit/d0eaf3ec1caf893d0bccae8a20c0445c3b35010b

mdavidsaver:
> ... We need to update our tool-chains for CI builds to support this target, which @mdavidsaver will probably have to do. ...

While I have it on my TODO list to publish another set of toolchain+kernel binaries for 4.9, 4.10, and 5.  It is an open question for me at what point it makes sense for me to start building/publishing toolchain+kernel+libbsd binaries for what will become RTEMS 6.  I don't follow RTEMS development enough to have a sense of the pace of change.  Would I be signing up to do weekly rebuilds to keep up with PRs?  monthly?

kiwichris:
> While I have it on my TODO list to publish another set of toolchain+kernel binaries for 4.9, 4.10, and 5. 

I was not aware EPICS only accepted changes based on releases of RTEMS. I am sorry this.

If EPICS supports RTEMS 4, 5 and then 6 there will be BSP support in EPICS that may not exist in each of these releases. This means EPICS does hold BSPs that are not available to a specific release a user is using. 😄 

The availability of BSPs not yet released does help our testing of new BSP in RTEMS. For example this is first BSP I have seen on EPICS RTEMS with 64bit support. I can also test SMP at some point.

> It is an open question for me at what point it makes sense for me to start building/publishing toolchain+kernel+libbsd binaries for what will become RTEMS 6. I don't follow RTEMS development enough to have a sense of the pace of change. Would I be signing up to do weekly rebuilds to keep up with PRs? monthly?

The RTEMS 6 tools updates have been GCC release branch fixes. RTEMS itself has been moving along but the changes have been either ESA qual related or new BSPs. Weekly rebuilds would be a nice from an RTEMS point of view but I am not sure it is needed. I am happy to ping you if we need the tools or packages updated?

Once RTEMS 6 is released these are the possible builds:
1. toolchain+kernel+libbsd+net-services
2. toolchain+kernel+net-legacy+net-services

Will EPICS CI support the legacy and libbsd stacks? RTEMS has maintained backwards compatibility with the older hardware where libbsd support is not at production quality or is not available.

RTEMS 6 is stable and in better shape compared to RTEMS 5. RTEMS 6 also has production quality SMP support. The BSP in this PR supports SMP so will EPICS CI build selected BSPs for SMP as well?

mdavidsaver:
> I was not aware EPICS only accepted changes based on releases of RTEMS. I am sorry this.

We are still solidifying these ideas.  OS development isn't something we have had to consider in the recent past, nor with other targets in the present.  eg. we are far behind the bleeding edge of Linux kernel development, and generally comfortable there.

> ... This means EPICS does hold BSPs that are not available to a specific release a user is using.

Quite so.  Our `configure/os/` tree is large, and grower faster than we can prune it.  So we have some reluctance to casually adding files.

> ... For example this is first BSP I have seen on EPICS RTEMS with 64bit support. I can also test SMP at some point.

This is part of the reason for insisting on CI testing for this target as it is a new CPU+OS combination for epics-base.  (though we do have some experience with linux/aarch64)

> ... Weekly rebuilds would be a nice from an RTEMS point of view but I am not sure it is needed. I am happy to ping you if we need the tools or packages updated?

For this regularity, I would want help.  (or someone to come up with an alternative process)  My process is fairly automated, but between github.com and RTEMS build process changes, that automation has needed non-trivial maintenance on every past iteration.

fyi.

https://github.com/mdavidsaver/rsb/blob/4.9/.github/workflows/build-4.9.yml

https://github.com/mdavidsaver/rsb/blob/4.10/.github/workflows/build-4.10.yml

https://github.com/mdavidsaver/rsb/blob/5/.github/workflows/build-5.yml

> Will EPICS CI support the legacy and libbsd stacks?

As long as we support RTEMS 4.9 and 4.10, I expect that this will be necessary. 

> RTEMS 6 is stable and in better shape compared to RTEMS 5.

I at least am open to talking about dropping support for 5.x as a "stepping stone" release.  However, I would find it hard to justify dropping support for all released versions of RTEMS, and recommending use of an active development branch in production.  Some may be ok with this "post-release" mindset, but I am not.

kiwichris:
> > I was not aware EPICS only accepted changes based on releases of RTEMS. I am sorry this.
> 
> We are still solidifying these ideas. OS development isn't something we have had to consider in the recent past, nor with other targets in the present. eg. we are far behind the bleeding edge of Linux kernel development, and generally comfortable there.

I understand. Rules are hard with so many variables in play. I am happy to work in any way that helps you.
 
> > ... This means EPICS does hold BSPs that are not available to a specific release a user is using.
> 
> Quite so. Our `configure/os/` tree is large, and grower faster than we can prune it. So we have some reluctance to casually adding files.

RTEMS carries BSPs until they do not compile and no one fixes them. We have a lot of BSPs the core dev can not run because we have no hardware.

> > ... For example this is first BSP I have seen on EPICS RTEMS with 64bit support. I can also test SMP at some point.
> 
> This is part of the reason for insisting on CI testing for this target as it is a new CPU+OS combination for epics-base. (though we do have some experience with linux/aarch64)

Sound reasonable.

> > ... Weekly rebuilds would be a nice from an RTEMS point of view but I am not sure it is needed. I am happy to ping you if we need the tools or packages updated?
> 
> For this regularity, I would want help. (or someone to come up with an alternative process) My process is fairly automated, but between github.com and RTEMS build process changes, that automation has needed non-trivial maintenance on every past iteration.
> 
> fyi.
> 
> https://github.com/mdavidsaver/rsb/blob/4.9/.github/workflows/build-4.9.yml
> 
> https://github.com/mdavidsaver/rsb/blob/4.10/.github/workflows/build-4.10.yml
> 
> https://github.com/mdavidsaver/rsb/blob/5/.github/workflows/build-5.yml
> 
> > Will EPICS CI support the legacy and libbsd stacks?
> 
> As long as we support RTEMS 4.9 and 4.10, I expect that this will be necessary.

I think 4.9 and 4.10 are stable because no one will be updating the releases anytime soon.

For 5 and later we have provided building support in the RSB to help with deployment. The RSB can build all the needed packages given suitable built sets. The build set for RTEMS 5 for the MVME2307 is:

```
%define with_rtems_bsp powerpc/mvme2307 powerpc/beatnik powerpc/mvme3100
%define with_rtems_legacy_network
%{rtems_version}/rtems-powerpc
%{rtems_version}/rtems-kernel
```

and for RTEMS 6 is it:

```
%define with_rtems_bsp powerpc/mvme2700
%{rtems_version}/rtems-kernel
%{rtems_version}/rtems-net-legacy
net/net-services
```

If you provide the RSB the options `--bset-tar` and `--no-install` plus the build set file a tar file is created for you with the single command. The RSB is a flexible tool but it has a lot of options and features and the type of deployment you are after needs extra support to wrap it. Gemini is working this way and we have had good success moving from RTEMS 5 to 6 with minimal fuss.

Gemini uses the [rtems-deployment](https://git.rtems.org/chrisj/rtems-deployment.git) repo. It  handles the build sets for different RTEMS versions placing them in a single location. The interface for using the repo is consistent across versions. Gemini has a script called [rtems-setup.sh](https://gitlab.com/nsf-noirlab/gemini/rtsw/epics-base/rtems/-/blob/unstable/2023q2/rtems-setup.sh#L56) that builds their RPMs. You can build tar files and not RPMs.

I am happy to work with you on getting this working so you only need the high level pieces and not all the detail. I know nothing about the github CI set up.

> > RTEMS 6 is stable and in better shape compared to RTEMS 5.
> 
> I at least am open to talking about dropping support for 5.x as a "stepping stone" release. However, I would find it hard to justify dropping support for all released versions of RTEMS, and recommending use of an active development branch in production. Some may be ok with this "post-release" mindset, but I am not.

I think we can avoid dropping any support and we maybe able to have RTEMS 5. It may not have `ntpd` support but 6 will.

joelsherrill:
I concur with everything @kiwichris has said. A weekly build at least 1 BSP for RTEMS 6 would at least catch major mismatches. But  doing one BSP in active use for each architecture used in the EPICS community might catch more. I think that really means only 3-4 BSPs for now.

In spite of being the development master, it is more stable than you would expect. We do not like to see things break. API changes do occur but that's not that common. 

And I sympathize with the impact of changing from autoconf/automake to waf. We had to change all the release scripts and process (thanks @kiwichris), my testing scripts had to be fixed, etc. We all have workflows and changes like build instructions are painful.

anjohnson:
@kiwichris This BSP supports SMP, and is nice with libdebugger.

Core Group: Fine with merging this if/when we can make use of the Gemini tool-chains to support this in our CI builds. Requires fix for #399 to allow SMP to work.

PR #396, opened 2023-06-26
feat: menuEpicsTypeSizes added and compatible to pRec->ftx.

jerzyjamroz:
I am considering the implementation of the function (menuEpicsTypeSizes) that returns the size of all the type possibilities of the record (I use it in another project).
Please, review it and let me know if it is already implemented or not and if it is useful in general. If yes, then what would be the best location for this implementation? I wrote it in this way to show the idea and I am considering refining it after your answer (or cancel).

### DIFF
--- a/modules/database/src/ioc/db/menu.h
+++ b/modules/database/src/ioc/db/menu.h
@@ -1,0 +1,52 @@
+/*************************************************************************\
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+ \*************************************************************************/
+
+/** @file menu.h
+ * @brief Adds db menu capabilities
+ * @see @ref menu
+ */
+
+#ifndef INCmenuH
+#define INCmenuH
+
+#include "menuFtype.h"
+#include "epicsTypes.h"
+
+/*Returns the variable size of the menuFtype*/
+epicsEnum16 menuEpicsTypeSizes(menuFtype type)
+{
+	switch (type)
+	{
+	case menuFtypeSTRING:
+		return sizeof(epicsString);
+	case menuFtypeCHAR:
+		return sizeof(epicsInt8);
+	case menuFtypeUCHAR:
+		return sizeof(epicsUInt8);
+	case menuFtypeSHORT:
+		return sizeof(epicsInt16);
+	case menuFtypeUSHORT:
+		return sizeof(epicsUInt16);
+	case menuFtypeLONG:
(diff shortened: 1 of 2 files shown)


anjohnson:
The function `dbValueSize()` declared in dbAccessDefs.h and found in the dbCore library already provides this, although you have to be very careful if you want it for a CA client program and not for some code that will live inside an IOC.

modules/database/src/ioc/db/menu.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#ifndef INCmenuH
+#define INCmenuH
+
+#include "menuFtype.h"
+#include "epicsTypes.h"
+
+/*Returns the variable size of the menuFtype*/
+epicsEnum16 menuEpicsTypeSizes(menuFtype type)
```
mdavidsaver, line 19: For future reference... 

Qualifying with `static inline` would be necessary to avoid duplicate symbol errors during linking.  Something which would likely become apparent were this header included/used from one or more object files.  (eg. like some unit test code, hint hint)

While unnecessary with `static inline`, the usual `#ifdef __cplusplus` ... `extern "C" {` would not be out of place in this header.  For that matter, a new header doesn't seem necessary to me.  This function could be added in `menuFtype.dbd.pod`.  See eg. `struct aidset` defined in `aiRecord.dbd.pod`.

Also, I think that `size_t` should be returned.  This is the type of `sizeof()`, and I doubt a narrowing cast would have any practical benefit.

mdavidsaver (commented):
fyi. https://github.com/epics-base/epics-base/blob/42604fc794b6f26b60db953a230b52c1118d3039/modules/database/src/ioc/db/dbAccess.c#L742

I agree with @anjohnson that `menuEpicsTypeSizes()` would be redundant to `dbValueSize()`.  Of course, so far unstated is the fact that `DBF_*` can be cast to/from `menuFtype`.  (with respect to `dbValueSize()` at least)

jerzyjamroz:
The function "dbValueSize()" delivers the same:
```
pRec->fta 10 dbValueSize(pRec->fta) 8 menuEpicsTypeSizes((menuFtype)pRec->fta) 8
pRec->fta 2 dbValueSize(pRec->fta) 1 menuEpicsTypeSizes((menuFtype)pRec->fta) 1
pRec->fta 6 dbValueSize(pRec->fta) 4 menuEpicsTypeSizes((menuFtype)pRec->fta) 4
```
I will investigate the "dbAccessDefs.h" capabilities. Thank you for the comments.

PR #401, opened 2023-07-07
Com: Make STATIC_ASSERT macro typedefs unique

JJL772:
When building in C++03 mode, `STATIC_ASSERT` generates a typedef in the form `static_assert_failed_at_line_N` where `N` is the line number. This can trigger shadowing warnings if you have a `STATIC_ASSERT` in function scope on the same line as a global-scope `STATIC_ASSERT` in some included header. EPICS has some `STATIC_ASSERT`s in various header files, which ended up triggering a shadowing warning in some code I wrote.

Something like this would fail to compile with `-std=c++03 -Wshadow -Werror=shadow`
```
// myheader.h

STATIC_ASSERT(sizeof(int) == 4); // Line 3

// myfile.cpp
void test() {
    STATIC_ASSERT(sizeof(bool) == 1); // Line 3
}
```

To fix this, I just inserted `__COUNTER__` into the typedef. Sufficiently old compilers, like GCC 4.1 and earlier, don't support this macro and will have to eat the warning.

### DIFF
--- a/modules/libcom/src/osi/epicsAssert.h
+++ b/modules/libcom/src/osi/epicsAssert.h
@@ -78,14 +78,22 @@ LIBCOM_API void epicsAssert (const char *pFile, const unsigned line,
 #if __cplusplus>=201103L
 #define STATIC_ASSERT(expr) static_assert(expr, #expr)
 #else
-#define STATIC_JOIN(x, y) STATIC_JOIN2(x, y)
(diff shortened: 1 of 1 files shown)


anjohnson:
Core Group: Approved as long as the re-run Appveyor check passes, thanks!

mdavidsaver:
Applied as 395015aac4d24fdc81e757b4db7e162a8b52a9fc

PR #403, opened 2023-07-13
rtems: Allow site local override of the default tick period

kiwichris:
Add to `configure/CONFIG_SITE.local`:
```
 OP_SYS_CFLAGS += -DCONFIGURE_MICROSECONDS_PER_TICK=1000
```
This change lets us avoid a local patch to change the default tick period.


anjohnson:
Core Group: Please just change the default for Posix to 1000 (1KHz ticks), and remove the ifdef.

anjohnson:
@kiwichris Is this PR still valid and needed? If so, did the Core Group's request from 2023/08/09 make sense?

kiwichris:
Yes. It lets you override the default tick period without carrying a private patch around. Some systems needs 1 millisecond.

anjohnson:
Core Group: Approved, although we might change the default tick rate to 1KHz before release.

PR #404, opened 2023-07-17
Fix issues detected by clang-tidy

JJL772:
Inspired by #392, I decided to run clang-tidy on epics-base and it pointed out these issues.

* Fix possible nullptr de-reference in `dbLexRoutines.c`. There was just a missing else block in some of these functions.
* Fix possible memory leak in `lnkConst.c`

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -856,7 +856,9 @@ static void dbDriver(char *name)
     if(!pgphentry) {
         yyerrorAbort("gphAdd failed");
     }
-    pgphentry->userPvt = pdrvSup;
+    else {
+        pgphentry->userPvt = pdrvSup;
+    }
     ellAdd(&savedPdbbase->drvList,&pdrvSup->node);
 }
 
@@ -876,7 +878,9 @@ static void dbLinkType(char *name, char *jlif_name)
     if (!pgphentry) {
         yyerrorAbort("gphAdd failed");
     }
-    pgphentry->userPvt = pLinkSup;
+    else {
+        pgphentry->userPvt = pLinkSup;
+    }
     ellAdd(&savedPdbbase->linkList, &pLinkSup->node);
 }
 
@@ -899,7 +903,9 @@ static void dbRegistrar(char *name)
     if(!pgphentry) {
         yyerrorAbort("gphAdd failed");
     }
-    pgphentry->userPvt = ptext;
+    else {
+        pgphentry->userPvt = ptext;
+    }
     ellAdd(&savedPdbbase->registrarList,&ptext->node);
 }
 
@@ -922,7 +928,9 @@ static void dbFunction(char *name)
     if(!pgphentry) {
        yyerrorAbort("gphAdd failed");
     }
-    pgphentry->userPvt = ptext;
+    else {
+        pgphentry->userPvt = ptext;
+    }
     ellAdd(&savedPdbbase->functionList,&ptext->node);
 }
 
@@ -946,7 +954,9 @@ static void dbVariable(char *name, char *type)
     if(!pgphentry) {
         yyerrorAbort("gphAdd failed");
     }
-    pgphentry->userPvt = pvar;
+    else {
+        pgphentry->userPvt = pvar;
+    }
     ellAdd(&savedPdbbase->variableList,&pvar->node);
 }
 
--- a/modules/database/src/std/link/lnkConst.c
+++ b/modules/database/src/std/link/lnkConst.c
@@ -225,8 +225,10 @@ static jlif_result lnkConst_string(jlink *pjlink, const char *val, size_t len)
         if (!vec)
             return jlif_stop;
         str = malloc(len+1);
-        if (!str)
+        if (!str) {
+            free(vec);
             return jlif_stop;
+        }
 
         strncpy(str, val, len);
         str[len] = '\0';
--- a/modules/libcom/src/calc/postfix.c
+++ b/modules/libcom/src/calc/postfix.c
@@ -335,6 +335,10 @@ LIBCOM_API long
             break;
 
         case SEPERATOR:
+            if (pstacktop == stack) {
+                *perror = CALC_ERR_BAD_SEPERATOR;
+                goto bad;
+            }
             /* Move operators to the output until open paren */
             while (pstacktop->name[0] != '(') {
                 if (pstacktop <= stack+1) {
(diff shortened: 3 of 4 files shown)


modules/libcom/src/calc/calcPerform.c
```diff
@@ -46,6 +46,7 @@ LIBCOM_API long
     calcPerform(double *parg, double *presult, const char *pinst)
 {
     double stack[CALCPERFORM_STACK+1];  /* zero'th entry not used */
+    memset(stack, 0, sizeof(stack));
```
anjohnson: I don't believe this is the correct fix. I get a seg-fault when `postfix()` parses the expressions `1,` which may be the case that clang-tidy has found, but adding this `memset()` doesn't prevent that. Did you find a different expression to trigger this? Most of my attempts such as a single `,` got rejected by the parser anyway without triggering the seg-fault.

I think this calls for a separate bug report against the calc parser, `0)` is also causing seg-faults, and my attempts at simple fixes don't resolve that properly.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -856,7 +856,9 @@ static void dbDriver(char *name)
     if(!pgphentry) {
         yyerrorAbort("gphAdd failed");
     }
-    pgphentry->userPvt = pdrvSup;
+    else {
+        pgphentry->userPvt = pdrvSup;
+    }
```
anjohnson, line 861: I guess we've not had a `gphAdd()` failure in a long time — that should only happen in an out-of-memory situation I suppose.

modules/database/src/std/link/lnkConst.c
```diff
@@ -225,8 +225,10 @@ static jlif_result lnkConst_string(jlink *pjlink, const char *val, size_t len)
         if (!vec)
             return jlif_stop;
         str = malloc(len+1);
-        if (!str)
+        if (!str) {
+            free(vec);
```
anjohnson, line 229: Agreed.

anjohnson (changes requested):
This looks good to me if you undo your changes to the libcom/src/calc files, and I'd appreciate if you would instead file a bug report for the failing calc expressions I quoted below and any others you've discovered.

modules/libcom/src/calc/calcPerform.c
anjohnson: FYI the Appveyor build failures from the VS2012 compiler for this PR were due to your `memset()` calls preceding some variable declarations — all our C code still puts all declarations before statements inside a block, which was required by older C standards.

JJL772: Appreciate you taking the time to look over this!

You're definitely right- this is not the correct solution, I should've dug into the code a bit more. 

I believe that adding a stack position check before dereferencing `pstacktop->name` would fix these issues. With the following change, all calc tests pass, and valgrind no longer complains about uninitialized memory use when parsing `0)`
```diff
diff --git a/modules/libcom/src/calc/postfix.c b/modules/libcom/src/calc/postfix.c
index 7457abaab..d9dce515d 100644
--- a/modules/libcom/src/calc/postfix.c
+++ b/modules/libcom/src/calc/postfix.c
@@ -335,6 +335,10 @@ LIBCOM_API long
             break;
 
         case SEPERATOR:
+            if (pstacktop == stack) {
+                *perror = CALC_ERR_BAD_SEPERATOR;
+                goto bad;
+            }
             /* Move operators to the output until open paren */
             while (pstacktop->name[0] != '(') {
                 if (pstacktop <= stack+1) {
@@ -353,6 +357,10 @@ LIBCOM_API long
             break;
 
         case CLOSE_PAREN:
+            if (pstacktop == stack) {
+                *perror = CALC_ERR_PAREN_NOT_OPEN;
+                goto bad;
+            }
             /* Move operators to the output until matching paren */
             while (pstacktop->name[0] != '(') {
                 if (pstacktop <= stack+1) {
```

For now I've dropped the problematic commit from this PR. If this fix looks acceptable, I can push it.

anjohnson: Those changes do seem to work, if you want to add them back would you also add tests for `0)` and `0,` to the modules/libcom/test/epicsCalcTest.cpp file, there are a bunch of calls to `testBadExpr()` towards the end of the `MAIN()` routine and these belong there. You would also need to change the number of tests in the call to `testPlan()`.

Thanks!

JJL772: Alright, added in the latest diff!

anjohnson: Thanks, looks fine by me now. The Core Developers group will probably take a look at this sometime, we're a bit behind in approving PRs at the moment though so I can't predict when it will actually get merged.

anjohnson (approved):

modules/database/src/std/link/lnkConst.c
mdavidsaver: Looking at this logic, I don't think it was or is handling OOM correctly.  By the time of the `malloc()`, the prceeding `realloc()` may have invalidated `clink->value.pmem`.  The added `free(vec)` would now make certain that `clink->value.pmem` is invalid, and left as a dangling pointer.

A better scheme might be to do the `malloc()` first.  like:

```c
str = malloc(len+1);
if(!str)
    return jlif_stop;
vec = realloc(clink->value.pmem, newElems * sizeof(char *));
if(!vec) {
    free(str);
    return jlif_stop; /* clink->value.pmem remains valid */
}
```

mdavidsaver (changes requested):
I have picked off the calc change as 60fa2d31da2deaaa89636b4a1f5b852efe0b2699.  The const link parsing still needs some work.

modules/database/src/std/link/lnkConst.c
anjohnson: How can `realloc()` invalidate that buffer in an OOM situation? If it can't supply the additional memory requested it [must](https://en.cppreference.com/w/c/memory/realloc) return `NULL` without freeing or modifying the old block. The `NULL` will then stop the parsing (line 225).

By this point `clink->type == ac40` so `lnkConst_free()` will eventually `free()` each of the already-saved strings and the `clink->value.pmem` buffer. The `clink->nElems` value is not incremented until later when `lnkConst_string()` is about to `return jlif_continue;` so `lnkConst_free()` won't try to free the uninitialized value that has been appended to the vector.

I agree that the original change that @JJL772 proposed is not correct either, the solution should be to move the assignment of `clink->value.pstrings = vec;` up to just before the call to `malloc()`, which will ensure that the new vector and all of the previously assigned string elements in it get properly released in `lnkConst_free()`.

PR #408, opened 2023-07-31
libCom: detect support for backtrace() with __has_include.

ericonr:
This is necessary in order to build epics-base with musl libc, for example, and any other C libraries which don't include this functionality. In order to not regress builds with older compilers, we still support the uclibc check.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdExecinfoBackTrace.cpp
+++ b/modules/libcom/src/osi/os/posix/osdExecinfoBackTrace.cpp
@@ -12,10 +12,13 @@
 #include <stdlib.h>
 
 // execinfo.h may not be present if uclibc is configured to omit backtrace()
-#if !defined(__UCLIBC_MAJOR__) || defined(__UCLIBC_HAS_EXECINFO__)
+// some C libraries, such as musl, don't have execinfo.h at all
+#if defined(__has_include)
+#  if __has_include(<execinfo.h>)
+#    define HAS_EXECINFO 1
+#  endif
+#elif !defined(__UCLIBC_MAJOR__) || defined(__UCLIBC_HAS_EXECINFO__)
 #  define HAS_EXECINFO 1
-#else
-#  define HAS_EXECINFO 0
 #endif
 
 #if HAS_EXECINFO


anjohnson:
Core Group: @mdavidsaver suggested that some of the repetition in this change could be removed.

@anjohnson isn't convinced, noting that the `__has_include` special operator is [only valid](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html) in `#if` or `#elif` expressions (maybe Michael was thinking of a different change though?).

On the other hand if uclibc is configured to omit `backtrace()`, might it still install the execinfo.h header file? If so that would break this change if the compiler used supports `__has_include`. This does seem to be the only file which tests for uclibc. Hopefully Michael can answer that, since he added this conditional in 2019, possibly for Fermilab.

mdavidsaver:
> ... some of the repetition in this change could be removed.

I was referring to the fact that

```c
#if HAS_EXECINFO
```

is well behaved even if `HAS_EXECINFO` is not defined.  Specifically, undefined names in `#if` will be replaced with `0`.  So:

```c
#  else
#    define HAS_EXECINFO 0
```

is redundant, and could be omitted.

mdavidsaver:
> ... `__has_include` special operator is [only valid](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html) in `#if` or `#elif` ...

This was my understanding as well.  I am not sure what you understood me to be suggesting.

> ... uclibc ...

I looked at uclibc previously in the context of [buildroot](https://buildroot.org/).  However, I ended up using glibc.

Also, it is worth noting that while we may talk about "uclibc", the original [uclibc](https://en.wikipedia.org/wiki/UClibc) is defunct, and has been superseded by [uclibc-ng](https://uclibc-ng.org/).  The former had 0.x.y version numbers, and the later started with 1.x.y.


> ... he added this conditional in 2019 ...

I recall this was in response to some tech-talk reports of compilation failures.

mdavidsaver:
> On the other hand if uclibc is configured to omit `backtrace()`, might it still install the execinfo.h header file?

The header is conditionally installed.

https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/Makefile.in?id=cdb07d2cd52af39feb425e6d36c02b30916b9f0a#n224

https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/Makefile.in?id=cdb07d2cd52af39feb425e6d36c02b30916b9f0a#n277

ericonr:
@mdavidsaver does this look alright to you?

mdavidsaver:
> @mdavidsaver does this look alright to you?

Yes, I think so.

mdavidsaver:
Applied as 7c4a21eab44183a84f25ea234ce2fde8ad08c4ed

PR #409, opened 2023-08-01
Fix libcom test failure on RTEMS

zhaynabb:
LibCom test started to fail on RTEMS after merging #310 to base. Fixed by creating a writable area for file operations.

### DIFF
--- a/modules/libcom/test/epicsStringTest.c
+++ b/modules/libcom/test/epicsStringTest.c
@@ -14,6 +14,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <fcntl.h>
 
 #include "epicsUnitTest.h"
 #include "epicsString.h"
--- a/modules/libcom/test/epicsStringTest.c
+++ b/modules/libcom/test/epicsStringTest.c
@@ -198,10 +198,22 @@ void testStrTok(void)
 static
 void testEpicsStrPrintEscaped(void)
 {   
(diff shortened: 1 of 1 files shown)


mdavidsaver:
@Zhaynabb What error message(s) are you seeing?  In what circumstances?  eg. are you running tests from a read-only filesystem?

zhaynabb:
```# testEpicsStrPrintEscaped()
not ok 440 - epicsStrPrintEscaped(testFile, "1234", 4) == 4
not ok 441 - epicsStrPrintEscaped(testFile, "\a\b\f\n\r\t\v\\\'\"", 10)
*** FATAL ***
fatal source: 9 (RTEMS_FATAL_SOURCE_EXCEPTION)
exception vector 18 (0x12)```

I was getting the error above when running the test on MVME3100. Epics base 7 was built on RTEMS 6.

anjohnson:
Note that this is the same solution that was implemented for [epicsStdioTest](https://github.com/epics-base/epics-base/blob/524f81b8bd147bb714c9ea7b7462b8912a134246/modules/libcom/test/epicsStdioTest.c#L150). There are build failures which seem to be because earlier RTEMS versions need a definition of `mkdir()` from some additional include file(s).

anjohnson:
Core Group: Approved.

mdavidsaver:
Applied as 7fd690e53e597ddca650e1fad04721b61eba973b

PR #413, opened 2023-08-03
Support fully static builds on Linux

ericonr:
This is working just fine for my (native, x86_64) builds, I'd love suggestions for broader testing.

I first brought this up in https://epics.anl.gov/tech-talk/2022/msg01637.php

@mdavidsaver brought up a few concerns in private, some of which I'm addressing here:

- this initial PR doesn't handle things such as static linking for libreadline. Since full static builds seem like a kind of "expert" option, I'd be okay with requiring extra setup in `configure/CONFIG_SITE` when building IOCs. The main goal is not having to patch epics-base to do it.
- musl (#408) explicitly supports static linking, unlike glibc, making it a reliable option (I faced some segfaults when using statically linked glibc, which depended on the host's configuration files, like `/etc/nsswitch.conf`)

I am entirely open to a better naming scheme, this was simply an initial suggestion.


ericonr:
The build failures, at least for the ubuntu jobs, seem to be in:

```
2023-08-03T19:15:04.7157095Z g++ -o antelope -static -L/home/runner/work/epics-base/epics-base/lib/linux-x86_64 -Wl,-rpath,/home/runner/work/epics-base/epics-base/lib/linux-x86_64           -rdynamic -m64         closure.o error.o lalr.o lr0.o antelope.o mkpar.o output.o reader.o skeleton.o symtab.o verbose.o warshall.o     -Wl,-Bdynamic  -lpthread   -lreadline -lm -lrt -ldl -lgcc
2023-08-03T19:15:04.7158442Z make[4]: Leaving directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:04.7165217Z make[4]: Entering directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:04.7166073Z Installing created executable ../../../../bin/linux-x86_64/antelope
2023-08-03T19:15:04.7166974Z make[4]: Leaving directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:04.7414792Z /bin/sh: 1: ../../../../bin/linux-x86_64/antelope: not found
2023-08-03T19:15:04.7415487Z make[4]: Entering directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:04.7419750Z "../../../../bin/linux-x86_64/antelope" -bparse -l -d ../flex/parse.y
2023-08-03T19:15:04.7420608Z make[4]: Leaving directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:05.7284620Z make[4]: Entering directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:05.7285613Z "../../../../bin/linux-x86_64/antelope" -bparse -l -d ../flex/parse.y
2023-08-03T19:15:05.7286561Z make[4]: Leaving directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:05.7291791Z /bin/sh: 1: ../../../../bin/linux-x86_64/antelope: not found
2023-08-03T19:15:05.7342357Z make[4]: Entering directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:05.7343218Z Installing PERL_MODULES file ../../../../lib/perl/EPICS/macLib.pm
2023-08-03T19:15:05.7344415Z make[4]: Leaving directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.linux-x86_64'
2023-08-03T19:15:05.7352368Z make[4]: *** No rule to make target 'parse.c', needed by 'parse.d'.  Stop.
2023-08-03T19:15:05.7352920Z make[4]: *** Waiting for unfinished jobs....
```

I wouldn't expect my changes to affect this at all... Do you have any suggestions?

anjohnson:
Don't worry about the Appveyor build failures, I told it to re-run them and I expect them to pass this time, that happens often.

The "Ub-20 gcc-9 + MinGW" check CI-build used this command to successfully link antelope for the linux-x86_64 host:
```
g++ -o antelope  -L/home/runner/work/epics-base/epics-base/lib/linux-x86_64 -Wl,-rpath,/home/runner/work/epics-base/epics-base/lib/linux-x86_64           -rdynamic -m64         closure.o error.o lalr.o lr0.o antelope.o mkpar.o output.o reader.o skeleton.o symtab.o verbose.o warshall.o
```
The "Ub-20 gcc-9 + MinGW, static" check CI-build that failed used this command to link antelope for the host:
```
g++ -o antelope -static -L/home/runner/work/epics-base/epics-base/lib/linux-x86_64 -Wl,-rpath,/home/runner/work/epics-base/epics-base/lib/linux-x86_64           -rdynamic -m64         closure.o error.o lalr.o lr0.o antelope.o mkpar.o output.o reader.o skeleton.o symtab.o verbose.o warshall.o     -Wl,-Bdynamic  -lpthread   -lreadline -lm -lrt -ldl -lgcc
```
Whatever binary the link actually created got installed, but it isn't being seen as something which can be executed as shown in your log excerpt above. Does that link command look right to you?

I don't remember how our CI scripts configure the different builds above, but they might be doing something which your modifications aren't expecting.

ralphlange:
I know this is about doing a fully static build of EPICS, but does that have to include the tools?

ericonr:
> Whatever binary the link actually created got installed, but it isn't being seen as something which can be executed as shown in your log excerpt above. Does that link command look right to you?

It does, but I will try and dig a bit deeper.

> I know this is about doing a fully static build of EPICS, but does that have to include the tools?

My intention had been not to include the tools, in fact. The `FULL_STATIC_BUILD` flag would be defined when building things which depend on epics-base, such as IOCs, but epics-base itself shouldn't have been affected.

configure/os/CONFIG.Common.linuxCommon
```diff
@@ -20,8 +20,10 @@ OP_SYS_LDFLAGS += -rdynamic
 
 # Linker flags for static & shared-library builds
 STATIC_LDFLAGS_YES= -Wl,-Bstatic
+STATIC_LDFLAGS_FULL= -Wl,-Bstatic
```
ericonr, line 23: I realized I'm not using this in `CONFIG_COMMON`

ericonr:
Comparing successful (#408) and failed runs, we get:

Success: `g++ -o antelope -Wl,-Bstatic -L/home/runner/work/epics-base/epics-base/lib/linux-x86_64 -Wl,-rpath,/home/runner/work/epics-base/epics-base/lib/linux-x86_64           -rdynamic -m64         closure.o error.o lalr.o lr0.o antelope.o mkpar.o output.o reader.o skeleton.o symtab.o verbose.o warshall.o     -Wl,-Bdynamic  -lpthread   -lreadline -lm -lrt -ldl -lgcc`

Failed: `g++ -o antelope -static -L/home/runner/work/epics-base/epics-base/lib/linux-x86_64 -Wl,-rpath,/home/runner/work/epics-base/epics-base/lib/linux-x86_64           -rdynamic -m64         closure.o error.o lalr.o lr0.o antelope.o mkpar.o output.o reader.o skeleton.o symtab.o verbose.o warshall.o     -Wl,-Bdynamic  -lpthread   -lreadline -lm -lrt -ldl -lgcc`

ericonr:
@anjohnson the issue seems to be due to my commit removing the "duplicated" info from the arch specific files... `configure/CONFIG.gnuCommon` defines `STATIC_LDFLAGS_YES= -static`, while `configure/os/CONFIG.Common.linuxCommon` defines it to (the working) `-Wl,-Bstatic`. However, `linuxCommon` is overridden by `gnuCommon`, and not the opposite (which is what I would have expected).

I am testing this in CI now.

ericonr:
```
make[4]: Entering directory '/home/erico.rolim@abtlus.org.br/projects/epics/musl/epics-7/modules/libcom/src/O.linux-x86_64'
/usr/bin/g++ -o antelope -static -L/home/erico.rolim@abtlus.org.br/projects/epics/musl/epics-7/lib/linux-x86_64 -Wl,-rpath,/home/erico.rolim@abtlus.org.br/projects/epics/musl/epics-7/lib/linux-x86_64          -rdynamic -m64         closure.o error.o lalr.o lr0.o antelope.o mkpar.o output.o reader.o skeleton.o symtab.o verbose.o warshall.o     -Wl,-Bdynamic  -lpthread    -lm -lrt -ldl -lgcc
/usr/sbin/ld: attempted static link of dynamic object `/usr/lib64/gcc/x86_64-linux-musl/12.2.0/../../../../lib64/libstdc++.so'
/usr/sbin/ld: attempted static link of dynamic object `/usr/lib64/gcc/x86_64-linux-musl/12.2.0/../../../../lib64/libc.so'
collect2: error: ld returned 1 exit status
make[4]: *** No rule to make target 'antelope', needed by '../../../../bin/linux-x86_64/antelope'.  Stop.
```

It would seem that linking command is actually incorrect... Which makes some sense: `-static` is about the binary being linked. The issue with gcc-9/binutils in ubuntu-20 is that they don't diagnose this bad condition and seem to instead simply spit out a defective binary.

Do you have any suggestions for changing the "inheritance" order for these properties? Or some other workaround? Or even just keep the revert from the last commit of this PV?

mdavidsaver:
> ... The issue with gcc-9/binutils in ubuntu-20 is ...

Can you link to some reference on this issue?  (bugzilla ticket or similar?)

Also, can you add a github actions job to test this new configuration?

ericonr:
> Can you link to some reference on this issue? (bugzilla ticket or similar?)

Will have to search for it; this was observed behavior, nothing else.

Do you think fixing the overriding behavior I pointed out

> However, linuxCommon is overridden by gnuCommon, and not the opposite (which is what I would have expected).

would be interesting? Or should I just remove the `Remove duplicated STATIC_{LDFLAGS,LDLIBS}_{YES/NO} definitions.` commit?

> Also, can you add a github actions job to test this new configuration?

I will have to check if the build can fully succeed when the tools are fully static, especially due to being on glibc... Otherwise, will have to add something like an Alpine build for it. This will probably only happen after ICALEPCS, though.

ericonr:
@mdavidsaver so, I've been poking at this for a bit...

If I add the following to `configure/CONFIG_SITE.local`:

```
STATIC_BUILD=YES
FULL_STATIC_BUILD=YES
SHARED_LIBRARIES=NO
PROD_LDFLAGS = -static

COMMANDLINE_LIBRARY=
```

I can get everything to build **if** I also remove `Cap5` from the build, modulo glibc warnings about linking certain functions into static executables. Trying to build dynamic libraries with `-Wl,-Bstatic` in LDFLAGS breaks things (as explained in <https://stackoverflow.com/questions/51998912/when-statically-linking-a-library-getting-linker-error-cannot-find-lgcc-s>; would the suggested fix be easy to apply to the flags here?):

```
make[5]: Entering directory '/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/modules/ca/src/perl/O.linux-x86_64'
clang++ -o libCap5.so  -shared -fPIC -Wl,-hlibCap5.so -L/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64 -Wl,-rpath,/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64          -rdynamic -m64         Cap5.o    -lca -lCom -Wl,-Bstatic  -lpthread    -lm -lrt -ldl
/usr/bin/ld: cannot find -lgcc_s: No such file or directory
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x22
/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64/libCom.a(osdFindSymbol.o): in function `epicsLoadLibrary':
osdFindSymbol.c:(.text+0x6): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x22
/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64/libCom.a(osdProcess.o): in function `osiGetUserName':
osdProcess.c:(.text+0x15): warning: Using 'getpwuid' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x22
/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64/libCom.a(osdSock.o): in function `ipAddrToHostName':
osdSock.c:(.text+0x208): warning: Using 'gethostbyaddr' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64/libCom.a(osdSock.o): in function `hostToIPAddr':
osdSock.c:(.text+0x2ad): warning: Using 'gethostbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: cannot find -lgcc_s: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[5]: *** [../../../../../configure/RULES_BUILD:329: libCap5.so] Error 1
```

Removing Cap5 was done with:

```
diff --git a/modules/ca/src/perl/Makefile b/modules/ca/src/perl/Makefile
index 05500ad05..c2850e3bd 100644
--- a/modules/ca/src/perl/Makefile
+++ b/modules/ca/src/perl/Makefile
@@ -49,12 +49,12 @@ ifeq ($(wildcard $(PERL_h)),)
   $(warning The Perl CA module will not be built.)
 else
 ifeq ($(findstring $(OS_CLASS),WIN32 cygwin32),) # Doesn't build on WIN32
-  LOADABLE_LIBRARY_HOST = Cap5
+  #LOADABLE_LIBRARY_HOST = Cap5
 
   PERL_SCRIPTS += capr.pl
 
   PERL_MODULES += CA.pm
-  PERL_MODULES += $(PERL_ARCHPATH)/$(LOADABLE_SHRLIB_PREFIX)Cap5$(LOADABLE_SHRLIB_SUFFIX)
+  #PERL_MODULES += $(PERL_ARCHPATH)/$(LOADABLE_SHRLIB_PREFIX)Cap5$(LOADABLE_SHRLIB_SUFFIX)
 
   HTMLS_DIR = .
   HTMLS = CA.html
```

Would it be acceptable to gate building Cap5 on the value of `SHARED_LIBRARIES`? Given that it is one

ericonr:
Either way,

> the issue seems to be due to my commit removing the "duplicated" info from the arch specific files... configure/CONFIG.gnuCommon defines STATIC_LDFLAGS_YES= -static, while configure/os/CONFIG.Common.linuxCommon defines it to (the working) -Wl,-Bstatic. However, linuxCommon is overridden by gnuCommon, and not the opposite (which is what I would have expected).

is still an issue in that building static executables on linux outside of the cases covered by the files in https://github.com/epics-base/epics-base/pull/413/commits/c488c1ed9e19a4222617c91c2d4a9cfd0b28e23d won't work properly. I imagine the precedence for gnuCommon over the OS might be due to OSs where GNU isn't the default toolchain. I don't know what to do about it though; it could be left untouched for this PR.

ralphlange:
See also #441 - different aspect, same thing.

anjohnson:
Strictly speaking Cap5 is a loadable module rather than a shared library, it's designed to be loaded by another program at run-time (in this case the Perl executable) instead of lazily-linked into it at build-time. On Linux with ELF binaries both use the same file format, but on some OSs such as macOS they are different things, see [here](https://stackoverflow.com/questions/2339679/what-are-the-differences-between-so-and-dylib-on-macos) for an explanation.

The EPICS build system should be able to compile loadable modules without building shared libraries, although that might not have been tested recently. Do you know if Perl on your OS is built with module support? Apparently musl libc does provide the `dlopen()`/`dlclose()` APIs, although `dlclose()` is a no-op there. Maybe we should support a configuration flag in a CONFIG_SITE.Common.<arch> file to indicate whether loadable modules are supported on this target, and not try to build any if not.

ericonr:
> but on some OSs such as macOS they are different things

huh, I had no idea about that, thanks!

> Do you know if Perl on your OS is built with module support? Apparently musl libc does provide the dlopen()/dlclose() APIs, although dlclose() is a no-op there. Maybe we should support a configuration flag in a CONFIG_SITE.Common. file to indicate whether loadable modules are supported on this target, and not try to build any if not.

Oh, this wasn't even a build attempt on musl. I was just using my ubuntu machine for it, because if building everything statically on glibc works fine, it's easier on my side to add it to CI :)

(my epics-base build on musl, which I don't do with `FULL_STATIC_BUILD`, since I only use the option for IOCs, builds the Perl module without issues)

What happens is that the command to create `libCap5.so` has the `-shared` flag (needed to create the so), but it also has `-Wl,-Bstatic` before the system libraries (from the `FULL_STATIC_BUILD` option)

```
clang++ -o libCap5.so  -shared -fPIC -Wl,-hlibCap5.so -L/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64 -Wl,-rpath,/home/erico.rolim@abtlus.org.br/projects/epics/epics-7/lib/linux-x86_64          -rdynamic -m64         Cap5.o    -lca -lCom -Wl,-Bstatic  -lpthread    -lm -lrt -ldl
```

Per this [SO question](https://stackoverflow.com/questions/51998912/when-statically-linking-a-library-getting-linker-error-cannot-find-lgcc-s), what happens is the compiler frontend adds `-lgcc_s` automatically to the linker command line, because it's a shared object, but the linker was told by `-Bstatic` to look only for static libraries, and it can't find `libgcc_s.a` (which doesn't exist).

A possible solution here would be fixing the command line somehow so that the compiler gets something like: `... -Wl,-Bstatic  -lpthread    -lm -lrt -ldl -Wl,-Bdynamic`, but it would be a special case for the shared objects only, otherwise it would break the build for the actual executables. That's why I wondered about having an option to simply disable Cap5 entirely, because that'd be something I actually know how to do.

anjohnson:
If you aren't going to/can't build the Cap5 loadable module there's no point in letting GNUmake descend into the `modules/ca/src/perl` directory at all. I would just apply this patch in that case:
```patch
diff --git a/modules/ca/src/Makefile b/modules/ca/src/Makefile
index 335cec67e..2796c8abb 100644
--- a/modules/ca/src/Makefile
+++ b/modules/ca/src/Makefile
@@ -17,7 +17,7 @@ DIRS += client
 DIRS += tools
 tools_DEPEND_DIRS = client
 
-DIRS += perl
+#DIRS += perl
 perl_DEPEND_DIRS = client
 
 DIRS += template
```

ericonr:
@anjohnson is that a patch I should apply for the CI run for fully static binaries?

anjohnson:
@ericonr Applying that patch for CI builds that aren't able to build the loadable library would be the simplest way to prevent them from breaking due to that Cap5 module, the patch should work on most EPICS 7.0.x releases to date.

However that isn't really solving this issue with the PR, that your fully static build configuration can't build loadable modules. There are a few EPICS extensions which build them (bindings to the languages Python and IDL), but you might not need them yourself. If you want standard EPICS releases to support fully static builds it would be better to fix this at the same time, but to be honest I haven't heard anyone else calling for this.

ericonr:
@anjohnson I'm struggling a bit to understand how to patch base for CI... Could you give me some pointers?

I'm looking at ci-scripts, and from what I can see in [cue.py](https://github.com/epics-base/ci-scripts/blob/130e88b7095812da93e423c602651e30f39da11a/cue.py#L301), when building base we don't define `modlist` to return `BASE` (it's empty instead), so I can't pass a `BASE_HOOK` with the desired patch. I can't even see it [skipping the bad hook](https://github.com/epics-base/ci-scripts/blob/130e88b7095812da93e423c602651e30f39da11a/cue.py#L635).

I found https://github.com/epics-base/epics-base/pull/168 , which mentioned disabling Cap5, but didn't actually do it.

> However that isn't really solving this issue with the PR, that your fully static build configuration can't build loadable modules.

I honestly didn't expect to build epics-base itself with this configuration, it was mainly about specific IOCs. I'm using epics-base in order to have a CI job testing it.

ralphlange:
> I'm looking at ci-scripts, and from what I can see in [cue.py](https://github.com/epics-base/ci-scripts/blob/130e88b7095812da93e423c602651e30f39da11a/cue.py#L301), when building base we don't define modlist to return BASE (it's empty instead), so I can't pass a BASE_HOOK with the desired patch. I can't even see it [skipping the bad hook](https://github.com/epics-base/ci-scripts/blob/130e88b7095812da93e423c602651e30f39da11a/cue.py#L635).

Modlist is the list of dependencies. When building base as the main module, it is not a dependency.
For the main module, you can directly do any necessary modification in your job definition and don't have to use the hooks that are needed for modifying dependencies.

ericonr:
@ralphlange

> For the main module, you can directly do any necessary modification in your job definition and don't have to use the hooks that are needed for modifying dependencies.

To do that, do I simply define a job at the same level as `native` and `docker` in the [CI file](https://github.com/epics-base/epics-base/blob/7.0/.github/workflows/ci-scripts-build.yml)? I don't know another way to make changes to the files...

I suppose I could add another matrix value for a conditional step that applies the patch, but I don't know if that'd be desired...

ralphlange:
I was thinking conditional step, yes.

Maybe there is justification for running hooks on the main module... I just didn't see a use case yet.
Please add an issue on `ci-scripts` if you would like to see that feature.

mdavidsaver:
Sorry, I'm just catching up to this PR.  In my experience, tracking .patch files in VCS repositories is not sustainable.  Nor for CI, and certainly not for users.

@anjohnson Did you mean for the `#DIRS += perl` to be applied as a .patch file during CI runs?

anjohnson:
I was only intending to reply to the part of [this comment](https://github.com/epics-base/epics-base/pull/413#issuecomment-1832571203) where @ericonr provided a diff showing how he was disabling building the Cap5 library for fully-static builds. I have no idea how he was applying his change. My suggestion, to not build the Perl bindings at all in that case is a better way to achieve the same end by being smaller, faster, and more robust against future modifications to those bindings.

I was not suggesting we should be doing that for our CI builds.

ralphlange:
This is about doing that step for exactly one CI build (the one that does the fully static setup), not all of them.

And that's where I'm saying:
Instead of defining a whole new job for this one additional step, maybe being able to define a hook script for the main module - similar to the way they are implemented for dependent modules - is a better way to do that.

ericonr:
Could we define a `BUILD_PERL=NO` or similar build option? It's not future-proof if there's ever a decision to add some other loadable module to epics-base, but (at least to me) that doesn't seem likely. And we avoid complicating the CI setup, since this would just be another `CONFIG_SITE` option.

> Sorry, I'm just catching up to this PR. In my experience, tracking .patch files in VCS repositories is not sustainable. Nor for CI, and certainly not for users.

`modules/ca/src/Makefile` changes very rarely, so a patch against should remain valid for a while, but I understand and agree with not wanting to carry patches here.

anjohnson:
I'd be okay with changing `modules/ca/src/Makefile` to do this:

```Makefile
ifneq ($(PERL_BINDINGS),NO)
  DIRS += perl
  perl_DEPEND_DIRS = client
endif
```

The CI build then just needs to add `PERL_BINDINGS=NO` to either the environment or the `make` command-line to disable compiling in that source subdirectory.

ericonr:
@anjohnson thank you for the suggestion! I added that commit with you as co-author, I hope that's okay.

Will see how CI behaves now.

anjohnson:
Several tests failed in your fully-static CI build here that aren't reflected in the test results shown at the bottom, not sure why. I wouldn't expect you to investigate that particular failing of the CI build, but the test failures themselves are I think in your court to investigate. After building that configuration, please run `make -sj tapfiles` and then `make -s test-results` to get the same summary that the CI-build reports and see if you get the same output. You can also run the failing test programs individually by hand and report whether they fail for you — for example, `cd modules/libcom/test/O.linux-x86_64; ./epicsLoadTest; ./epicsStackTraceTest`

ericonr:
Took me a while to get back to this, but what I have so far:

```
epicsMessageQueueTest.tap ....... 
Failed 1/74 subtests 
[...]
Test Summary Report
-------------------
epicsMessageQueueTest.tap     (Wstat: 0 Tests: 73 Failed: 0)
  Parse errors: Bad plan.  You planned 74 tests but ran 73.
```

I got this once, but couldn't reproduce it for the subsequent run. I think there's a race condition somewhere and the receiver doesn't get the last `ok 74` (it was missing from the tapfile). It seems unrelated, anyway.

```
epicsStackTraceTest.tap ......... 
not ok  2 - dumping symbol from library
not ok  3 - dumping global symbols
not ok  4 - dumping local symbols
Failed 3/5 subtests 
```

This case doesn't seem to be supported for fully static binaries: <https://stackoverflow.com/questions/62804304/getting-in-process-backtraces-in-a-static-linked-binary> and <https://stackoverflow.com/questions/13307820/backtrace-symbols-with-both-static-and-rdynamic>.

```
epicsLoadTest.tap ............... 
not ok  2 - (nil) == 0x4098a0 (epicsThreadGetCPUs) : epicsLoadTest: undefined symbol: epicsThreadGetCPUs
Failed 1/4 subtests 
```

This case doesn't seem to be supported for fully static binaries either: <https://stackoverflow.com/questions/14289488/use-dlsym-on-a-static-binary>.

These two tests make me think it might make sense to have a `-DLINKING_STATIC_FULL` compile flag for tests (akin to the existing `-DLINKING_STATIC`), but that just keeps piling up on the changes from this PR, and I don't love that. Would adding that flag be ok?

ericonr:
Fixed the failing tests with a new macro for the two test files.

There's now a segfault in `testChannelAccess` from `pvAccess` (both in Ubuntu 20.04 and 22.04), which seems to be caused by issues with `getpwnam` (unsurprising - this is one of the reasons I use musl when static linking). It doesn't fail locally for me -- outside of a container -- so it seems the error is caused by some difference in the environment. I will see if I can fix it in the test environment, or if it will be necessary to add Alpine to the testing suite.

anjohnson:
Core Group: If there is still interest in including this please rebase and clear the conflict, and remind us of the current state of the segfault mentioned above. We've converted this to draft status for now, it can be switched back later.

Issue #416, opened 2023-08-11
epicsErrlog does not longer print to console on VxWorks


dirk-zimoch:
Since the major errorlog code refactoring, errlog messages are no longer printed to the console.

I think the logic of this line is wrong:
https://github.com/epics-base/epics-base/blob/00dc55b8a2f8426e531b99b66264c6eeff1e340b/modules/libcom/src/error/errlog.c#L163

It only allows printing to the console during shutdown.

I think is was supposed to mean: `if((localEcho && isOkToBlock) || atExit)` ?
or maybe ` if((localEcho || atExit) && isOkToBlock)` ?

@mdavidsaver, your comment please?

ralphlange:
I have seen a somewhat similar behavior in tests.

If IOCs are run under python, the errlog message queue doesn't get flushed.
In my case, the tests fail because the expected output does not appear. However, when the IOC shuts down, all the queued error messages are printed.

Workaround that causes the errlog queue to be flushed:
On the console, run 'eltc' or 'errlog x'.

dirk-zimoch:
Ralph, that may something else. Usually stdout is in line mode when connected to a console and ins full buffered mode otherwise (e.g. when writing to a pipe to your python script. But stderr should always be line buffered, thus flushing automatically with every `\n`. You may try to call the IOC through something like "expect" which fakes to be a tty.

ralphlange:
The change that Michael's PR introduced is double-buffering the error output.

The logic that you are pointing at is making sure that direct synchronous writes indeed may only happen during exit (because the double-buffering is being torn down).

mdavidsaver:
Part of #113 is that calls to `errlogPrintf()` no longer block for I/O except during IOC shutdown.  My goal being to reduce the impact of logging on callers.

To that end, the errlog message buffer also contains a byte with flags.  The `ERL_*` bit masks.

https://github.com/epics-base/epics-base/blob/00dc55b8a2f8426e531b99b66264c6eeff1e340b/modules/libcom/src/error/errlog.c#L48-L57

Where `ERL_LOCALECHO` marks whether a message should be printed to the local console.  This flag is captured just below the line quoted above.

https://github.com/epics-base/epics-base/blob/00dc55b8a2f8426e531b99b66264c6eeff1e340b/modules/libcom/src/error/errlog.c#L163-L170

This flag is consumed by the errlog thread

[...]

PR #417, opened 2023-08-15
On VxWorks, `stderr` cannot be passed between threads

dirk-zimoch:
On VxWorks, the `stderr` used during initialization may no longer be valid in  `errlogThread`. Thus, do not pass `stderr` directly but initialize `console` to `NULL` instead and later in `errlogThread` use the current `stderr` when `NULL` is found. That means that also the test `isATTY()` can only be done in the `errlogThread` once the current `console` has been decided on.

### DIFF
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -95,7 +95,6 @@ static struct {
     int          atExit;
     int          sevToLog;
     int          toConsole;
-    int          ttyConsole;
     FILE         *console;
 
     /* A loop counter maintained by errlogThread. */
@@ -164,7 +163,8 @@ size_t msgbufCommit(size_t nchar, int localEcho)
         /* errlogThread is not running, so we print directly
          * and then abandon the buffer.
          */
-        fprintf(pvt.console, "%s", start);
+        FILE* console = pvt.console ? pvt.console : stderr;
+        fprintf(console, "%s", start);
 
     } else if(!atExit) {
         start[0u] = ERL_STATE_READY | (localEcho ? ERL_LOCALECHO : 0);
@@ -476,8 +476,7 @@ int errlogSetConsole(FILE *stream)
 {
     errlogInit(0);
     epicsMutexMustLock(pvt.msgQueueLock);
-    pvt.console = stream ? stream : stderr;
-    pvt.ttyConsole = isATTY(pvt.console);
+    pvt.console = stream;
     epicsMutexUnlock(pvt.msgQueueLock);
     /* make sure worker has stopped writing to the previous stream */
     errlogSequence();
@@ -551,8 +550,7 @@ static void errlogInitPvt(void *arg)
     pvt.maxMsgSize = pconfig->maxMsgSize;
     ellInit(&pvt.listenerList);
     pvt.toConsole = TRUE;
-    pvt.console = stderr;
-    pvt.ttyConsole = isATTY(stderr);
+    pvt.console = NULL;
     pvt.waitForWork = epicsEventCreate(epicsEventEmpty);
     pvt.listenerLock = epicsMutexCreate();
     pvt.msgQueueLock = epicsMutexCreate();
@@ -641,8 +639,8 @@ static void errlogThread(void)
         } else {
             /* snapshot and swap buffers for use while unlocked */
             size_t nLost = pvt.nLost;
-            FILE *console = pvt.toConsole ? pvt.console : NULL;
-            int ttyConsole = pvt.ttyConsole;
+            FILE *console = pvt.toConsole ? pvt.console ? pvt.console : stderr : NULL;
+            int ttyConsole = console ? isATTY(console) : 0;
             size_t pos = 0u;
             buffer_t *print;
 


dirk-zimoch:
Fixes #416.

modules/libcom/src/error/errlog.c
```diff
@@ -476,8 +476,7 @@ int errlogSetConsole(FILE *stream)
 {
     errlogInit(0);
     epicsMutexMustLock(pvt.msgQueueLock);
-    pvt.console = stream ? stream : stderr;
-    pvt.ttyConsole = isATTY(pvt.console);
+    pvt.console = stream;
```
mdavidsaver, line 479: What happens with `errlogSetConsole(stdout)` or `errlogSetConsole(stderr)`?

I think an explicit test for the three special `std*` is necessary, with the result stored alongside `pvt.console=NULL`.  (passing `stdin` should be an error)

Also, please do not remove `pvt.ttyConsole`.  `isATTY()` is a relatively expensive function, so I think caching the result makes sense.  (or look at speeding up `isATTY()`)

modules/libcom/src/error/errlog.c
```diff
@@ -95,7 +95,6 @@ static struct {
     int          atExit;
     int          sevToLog;
     int          toConsole;
-    int          ttyConsole;
```
mdavidsaver, line 98: Again, keep `ttyConsole`.  Also add an enum like:

```c
enum consoleType_t {
    consoleFILE, /* use pvt.console */
    consoleStdout, /* use stdout */
    consoleStderr, /* use stderr */
};
```

modules/libcom/src/error/errlog.c
```diff
@@ -641,8 +639,8 @@ static void errlogThread(void)
         } else {
             /* snapshot and swap buffers for use while unlocked */
             size_t nLost = pvt.nLost;
-            FILE *console = pvt.toConsole ? pvt.console : NULL;
-            int ttyConsole = pvt.ttyConsole;
+            FILE *console = pvt.toConsole ? pvt.console ? pvt.console : stderr : NULL;
+            int ttyConsole = console ? isATTY(console) : 0;
```
mdavidsaver, line 643: Rather than evaluating `stderr` and `isATTY(stderr)` every time, do so once on entry to `errlogThread()` as local variables.  The same for `stdout`.

mdavidsaver (changes requested):
I was able to find a... resource on vxworks showing that `stdin`, `stdout`, and `stderr` can evaluate to thread local storage.  These `FILE` reference FDs 0, 1, and 2.  And so would also be effected by `ioTaskStdSet()`.

So I think this PR is necessary.

However, I think it need to go further to ensure that `errlogSetConsole()` does something sensible with `std*`.

mdavidsaver:
A note for the future on an approach I am **not** suggesting.  I looked into what it would take to "clone" a `FILE*`.  It appears that MSVC (now?) provides the necessary POSIX functions `fileno()`, `dup()`, `fdopen()`, and `setvbuf()`.  This would improve the safety of `errlogSetConsole()`, but seems like too much effort.

fyi. I do not find a "getter" to match `setvbuf()`.  So I do not think it possible to "clone" the buffer mode.  However, in the context of errlog, forcing `_IOLBF` would be reasonable.

dirk-zimoch:
My previous idea to use FDs instead of FILE* is blocked by Windows (as usual) not being able to *printf to a FD. GNU/[POSIX.1-2008](http://www.opengroup.org/onlinepubs/9699919799/functions/dprintf.html) has `dprintf` and VxWorks has `fdprintf` for this but I found nothing similar for Windows.

dirk-zimoch:
Maybe we only need something special for VxWorks. Something like
```console = fileno(stream)<=2 ? fdopen(fileno(stream)) : stream; ```
Of course, also making sure to `if(fileno(console)<=2) fclose(console);` when done.

dirk-zimoch:
... on the other hand, if `fileno(stream)` returns 2, that really means nothing because the mapping can change. Probably more something like `fdopen(ioTaskStdGet(0,fileno(stream)),mode)`. But then cleaning up is more difficult.

dirk-zimoch:
A less invasive approach: #418

dirk-zimoch:
Why cloning is necessary: Imagine the calling thread (e.g. the interactive VxWorks shell) calls `errlogSetConsole(NULL)` while having stderr redirected to a file. Simply setting a flag like `consoleStderr` would make the `errlogThread` to use *its own* `stderr` which will *not* use the redirection that was active during `errlogSetConsole()`! This is not what I would expect and it is not what happens on other implementations. Thus, this PR cannot work. Neither with my original approach nor with any other approach that uses any flags and not a clone of the stdout/err `FILE*` in use during `errlogSetConsole()`.
Thus, I close this PR.

mdavidsaver:
> ... calls `errlogSetConsole(NULL)` while having stderr redirected to a file. ... make the `errlogThread` to use _its own_ `stderr` ...

This is ~what I had in mind.  Although I may be missing some subtlety of vxworks.

To my mind, `errlogSetConsole(NULL)` is expressing something like "log output to default", where "default" is the (global) FD 2.  So I think it is valid that vxWorks per-thread redirection is not used.  To capture errlog output, it would be necessary to either:

- `dup2()` to replace FD 2  (globally)
- `errlogSetConsole()` with a `FILE*` usable by the errlog thread.

By passing a non-NULL to `errlogSetConsole()`, the caller is taking responsibility for ensuring that the `FILE*` remains valid until `errlogSetConsole()` is called again.

PR #418, opened 2023-08-21
On Vxworks clone std* file handles because they are thread specific

dirk-zimoch:
Attempts to address #416

### DIFF
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -92,10 +92,11 @@ static struct {
     epicsMutexId msgQueueLock;
 
     /* guarded by msgQueueLock */
-    int          atExit;
     int          sevToLog;
-    int          toConsole;
-    int          ttyConsole;
+    unsigned int atExit:1;
+    unsigned int toConsole:1;
+    unsigned int ttyConsole:1;
+    unsigned int closeConsole:1;
     FILE         *console;
 
     /* A loop counter maintained by errlogThread. */
@@ -466,7 +467,7 @@ int eltc(int yesno)
 {
     errlogInit(0);
     epicsMutexMustLock(pvt.msgQueueLock);
-    pvt.toConsole = yesno;
+    pvt.toConsole = yesno ? 1 : 0;
     epicsMutexUnlock(pvt.msgQueueLock);
     errlogFlush();
     return 0;
@@ -474,9 +475,30 @@ int eltc(int yesno)
 
 int errlogSetConsole(FILE *stream)
 {
+    int closeConsole = 0;
     errlogInit(0);
+    if (!stream) {
+        stream = stderr;
+    }
+#ifdef vxWorks
+    /* The stderr/stdout we get may become invalid as they are thread specific.
+     * Also, the FDs are actually maps to some "real" FDs.
+     * Thus, we clone the stream and link it with a dup of the "real" FD.
+     * The dup'ed FD will be closed when the FILE handle is fclosed. */
+    if (fileno(stream) <= 2) {
+        closeConsole = 1;
+        stream = fdopen(dup(ioTaskStdGet(0, fileno(stream))),"w");
+        if (!stream) {
+            return -1;
+        }
+    }
+#endif
     epicsMutexMustLock(pvt.msgQueueLock);
-    pvt.console = stream ? stream : stderr;
+    if (pvt.closeConsole) {
+        fclose(pvt.console);
+    }
+    pvt.closeConsole = closeConsole;
+    pvt.console = stream;
     pvt.ttyConsole = isATTY(pvt.console);
     epicsMutexUnlock(pvt.msgQueueLock);
     /* make sure worker has stopped writing to the previous stream */
@@ -551,7 +573,13 @@ static void errlogInitPvt(void *arg)
     pvt.maxMsgSize = pconfig->maxMsgSize;
     ellInit(&pvt.listenerList);
     pvt.toConsole = TRUE;
+#ifdef vxWorks
+    pvt.console = fdopen(dup(ioTaskStdGet(0, 2)),"w");
+    pvt.closeConsole = 1;
+#else
     pvt.console = stderr;
+    pvt.closeConsole = 0;
+#endif
     pvt.ttyConsole = isATTY(stderr);
     pvt.waitForWork = epicsEventCreate(epicsEventEmpty);
     pvt.listenerLock = epicsMutexCreate();


modules/libcom/src/error/errlog.c
```diff
@@ -92,10 +92,11 @@ static struct {
     epicsMutexId msgQueueLock;
 
     /* guarded by msgQueueLock */
-    int          atExit;
     int          sevToLog;
-    int          toConsole;
-    int          ttyConsole;
+    int          atExit:1;
+    int          toConsole:1;
+    int          ttyConsole:1;
```
dirk-zimoch: These are all booleans.

anjohnson: If they are booleans, please make these bitfields `unsigned int` — a signed bitfield of width 1 might have a true value of either 1 or -1 on different architectures, so how do you set it portably?

modules/libcom/src/error/errlog.c
```diff
@@ -476,8 +477,22 @@ int errlogSetConsole(FILE *stream)
...
+        fclose(pvt.console);
+    }
     pvt.console = stream ? stream : stderr;
     pvt.ttyConsole = isATTY(pvt.console);
+#ifdef vxWorks
+    /* The stderr/stdout we get may become invalid as they are thread specific.
+     * Also, the FDs are actually maps to some "real" FDs.
+     * Thus, we clone the pvt.console and link it with a dup of the "real" FD.
+     * The dup'ed FD will be closed when the FILE handle is fclosed. */
+    if (fileno(pvt.console) <= 2) {
```
mdavidsaver: I think this condition should be checked earlier, before overwriting `pvt.console`.  Doing so gives a simpler path on error.  I expect that `fdopen()`, `dup()`, and perhaps `ioTaskStdGet()` can fail.

Also, would a more direct test for the thread specific case be better?

```c
if (stream==stdin) return EBADF
if(stream==stdout || stream==stderr) ... clone
```

dirk-zimoch: If any of the used functions fails, we will get a `NULL` for the `FILE*`. What to do in that case? Keep the old setting?

mdavidsaver: Can you think of anything more sensible than keeping the previous `FILE*`, and returning non-zero?

dirk-zimoch: The `errlogThread` simply does nothing if `pvt.console` is `NULL`.
But `fclose()` may not like it, as well as the direct `fprintf()` at exit.
But I will push a new solution with proper error check.

Also I think that checking the FDs is better (in case someone passes a std* from a foreign thread).

modules/libcom/src/error/errlog.c
```diff
@@ -466,17 +467,38 @@ int eltc(int yesno)
...
+        stream = stderr;
+    }
+#ifdef vxWorks
+    /* The stderr/stdout we get may become invalid as they are thread specific.
+     * Also, the FDs are actually maps to some "real" FDs.
+     * Thus, we clone the stream and link it with a dup of the "real" FD.
+     * The dup'ed FD will be closed when the FILE handle is fclosed. */
+    if (fileno(stream) <= 2) {
+        closeConsole = 1;
+        stream = fdopen(dup(ioTaskStdGet(0, fileno(stream))),"w");
```
dirk-zimoch, line 490: I confirmed that `fileno()` of an invalid stream returns `-1`, that `ioTaskStdGet()` of an invalid FD (not 0,1 or 2) returns `-1`, that `dup(-1)` returns `-1` and that `fdopen(-1,"w")` returns `NULL`. 
Also `fclose(NULL)` simply returns `-1` as well as `fprintf(NULL,"x")`.

dirk-zimoch: ... however the functions taking a `FILE*` argument do not explicitly check for `NULL` arguments but instead access the `unsigned long` at that address and check it against the pointer itself, `0` in this case (on most, maybe all supported archs, depending how `NULL` gets encoded), and then the byte at offset `0x0a` against an enum with value`102`. Not the safest thing in the world....

modules/libcom/src/error/errlog.c
```diff
@@ -466,17 +467,38 @@ int eltc(int yesno)
 {
     errlogInit(0);
     epicsMutexMustLock(pvt.msgQueueLock);
-    pvt.toConsole = yesno;
+    pvt.toConsole = yesno ? 1 : 0;
```
ericonr, line 470: Wouldn't it be clearer to do a cast to `bool` here or use `!!yesno`?

Were you fixing a warning with this change? From what I can see, all usage of `eltc` is with 0 or 1 already.

modules/libcom/src/error/errlog.c
```diff
@@ -551,7 +573,13 @@ static void errlogInitPvt(void *arg)
     pvt.maxMsgSize = pconfig->maxMsgSize;
     ellInit(&pvt.listenerList);
     pvt.toConsole = TRUE;
+#ifdef vxWorks
+    pvt.console = fdopen(dup(ioTaskStdGet(0, 2)),"w");
+    pvt.closeConsole = 1;
+#else
     pvt.console = stderr;
+    pvt.closeConsole = 0;
```
ericonr, line 581: Setting `pvt.closeConsole` in the `#else` case shouldn't be necessary, right? Since nothing in the non-vxworks case writes into it.

modules/libcom/src/error/errlog.c
```diff
@@ -466,17 +467,38 @@ int eltc(int yesno)
...
+        if (!stream) {
+            return -1;
+        }
+    }
+#endif
     epicsMutexMustLock(pvt.msgQueueLock);
-    pvt.console = stream ? stream : stderr;
+    if (pvt.closeConsole) {
+        fclose(pvt.console);
+    }
```
ericonr, line 499: It bothers me slightly that this block has a conditional for a variable that's never going to be touched outside of vxworks; the "performance" isn't really relevant, but it still can make the code more confusing to readers. I wonder if it wouldn't make sense to add it under another `#ifdef` block, or something of the sort...

modules/libcom/src/error/errlog.c
mdavidsaver: > But `fclose()` may not like it ...

I have found some claims that `fclose(NULL)` is not portable.  Some (older?) glibc triggers [`*NULL`](https://stackoverflow.com/questions/16922871/why-glibcs-fclosenull-cause-segmentation-fault-instead-of-returning-error) and windows may popup [abort/retry/fail](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fclose-fcloseall?view=msvc-170).

modules/libcom/src/error/errlog.c
mdavidsaver: > Were you fixing a warning with this change?

I'm guessing that Dirk is working with visual studio?  MSVC likes to complain about implicit cast to boolean.  I personally consider that warning to be too noisy, but I know that any compiler warning makes Dirk [angry](https://en.wikipedia.org/wiki/You_wouldn%27t_like_me_when_I%27m_angry).

modules/libcom/src/error/errlog.c
mdavidsaver: I'm not bothered about an `#ifdef`.  However, since `fclose(NULL)` is not portable, paranoia would motivate me test `if(pvt.closeConsole && pvt.console)`.

anjohnson:
Core Group: Closing due to lack of interest. Sites other than PSI and ANL should reopen if they want it to be included.

PR #420, opened 2023-09-01
Read the Docs

minijackson:
This PR adds a Read the Docs build (and everything needed), which allows us to serve the Record reference (from `.dbd.pod` files), and the C/C++ API Reference (from Doxygen + Breathe).

Currently, this introduces a new `doc` make target, which currently only installs markdown file generated from `.pod` and `.dbd.pod` files. This allows not disturbing "normal" builds of the base.

cc @ttkorhonen @mdavidsaver @simon-ess


minijackson:
How it looks like: https://epics-docs-mj-test.readthedocs.io/projects/epics-base/en/latest/

How I plan on integrating it in the epics-docs website: https://epics-docs-mj-test.readthedocs.io/en/latest/

Makefile
```diff
@@ -22,4 +22,13 @@ test_DEPEND_DIRS = src
...
 modules_DEPEND_DIRS = src
 
+DIRS += documentation
+documentation_DEPEND_DIRS = modules
+
 include $(TOP)/configure/RULES_TOP
+
+readthedocs: inc doc
+	$(MAKE) -C documentation/ doxygen T_A=any
+	sphinx-build documentation _readthedocs/html
```
mdavidsaver: fyi.  In other projects I use `$(PYTHON) -m sphinx ...`

documentation/conf.py
```diff
@@ -0,0 +1,147 @@
...
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+
+extensions = [
+    "hoverxref.extension",
+    "breathe",
+    # "exhale",
```
mdavidsaver: Can the `exhale` debris here and below be removed?

documentation/generate-doxygen-header.py
```diff
@@ -0,0 +1,24 @@
...
+    sys.exit(1)
+
+os.makedirs("doxygen", exist_ok=True)
+
+header_name = sys.argv[1]
+doc_file = f"doxygen/{header_name}.rst"
+
+with open(doc_file, "w") as f:
+    print(f"Generating {doc_file}")
+    underline = "".join(["=" for _ in range(len(header_name) + 2)])
```
mdavidsaver:
```py
underline = "=" * (len(header_name) + 2)
```

src/tools/podToMD.pl
```diff
@@ -0,0 +1,103 @@
...
+$SIG{__DIE__} = sub {
+    die @_ if $^S;  # Ignore eval deaths
+    close $out;
+    unlink $opt_o;
+};
+
+my $podMD = Pod::Markdown->new();
+
+$podMD->perldoc_url_prefix('');
+# $podMD->perldoc_url_postfix('.md');
```
mdavidsaver, line 89: As I understand it, this method does not exist?  So the links to other .md are broken.

minijackson: Yes, that is true. Currently, the URLs that are broken are URLs to another file, with a fragment. So a URL like this are broken: `dbCommonRecord#Alarm-Fields`.

URLs pointing to just another file, or just another section are fine.

.gitignore
```diff
@@ -17,3 +17,5 @@ O.*/
 *.log
 .*.swp
 .DS_Store
+venv
+_readthedocs
```
anjohnson: Can these files/directories be created anywhere in the tree, or only at the top level? If the latter they should be spelled `/venv/` and `/_readthedocs/` (and please move them above the other patterns that are more general).

modules/database/src/std/link/Makefile
```diff
@@ -16,4 +16,5 @@ dbRecStd_SRCS += lnkCalc.c
 dbRecStd_SRCS += lnkState.c
 dbRecStd_SRCS += lnkDebug.c
 
+MARKDOWN += links.md
```
anjohnson: Should this be `MARKDOWNS += links.md` instead of the singular version?

modules/database/src/tools/dbdToMD.pl
```diff
@@ -0,0 +1,338 @@
...
+use DBD::Parser;
+use EPICS::Getopts;
+use EPICS::macLib;
+use EPICS::Readfile;
+
+use Pod::Usage;
+
+=head1 NAME
+
+dbdToRst.pl - Convert DBD file with POD to Markdown
```
anjohnson: This file got renamed (line 40 also still uses the old name).

modules/database/src/tools/dbdToMD.pl
```diff
@@ -0,0 +1,338 @@
...
+    } $dbd->pod;
+
+$podRst->output_fh($out);
+$podRst->parse_string_document($pod);
+close $out;
+
+sub menuToPod {
+    my ($menu) = @_;
+    my $index = 0;
+    return "| Index | Identifier | Choice String |\n| ----- | ---------- | ------------- |",
```
anjohnson: Instead of embedding `\n` and multiple lines of text into a single string, please pass a separate list item for each line like the dbdToHtml.pl script does — the top-level parsing code joins all of the list items together with a newline, and returning multiple items makes it easier to understand the Perl code (more so with the instance in `rtypeToPod()` below).

modules/database/src/tools/dbdToMD.pl
```diff
@@ -0,0 +1,338 @@
...
+            my @names = split /\s*,\s*/, $1;
+            # Look up the named fields
+            my @fields = map {
+                    my $field = $rtyp->field($_);
+                    die "Unknown field name '$_' in $infile POD\n"
+                        unless $field;
+                    $field;
+                } @names;
+            # Generate Pod for the table
+   "| Field | Summary | Type | DCT | Default |  Read | Write | CA PP |\n| ----- | ------- | ---- | --- | ------- | ---- | ---- | ----- |",
```
anjohnson: Returning multiple list items also lets you split them into multiple source lines and avoids developers having to scroll sideways.

modules/database/src/tools/dbdToMD.pl
```diff
@@ -0,0 +1,338 @@
...
+        $md .= " | ",;
+    } else {
+        $md .= " | ",;
+    }
+    $md .= $fld->attribute('promptgroup') ? 'Yes' : 'No'; $md .= " | ",;
+    $md .= $fld->attribute('initial') || ' '; $md .= " | ",;
+    $md .= $fld->readable; $md .= " | ",;
+    $md .= $fld->writable;$md .= " | ",;
+    $md .= $fld->attribute('pp') eq 'TRUE' ? 'Yes' : 'No'; $md .= " | ",;
+    return $md;
```
anjohnson: Given all of the identical `$md .= " | ";` separators this code has to append, I would change it to push items onto an array and `join` them all with that string at the end, something like this:

```perl
    my @md;
    push @md, $fld->name, $fld->attribute('prompt');
    my $type = $fld->public_type;
    push @md, $type;
    push @md, ' [' . $fld->attribute('size') . ']'
        if $type eq 'STRING';
…
    push @md, $fld->attribute('initial') || ' ';
    push @md, $fld->readable, $fld->writable;
    push @md, $fld->attribute('pp') eq 'TRUE' ? 'Yes' : 'No';
    return '| ' . join(' | ', @md) . ' |';
```

This is slightly more efficient than continually concatenating to a string (as is using single-quoted literal strings when there are no escape characters or interpolated variables in the value).

modules/database/src/tools/dbdToMD.pl
```diff
@@ -0,0 +1,338 @@
...
+    return 'Maybe'
+        if $special eq "SPC_DBADDR";
+    return $fld->dbf_type eq "DBF_NOACCESS" ? 'No' : 'Yes';
+}
+
+=pod
+
+=head1 Converting Wiki Record Reference to POD
+
+If you open the src/std/rec/aiRecord.dbd.pod file in your favourite plain text
```
anjohnson, line 256: I recommend removing this text from the file since it's just duplicating what's in dbdToHtml.pl. If you're thinking that converter might get deleted in the future though maybe it should be kept here and you'd need to correct the output quoted below, replacing the HTML with the Markdown.

src/tools/podToMD.pl
anjohnson: I had to fix issues like this for the HTML processors, that's why I added the two modules EPICS::PodHtml and EPICS::PodXHtml which are in the src/tools/EPICS directory. I suggest you create similar EPICS::PodMarkdown module to do the link mangling.

configure/RULES_TOP
```diff
@@ -35,7 +35,7 @@ ifndef DISABLE_TOP_RULES
     endif
   endif
 
-  UNINSTALL_DIRS += $(INSTALL_DB) $(INSTALL_DBD) $(INSTALL_DOC) $(INSTALL_HTML)
+  UNINSTALL_DIRS += $(INSTALL_DB) $(INSTALL_DBD) $(INSTALL_DOC) $(INSTALL_HTML) $(INSTALL_MARKDOWN)
```
anjohnson: The reason for constructing `UNINSTALL_DIRS` in multiple lines using `+=` was to avoid making the source lines wider than 80 characters...

documentation/Doxyfile@
```diff
@@ -1986,7 +1985,17 @@ INCLUDE_FILE_PATTERNS  =
...
+                         LIBCOM_API \
+						 DBCORE_API \
+						 LIBCA_API \
+						 DBRECSTD_API \
+						 EPICS_ALWAYS_INLINE \
+						 epicsShareExtern \
+						 epicsShareClass \
+						 epicsShareFunc \
+						 epicsShareAPI \
+						 epicsStdCall
```
anjohnson: Please use spaces instead of tabs for indentation.

configure/CONFIG_COMMON
```diff
@@ -66,6 +66,7 @@ INSTALL_HOST_LIB            = $(INSTALL_LOCATION_LIB)/$(EPICS_HOST_ARCH)
 
 INSTALL_INCLUDE             = $(INSTALL_LOCATION)/include
 INSTALL_DOC                 = $(INSTALL_LOCATION)/doc
+INSTALL_MARKDOWN            = $(TOP)/documentation/generated
```
anjohnson: This line has a subtle [code smell](https://en.wikipedia.org/wiki/Code_smell), it's different from its neighbors. I'm not keen on installing files inside the source tree like this, could we find some other way (or place perhaps, maybe just use `$(INSTALL_LOCATION)/markdown` instead?) to collect the generated markdown files? I think it's partly the non-locality that bothers me, that it's pointing into some other place inside the sources.

configure/RULES_BUILD
```diff
@@ -111,7 +111,7 @@ endif
 #---------------------------------------------------------------
 # Read dependency files
 
-ifneq (inc,$(strip $(MAKECMDGOALS)))
+ifneq ($(filter-out inc doc,$(strip $(MAKECMDGOALS))),)
```
anjohnson: This condition appears in a couple of other places too, I wonder whether it would make sense to turn it into a variable so the complex definition only has to appear once somewhere in the configure directory?

modules/database/src/tools/dbdToMD.pl
```diff
@@ -0,0 +1,338 @@
...
+#     National Laboratory.
+# SPDX-License-Identifier: EPICS
+# EPICS BASE is distributed subject to a Software License Agreement found
+# in file LICENSE that is included with this distribution.
+#*************************************************************************
+
+use strict;
+
+
+use base 'Pod::Markdown';
```
anjohnson: `Pod::Markdown` doesn't come installed with standard Perl so will need installing to use this. I'm sure it's on CPAN, but it isn't listed in the documentation/requirements.txt file that you're adding — it's not clear to me whether that file is intended for human consumption or for a machine.

anjohnson (changes requested):
Various mostly minor issues.

Overall I'm impressed though, thanks!

modules/database/src/tools/dbdToMD.pl
ttkorhonen: This snippet is from me...and shows my clumsiness with Perl. I tried to use list items and join them together, but newlines etc. were inserted where I did not expect them and so on. I resorted to brute-force and this (as well as the code below) was the only way I could make this work. I try to get back to this and figure out a cleaner way.

anjohnson: The sub already returns a list comprising this text string and the output from the `map()` below it. Each list item will have a newline added by the caller, so I'm just suggesting this instead:
```perl
   return "| Index | Identifier | Choice String |",
          "| ----- | ---------- | ------------- |",
        map({choiceTableRow($_, $index++)} $menu->choices);
```

modules/database/src/tools/dbdToMD.pl
anjohnson:
```perl
            # Generate Markdown for the table
            "| Field | Summary | Type | DCT | Default | Read | Write | CA PP |",
            "| ----- | ------- | ---- | --- | ------- | ---- | ----- | ----- |",
            map({fieldTableRow($_, $dbd)} @fields);
        }
```

anjohnson (commented):
Should the subroutines `rtypeToPod` and `menuToPod` be renamed to `…ToMD` instead?

modules/database/src/tools/dbdToMD.pl
ttkorhonen: All right, this works. I thought I had tried something like this but I guess I was just confused. Thanks!

configure/CONFIG_COMMON
ttkorhonen: Maybe @minijackson could comment on this? I do not remember exactly but I thought this target was for RTD build only and would not be installed in a normal build. I think that RTD expects to find the markdown files in the source tree; maybe there is a way to work around this, if necessary. 
For RTD builds, markdown files are the sources ;-)

src/tools/podToMD.pl
ttkorhonen: @minijackson would probably solve this much quicker, but if he does not find the time, I can take this as a challenge to improve my Perl skills. I cannot guarantee quick (or any, for that matter) results though...

anjohnson:
Core Group: Not ready to merge yet, some comments above still need discussing.

src/tools/podToMD.pl
minijackson: I do *not* feel confident about my Perl skill, nor about maintaining an in-house fork of Pod::Markdown.

If possible, I'd like to leave this issue for later, as to not block this PR further.

If it's ok, a workaround might be to replace these links from this: `L<Alarm Fields|dbCommonRecord/Alarm Fields>` to this: `L<Alarm Fields|dbCommonRecord.md/Alarm Fields>`.

This will fix the links for Read the Docs, which if I'm not mistaken should be the default documentation entry point going forward.

ttkorhonen: I am not confident in Perl either, but I looked at the code that Andrew pointed out, and I think I should be able to figure it out. The code that Andrew referred to is not meant to be an implementation of POD to MD conversion, but a rather short snippet to do the link mangling.
However, if I do not manage to solve this in a couple of days, I would be OK to leave this issue to be solved later so that the PR will not be delayed further. It would also be easier (for me at least) to fix this issue separately.
I am a bit hesitant to do the workaround with the links, at least yet. It would solve the issue for RTD, but break the current RRM builds. And, there would be quite many files to change. 
If you give me a couple of days to try this out, I will comment again by the end of the week if I managed.

modules/database/src/tools/dbdToMD.pl
minijackson: I've replaced HTML with Markdown, since ReadTheDocs might replace the raw HTML generation in the future. I've also removed the HTML excerpt.

configure/CONFIG_COMMON
minijackson: I agree with @ttkorhonen, I think the reason I chose TOP and not INSTALL_LOCATION, is that the markdown files are not that useful as installed documentation, and would make the build more complicated when we need to run readthedocs afterwards.

configure/RULES_BUILD
minijackson: Where do you suggest we put it? Should it be named something like "BUILDGOALS"?

modules/database/src/tools/dbdToMD.pl
minijackson: The `requirements.txt` file is for creating Python virtual environments, I think it's not suitable for human consumption. I'll change the `documentation/README.md` file, which also needs an update that I missed.

minijackson:
I've now included the README and KnownProblems in the documentation. I'm sure there is some content to be moved between the README and the epics-docs repository, but if possible I'd like that to be for another PR.

Let me know when you think the PR is acceptable, so that I can `git rebase -i --autosquash` my fixup commits.

src/tools/podToMD.pl
minijackson: Thanks a lot @ttkorhonen! If you don't mind, I do prefer leaving this issue for another PR. Feel free to tag me on that one if you want.

ttkorhonen: Fine with me! I think it is even better that way, makes things simpler. From my point of view, this PR would be good to go.

anjohnson: @ttkorhonen I'd be happy to help if you get stuck. If you haven't already take a look at [this code](https://github.com/rwstauner/Pod-Markdown/blob/9866b1c8d3c6193e91417668ae2f8e74a94e8467/lib/Pod/Markdown.pm#L1221) and below it in the `Pod::Markdown` module which seems to be the part that formats links. Your new `EPICS::PodMarkdown` module would be derived from `Pod::Markdown` and provide its own implementation of one of the routines used there (which it can call using `$self->SUPER::...()` for other link types), similar to the `sub do_pod_link` in my `EPICS::PodHtml` module. Unfortunately every Pod processing module does things slightly differently so you'll have to discover exactly how to detect the link type and process the result appropriately.

configure/CONFIG_COMMON
anjohnson: Since the Markdown files are not being installed under `$(INSTALL_LOCATION)` please rename the new variables to not start with `INSTALL_` and move their definitions (and possibly the rules themselves) so they aren't in the middle of other `INSTALL_...` definitions. That helps to stop future maintainers from thinking that these files *should* be installed like the others are.

The other scent in the code smell is that `$(TOP)/documentation/generated` doesn't match any of the other places where we create or put generated files, and this directory won't be deleted by either a `make clean` or `make distclean` either.

configure/CONFIG_COMMON
```diff
@@ -411,6 +412,7 @@ INSTALL_SCRIPTS = $(SCRIPTS:%= $(INSTALL_BIN)/%)
 INSTALL_OBJS = $(OBJSNAME:%= $(INSTALL_BIN)/%)
 
 INSTALL_DOCS = $(DOCS:%= $(INSTALL_DOC)/%)
+INSTALL_MARKDOWNS = $(MARKDOWNS:%= $(INSTALL_MARKDOWN)/$(MARKDOWNS_DIR)/%)
```
anjohnson: Is `MARKDOWNS_DIR` supposed to be defined somewhere? It's used several times but not set anywhere in this PR.

Makefile
```diff
@@ -23,3 +23,9 @@ DIRS += modules
 modules_DEPEND_DIRS = src
 
 include $(TOP)/configure/RULES_TOP
+
+readthedocs: inc doc
+	$(MAKE) -C documentation/ doxygen T_A=any
+	$(PYTHON) -m sphinx documentation _readthedocs/html
+
+.PHONY: readthedocs
```
anjohnson: The EPICS build rules get re-used by every downstream module, and we try to provide rules that can be used by them too. These targets really belong in the `configure/RULES_TOP` file to allow that.

anjohnson (commented):
I started another review but I don't have time to finish it now and I'm off work for the rest of this week for the US Thanksgiving holiday, so I'll probably have some more comments next week. Sorry!

configure/CONFIG_COMMON
minijackson: Good catch, that's probably dead code.

configure/CONFIG_COMMON
minijackson: I've changed the naming, but I'm not sure what to do about the `clean` / `distclean`, since the `documentation` is not included via the `DIRS` variable. I've had issues trying that before. I'm not sure this is a big issue, since the documentation build is not run by default.

Makefile
minijackson: I'm not sure we should expose these rules, they are quite specific to the way epics-base is architectured. Maybe we can refactor then if and when people ask for it?

ttkorhonen: I agree with @minijackson ; I think these rules would be of very limited, if any, use for downstream modules.

anjohnson: Where `RULES_TOP` is concerned, downstream also includes the pvData, pvAccess and other external Base submodules; are there plans to eventually run Sphinx and RTD on them (which I think would have to happen in a separate stand-alone build, not when checked out as a submodule like this build is)?

I'm fine with the idea of refactoring the rules for future releases, but I disagree that the change would have limited use downstream. External support modules such as [Asyn](https://github.com/epics-modules/asyn/blob/master/docs/Makefile) and [AreaDetector](https://github.com/areaDetector/areaDetector/blob/master/docs/Makefile) are already using Sphinx, but have developed their own rules because they added them before Base did.

mdavidsaver: > ... I'm fine with the idea of refactoring the rules for future releases ...

Would this include incompatible changes, and eventual removal?

I am also skeptical of the desirability of expanding the already expansive interface of our Makefiles.  At least not right away.  Also, I hesitate to enshrine `readthedocs` as a special name in a public interface.  There are alternatives, and RTD won't be around forever.

documentation/Makefile
```diff
@@ -8,17 +8,79 @@ EXPAND_ME += EPICS_MODIFICATION
...
-TARGETS += doxygen
-
 DOXYGEN = doxygen
 ME = documentation/O.$(T_A)/html
 GH_FILES = $(ME)/ $(ME)/.nojekyll $(ME)/*.* $(ME)/*/*.*
 
+DOC_HEADERS += alarm
+DOC_HEADERS += asTrapWrite
+DOC_HEADERS += bucketLib
+DOC_HEADERS += cantProceed
```
anjohnson: How was this `DOC_HEADERS` list generated, what is the criteria for inclusion, and what is it used for? If I do a `grep -p '[@\\]file' *.h */*/*.h` in my `base-7.0/include` directory I get a list of 111 files, many more than are listed here. Note that the headers generated from the `menu*.dbd.pod` and `*Record.dbd.pod` files have contained Doxygen annotations for several releases now.

documentation/KnownProblems.md
```diff
@@ -0,0 +1,12 @@
+# EPICS 7.0.3.1: Known Problems
+
+Any patch files linked below should be applied at the root of the base-7.0.3.1 tree.
```
anjohnson: I haven't actually published a Known Problems page online since EPICS 7.0.1, maintaining it was becoming too time-consuming. I suggest you delete this .md file too (and any links to it, but it looks like only the README.md here actually has one).

documentation/README.md
```diff
@@ -169,12 +155,12 @@ configuration variable if that has been set.
...
 #### `base/documentation` Directory
 
-This contains documents on how to setup, build, and install EPICS.
+This contains EPICS reference documentation.
 
 ```
     README.md           This file
     RELEASE_NOTES.md    Notes on release changes
-    KnownProblems.html  List of known problems and workarounds
+    KnownProblems.md  List of known problems and workarounds
```
anjohnson: Delete this link (discussed above).

documentation/c-api.rst
```diff
@@ -0,0 +1,64 @@
...
+===================
+
+.. toctree::
+   :maxdepth: 1
+   :caption: C/C++ API Reference:
+
+   doxygen/alarm
+   doxygen/asTrapWrite
+   doxygen/bucketLib
+   doxygen/cantProceed
```
anjohnson: Couldn't this list be generated by the Makefile from the contents of `DOC_HEADERS` instead of  having to keep both copies up to date? Doing that by hand is a maintenance headache. I'd be happy to work out how, it should be relatively easy.

documentation/index.rst
```diff
@@ -0,0 +1,28 @@
...
+
+   Back <https://docs.epics-controls.org/en/latest/>
+
+.. toctree::
+   :maxdepth: 2
+   :caption: General Information
+
+   README
+   RELEASE_NOTES
+   KnownProblems
```
anjohnson: Another `KnownProblems` reference to be removed.

configure/RULES_BUILD
anjohnson: I withdraw this suggestion, leave them as they are.

src/tools/podToMD.pl
ttkorhonen: Thanks to a bit slower business before holidays I found some free time to figure out this issue - and in fact, I managed to find a solution. :-) I followed your instructions, @anjohnson , and derived a new module for Pod::Markdown which overrides one function of the parent module. 
I would still prefer to do that change in a follow-up PR, not to make this one more complicated than it already is.

minijackson:
I just rebased the PR. @anjohnson can I bother you to have a look at the last few comments?

src/tools/podToMD.pl
ttkorhonen: I would be grateful if this PR gets attention relatively soon, before I forget the solution I mentioned above... ;-)

anjohnson: @ttkorhonen Ack, I will review Rémi's rebase as soon as I can fit it in.

anjohnson (commented):
Apparently GitHub has 7 review comments saved here that I wrote I-don't-remember-how-long-ago and haven't actually submitted yet, possibly because I didn't get to the end in this round. Sorry for that, the APS upgrade is now in a pre-commissioning Test & Checkout phase, I'm looking forward to having a bit more spare time soon. I'm submitting those comments now so they don't get lost, and I'll take another look.

-----
Later — I didn't see anything else obvious.

documentation/Makefile
minijackson: That's a really good question, I can't remember for the life of me what I used to come up with that list. I've now included every file containing a `@file` / `\file` command.

I've segmented this into three sections, "Generic C/C++ API", "Record C/C++ API", and "Menu C/C++ API", to make the list more "browsable".

documentation/c-api.rst
minijackson: Done in 92d2885a9200f3a2cadc8790ed725e5ac299cf8b

Makefile
anjohnson: I agree that the `readthedocs` rule shouldn't be added to `configure/RULES_TOP`, or to the top-level `Makefile` either. I would guess that only those working on the RTD processing would normally need to run itt, can it appear in the `documentation/Makefile` instead?

documentation/Makefile
```diff
@@ -8,20 +8,161 @@ EXPAND_ME += EPICS_MODIFICATION
...
-
 DOXYGEN = doxygen
 ME = documentation/O.$(T_A)/html
 GH_FILES = $(ME)/ $(ME)/.nojekyll $(ME)/*.* $(ME)/*/*.*
 
+c_TITLE = "Generic C/C++ API"
+
+c_HEADERS += adjustment
+c_HEADERS += alarm
+c_HEADERS += alarmString
```
anjohnson: These lists need to be generated by a program. The existing Doxyfile just processes all of the files installed under `$(TOP)/include` (there are some headers in subdirectories of that which are relevant). If you don't want to include all of them, you should search for those that contain `@file` or `\file`, but I don't know how you would group them under different header types.

documentation/record-reference.md
```diff
@@ -0,0 +1,109 @@
...
+These are the record types supplied with EPICS Base:
+
+```{eval-rst}
+.. toctree::
+   :maxdepth: 1
+
+   generated/aaiRecord
+   generated/aaoRecord
+   generated/aiRecord
+   generated/aoRecord
```
anjohnson: Another list that I would like to be generated if possible, plus the menu files below.

configure/CONFIG_COMMON
```diff
@@ -81,6 +81,11 @@ FINAL_LOCATION              = $(INSTALL_ABSOLUTE)
 # IOC's view of install path
 IOCS_APPL_TOP               = $(INSTALL_ABSOLUTE)
 
+#-------------------------------------------------------
+# Temporary build products
+
+GENERATED_MARKDOWN_LOCATION = $(TOP)/documentation/generated
```
anjohnson: I'm still not keen on this semi-installation directory path. Maybe using the `$(INSTALL_DOC)` path would be appropriate and wouldn't need as many additions to the configure files. Markdown is now a common documentation file format, all-be-it more normally used as a source format.

configure/RULES_BUILD
```diff
@@ -167,6 +167,8 @@ build: $(OBJSNAME) $(LIBTARGETS) $(PRODTARGETS) $(TESTPRODTARGETS) \
 inc: $(COMMON_INC) $(INSTALL_INC) $(INSTALL_CONFIGS) $(INSTALLS_CFG) \
 	$(INSTALL_HTMLS) $(INSTALLS_PERL_MODULES) $(INSTALL_SCRIPTS)
 
+doc: inc $(INSTALL_MARKDOWNS)
```
anjohnson: How much extra time does generating the `$(INSTALL_MARKDOWNS)` files add to the length of a build? If it isn't very much I'd be inclined to drop the new `doc` action type, move `$(INSTALL_DOCS)` from `buildInstall` to `inc` and use that to install the generated markdown files (you would need to add an `$(INSTALL_DOC)` rule from `$(COMMON_DIR)` though).

anjohnson (changes requested):
I need to stop now, @ttkorhonen I think I'm ready to discuss this.

ttkorhonen:
@minijackson , would you have some time to look at Andrew's comments? I would appreciate your input, I do not feel confident enough to answer them myself. At least without substantial work.

documentation/record-reference.md
ttkorhonen: This can of course be auto-generated, but would need some scripting to glue parts together. Markdown does not support including other files, at least trivially. I would be happy if auto-generation could be considered as a future enhancement.

documentation/Makefile
ttkorhonen: I agree, but it is hard to get everything perfect in one go. This is a big and complicated PR, but at the same time difficult to split up in a meaningful way. I fully agree that we should not put in breaking changes or things that are difficult to revert. But this one seems to me like one that can be incrementally improved - unless I am missing something.
I would be happy to work on improvements as soon as we get this basic infrastructure in. Otherwise I start slowly to lose the hope that this will ever happen.

ralphlange: +1 for gradual improvement

minijackson:
@ttkorhonen & @anjohnson I'm on vacation right now, so I won't be able to do any work this week. I can answer some questions here and there, but don't expect much

ttkorhonen:
Fair enough, @minijackson . Enjoy your vacation, and no stress!

configure/RULES_BUILD
ttkorhonen: I am a bit rusty on the details, but unless I am missing some details, this rule is about markdown files that are auto-generated from .dbd.pod during the build. Those markdown files are of interest to the RTD build only, and when RTD build is done, they are scrapped - as they should IMHO. Installation elsewhere "permanently" does not add any value, if that is what is meant with this change.
@minijackson should correct if I got that wrong.

anjohnson: Hi @ttkorhonen I believe you're right, but what I'm suggesting is that those files could be useful for developers too. It wouldn't add much time to the build if we were to run the pod-to-markdown conversion from the existing `make inc` target, and install the markdown files into the existing `$(TOP)/doc` where they're readable as a text file for anyone who doesn't want to open a browser to read the HTML versions.

I'm trying to avoid adding the additional `doc` target and the new install rules and directory by reusing an existing target, set of rules and install directory for these generated files. I have this branch checked out, if I can find enough time on Thursday or Friday this week (maybe on the plane home Friday pm) I will try making the changes as I see them.

ttkorhonen: OK, that is of course fair enough. My only worry is that we try to achieve too much in one go, as I already wrote in other comments.

Makefile
minijackson: I don't think this rule can be moved to `documentation/Makefile`, unfortunately. The RTD build needs the header files and Markdown files generated from `.pod` files. I've tried to put as much as possible in `documentation/Makefile` and related.

documentation/Makefile
minijackson: I'm also in favor of leaving it for a later PR. I don't have a lot of time these days, so this would delay the PR further. Listing all header files is also in line with the Sphinx philosophy, IMHO, where we already have to list every document to include.

configure/CONFIG_COMMON
minijackson: That would mean quite a bit refactor, right now. These Markdown file are installed in this private directory because they need to be included during the RTD build, whose root is `$(TOP)/documentation`. I won't have time to rethink that architecture in the following months, unfortunately…

configure/RULES_BUILD
minijackson: > I'm trying to avoid adding the additional doc target and the new install rules and directory by reusing an existing target, set of rules and install directory for these generated files. I have this branch checked out, if I can find enough time on Thursday or Friday this week (maybe on the plane home Friday pm) I will try making the changes as I see them.

Do you want to do these changes in this PR or in a follow-up PR?

Note: if you try to install the markdown files in `$(TOP)/doc` only, they won't be included in the RTD build. IIRC, RTD refuses to include external files, i.e. files that are not in its root directory, which is `$(TOP)/documentation` in this case.

anjohnson (approved):
I made various changes to this PR:
- Rebased onto the current 7.0 branch.
- Markdown files are added to the `DOCS` variable and get installed into the `$(INSTALL_LOCATION)/docs/` directory.
- Markdown generation and installation happens in the `make inc` build step.
- The Perl `Pod::Markdown` module and its dependency are now included in the Base sources, so no additional Perl modules are needed.
- I added Timo's `EPICS::PodMD.pm` module which resolves the issues with POD links.
- The documentation directory's `make commit` target and `commit-gh.sh` script were removed.
- I added some text updates to the `README.md` file.
- `ComponentReference.pod` was moved into `modules/database/src/std`.
- The scripts that generated the `.rst` files have been replaced with templates: 
  - Each API now has its own template file, which allows adding an intro/description of those headers.
  - The Generic C/C++ APIs have been split into separate templates for the libcom and database headers.
  - The files generated for each header are now named `_h.rst` so the record and menu files don't clash with the `.md` files generated from their `.dbd.pod` sources.
- Sphinx now runs in the `documentation/O.<host>` directory.
  - To build it manually, run `make inc` (or just `make` which takes longer) from the Base top directory, then `cd documentation` and `make rtd`, or from the Base top directory `make -C documentation rtd`.
  - To use a non-system Python that has the Sphinx modules installed, add a `configure/CONFIG_SITE.local` file and set the `PYTHON` variable there.
  - The `rtd` target runs `doxygen` and generates all the `.rst` files, then copies all the necessary source files into an `rtd-src` subdirectory for Sphinx to use.

I have no idea whether RTD needs to be told that the Sphinx output has moved, that location doesn't appear in the `.readthedocs.yaml` file so maybe it finds it automatically.

minijackson:
Thanks!

> I have no idea whether RTD needs to be told that the Sphinx output has moved, that location doesn't appear in the `.readthedocs.yaml` file so maybe it finds it automatically.

It doesn't, unfortunately. With the current `.readthedocs.yaml` and the line `configuration: documentation/conf.py`, we're specifying that the RTD server will run its own builtin Sphinx build.

If we want the build to be run by make, and bypass the builtin Sphinx build, the `.readthedocs.yaml` file needs to be partially rewritten: https://docs.readthedocs.io/en/stable/build-customization.html

But completely overriding the build process comes with a big caveat, we won't be able to use `apt_packages` (see https://docs.readthedocs.io/en/stable/config-file/v2.html#build-apt-packages).

From the top of my head, in the `.readthedocs.yaml` file, I think either Make does the RTD build, but with some hackiness to satisfy the build process, or we only use Make to create a Sphinx source directory that satisfies the builtin RTD build process.

I have setup a RTD build on this branch to test it, you can see it here: https://readthedocs.org/projects/epics-base-dbd-test/builds/
It currently fails because it tries to build the documentation from the not-generated source directory.

ttkorhonen:
I have not had a lot of time to look at this yet, but one comment nevertheless:

> From the top of my head, in the `.readthedocs.yaml` file, I think either Make does the RTD build, but with some hackiness to satisfy the build process, or we only use Make to create a Sphinx source directory that satisfies the builtin RTD build process.

The latter option (use builtin RTD process) was what I originally had in mind. I did look at build customization at one stage but it creates one more thing to maintain and comes with the caveat that @minijackson mentioned.

> I have setup a RTD build on this branch to test it, you can see it here: https://readthedocs.org/projects/epics-base-dbd-test/builds/ It currently fails because it tries to build the documentation from the not-generated source directory.

Thanks, I will take a look as soon as I can.

anjohnson:
I'm fine with letting RTD run Sphinx after our build process. Do we just need a build target in `documentation/Makefile` which runs `doxygen` and copies everything Sphinx needs into the `O.<host>/rtd-src` directory? How do we tell it where those sources are, or do we need to copy them to some standard location?

minijackson:
@anjohnson I haven't tried it, but I think setting this should work:

```yaml
sphinx:
  configuration: documentation/O.linux-x86_64/rtd-src/conf.py
```

ttkorhonen:
> @anjohnson I haven't tried it, but I think setting this should work:
> 
> ```yaml
> sphinx:
>   configuration: documentation/O.linux-x86_64/rtd-src/conf.py
> ```

Apart from me being not too enthusiastic about tying this to a specific architecture (linux-x86_64; I know that this will exist for a long time, but it just seems wrong), isn't this a post-build directory, and conf.py should be available at build start?
I may be wrong though. I try to find some time today to check this.

ttkorhonen:
> Apart from me being not too enthusiastic about tying this to a specific architecture (linux-x86_64; I know that this will exist for a long time, but it just seems wrong), isn't this a post-build directory, and conf.py should be available at build start? I may be wrong though. I try to find some time today to check this.

I was indeed wrong, I think what @minijackson proposed should work. (This should go into .readthedocs.yml, to be clear.)
 I suggest we try this out and see how it works. Maybe we can find a nicer-looking setup later.

(Of course, when I did a local build on my Mac, the rtd files land into the O.darwin86 directory; M1 chip but running under Rosetta.)

ralphlange:
Yes. should be done on `HOST_ARCH` only, I agree.

anjohnson:
If we want to avoid putting the host-arch in the conf.py file I could move `rtd-src` into the `O.Common` directory, which would make that path independent of the host-arch.

ttkorhonen:
> If we want to avoid putting the host-arch in the conf.py file I could move `rtd-src` into the `O.Common` directory, which would make that path independent of the host-arch.

I would prefer that (O.Common). Host-arch would need to be in the .readthedocs.yml file. To get that processed before the build starts would be a bit tricky to do. But using O.Common is rather straightforward, and unless the build system drastically changes, should work without issues.
 
We need two modifications here: move the rtd-src to O.Common and add that path in the .readthedocs.yml file:

```
sphinx:
  configuration: documentation/O.Common/rtd-src/conf.py
```

anjohnson:
I just pushed the changes to move `rtd-src` into the `O.Common` directory.
I think an RTD build of this commit should now succeed.

For manual runs, a new build target `make sphinx` in the `documentation` directory runs all the build steps including Sphinx, and installs the output into `$(INSTALL_HTML)/readthedocs/`.

Any reason to not change the first RTD `build: jobs: pre_build` command in the .yaml file to `make -sj inc` to parallelize that part?

Note: The `templates_path` setting in conf.py is for Sphinx theme templates, not C++ ones.

I'm seeing a load of warnings from Sphinx, some of which seem to be systematic implying it isn't parsing our headers properly. It would be nice to clean them up if we can, but that can happen after this PR is merged.

modules/database/src/std/ComponentReference.pod
minijackson, line 1: I think this file should be rewritten in Markdown, maybe into several files.

"EPICS Component Reference Manual" isn't very telling, and "IOC Component Reference" as already the title in the sidebar.

This file links to other references, but using "raw" links instead of toctrees, which makes Sphinx output a lot of warnings, and make the page not appear in the sidebar.

I also think that the record reference is the most important piece of EPICS documentation, so it should appear at the top-level in the sidebar.

documentation/HEADER_h.rst
```diff
@@ -0,0 +1,4 @@
+@HEADER@.h
+=======================================
```
minijackson: This will almost always generate a warning with Sphinx, because it expects that the underline is of the same length as the title.

I think it's possible to bypass that by converting the file into markdown:

````markdown
# @HEADER@.h

```{doxygenfile} @HEADER@.h
```
````

**Review by minijackson — COMMENTED** [review]
The build pass :tada: 

- build logs: https://readthedocs.org/projects/epics-base-dbd-test/builds/25415766/
- rendered: https://epics-docs-mj-test.readthedocs.io/projects/epics-base/en/latest/index.html

File: modules/database/src/std/ComponentReference.pod
Comment (ttkorhonen) [reply modules/database/src/std/ComponentReference.pod:1]: I agree with all points.
Navigation is also rather cumbersome with this setup. If you open one record's reference, the menu collapses back again and you need several clicks to get back to the record list.
However, maybe we can do this in subsequent update steps, not to delay the merge.

Comment (minijackson) [reply modules/database/src/std/ComponentReference.pod:1]: I'm okay with doing this in a subsequent PR, but for documentation structure change, I think it's best to do it as soon as possible, to avoid breaking URLs

Comment (ttkorhonen) [reply modules/database/src/std/ComponentReference.pod:1]: Right, agreed. I guess I am just getting too impatient.

Comment (ralphlange) [reply modules/database/src/std/ComponentReference.pod:1]: > If you open one record's reference, the menu collapses back again and you need several clicks to get back to the record list.

Would that change to something more reasonable when rewriting to Markdown as @minijackson suggests?

Comment (ttkorhonen) [reply modules/database/src/std/ComponentReference.pod:1]: > > If you open one record's reference, the menu collapses back again and you need several clicks to get back to the record list.
> 
> Would that change to something more reasonable when rewriting to Markdown as @minijackson suggests?

Using toctrees, yes, I think it would.

File: documentation/HEADER_h.rst (hunk 28 above)
Comment (anjohnson) [reply documentation/HEADER_h.rst]: Fine by me. The documentation for RST said the underlines had to be "as least as long as the title", and I didn't know what the Markdown syntax should be if it was even possible.

Comment (anjohnson) [reply documentation/HEADER_h.rst]: I pushed the change to generate Markdown instead.

File: modules/database/src/std/ComponentReference.pod
Comment (anjohnson) [reply modules/database/src/std/ComponentReference.pod:1]: I agree about the cumbersome navigation that the POD version causes. I'm not sure how to do this conversion properly though, could someone with more experience handle that? The current build creates a Markdown version which can be used as a starting point, that file can be found in the `modules/database/src/std/O.<host>/` directory after running `make inc`.

I do want to use "IOC Component Reference" instead of the original "Record Reference" name for this document, although if there's a sub-document for just the Record types that part of it could re-use the older name.

While editing the file, the "Corrections and Updates" section at the bottom can be removed, as could the second sentence at the top and the one below that about external links.

Comment (minijackson) [reply modules/database/src/std/ComponentReference.pod:1]: I've pushed the markdown version of `ComponentReference.pod`, but left the old one as-is (see the commit message for more information). I've also pushed two minor warning fixes.

I personally think that in the future, it would be good to sunset the current `html` build product, in favor of only the Sphinx output, and we could remove the `ComponentReference.pod` then. This would make the documentation more unified, and would reduce maintenance costs.

On the topic of the "IOC Component Reference" title and navigation, I'll open a subsequent PR to show you what I mean. If that's okay, we can discuss it there.

**Comment by minijackson** [comment]
I've rebased the PR to fix a merge conflict

File: modules/database/src/std/ComponentReference.pod
Comment (anjohnson) [reply modules/database/src/std/ComponentReference.pod:1]: I deliberately moved the ComponentReference source into `modules/database/src/std` to put it closer to all the things it's documenting, and I want the Markdown copy to live in the same place. The build rules should install an existing `.md` file that they find in the source directory in preference to converting from the `.pod` version, so there shouldn't be a problem with that.

I am _not_ in favor of stopping building the HTML files. We shouldn't assume that our users will have always-on access to the Internet, and users shouldn't have to install Doxygen and/or Sphinx to get a readable local copy of the documentation. I have personally used these files on long airplane flights many times, and the local generation also makes it easier for Codeathon participants to check their edits to the document sources.

I do accept that having two ComponentReference index sources isn't good, but we should be able to get around that soon (after this PR) by generating the index at build-time; the `modules/database/src/std/Makefile` already has lists of the source files that the index links to.

Comment (minijackson) [reply modules/database/src/std/ComponentReference.pod:1]: I've moved the `ComponentReference.md` back into `modules/database/src/std`.

File: documentation/Makefile [hunk 29]
```diff
@@ -7,25 +7,168 @@ EXPAND_ME += EPICS_REVISION
...
+	@$(MKDIR) $(RTD_SRC)
 	$(DOXYGEN)
-	rsync -av $(TOP)/html/ html/
 
-.PHONY: doxygen
+rtd: doxygen $(API_RST_FILES) $(HEADER_MD_FILES)
+	rsync -av $(INSTALL_DOC)/ $(RTD_SRC)/
+	rsync -av $(HEADER_MD_FILES) $(RTD_SRC)/
+	rsync -av $(API_RST_FILES) $(RTD_SRC)/
+	rsync -av ../index.rst ../../modules/database/src/std/ComponentReference.md ../conf.py $(RTD_SRC)/
```
anjohnson: It should not be necessary to list the `ComponentReference.md` file here, and I especially don't want its relative path to appear. The top-level `make inc` step completes before `make rtd` runs here, and when `make inc` reached `modules/database/src/std` the file will have been installed into the `$(INSTALL_DOC)` directory, so the first `rsync` above should have already copied it into `$(RTD_SRC)`.

anjohnson (changes requested):
The one change in `documentation/Makefile` I mention below is the only issue I have left with this PR.

Thanks!

documentation/Makefile
minijackson: Done.

anjohnson (approved):

anjohnson:
Except I'm now getting an exception when running Sphinx locally (although I do have a later version of that installed):
```
Exception occurred:
  File "/Users/anj/.local/pipx/venvs/sphinx/lib/python3.12/site-packages/breathe/renderer/sphinxrenderer.py", line 1588, in visit_docsectN
    section += nodes.title(node.title, node.title)
                           ^^^^^^^^^^
AttributeError: 'docSect2TypeSub' object has no attribute 'title'
The full traceback has been saved in /var/folders/pw/1nmxkq190q760fflx9cg8m9h0000gr/T/sphinx-err-bzyiksac.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
gmake[2]: *** [../Makefile:170: sphinx] Error 2
gmake[1]: *** [Makefile:8: sphinx] Error 2
gmake: *** [Makefile:156: sphinx] Error 2
```

minijackson:
@anjohnson which version of sphinx / breathe do you have? Did you install the versions as specified in `documentation/requirements.txt`?

Do you consider this blocking for this PR?

I can't reproduce your issue with this setup:

```bash
virtualenv venv
source venv/bin/activate
pip install -r documentation/requirements.txt
make inc
make -C documentation sphinx
```

The [ReadTheDocs build](https://readthedocs.org/projects/epics-base-dbd-test/builds/25520442/) also doesn't seem to have an issue.

anjohnson:
@minijackson No I did not install the versions specified, so I don't regard my errors as a blocker at all (I only remarked because I didn't get errors before your last push).

I'm running on a Mac and use Homebrew to install open-source packages; it had Python 3.12 installed. Somehow (pipx?) I created a venv for my Sphinx installation, and installed the Python packages into that. From the `lib/python3.12/site-packages` directory in that venv it apparently has sphinx-7.4.7, myst_parser-3.0.1, breathe-4.35.0, sphinx_rtd_theme-2.0.0 etc.

If would probably be a good idea to add some documentation explaining how to set up an environment for running Sphinx. Your setup instructions above might work, although I haven't tried them. Will they run on MacOS and/or Windows? I don't think they need to, but should say what OS(s) you expect them to work. Maybe that should be included in a Release Notes entry describing the Sphinx addition?

ttkorhonen:
Is there still something to be done for this PR before merging?

minijackson:
@ttkorhonen nothing that I'm aware of.

@anjohnson:

> If would probably be a good idea to add some documentation explaining how to set up an environment for running Sphinx. Your setup instructions above might work, although I haven't tried them. Will they run on MacOS and/or Windows? I don't think they need to, but should say what OS(s) you expect them to work. Maybe that should be included in a Release Notes entry describing the Sphinx addition?

I'll do it on another PR, then. The commands that I commented earlier should work on MacOS and Linux. I'm not completely sure about Windows, but mostly because I have no experience with using Python on it.

dirk-zimoch:
This PR (I think) introduced bug #583. Can you please have a look and fix it?

anjohnson:
Since @dirk-zimoch raised this PR again, I need a Release Notes entry for the new RTD additions to Base. I can write something myself, but if someone closer to the Sphinx work would be willing to do that I would be grateful. Posting it as a comment in this PR would be sufficient instead of creating a separate PR if you like.

minijackson:
Sorry for the delay and the breakage… It seems I missed the train for writing the release notes…

On the Dirk Zimoch issue, I have no idea where I can help. Is it still relevant?

ttkorhonen:
> Sorry for the delay and the breakage… It seems I missed the train for writing the release notes…
> 
> On the Dirk Zimoch issue, I have no idea where I can help. Is it still relevant?

It seems still to be relevant, see issue #583. I have not managed to look into the issue in detail, your help would be greatly appreciated, @minijackson.
The issue report says: _make $EPICS_HOST_ARCH is broken since commit https://github.com/epics-base/epics-base/commit/615ba73044dd5485d7dea21d4f0df645fea3f008._ 
This commit included the RTD documentation build.

PR #424, opened 2023-09-14
Fix db event double cancel

### LINKED ISSUE #423: dbEvent: double cancel causes hang
A logic hole in dbEvent.c can combine with a race between concurrent `db_cancel_event()` and an event callback to cause `db_cancel_event()` to hang.  The following assertion will occasionally fail if a thread calls `db_cancel_event()` on two different subscriptions.  Not a double free, two separate free's of subscriptions which share the same dbEventCtx.

While I am not sure I can completely explain the sequence of events, I think this issue arises from an attempt to handle self-cancel and concurrent cancel with the same logic.  One fix I will explore is to move self-cancel detection from the common `struct event_user` to the individual `struct evSubscrip`.

```diff
diff --git a/modules/database/src/ioc/db/dbEvent.c b/modules/database/src/ioc/db/dbEvent.c
index 6dab45104..7cbbda91b 100644
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -620,6 +620,7 @@ void db_cancel_event (dbEventSubscription event)
     assert ( pevent->npend == 0u );
 
     if ( pevent->ev_que->evUser->taskid == epicsThreadGetIdSelf() ) {
+        assert(!pevent->ev_que->evUser->pSuicideEvent);
         pevent->ev_que->evUser->pSuicideEvent = pevent;
     }
     else {
```

```
A call to 'assert(!pevent->ev_que->evUser->pSuicideEvent)'
    by thread 'qsrvSingle' failed in ../db/dbEvent.c line 623.
Dumping a stack trace of thread 'qsrvSingle':
[    0x7f5dcc2a352a]: /home/dev/epics/pvxs/test/O.linux-x86_64-debug/../../../base-git/lib/linux-x86_64-debug/libCom.so.3.22.1(epicsBackTrace+0x20)
[    0x7f5dcc2a33ac]: /home/dev/epics/pvxs/test/O.linux-x86_64-debug/../../../base-git/lib/linux-x86_64-debug/libCom.so.3.22.1(epicsStackTrace+0x6f)
[    0x7f5dcc299d21]: /home/dev/epics/pvxs/test/O.linux-x86_64-debug/../../../base-git/lib/linux-x86_64-debug/libCom.so.3.22.1(epicsAssert+0x5e)
[    0x7f5dcc31dc6b]: /home/dev/epics/pvxs/test/O.linux-x86_64-debug/../../../base-git/lib/linux-x86_64-debug/libdbCore.so.3.22.1(db_cancel_event+0x1fb)
[    0x7f5dcc6882a1]: /home/dev/epics/pvxs/test/O.linux-x86_64-debug/../../lib/linux-x86_64-debug/libpvxsIoc.so.1.2(_ZZN4pvxs3ioc12Subscription9subscribeEPvRKNS0_7ChannelEPFvS2_P9dbChanneliP12db_field_logES2_jENUlS2_E_clES2_+0x2b)
```

mdavidsaver:
Attempt to fix #423.

Allow for multiple db_event_cancel() (concurrent or self-cancel) prior to event_task wakeup.

In db_event_cancel(), immediate free() only if idle (not queued or in progress).  Otherwise, defer free() to event task.  Avoids need to immediately expunge canceled event from queue.  Allow event task to process canceled events as normal (except no user_sub) until npend==0.

### DIFF
--- a/modules/database/src/ioc/db/dbChannel.h
+++ b/modules/database/src/ioc/db/dbChannel.h
@@ -41,7 +41,12 @@ extern "C" {
 /**
  * event subscription
  */
-typedef struct evSubscrip {
+struct evSubscrip;
+
+typedef struct evSubscrip evSubscrip;
+
+#ifdef EPICS_PRIVATE_API
+struct evSubscrip {
     ELLNODE             node;
     struct dbChannel  * chan;
     EVENTFUNC         * user_sub;
@@ -54,7 +59,8 @@ typedef struct evSubscrip {
     char                useValque;
     char                callBackInProgress;
     char                enabled;
-} evSubscrip;
+};
+#endif
 
 typedef struct chFilter chFilter;
 
--- a/modules/database/src/ioc/db/dbChannel.h
+++ b/modules/database/src/ioc/db/dbChannel.h
@@ -49,15 +49,24 @@ typedef struct evSubscrip evSubscrip;
 struct evSubscrip {
     ELLNODE             node;
     struct dbChannel  * chan;
+    /* user_sub==NULL used to indicate db_cancel_event() */
     EVENTFUNC         * user_sub;
     void              * user_arg;
+    /* associated queue, may be shared with other evSubscrip */
     struct event_que  * ev_que;
+    /* NULL if !npend.  if npend!=0, pointer to last event added to event_que::valque */
     db_field_log     ** pLastLog;
-    unsigned long       npend;      /**< n times this event is on the queue */
-    unsigned long       nreplace;   /**< n times replacing event on the queue */
+    /* n times this event is on the queue */
+    unsigned long       npend;
+    /* n times replacing event on the queue */
+    unsigned long       nreplace;
+    /* DBE mask */
     unsigned char       select;
+    /* if set, subscription will yield dbfl_type_val */
     char                useValque;
+    /* event_task is handling this subscription */
     char                callBackInProgress;
+    /* this node added to dbCommon::mlis */
     char                enabled;
 };
 #endif
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -76,7 +76,6 @@ struct event_que {
     unsigned short          getix;
     unsigned short          quota;          /* the number of assigned entries*/
     unsigned short          nDuplicates;    /* N events duplicated on this q */
-    unsigned short          nCanceled;      /* the number of canceled entries */
     unsigned                possibleStall;
 };
 
@@ -92,7 +91,7 @@ struct event_user {
     void                *extralabor_arg;/* parameter to above */
 
     epicsThreadId       taskid;         /* event handler task id */
-    struct evSubscrip   *pSuicideEvent; /* event that is deleting itself */
+    epicsUInt32         pflush_seq;     /* worker cycle count for synchronization */
     unsigned            queovr;         /* event que overflow count */
     unsigned char       pendexit;       /* exit pend task */
     unsigned char       extra_labor;    /* if set call extra labor func */
@@ -123,10 +122,9 @@ static void *dbevFieldLogFreeList;
 
 static char *EVENT_PEND_NAME = "eventTask";
 
-static struct evSubscrip canceledEvent;
-
 static epicsMutexId stopSync;
 
+/* unused space in queue (EVENTQUESIZE when empty) */
 static unsigned short ringSpace ( const struct event_que *pevq )
 {
     if ( pevq->evque[pevq->putix] == EVENTQEMPTY ) {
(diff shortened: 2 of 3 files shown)


ralphlange:
I like the approach: indicate instead of screw-around-with. Keeping a single consumer.

mdavidsaver:
we have realized that `db_event_cancel()` still needs to wait for an in-progress callback to complete.

modules/database/src/ioc/db/dbEvent.c
```diff
@@ -1016,25 +973,15 @@ static int event_read ( struct event_que *ev_que )
...
-            }
-            else {
-                pevent->callBackInProgress = FALSE;
-                if ( pevent->user_sub==NULL && pevent->npend==0u ) {
-                    epicsEventSignal ( ev_que->evUser->pflush_sem );
-                }
-            }
+            pevent->callBackInProgress = FALSE;
+        }
+        /* callback may have called db_cancel_event(), so much check user_sub again */
```
anjohnson: typo `s/much/must/`

modules/database/src/ioc/db/dbChannel.h
```diff
@@ -41,20 +41,30 @@ extern "C" {
...
+    /* user_sub==NULL used to indicate db_cancel_event() */
     EVENTFUNC         * user_sub;
     void              * user_arg;
+    /* associated queue, may be shared with other evSubscrip */
     struct event_que  * ev_que;
+    /* NULL if !npend.  if npend!=0, pointer to last event added to event_que::valque */
     db_field_log     ** pLastLog;
     unsigned long       npend;      /**< n times this event is on the queue */
     unsigned long       nreplace;   /**< n times replacing event on the queue */
+    /* DBE mask */
```
anjohnson: Could these new comments be formatted as Doxygen annotations please? I see you've taken the structure definition private so it won't now appear in future output, but other data members are commented that way already and the `/**<` prefix makes it easier to see immediately which field a comment applies to. Suggestions for shorter wording:
```C
    EVENTFUNC         * user_sub;   /**< NULL for db_cancel_event() */

    struct event_que  * ev_que;     /**< May be shared with other evSubscrip */

    db_field_log     ** pLastLog;   /**< NULL or ptr to last event on ev_que->valque */

    unsigned char       select;     /**< DBE mask */
```

anjohnson (commented):
I agree with Ralph, this approach looks better.

A couple of minor points about commenting.

modules/database/src/ioc/db/dbChannel.h
mdavidsaver: If docygen markup is a question of all or nothing, then I am inclined towards nothing.  I could also move this definition to some *Pvt.h if you would prefer.

anjohnson: Since they won't get processed they don't need to be Doxygen markup as such, it's just harder to read data member comments that are above or below the member instead of on the same line, my brain always goes "is this comment about the one above or below?".

mdavidsaver: Ok, I understand a desire for consistency.  I have move the remaining suffix comments to the previous line.

mdavidsaver:
> we have realized that `db_event_cancel()` still needs to wait for an in-progress callback to complete.

I think this sync. issue has been resolved.  So this change should be ready to go.

Another change I have made is to how this sync. is done.  In both cases, the epicsEvent used is in the `event_user` (aka `dbEventCtx`).  However, previously the `evSubscript::callBackInProgress` flag was tested under the `event_que::writelock`.  Now, a new `event_user::pflush_seq` sequence counter is tested under `event_user::lock`.  So all parts of `event_user`.

The `pflush_seq` counter is incremented at the end of the main loop in `event_task()`, instead of individually at the end of `event_read()`.  This will be a little later, but I don't think this will have practical consequence.

Also, with `event_user::pflush_seq` I am "cheating" a little when `db_cancel_event()` re-broadcasts sequence counter changes.  I think the only effect is that the next call to `db_cancel_event()` will take an extra loop through as the first `epicsEventWait()` will return immediately.  This arises because I am not tracking the number of waiters, and re-broadcasting only when there is more than one concurrent `db_cancel_event()`.  This condition would be quite hard to test, so I think a slight speedup in what should be an uncommon case does not justify the added code complexity.

(imo. the proper way to do this would be an OSI abstraction equivalent to `pthread_cond_t`)

anjohnson:
Core Group: Approved.

mdavidsaver:
This morning I didn't think that [lp:541371](https://bugs.launchpad.net/bugs/541371) would be affected by this change.  Now I am not so sure.

https://github.com/epics-base/epics-base/blob/5aca4c684cc87158ded4c4d4c3303b4c018e43fa/modules/database/src/ioc/db/dbEvent.c#L814-L816
https://github.com/epics-base/epics-base/blob/5aca4c684cc87158ded4c4d4c3303b4c018e43fa/modules/database/src/ioc/db/dbEvent.c#L835-L836

Could that assert be tripped if `npend==0` and the buffer is full of "cancelled" events?

I think the assumption behind the reservation logic (4 events per subscriber * 36 subscribes per ev_que) is violated between a call to `db_cancel_event()` when `quota` is decremented, and `event_task` cleaning out the cancelled events to make that capacity available.

If so, this PR should fix lp:541371 as well.

mdavidsaver:
Applied as fab8fd7102143a661a81cf7163dd889601eb015c

PR #425, opened 2023-09-23
Modify libca and libcas to address problem with CA Nameserver and CA_V413 protocol

bsbevins:
Modifications to ca/client and ca/legacy/pcas to support handling CA protocol minor version numbers in CA nameservers.

### DIFF
--- a/src/ca/client/cacChannel.cpp
+++ b/src/ca/client/cacChannel.cpp
@@ -130,6 +130,13 @@ unsigned cacChannel::getHostName (
     return 0u;
 }
 
+unsigned cacChannel::getHostMinorProtocol ( 
+    epicsGuard < epicsMutex > &) const throw ()
+{
+    epicsThreadOnce ( & cacChannelIdOnce, cacChannelSetup, 0);
+    return 99u;
+}
+
 // the default is to assume that it is a locally hosted channel
 const char * cacChannel::pHostName (
     epicsGuard < epicsMutex > & ) const throw ()
--- a/src/ca/client/cacChannel.cpp
+++ b/src/ca/client/cacChannel.cpp
@@ -134,7 +134,7 @@ unsigned cacChannel::getHostMinorProtocol (
     epicsGuard < epicsMutex > &) const throw ()
 {
     epicsThreadOnce ( & cacChannelIdOnce, cacChannelSetup, 0);
-    return 99u;
+    return 0u;
 }
 
 // the default is to assume that it is a locally hosted channel
--- a/src/ca/client/netiiu.cpp
+++ b/src/ca/client/netiiu.cpp
@@ -116,6 +116,12 @@ const char * netiiu::pHostName (
     return pHostNameNetIIU;
 }
 
+unsigned netiiu::getHostMinorProtocol ( 
+    epicsGuard < epicsMutex > & ) const throw ()
+{
+    return 0u;
+}
+
 osiSockAddr netiiu::getNetworkAddress (
     epicsGuard < epicsMutex > & ) const
 {
--- a/src/ca/legacy/pcas/generic/casDGClient.cc
+++ b/src/ca/legacy/pcas/generic/casDGClient.cc
@@ -332,7 +332,12 @@ caStatus casDGClient::searchResponse ( const caHdrLargeArray & msg,
     //
     if ( status == S_cas_success ) {
         AlignedWireRef < epicsUInt16 > tmp ( *pMinorVersion );
-        tmp = CA_MINOR_PROTOCOL_REVISION;
+        if (retVal.getMinorProtocol()) {
+          tmp = retVal.getMinorProtocol();
+        }
+        else {
+          tmp = CA_MINOR_PROTOCOL_REVISION;
+        }
         this->out.commitMsg ();
     }
     
--- a/src/ca/legacy/pcas/generic/casDGClient.cc
+++ b/src/ca/legacy/pcas/generic/casDGClient.cc
@@ -332,10 +332,8 @@ caStatus casDGClient::searchResponse ( const caHdrLargeArray & msg,
     //
     if ( status == S_cas_success ) {
         AlignedWireRef < epicsUInt16 > tmp ( *pMinorVersion );
-        if (retVal.getMinorProtocol()) {
-          tmp = retVal.getMinorProtocol();
-        }
-        else {
+        tmp = retVal.getMinorProtocol();
+        if ( tmp < 1 ) {
           tmp = CA_MINOR_PROTOCOL_REVISION;
         }
         this->out.commitMsg ();
--- a/src/ca/legacy/pcas/generic/casDGClient.cc
+++ b/src/ca/legacy/pcas/generic/casDGClient.cc
@@ -333,7 +333,7 @@ caStatus casDGClient::searchResponse ( const caHdrLargeArray & msg,
     if ( status == S_cas_success ) {
         AlignedWireRef < epicsUInt16 > tmp ( *pMinorVersion );
         tmp = retVal.getMinorProtocol();
-        if ( tmp < 1 ) {
+        if ( ( tmp < 1 ) || ( tmp > CA_MINOR_PROTOCOL_REVISION ) ) {
           tmp = CA_MINOR_PROTOCOL_REVISION;
         }
         this->out.commitMsg ();
--- a/src/ca/legacy/pcas/generic/casdef.h
+++ b/src/ca/legacy/pcas/generic/casdef.h
@@ -134,6 +134,8 @@ private:
     ca_uint16_t minorProtocolVersion;
 };
 
+#define HAS_CAS_RETURN_MINOR_PROTOCOL
+
 //
 // pvAttachReturn
 //
--- a/src/ca/legacy/pcas/generic/casdef.h
+++ b/src/ca/legacy/pcas/generic/casdef.h
@@ -130,8 +130,8 @@ public:
     unsigned getMinorProtocol() const;
 private:
     caNetAddr address;
-    ca_uint16_t minorProtocolVersion;
     pvExistReturnEnum status;
+    ca_uint16_t minorProtocolVersion;
 };
 
 //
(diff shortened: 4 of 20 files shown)


mdavidsaver:
From packet captures provided by @bsbevins via. private mail, this CA name server is returning a reply with a wildly invalid version number of 0x7ffd.

```
Channel Access
    Command: Search (0x0006)
    Payload Size: 8
    Server Port: 5064
    Server IP: XXX.XXX.XXX.XXX
    Client Channel ID: 1
    Version: 32765
```

Where is that number coming from?

mdavidsaver:
Also, some windows builds fail with:

```
cl -EHsc -GR               -nologo -FC -D__STDC__=0 -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE   -Ox -GL -Oy-   -W3 -w44355 -w44344 -w44251        -MD -DEPICS_BUILD_DLL -DEPICS_CALL_DLL -TP  -I. -I../O.Common -I. -I. -I.. -I../../../../src/ca/client/test -I../../../../include/compiler/msvc -I../../../../include/os/WIN32 -I../../../../include         -c ../access.cpp
access.cpp
C:\projects\epics-base\src\ca\client\oldAccess.h(64): error C2373: 'ca_host_minor_protocol': redefinition; different type modifiers
C:\projects\epics-base\src\ca\client\cadef.h(663): note: see declaration of 'ca_host_minor_protocol'
```

src/ca/client/cacChannel.cpp
```diff
@@ -130,6 +130,13 @@ unsigned cacChannel::getHostName (
     return 0u;
 }
 
+unsigned cacChannel::getHostMinorProtocol ( 
+    epicsGuard < epicsMutex > &) const throw ()
+{
+    epicsThreadOnce ( & cacChannelIdOnce, cacChannelSetup, 0);
+    return 99u;
```
mdavidsaver: What is the significance of `99`???

src/ca/client/netiiu.cpp
```diff
@@ -116,6 +116,12 @@ const char * netiiu::pHostName (
     return pHostNameNetIIU;
 }
 
+unsigned netiiu::getHostMinorProtocol ( 
+    epicsGuard < epicsMutex > & ) const throw ()
+{
+    return 0u;
+}
```
mdavidsaver, line 123: Why implement this pure virtual method?

If this is meant to be a default, wouldn't `CA_MINOR_PROTOCOL_REVISION` be more appropriate?  Or `cantProceed()` if it is not meant to be called?

src/ca/legacy/pcas/generic/casDGClient.cc
```diff
@@ -332,7 +332,12 @@ caStatus casDGClient::searchResponse ( const caHdrLargeArray & msg,
     //
     if ( status == S_cas_success ) {
         AlignedWireRef < epicsUInt16 > tmp ( *pMinorVersion );
-        tmp = CA_MINOR_PROTOCOL_REVISION;
+        if (retVal.getMinorProtocol()) {
+          tmp = retVal.getMinorProtocol();
+        }
+        else {
+          tmp = CA_MINOR_PROTOCOL_REVISION;
+        }
```
mdavidsaver, line 338: The value returned bt `getMinorProtocol()` needs to be bounds checked to the range `[1, CA_MINOR_PROTOCOL_REVISION]` inclusive.

Also, it would be nice to avoid calling `getMinorProtocol()` twice.

src/ca/legacy/pcas/generic/casdef.h
```diff
@@ -119,14 +120,17 @@ class epicsShareClass pvExistReturn {
     pvExistReturn ( pvExistReturnEnum s = pverDoesNotExistHere );
     // directory service server tools will use this (see caNetAddr.h)
     pvExistReturn ( const caNetAddr & );
+    pvExistReturn ( const caNetAddr &, ca_uint16_t );
```
mdavidsaver, line 123: If this API change is to be introduce into PCAS in base 3.15, as opposed to the [current home for PCAS](https://github.com/epics-modules/pcas), then some kind of ID macro is necessary.

src/ca/legacy/pcas/generic/pvExistReturn.cc
```diff
@@ -30,13 +34,15 @@ const pvExistReturn & pvExistReturn::operator = ( pvExistReturnEnum rhs )
 {
 	this->status = rhs;
 	this->address.clear ();
+   this->minorProtocolVersion= 0;
```
mdavidsaver: Please use consistent indentation in this file.

mdavidsaver (changes requested):

mdavidsaver:
@anjohnson @ralphlange Should the version of PCAS w/ base 3.15 be getting new API (or changing at all)?

@bsbevins Certainly https://github.com/epics-modules/pcas needs the same change.

mdavidsaver:
I have opened an issue #426

src/ca/client/netiiu.cpp
bsbevins: I coded this method exactly parallel to how getHostName() and getNetworkAddress() are done. They are all declared pure virtual, but then defined.

src/ca/client/cacChannel.cpp
bsbevins: I will fix this. 99 was a debug value. I will change this to 0u to indicate an error.

src/ca/legacy/pcas/generic/casDGClient.cc
bsbevins: I will fix calling twice and check for values <1. I'm not sure I agree about capping the value at CA_MINOR_PROTOCOL_REVISION. The whole point of this change is to allow a PCAS (like a name server) to return a different server's version. The specific problem being addressed is when the "real" server has a version lower than the nameserver, but in the future the real server could also have a higher version than the PCAS.

src/ca/legacy/pcas/generic/casdef.h
bsbevins: I will be happy to do this if you can point me at an example of what you're looking for.

src/ca/legacy/pcas/generic/pvExistReturn.cc
bsbevins: Will fix.

bsbevins:
> ❌ [Build EPICS Base 3.15 base-3.15-102 failed](https://ci.appveyor.com/project/epics-base-3/epics-base/builds/49304123) (commit [e384aa394b](https://github.com/epics-base/epics-base/commit/e384aa394b) by @)

The build failures all complain of a missing "distutils" module. That doesn't sound like it has anything to do with my code. Perhaps there is a configuration issue with the test setup?

anjohnson:
Hi Brian, yes don't worry about the missing "distutils" build errors. I just triggered the GitHub CI builds as well, the MacOS clang-12 failure there isn't likely to be your code either although I haven't actually checked yet.
@ralphlange do we need to update the ci-scripts version for the 3.15 branch?

I think by an ID macro @mdavidsaver means that the header files that declare the modified APIs should have a new macro added to each, something like
```C
#define HAS_CA_HOST_MINOR_PROTOCOL
```
in cadef.h and
```C
#define HAS_CAS_RETURN_HOST_PROTOCOL
```
in casdef.h so that code which includes those headers can be built against both old and new Base versions.

ralphlange:
> @ralphlange do we need to update the ci-scripts version for the 3.15 branch?

I did that about an hour ago.

src/ca/legacy/pcas/generic/casdef.h
bsbevins: I added

#define HAS_CA_HOST_MINOR_PROTOCOL

in cadef.h and 

#define HAS_CAS_RETURN_MINOR_PROTOCOL

in cadef.h.

src/ca/legacy/pcas/generic/casDGClient.cc
mdavidsaver: > ... I'm not sure I agree about capping the value at CA_MINOR_PROTOCOL_REVISION.  ...

I insist.

Using the lesser of the two peer versions has been the basis for introducing backwards compatible protocol changes to CA (and PVA for that matter).  To my mind it is imperative that every peer report a minor version between `CA_MINIMUM_SUPPORTED_VERSION` and the smaller of `CA_MINOR_PROTOCOL_REVISION` or the peer reported version.  Zero could serve to indicate a disconnected channel as it was never a valid version.

src/ca/client/netiiu.cpp
mdavidsaver: I see an implementation for netiiu and all three of its sub-classes (tcpiiu, udpiiu, and noopiiu).  The noopiiu overload does nothing but call the base (netiiu) implementation.  So I think that either the netiiu or noopiiu definition could (and should) be removed.

mdavidsaver (changes requested):
Completing a review which I apparently started many months ago.  (Apparently github remembers comments indefinitely)

src/ca/legacy/pcas/generic/casDGClient.cc
bsbevins: I concede. The only practical downside is that your nameserver must always be built with a protocol version at least as high as the highest of your IOCs, or clients risk potential unnecessary downgrades. In practice, updating the nameserver should be a small burden.

src/ca/client/netiiu.cpp
bsbevins: I have made getHostMinorProtocol() non-virtual in netiiu and removed the implementation from noopiiu.

anjohnson:
Core Group: We think this is the best solution, but please split this PR; move the CA Client changes to a PR against the 7.0 branch of Base, and the PCAS changes to a PR in [that stand-alone project](https://github.com/epics-modules/pcas). We also look forward to seeing the necessary changes to the name-server project in a PR there.

bsbevins:
I will resubmit against 7.0 and PCAS as time permits.

anjohnson:
Merged #711 so this is now moot in this project.

PR #427, opened 2023-10-02
Rationalize osdMutex

mdavidsaver:
I see this mainly a "cleanup" change.  Although it does:

* Remove the (semi)public `epicsMutexOsd*()` functions.

My github/google searches do not turn up any external usage.

If there is significant concern, I could re-add these as wrappers around regular `epicsMutedId`, with some obnoxious deprecation warnings.

* Remove the free list in epicsMutex.cpp.

I don't think this special case free-list has much value for Linux, OSX, or Windows.  I could be persuaded to keep it for the RTOSs specifically.

* Eliminates the separate `epicsMutexOSD` struct.

Avoids allocation of a second struct on posix and windows.

* Consolidate OSD in private `epicsMutexImpl.h`.

Yes, I could have added 5 OSD headers instead of the `#ifdef`, but I don't think all of the extra boilerplate would clarify the situation.

* WIN32: eliminate pre-XP code

The current `_WIN32_WINNT < 0x0501` condition already prevents anyone from building with older windows.

* RTEMS-score: eliminate unused non-fast code.

Now that our use of RTEMS/score is solidly legacy, I don't think there is any benefit to preserving the untested non-fast version.  I didn't check if this "alternative" works currently.

* Eliminate unused `LOG_LAST_OWNER`.

Again, I didn't test to see if it works.

* vxworks...

untested.

### DIFF
--- a/modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
+++ b/modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
@@ -26,6 +26,7 @@
 
 #include "epicsEvent.h"
 #include "epicsThread.h"
+#include "rtemsNamePvt.h"
 #include "errlog.h"
 
 /* #define EPICS_RTEMS_SEMAPHORE_STATS */
@@ -47,12 +48,9 @@ epicsEventCreate(epicsEventInitialState initialState)
 {
     rtems_status_code sc;
     rtems_id sid;
-    rtems_interrupt_level level;
-    static char c1 = 'a';
-    static char c2 = 'a';
-    static char c3 = 'a';
+    static uint32_t name;
 
-    sc = rtems_semaphore_create (rtems_build_name ('B', c3, c2, c1),
+    sc = rtems_semaphore_create (next_rtems_name ('B', &name),
         initialState,
         RTEMS_FIFO | RTEMS_SIMPLE_BINARY_SEMAPHORE |
             RTEMS_NO_INHERIT_PRIORITY | RTEMS_NO_PRIORITY_CEILING | RTEMS_LOCAL,
@@ -62,26 +60,6 @@ epicsEventCreate(epicsEventInitialState initialState)
         errlogPrintf ("Can't create binary semaphore: %s\n", rtems_status_text (sc));
         return NULL;
     }
-    rtems_interrupt_disable (level);
-    if (c1 == 'z') {
-        if (c2 == 'z') {
-            if (c3 == 'z') {
-                c3 = 'a';
-            }
-            else {
-                c3++;
-            }
-            c2 = 'a';
-        }
-        else {
-            c2++;
-        }
-        c1 = 'a';
-    }
-    else {
-        c1++;
-    }
-    rtems_interrupt_enable (level);
     return (epicsEventId)sid;
 }
 
--- a/modules/libcom/src/osi/os/vxWorks/osdMutex.c
+++ b/modules/libcom/src/osi/os/vxWorks/osdMutex.c
@@ -13,6 +13,7 @@
 
 #include <vxWorks.h>
 #include <semLib.h>
+#include <errno.h>
 #include <time.h>
 #include <objLib.h>
 #include <sysLib.h>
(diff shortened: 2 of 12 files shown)


mdavidsaver:
Since this change is intended primarily as a cleanup, I don't see any urgency to merge it before the next (7.0.8?) release.

modules/libcom/src/osi/os/vxWorks/osdMutex.c
```diff
@@ -23,30 +24,47 @@ int sysClkRateGet(void);
...
+void epicsMutexOsdSetup(void)
+{
+    if(!epicsMutexGlobalLock.osd) {
+        epicsMutexOsdPrepare(&epicsMutexGlobalLock);
+    }
+}
+
+long epicsMutexOsdPrepare(struct epicsMutexParm *mutex)
+{
+    mutex->osd = semMCreate(SEM_DELETE_SAFE|SEM_INVERSION_SAFE|SEM_Q_PRIORITY));
```
dirk-zimoch: Syntax error: Double `)`

dirk-zimoch:
After fixing the above typo, I tested it on MVME 5100 running VxWorks 6.9.4 and found no problem. I stressed the system with command line caget in a loop and found no memoy fragmentation.

mdavidsaver:
> After fixing the above typo ... no problem

Thank you Dirk.

> ... found no memoy fragmentation.

I know this is a digression, but you have me by the curiosity.  How are you measuring fragmentation?

dirk-zimoch:
A quite simple approach to measure memory fragmentation is: 1-largest_free_memory_chunk/total_free_memory.

The IOC was not producing more free memory chunks and the largest one was not shrinking. It seems, it re-uses the same memory over and over, just as a good free-list implementation would do.

anjohnson:
@dirk-zimoch I believe you said recently that PSI upgraded their VxWorks IOCs to VxWorks 6.x. In the early version 6.x releases Wind River replaced their old first-fit memory allocator with a best-fit allocator, so where the previous one would fragment badly by breaking up the first block it found that was larger then the request, the new one behaves reasonably. I think they had to do that for C++ code using the standard C++ template library, which allocates a lot of small and short-lived memory blocks.

mdavidsaver:
Fixed the syntax error in vxWorks/osdMutex.c and rebased onto 7.0.8.

modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
```diff
@@ -48,11 +49,9 @@ epicsEventCreate(epicsEventInitialState initialState)
     rtems_status_code sc;
     rtems_id sid;
     rtems_interrupt_level level;
-    static char c1 = 'a';
-    static char c2 = 'a';
-    static char c3 = 'a';
+    static uint32_t name;
```
kiwichris, line 51: Is using static here thread safe?

modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
```diff
@@ -48,11 +49,9 @@ epicsEventCreate(epicsEventInitialState initialState)
     rtems_status_code sc;
     rtems_id sid;
     rtems_interrupt_level level;
```
kiwichris: Is this used?

kiwichris (commented):
I have not reviewed the RTEMS 4.* (`rtems-score`) changes in detail other than points raised.

modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
mdavidsaver: No.  Neither the new nor the current c1,c2,c3 implementation.

My understanding is that duplicate names would be confusing, but not a functional issue.  So I wasn't going to bother with the "old" RTEMS series.

modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
mdavidsaver: No, `level` is no longer used and should be removed.

modules/libcom/src/osi/os/RTEMS-score/osdEvent.c
mdavidsaver: Actually, scratch this.  Both implementations are guarded by `rtems_interrupt_enable()`.  This PR moves that logic into `next_rtems_name ()` to avoid duplication.

kiwichris: OK

anjohnson:
Core Group: Approved.

mdavidsaver:
Simple re-base to re-test

mdavidsaver:
Merged as e4ad4becde5ebe76b2251bef0fdfccd4fd1101ea

PR #430, opened 2023-10-05
iocsh: keep history file

mdavidsaver:
It turns out to be easy to have readline read/write a command history file like eg. `~/.bash_history`.

What should the default history file name be?

I have initially picked a default of `.iocsh_history` (aka in the current directory) as I think a per-IOC history makes the most sense.

This default can be overridden by setting `$HISTFILE` (like bash and others).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ O.*/
 *.log
 .*.swp
 .DS_Store
+.iocsh_history
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -16,6 +16,7 @@
 #include <vector>
 #include <map>
 #include <string>
+#include <sstream>
 
 #include <stddef.h>
 #include <string.h>
@@ -43,6 +44,7 @@
 
 #if EPICS_COMMANDLINE_LIBRARY == EPICS_COMMANDLINE_LIBRARY_READLINE
 #  include <readline/readline.h>
+#  include <readline/history.h>
 #  define USE_READLINE
 /* libedit also provides readline.h, but isn't fully compatible with
  * GNU readline.  It also doesn't specifically identify itself.
@@ -633,6 +635,15 @@ struct ReadlineContext {
             rl_completer_quote_characters = (char*)"\"";
             rl_attempted_completion_function = &iocsh_attempt_completion;
             rl_bind_key('\t', rl_complete);
+            compute_hist_file();
+            if(!hist_file.empty()) {
+                if(int err = read_history(hist_file.c_str())) {
+                    if(err!=ENOENT)
+                        fprintf(stderr, "Error %s (%d) loading '%s'\n",
+                                strerror(err), err, hist_file.c_str());
+                }
+                stifle_history(1024); // some limit...
+            }
         }
 #endif
         return context;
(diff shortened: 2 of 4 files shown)


ralphlange:
Question at the EPICS Collab Meeting: Will this support backward-search?

mdavidsaver:
readline-search

Apparently, yes.  `Ctrl+r` starts a reverse history search.

mdavidsaver:
Actually, the reverse history search has probably been available for some time.

mdavidsaver:
> This default can be overridden by setting `$HISTFILE` (like bash and others).

On further thought.  Maybe using the same env. name as bash is not a good idea as it could result in mixing bash and iocsh history in the same file.  I guess `$EPICS_IOCSH_HISTFILE` would be consistent with our usually verbose names?

ralphlange:
There was a concern expressed in the EPICS Meeting as to where the history file resides. Their installation - ITER, too - have the IOC boot directory mounted read-only.

The environment variable to control the location resolves the potential problem. (And, yes, the EPICS_IOCSH prefix is a good idea.)

mdavidsaver:
> ... IOC boot directory mounted read-only.

As a note.  While I am not planning to pursue it, I think the readline history API gives enough feedback to implement some kind of a search path and/or list of fallback paths.

ralphlange:
Adding unnecessary complexity. Being able to set the name/location is perfectly fine.

mdavidsaver:
Updated to switch the environment variable to `EPICS_IOCSH_HISTFILE`, and add `.iocsh_history` to various `.gitignore` files.

mdavidsaver:
Core Meeting: Looks ok.

mdavidsaver:
Applied as 2ca70d3aa280610c768ae78668d2f62fc4ef5212

dirk-zimoch:
Some issues with this enhancement that I found when taking it into operation.
1. If the startup directory is not writable, iocsh claims an error:
`ERROR read-only file system (30) writing '.iocsh_history'`
First, this should not be an error (which scares users) but only a warning (or less), because the ioc is fully functional. Second, the error message does not mention the effective directory, thus the reader of the messages does not know which location is read-only. Consider using `realpath()`.
2. One can change the location, but not in the startup script. The environment variable `EPICS_IOCSH_HISTFILE` has to be set *before* the ioc is started (because it is only read once at initialization before reading the startup script). Consider making this modifiable from the startup script.
3. Not only manual user input is stored in the file but every line from the startup script, including all comments. This quickly overruns the file size, so that nothing from the previous run remains accessible but only (the last part of) the last boot. This defies the whole purpose of a persistent history. Consider only adding manual input but not input from startup scripts to the history (like bash does it). And do not store comment lines.

ralphlange:
Not sure how 2. would work.

Changing the history file mid-flight would drop the existing history which - as you write in 3. - defies the purpose of a persistent history. (Except in situations where you intentionally want to delete history.)

dirk-zimoch:
If only manual input would go into the history, then the history facility would not be initialized before the end of the startup script. Thus, in that case, one could set the environment variable in the script. In other words, solving item 3 automatically solves item 2.

mdavidsaver:
> 1. If the startup directory is not writable ...

It should be straightforward to ignore (or demote?) `EPERM` from `write_history()` as `ENOENT` from `read_history()` is ignored.

> 2. One can change the location, but not in the startup script. ...

readline history is not actually needed until `iocsh(NULL)`.  Currently, it is also being setup during `iocsh("./st.cmd")` as scripts are read through the readline context.  I think this is only for convenience?  In `ReadlineContext::setup()`, the condition `if(context) {` might be adjusted to `if(context && (!fp || fp==stdin)) {`.  Although this may run into problems with vxworks special handling of stdin...

> 3. Not only manual user input is stored in the file but every line from the startup script ...

The change for 2. would have the side effect of not storing script lines in history.

Personally, I see including file content in history as a minor feature.  But I can understand that others may see it as clutter.  I am fine with changing this behavior.

PR #432, opened 2023-10-25
Avoid hang during concurrent db_cancel_event()

### LINKED ISSUE #423: dbEvent: double cancel causes hang

mdavidsaver:
Intended to fix #423.  A further correction to #424 (fab8fd7102143a661a81cf7163dd889601eb015c).

My attempt to handle concurrent `db_cancel_event()` just didn't work.  I had been hoping to avoid having to allocate (temporarily) an `epicsEventId` for each waiter.  However, I can't see a way around this.  So the single `pflush_sem` is replaced with a list of temporary events.

When a concurrent cancel is detected, the caller will (hopefully) allocate an epicsEventId and add this to the list of waiters.  It will then poke the worker in case it has already gone idle, and then wait for the corresponding notification that the worker is going idle again.

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -82,9 +82,10 @@ struct event_que {
 struct event_user {
     struct event_que    firstque;       /* the first event que */
 
+    ELLLIST             waiters;        /* event_waiter::node */
+
     epicsMutexId        lock;
     epicsEventId        ppendsem;       /* Wait while empty */
-    epicsEventId        pflush_sem;     /* wait for flush */
     epicsEventId        pexitsem;       /* wait for event task to join */
 
     EXTRALABORFUNC      *extralabor_sub;/* off load to event task */
@@ -100,6 +101,11 @@ struct event_user {
     void                (*init_func)();
     epicsThreadId       init_func_arg;
 };
+
+typedef struct {
+    ELLNODE node; /* event_user::waiters */
+    epicsEventId wake;
+} event_waiter;
 
 /*
  * Reliable intertask communication requires copying the current value of the
@@ -306,9 +312,6 @@ dbEventCtx db_init_events (void)
     evUser->ppendsem = epicsEventCreate(epicsEventEmpty);
     if (!evUser->ppendsem)
         goto fail;
-    evUser->pflush_sem = epicsEventCreate(epicsEventEmpty);
-    if (!evUser->pflush_sem)
-        goto fail;
     evUser->lock = epicsMutexCreate();
     if (!evUser->lock)
         goto fail;
@@ -326,8 +329,6 @@ fail:
         epicsMutexDestroy (evUser->firstque.writelock);
     if(evUser->ppendsem)
         epicsEventDestroy (evUser->ppendsem);
-    if(evUser->pflush_sem)
-        epicsEventDestroy (evUser->pflush_sem);
     if(evUser->pexitsem)
         epicsEventDestroy (evUser->pexitsem);
     freeListFree(dbevEventUserFreeList,evUser);
@@ -391,7 +392,6 @@ void db_close_events (dbEventCtx ctx)
 
     epicsEventDestroy(evUser->pexitsem);
     epicsEventDestroy(evUser->ppendsem);
-    epicsEventDestroy(evUser->pflush_sem);
     epicsMutexDestroy(evUser->lock);
 
     epicsMutexUnlock (stopSync);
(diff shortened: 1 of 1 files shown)


modules/database/src/ioc/db/dbEvent.c
```diff
@@ -592,23 +592,32 @@ void db_cancel_event (dbEventSubscription event)
...
-             * So uncondtionally re-trigger.  This will result in one spurious
-             * wakeup for each cancellation.
-             */
-            epicsEventTrigger(evUser->pflush_sem);
+            epicsEventMustTrigger(evUser->ppendsem);
+
+            if(wait.wake) {
+                epicsEventMustWait(wait.wake);
+            } else {
+                epicsThreadSleep(0.01); /* ick. but better than cantProceed() */
```
mdavidsaver, line 613: I know this looks odd.  I didn't want to call `epicsEventMustCreate()`, and chance a `cantProceed()`.  This seemed like the least bad option.

modules/database/src/ioc/db/dbEvent.c
```diff
@@ -1041,11 +1050,18 @@ static void event_task (void *pParm)
         pendexit = evUser->pendexit;
 
         evUser->pflush_seq++;
+        if(ellCount(&evUser->waiters)) {
+            /* hold lock throughout to avoid race between event trigger and destroy */
+            ELLNODE *cur;
+            for(cur = ellFirst(&evUser->waiters); cur; cur = ellNext(cur)) {
+                event_waiter *w = CONTAINER(cur, event_waiter, node);
+                if(w)
+                    epicsEventMustTrigger(w->wake);
```
anjohnson: In the body of the `for (…; cur; …)` loop we know that `cur` can't be NULL, so how could `w` (which I think should point to the same address as `cur`) ever be NULL? Shouldn't that `if(w)` be `if (w->wake)` instead?

mdavidsaver: You are correct, it should.

mdavidsaver: Fixed.

anjohnson:
Core Group: Approved.

PR #433, opened 2023-10-26
Split up dbcar so that it can use a callback function

simon-ess:
This should have no functional changes, but it allows other utilities to read CA link data from an IOC. For example, recsync could export the CA links for given records which could be useful.

### DIFF
--- a/modules/database/src/ioc/db/dbCaTest.c
+++ b/modules/database/src/ioc/db/dbCaTest.c
@@ -45,29 +45,106 @@
 #include "dbCa.h"
 #include "dbCaPvt.h"
 #include "dbCaTest.h"
-#include "dbCommon.h"
 #include "db_convert.h"
 #include "dbLock.h"
 #include "link.h"
 
-
+typedef struct ca_counts {
+    int level;
+    int ncalinks;
+    int nconnected;
+    unsigned long nDisconnect;
+    unsigned long nNoWrite;
+    int noReadAccess;
+    int noWriteAccess;
+} ca_counts;
+
+long print_ca_data(dbCommon *precord, dbFldDes *pdbFldDes, DBLINK* plink, void* pvt) {
+    caLink* pca;
+    ca_counts *counts = (ca_counts*)pvt;
+
+    counts->ncalinks++;
+    pca = (caLink *)plink->value.pv_link.pvt;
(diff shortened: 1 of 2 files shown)


anjohnson:
I don't have a major objection to this, but it seems like a fair amount of churn to avoid having to add something like 25 lines of code (i.e. your `process_ca_record_links()` routine) to the utility that wants the functionality. Can you convince me why it's worth doing?

I'd like to see Doxygen annotations added to the dbCaTest.h header, create some unit tests for those APIs, and it would need an entry in the Release Notes about the new API routine. Those might push me to the "in favor" side.

mdavidsaver:
I don't see that `print_ca_data()` adds value other than to move that ugly dbStaticLib loop into base.  I understand a desire to avoid dealing with the arrays stored in `struct dbRecordType`.  However, that is (regrettably imo.) our supported public API.

Now, I would be open to a more general callback based alternative to such loops if such a design provided a place for a mutex to guard access to `pdbbase`.  The unsafe nature of dbStaticLib is one of the barriers to eg. allowing addition of records after `iocInit()`.  This would be a larger project though.

modules/database/src/ioc/db/dbCaTest.c
```diff
@@ -45,29 +45,106 @@
...
+    int noReadAccess;
+    int noWriteAccess;
+} ca_counts;
+
+long print_ca_data(dbCommon *precord, dbFldDes *pdbFldDes, DBLINK* plink, void* pvt) {
+    caLink* pca;
+    ca_counts *counts = (ca_counts*)pvt;
+
+    counts->ncalinks++;
+    pca = (caLink *)plink->value.pv_link.pvt;
```
mdavidsaver, line 67: Rather than accessing the private `struct caLink`, external code can use the link support functions.  eg. `dbIsLinkConnected()` and `dbLinkIsVolatile()`.  This has the benefit of also handling pva links.

simon-ess:
@anjohnson - that's a fair point. When I first thought of doing this I hadn't really thought of how little code would need to be copied into new places, and now that you mention it, it really isn't all that much that would be duplicated in other places.

Furthermore, as per @mdavidsaver 's comment: the function `dbcar` probably should only reference CA links, but in the use cases that I am thinking of it would be likely better to be somewhat more link-type agnostic.

With that said, I can accept closing this and moving the appropriate changes to e.g. recsync where I actually want to use them.

PR #434, opened 2023-10-27
DB parser hint for unknown field name

mdavidsaver:
Adds a "did you mean" hint when an unknown field name is encountered.  Prints the name and prompt string of the best match.

```
$ softIoc -d badfield.db 
Error: Failed to load: badfield.db
ai Record "foo" does not have a field "DYTP"
    Did you mean "DTYP"?  (Device Type)
Error at or before ')' in path "."  file "badfield.db" line 2
dbLoadRecords: failed to load 'badfield.db'
```

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1167,8 +1167,28 @@ static void dbRecordField(char *name,char *value)
     pdbentry = ptempListNode->item;
     status = dbFindField(pdbentry,name);
     if (status) {
-        epicsPrintf("Record \"%s\" does not have a field \"%s\"\n",
-            dbGetRecordName(pdbentry), name);
+        epicsPrintf("%s Record \"%s\" does not have a field \"%s\"\n",
+                    dbGetRecordTypeName(pdbentry), dbGetRecordName(pdbentry), name);
+        if(dbGetRecordName(pdbentry)) {
+            DBENTRY temp;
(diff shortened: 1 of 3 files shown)


modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1167,8 +1167,28 @@ static void dbRecordField(char *name,char *value)
     pdbentry = ptempListNode->item;
     status = dbFindField(pdbentry,name);
     if (status) {
-        epicsPrintf("Record \"%s\" does not have a field \"%s\"\n",
-            dbGetRecordName(pdbentry), name);
+        epicsPrintf("%s Record \"%s\" does not have a field \"%s\"\n",
+                    dbGetRecordTypeName(pdbentry), dbGetRecordName(pdbentry), name);
+        if(dbGetRecordName(pdbentry)) {
```
anjohnson, line 1172: We should probably avoid passing a NULL `dbGetRecordName(pdbentry)` to `epicsPrintf()` (most libc implementations do handle that okay, but it doesn't look good). Move the print inside the conditional?

mdavidsaver: I don't think it is actually possible to reach `dbRecordField()` unless the record exists.  I can't recall ever seeing `Record "(null)"`.  So the condition is I think mainly paranoia.  I was anyway going to have a separate block for the added locals.

anjohnson:
Core Group: Approved.

PR #435, opened 2023-10-27
Colorize stderr

mdavidsaver:
So far we have only included ANSI escape sequences through errlog, where they are filtered out when not supported.  This PR adds colorized errors to many `fprintf(stderr, ...`, where they would not be filtered out.

I think the benefit of colorized errors in eg. db file parse errors is obvious to anyone who has failed to notice such an error in the middle of the output of a long `st.cmd`.

Of course unconditionally printing ANSI escapes may annoy someone who is collecting stderr for viewing with a program which does not know about escapes.

I can think of several ways to avoid this, however these would introduce additional complexity and/or failure modes which I don't like.  eg. `epicsSnprintf()` into an allocated temporary buffer.

Some arguments in favor...

We could allow escapes in stderr, but not stdout.  imo. stdout is more likely to be redirected and consumed programmatically than stderr.

Captured stderr logs would be improved if also passed through tools which convert eg. [escapes to html](https://pypi.org/project/ansi2html/) tags.

So, would the extra overhead of striping escapes be worthwhile?

### DIFF
--- a/modules/ca/src/client/udpiiu.cpp
+++ b/modules/ca/src/client/udpiiu.cpp
@@ -549,7 +549,7 @@ void epicsStdCall caRepeaterRegistrationMessage (
             char sockErrBuf[64];
             epicsSocketConvertErrnoToString (
                 sockErrBuf, sizeof ( sockErrBuf ) );
-            fprintf ( stderr, "error sending registration message to CA repeater daemon was \"%s\"\n",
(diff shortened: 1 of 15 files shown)


anjohnson:
Core Group: Please add a Release Note about this, then merge is approved.

mdavidsaver:
Applied as a352865df9b0da81ef6dab6be48eb21ab98f0846

PR #436, opened 2023-11-03
bi "Raw Soft Channel" use MASK

mdavidsaver:
If set, apply MASK to the value which has just been read into RVAL.

### DIFF
--- a/modules/database/src/std/dev/devBiSoftRaw.c
+++ b/modules/database/src/std/dev/devBiSoftRaw.c
@@ -51,6 +51,9 @@ static long readLocked(struct link *pinp, void *dummy)
 
     if (status) return status;
 
+    if (prec->mask)
+        prec->rval &= prec->mask;
+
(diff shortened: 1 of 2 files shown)


anjohnson:
I don't particularly object to this as it's obviously useful, but should point out that all `MASK` fields in Base are marked `special(SPC_NOMOD)` so they can only be set by device support or from a .db file. It might be worth pointing this out in the record documentation change.

mdavidsaver:
> ... It might be worth pointing this out in the record documentation change.

Of course the (remotely) read-only status of MASK is reflect in a generated field table, and would be readily apparent to anyone attempting a PUT.

If you (@anjohnson) have specific wording in mind, you might try editing through the github.com web UI.

mdavidsaver:
> I don't particularly object to this as it's obviously useful ...

Since there is agreement on the function change, merged as f2fe9d12032569a8b1d3f09ddf7e07f2ce8c04dd.  Documentation will of course continue to be refined.

PR #437, opened 2023-11-03
Null check callback function in callbackRequest

JJL772:
We will get a segfault from one of the callback threads if pcallback->callback is NULL. Better to NULL check it and return error.


mdavidsaver:
This seems like a reasonable check.

PR #439, opened 2023-11-17
Set ASL0 for mbboDirect Bx fields.

ericonr:
Since the record's VAL field is ASL0, it doesn't make sense to gate writes into the Bx fields with ASL1.


anjohnson (approved):
I agree, this change makes sense to me.

mdavidsaver (approved):
Looks fine to me.

PR #440, opened 2023-11-17
Config: Fixed Hard coded LDFLAGS in MVME2500

ezeob002:
Updated ARCH_DEP_LDFLAGS for Universal Compatibility Across All RTEMS Versions.


anjohnson:
Merging without running the tests since this only affects an RTEMS target that we don't build, and I believe this target also requires at least one of @kiwichris' PRs for RTEMS-6. Existing RTEMS-5 builds of the qoriq_e500 target will not be affected.

PR #444, opened 2023-11-24
Don't include linux/ header in osdSockUnsentCount.

ericonr:
Per the manpage in the file's comment, SIOCOUTQ is equivalent to the TIOCOUTQ ioctl, whose value can be obtained by including <sys/ioctl.h>, which allows us to avoid including system-specific headers.

---

The `epicsSocketUnsentCount` function seems to only be used in logClient code (at least as far as epics-base goes). Is it critical that a working version be defined for as many platforms as possible?

I'm asking because, while investigating this ioctl, I came across the `FIONWRITE` one from the BSDs, which seems to be supported at least by [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=ioctl&apropos=0&sektion=0&manpath=FreeBSD+14.0-RELEASE+and+Ports&arch=default&format=html), [NetBSD](https://man.netbsd.org/ioctl.2) and [vxWorks](https://www.ecb.torontomu.ca/~courses/ee8205/Data-Sheets/Tornado-VxWorks/vxworks/ref/tyLib.html), which might have made it a candidate for the `posix` implementation? (I don't fully understand EPICS's policy for the osi code). However, [zOS](https://www.ibm.com/docs/en/zos/3.1.0?topic=functions-ioctl-control-device), for example, has a different meaning for it.

### DIFF
--- a/modules/libcom/src/osi/os/Linux/osdSockUnsentCount.c
+++ b/modules/libcom/src/osi/os/Linux/osdSockUnsentCount.c
@@ -4,7 +4,6 @@
 * in file LICENSE that is included with this distribution.
 \*************************************************************************/
 
-#include <linux/sockios.h>
 #define EPICS_PRIVATE_API
 #include "osiSock.h"
 
@@ -14,7 +13,7 @@
  */
 int epicsSocketUnsentCount(SOCKET sock) {
     int unsent;
-    if (ioctl(sock, SIOCOUTQ, &unsent) == 0)
+    if (ioctl(sock, TIOCOUTQ, &unsent) == 0)
         return unsent;
     return -1;
 }


anjohnson:
Core group: That particular file is only built on Linux anyway, other OSs may/should have their own implementations in different files (the osi/os/Linux in the path means this is a linux-only source file). Please look at the others for comparison.

ericonr:
> That particular file is only built on Linux anyway

I am aware. The motivation for the PR was that, unlike glibc, musl doesn't require kernel UAPI headers to be installed. So, with this PR, a basic setup for building epics-base on musl would work without installing a kernel header package.

I should have made that clear in the PR description. Should I include it in the commit?

> Please look at the others for comparison.

I'm not sure what I should be looking for.

ralphlange:
The "policy" (more a procedure) of the EPICS build for the OSI mechanism is:
- The Makefile (usually above) defines source files for a library/binary by simple name (without path)
- Make searches for each file in
  1. `os/<specific target OS>`
  2. `os/posix` (if the target OS claims to be posix compliant)
  3. `os/default`
- First matching file found is taken for the build.

mdavidsaver:
fyi. On Debian both definitions come from headers installed by the `linux-libc-dev` package.

```
$ grep -wR SIOCOUTQ /usr/include/
/usr/include/linux/sockios.h:#define SIOCOUTQ   TIOCOUTQ        /* output queue size (not sent + not acked) */
$ grep -wR TIOCOUTQ /usr/include/
/usr/include/asm-generic/ioctls.h:#define TIOCOUTQ      0x5411
...
$ dpkg -L linux-libc-dev | grep sockios.h
/usr/include/asm-generic/sockios.h
/usr/include/linux/sockios.h
...
```

So I guess this comes down to musl choosing to provide the TTY definition and not the socket definition.  TTY vs. socket is as I understand it, the origin of the different T vs. S names, which have now come to be synonyms.

@ericonr Have you checked Linux history to see if SIOCOUTQ and TIOCOUTQ were ever not equivalent?

mdavidsaver:
> I should have made that clear in the PR description.  ...

Please do.

Also, have you checked to see which macros uclibc provides?

ericonr:
> So I guess this comes down to musl choosing to provide the TTY definition and not the socket definition.

I believe that's because the TTY definition is more standard across OSs, while the socket one is Linux specific.

> Have you checked Linux history to see if SIOCOUTQ and TIOCOUTQ were ever not equivalent?

@mdavidsaver it's been that way since at least the first commit into git: https://github.com/torvalds/linux/commit/1da177e4c3f41524e886b7f1b8a0c1fc7321cac2

> Also, have you checked to see which macros uclibc provides?

uclibc doesn't seem to define either of these, but they include the kernel headers.

ericonr:
Ping for merging here, all tests have passed now.

ralphlange:
Noted. Thanks!
I'm sure the Core group will have another look at their next review session.

anjohnson:
Core: Apologies for the delay, approved!

PR #446, opened 2023-11-29
change PROC to ASL0

dirk-zimoch:
The `PROC` field is a "user" field like `VAL`, not a "configuration" field like `DRVH`.
But since the `PROC` field has the default `ASG1`, a Channel Access who wants to process the record via `PROC` field needs configuration rights.
This PR changes the `PROC` to `ASG0`, the same as `VAL` typically has. This allows a user with only write access but without configuration access to process the record.

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -22,6 +22,12 @@ should also be read to understand what has changed since earlier releases:
 
 ## Changes made on the 7.0 branch since 7.0.7
 
+### PROC field changed to ASL0
+
+The PROC field has been changed from access security level ASL1 to ASL0.
+This allows users to trigger processing a record without having the rights
+to reconfigure the records.
+
 ### bi "Raw Soft Channel" use MASK
 
 If MASK is non-zero, The raw device support will now apply MASK to the
--- a/modules/database/src/ioc/db/dbCommon.dbd.pod
+++ b/modules/database/src/ioc/db/dbCommon.dbd.pod
@@ -240,6 +240,7 @@ The B<SPVT> field is for internal use by the scanning system.
 	}
 	field(PROC,DBF_UCHAR) {
 		prompt("Force Processing")
+		asl(ASL0)
 		pp(TRUE)
 		interest(3)
 	}


mdavidsaver:
As I think about it, changing `PROC` to `ASL0` would *not* be granting additional access for records where `VAL` is `ASL0` and `pp(TRUE)`.  For these recordtypes, allowing ASL0 to PROC seems acceptable to me.  However, there are a couple of recordtypes in Base which don't meet this criteria:

* aSub
* calcout
* calc
* event
* histogram
* sel

Maybe the first question to ask is whether these (and other) recordtype definitions are correct?

eg. I am surprised that `fanout` has `VAL` with `pp(TRUE)` and ASL0.  Does this make `VAL` and `PROC` functionally equivalent for that recordtype?

dirk-zimoch:
As `PROC` is in dbCommon.dbd, it is not possible to set permissions different for different record types.
I don't think that we should abuse writing to `VAL` as a workaround for writing to `PROC`. In that case we could remove `PROC` altogether. For example CA `FLNK` should point to `PROC`, not `VAL`, according to documentation (and warning if done differently).

The question here is not "Is it necessary?" or "Does it grant more access?" The question is: "Does the current implementation make sense? And would a change make more sense?"

Background: After an unfortunate and still mysterious event of probably someone's script writing `1` to "everything" (including drive and alarm limits) we use now caPutLog to log every access to `ASL1` fields. But this also writes logs if records are processed via `PROC`.

anjohnson:
It should be possible to allow record types to override some attributes of fields defined in dbCommon, although not trivial — mostly Perl coding and build system changes I think.

However I agree with Dirk that we should first decide whether what we have now makes sense, if making PROC be ASL0 for all record types would be more sensible (or more dangerous for existing IOCs), or if we would confuse our users too much by letting different record types have different AS levels for PROC.

anjohnson:
For future reference, the following Perl changes make it possible for a Record.dbd.pod file to amend (override) the field attributes of a record field that was defined in dbCommon.dbd.

```patch
diff --git a/modules/database/src/tools/DBD/Parser.pm b/modules/database/src/tools/DBD/Parser.pm
index d93588465..9535e848e 100644
--- a/modules/database/src/tools/DBD/Parser.pm
+++ b/modules/database/src/tools/DBD/Parser.pm
@@ -310,7 +310,14 @@ sub parse_record {
 
 sub parse_field {
     my ($rtyp, $field_name, $field_type) = @_;
-    my $fld = DBD::Recfield->new($field_name, $field_type);
+    my $fld = $rtyp->field($field_name);
+    my $amendment = defined($fld);
+    if ($amendment) {
+       dieContext("Repeated field '$field_name' changes field type")
+           unless $fld->dbf_type eq $field_type;
+    } else {
+        $fld = DBD::Recfield->new($field_name, $field_type);
+    }
     pushContext("field($field_name, $field_type)");
     while(1) {
         parseCommon($fld);
@@ -321,7 +328,7 @@ sub parse_field {
         }
         elsif (m/\G \}/xgc) {
             print "  Field-End:\n" if $debug;
-            $rtyp->add_field($fld);
+            $rtyp->add_field($fld) unless $amendment;
             popContext("field($field_name, $field_type)");
             return;
         } else {
```
No build system changes are needed AFAICS.

dirk-zimoch:
Info: This change (in dbCommon and thus for all records types) is in use at PSI since 2023 without any negative effects.

mdavidsaver:
> Info: This change (in dbCommon and thus for all records types) is in use at PSI since 2023 without any negative effects.

imo. "Doesn't break" is the wrong criteria for evaluating a change to loosen permissions.

mdavidsaver:
> ```diff
> +    my $amendment = defined($fld);
> +    if ($amendment) {
> +       dieContext("Repeated field '$field_name' changes field type")
> +           unless $fld->dbf_type eq $field_type;
> ```

Maybe better for a change like this to use the same special `*` marker as record instance amendments.

```
recordtype(longin) {
    ...
    include "dbCommon.dbd"
    field(PROC, "*") {
        asl(ASL0)
    }

**Comment by dirk-zimoch** [comment]
> imo. "Doesn't break" is the wrong criteria for evaluating a change to loosen permissions.

It is actually tightening access because now I do not need to give ASL1 access to users just to allow them to process the record.

**Comment by anjohnson** [comment]
> use the same special `*` marker as record instance amendments.

Good idea, requiring `"*"` makes it obvious that we're changing a field we know was defined earlier, not trying to add a new field.

Changed and tested, with a better error message:

```patch
diff --git a/modules/database/src/tools/DBD/Parser.pm b/modules/database/src/tools/DBD/Parser.pm
index d93588465..ddf849f8d 100644
--- a/modules/database/src/tools/DBD/Parser.pm
+++ b/modules/database/src/tools/DBD/Parser.pm
@@ -310,7 +310,15 @@ sub parse_record {
 
 sub parse_field {
     my ($rtyp, $field_name, $field_type) = @_;
-    my $fld = DBD::Recfield->new($field_name, $field_type);
+    my $fld = $rtyp->field($field_name);
+    my $amendment = defined($fld);
+    if ($amendment) {
+        dieContext("Field '$field_name' already defined.",
+            "To amend a dbCommon field attribute, select the field using",
+            "    field($field_name, \"*\") { ... }")
+            unless $field_type eq '*';
+    } else {
+        $fld = DBD::Recfield->new($field_name, $field_type);
+    }
     pushContext("field($field_name, $field_type)");
     while(1) {
         parseCommon($fld);
@@ -321,7 +329,7 @@ sub parse_field {
         }
         elsif (m/\G \}/xgc) {
             print "  Field-End:\n" if $debug;
-            $rtyp->add_field($fld);
+            $rtyp->add_field($fld) unless $amendment;
             popContext("field($field_name, $field_type)");
             return;
         } else {
```

-----

Earlier @dirk-zimoch wrote:
> I don't think that we should abuse writing to VAL as a workaround for writing to PROC.

Unfortunately for almost any I/O record instance configured to read from its INP or DOL field, a client writing to VAL normally has the same effect as writing to PROC; it triggers record processing, which causes the input link to overwrite whatever value was written by the client. Some databases and displays for IOCs using Access Security may be using that equivalence to get around the fact that PROC is ASL1. That practice may also be why `dbPut()` skips the `db_post_events()` call for writes to VAL fields that are marked `pp(TRUE)`; someone probably asked for that after seeing double monitor updates.

Most Base record types define VAL as both `asl(ASL0)` and `pp(TRUE)`. @mdavidsaver already pointed out the ones that do _not_ define VAL as `pp(TRUE)`. These are all the fields marked `pp(TRUE)` in those record types (including dbCommon fields), none of which are also marked `asl(ASL0)`:
- aSubRecord: UDF, PROC, BRSV
- calcoutRecord: UDF, PROC, CALC, OCAL, A … L, HIHI … LOLO, HHSV … LLSV
- calcRecord: UDF, PROC, CALC, A … L, HIHI … LOLO, HHSV … LLSV
- eventRecord: UDF, PROC
- histogramRecord: UDF, PROC
- selRecord: UDF, PROC, A … L, HIHI … LOLO, HHSV … LLSV

Absent this PR there are no ASL0 fields in dbCommon, so the above record types cannot be triggered directly by clients with only ASL0 access permissions. There is thus a danger that that accepting this PR would allow some clients to do things they didn't used to be able to do.

PR #447, opened 2023-12-04
Print link types as text instead of numbers

dirk-zimoch:
Improve error message when using wrong link format.
Old: `ERROR: RECORD.INP: can't initialize link type 12 with "#C1 S2@SOME_TEXT" (type 2)`
New: `ERROR: RECORD.INP: can't initialize link type INST_IO with "#C1 S2@SOME_TEXT" (type VME_IO)`

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -2206,12 +2206,12 @@ long dbInitRecordLinks(dbRecordType *rtyp, struct dbCommon *prec)
              */
 
         } else if(dbCanSetLink(plink, &link_info, devsup)!=0) {
-            errlogPrintf(ERL_ERROR ": %s.%s: can't initialize link type %d with \"%s\" (type %d)\n",
-                         prec->name, pflddes->name, plink->type, plink->text, link_info.ltype);
+            errlogPrintf(ERL_ERROR ": %s.%s: can't initialize link type %s with \"%s\" (type %s)\n",
+                         prec->name, pflddes->name, pamaplinkType[plink->type].strvalue, plink->text, pamaplinkType[link_info.ltype].strvalue);
 
         } else if(dbSetLink(plink, &link_info, devsup)) {
-            errlogPrintf(ERL_ERROR ": %s.%s: failed to initialize link type %d with \"%s\" (type %d)\n",
-                         prec->name, pflddes->name, plink->type, plink->text, link_info.ltype);
+            errlogPrintf(ERL_ERROR ": %s.%s: failed to initialize link type %s with \"%s\" (type %s)\n",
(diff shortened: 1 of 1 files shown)


mdavidsaver (approved):
Looks good to me.

anjohnson:
Minor: There's already one newline inside those call arguments, can you add at least one more to each (before the last argument say) to keep the line width sane please.

I personally prefer a line width of 80 rather than the 100 which the previous edits left here, but this change takes those lines over 140 which is getting ridiculous IMHO.

mdavidsaver:
> I personally prefer a line width of 80 rather than the 100 which the previous edits left here, but this change takes those lines over 140 which is getting ridiculous IMHO.

@anjohnson Can you provide a some more concrete guidance?

```
ERROR: RECORD.INP: can't initialize link type INST_IO with "#C1 S2@SOME_TEXT" (type VME_IO)
```

maybe like the following, which splits the two variable length parts (record name and field value) onto separate lines.

```
RECORD.INP ERROR: can't initialize link type INST_IO
    with type VME_IO "#C1 S2@SOME_TEXT"
```

Having the source `record.FLD` first seems to be the more common convention at the moment.  Although it does put one variable length piece at the start of the first line, which makes indentation of the following lines imo. somewhat less clear.

anjohnson:
Sorry for being unclear, I was referring to the line width of the source code. I'm not too concerned about adding newlines to wide error messages, although if you do that I would ask that (as you did above) there should be at least a one space indentation after the newline since some log-file parsers take a leading space to indicate a continuation of the previous line.

dirk-zimoch:
Added line breaks and rebased because of ci problems.

dirk-zimoch:
Still fails in CentOS-7. Please someone fix the CI!

Insomnia1437:
It seems `ci-core-dumper` upgraded node version to 20 recently which is too new for CentOS7
https://github.com/mdavidsaver/ci-core-dumper/commit/6ae227ed6892d9b47edc18b3f88dc21284240748
I think by changing this line 
https://github.com/epics-base/epics-base/blob/cb1571783b4793cb9c545e5736659ee5ed7d03bb/.github/workflows/ci-scripts-build.yml#L286
to  `uses: mdavidsaver/ci-core-dumper@v4` may fix

ralphlange:
This will bring back the deprecation warnings and stop working soon.
(Which was the reason to update the ci-core-dumper.)

When building using containers, the GHA infrastructure mounts and runs the node binaries outside the container from inside the container.
If the GHA team builds their node installation on a machine that has a newer GLIBC, running it from inside a container with an older GLIBC will fail.

dirk-zimoch:
> When building using containers, the GHA infrastructure mounts and runs the node binaries outside the container from inside the container.

Great! Weren't such compatibility issues one of the things containers were designed to solve?

mdavidsaver:
wrt. CI failures.  See https://github.com/mdavidsaver/ci-core-dumper/issues/4#issuecomment-2027800172

PR #449, opened 2023-12-22
dbLoadTemplate: error propagation

mdavidsaver:
Propagate `dbLoadDatabase()` errors out from `dbLoadTemplate()`.  Also print more meaningful error messages.

Seems straightforward enough to me, although failing earlier is a behavior change.


anjohnson:
Makes sense to me. Should the error message(s) be using the color macros or are they included in the `yyerror()` output already?

mdavidsaver:
`yyerror()` does prefix `ERL_ERROR`.

anjohnson:
LGTM then.

mdavidsaver:
> LGTM then.

I will consider this approved then.

mdavidsaver:
Merged as 1cd141c540c5532f1618c538c3a81a62fc73c639

Issue #331, opened 2023-02-22
rsrv: `db_create_read_log()` w/o locking


mdavidsaver:
It looks to me like `db_create_read_log()` is called without locking the associated record.  So the target field value, and time+alarm meta-data are accessed without locking during a CA GET.

https://github.com/epics-base/epics-base/blob/718da5c9be96b7eccd7ce3a95f3b2d232c11ca32/modules/database/src/ioc/rsrv/camessage.c#L538-L548

mdavidsaver:
Appears to `valgrind --tool=helgrind` as:

```
...
Possible data race during write of size 2 at 0x5175D7C by thread #14
Locks held: 1, at address 0x51845D0
   at 0x48D4A28: recGblResetAlarms (recGbl.c:197)
   by 0x485E15B: monitor (calcRecord.c:401)
   by 0x485E15B: process (calcRecord.c:132)
   by 0x48B5E63: dbProcess (dbAccess.c:610)
   by 0x48CB2C3: scanList (dbScan.c:1012)
   by 0x48CB6D7: periodicTask (dbScan.c:804)
   by 0x496B7AC: start_routine (osdThread.c:441)
   by 0x483F876: mythread_wrapper (hg_intercepts.c:387)
   by 0x4F41EA6: start_thread (pthread_create.c:477)
   by 0x4CC9A2E: clone (clone.S:95)

This conflicts with a previous read of size 2 by thread #22
Locks held: 1, at addresses  (and 1 that can't be shown)
   at 0x48CE6C5: db_create_field_log (dbEvent.c:709)
   by 0x48CE6C5: db_create_read_log (dbEvent.c:764)
   by 0x48FBC0F: read_reply (camessage.c:540)
   by 0x48FD912: read_notify_action (camessage.c:745)
   by 0x48FF067: camessage (camessage.c:2546)
   by 0x48FAC5F: camsgtask (camsgtask.c:116)
   by 0x496B7AC: start_routine (osdThread.c:441)
   by 0x483F876: mythread_wrapper (hg_intercepts.c:387)
   by 0x4F41EA6: start_thread (pthread_create.c:477)
...
```

anjohnson:
Ack. The 3.15 version of `db_create_read_log()` didn't access the record at all, and the filters run by the `dbChannelRunP*Chain()` routines look to do their own locking, so unfortunately I suspect this points to 27fe3e4 as when this came in.

mdavidsaver:
[...]

PR #450, opened 2023-12-22
db: lock record for db_create_read_log() and dbChannelGetField()

mdavidsaver:
Resolve a data race of time/alarm meta-data during CA GET with server side filter.

Since 27fe3e4468ec62d4d17f775c4aced939a666ba36 db_create_read_log() accesses record fields, so callers must hold the record lock.  However, it is impossible for callers of `dbGetField()` to do so atomically.  So change `dbGetField()` and `dbChannel_get_count()` to create a local `db_field_log*` when useful.

### DIFF
--- a/modules/database/src/ioc/db/dbChannel.c
+++ b/modules/database/src/ioc/db/dbChannel.c
@@ -635,9 +635,21 @@ long dbChannelGetField(dbChannel *chan, short dbrType, void *pbuffer,
 {
     dbCommon *precord = chan->addr.precord;
     long status = 0;
+    unsigned char local_fl = 0;
 
     dbScanLock(precord);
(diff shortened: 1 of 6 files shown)


mdavidsaver:
I have been testing this change locally since Feb. 2023.  I should have put up a PR sooner, but I have to admit that I lost track of this issue.

anjohnson:
Seems reasonable to me. I wonder whether some recent tech-talk reports might be resolved by this?

anjohnson:
Core: Rebase to re-run tests, approved assuming they pass.

mdavidsaver:
Would address #331

> Core: Rebase to re-run tests, approved assuming they pass.

I did, and they did.  However, at this point the exercise needs to be repeated after 7.0.9

mdavidsaver:
Rebased.

PR #451, opened 2023-12-22
Avoid initMainThread() except on vxworks

mdavidsaver:
Move isOkToBlock tracking to osdThread.  Targets except vxworks can store this flag in epicsThreadOSD. Continue to use TLS w/ vxWorks.  Eliminates one c++ global constructor (cf. #298 and #286).

### DIFF
--- a/modules/libcom/src/osi/os/vxWorks/osdThread.c
+++ b/modules/libcom/src/osi/os/vxWorks/osdThread.c
@@ -127,7 +127,8 @@ static void epicsThreadInit(void)
         taskIdListSize = ID_LIST_CHUNK;
         atRebootRegister();
         ALLOT_JOIN(0);
-        done = 1;
+        done = 1; /* avoids recursive call */
+        epicsThreadSetOkToBlock(1);
     }
     lock = 0;
 }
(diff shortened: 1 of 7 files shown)


modules/libcom/src/osi/os/vxWorks/osdThread.c
```diff
@@ -127,6 +127,8 @@ static void epicsThreadInit(void)
         taskIdListSize = ID_LIST_CHUNK;
         atRebootRegister();
         ALLOT_JOIN(0);
+        (void)epicsThreadGetIdSelf();
```
anjohnson: This call to `epicsThreadGetIdSelf()` isn't necessary, the VxWorks implementation just forwards to the OS's `taskIdSelf()` to get the taskId of the shell task that's currently loading `st.cmd`, so it has no side-effects.

modules/libcom/src/osi/os/vxWorks/osdThread.c
```diff
@@ -127,6 +127,8 @@ static void epicsThreadInit(void)
...
+        epicsThreadSetOkToBlock(1);
         done = 1;
```
anjohnson: The `epicsThreadSetOkToBlock()` routine can't be called until after `done` has been set; it calls `epicsThreadPrivateCreate()` which calls `epicsThreadInit()` (this routine) which then deadlocks waiting for `lock` to be cleared.

The deadlock doesn't occur if I swap these two lines, which looks safe to do since all VxWorks systems are uni-processor, but I'd appreciate a second pair of eyes on that solution.

modules/libcom/src/osi/os/vxWorks/osdThread.c
```diff
@@ -577,3 +579,30 @@ LIBCOM_API int epicsThreadGetCPUs(void)
...
     return 1;
 }
+
+
+static epicsThreadOnceId okToBlockOnce = EPICS_THREAD_ONCE_INIT;
+static epicsThreadPrivateId okToBlockPrivate;
+static const int okToBlockNo = 0;
+static const int okToBlockYes = 1;
+
+static void epicsThreadOnceIdInit(void *)
```
anjohnson:
```
../osi/os/vxWorks/osdThread.c: In function 'epicsThreadOnceIdInit':
../osi/os/vxWorks/osdThread.c:589: error: parameter name omitted
```

```patch
diff --git a/modules/libcom/src/osi/os/vxWorks/osdThread.c b/modules/libcom/src/osi/os/vxWorks/osdThread.c
index 7a951fee9..b907a7f12 100644
--- a/modules/libcom/src/osi/os/vxWorks/osdThread.c
+++ b/modules/libcom/src/osi/os/vxWorks/osdThread.c
@@ -586,7 +586,7 @@ static epicsThreadPrivateId okToBlockPrivate;
 static const int okToBlockNo = 0;
 static const int okToBlockYes = 1;
 
-static void epicsThreadOnceIdInit(void *)
+static void epicsThreadOnceIdInit(void *not_used)
 {
     okToBlockPrivate = epicsThreadPrivateCreate();
 }
```

anjohnson (changes requested):
Three issues in the VxWorks code with solutions: one build-time, one run-time and one minor optimization.

modules/libcom/src/osi/os/vxWorks/osdThread.c
mdavidsaver: Removed.

modules/libcom/src/osi/os/vxWorks/osdThread.c
mdavidsaver: Changed as recommended.

> but I'd appreciate a second pair of eyes on that solution.

I think what you say is correct.  By this point, `epicsThreadInit()` is done.

modules/libcom/src/osi/os/vxWorks/osdThread.c
mdavidsaver: parameter name un-omitted.

mdavidsaver:
I think this is ready for another look.

mdavidsaver:
Rebased.

PR #452, opened 2023-12-22
Add typing to dbEvent.h opaque structs when -DUSE_TYPED_DBEVENT

mdavidsaver:
dbEvent.h avoid typedefs of `void*`.  Since this would be an incompatible API change, follow the pattern of device and driver support APIs with an opt-in macro `USE_TYPED_DBEVENT`.

This PR updates internal usages except `caservertask.c`, which is left as-is to "prove" backwards compatibility.

As an example of why this change could otherwise be backwards incompatible.  In PVXS I had to change:

```c++
std::shared_ptr<void> S(db_add_event(...));
// to
std::shared_ptr<std::remove_pointer<dbEventSubscription>::type> S(db_add_event(...));
```

### DIFF
--- a/modules/database/src/ioc/db/dbChannelIO.cpp
+++ b/modules/database/src/ioc/db/dbChannelIO.cpp
@@ -14,6 +14,8 @@
  *  505 665 1831
  */
 
+#define USE_TYPED_DBEVENT
+
 #include <string>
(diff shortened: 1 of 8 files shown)


anjohnson:
Core Group: Looks fine in principle.

PR #453, opened 2023-12-28
Rebase and update of the CodeQL migration branch

anjohnson:
The repo for PR #306 is no longer writable, this PR replaces that for future work.

The Core Group asked:

> Find how to control which tests get run and how to turn them off before merging this.

The answers to that are in the new `.github/codeql/config.yml` file which should now control the tests (called "queries" here) that it runs, assuming I've understood the documentation and configured the workflow correctly.


mdavidsaver:
After another quick look, I have to admit that I remain confused about how codeql is supposed to integrate with GHA.  eg. where are the full results?

@anjohnson Could you setup and demonstrate this with https://github.com/anjohnson/epics-base ?

I realize this is a big ask.  Before turning on any new static analysis, I would like to have an idea of what sorts of checks are enabled, and what they flag in epics-base.  I see this as necessary to feed a discussion of which of checks should be turned on initially, and if there are any checks which while too noisy now, should be targeted for future.

In short, before merging I think someone(s) need to go through the complete list of errors/warnings/... and do a first pass triage.  I can help with this, but first I need to know where to find the results...

anjohnson:
The full results from the CodeQL scan that this PR triggered can be found [here](https://github.com/epics-base/epics-base/security/code-scanning?query=is%3Aopen+pr%3A453). Once we merge this PR they should be accessible from the [Code Scanning](https://github.com/epics-base/epics-base/security/code-scanning) link under the project's Security page Vulnerability Alerts on the LHS, and I suspect the GHA action will also have a link showing any changes to the set of alerts found by each scan.

anjohnson:
Core group: Approved since it's CI-only.

PR #454, opened 2024-01-02
Fix Clang-15 warnings

anjohnson:
These commits clean up compiler warnings from Clang-15 when building the built-in modules. I developed and tested this on macOS, both Intel and ARM; I didn't check for warning reductions on gcc though.

### DIFF
--- a/modules/database/src/ioc/db/dbConvertFast.h
+++ b/modules/database/src/ioc/db/dbConvertFast.h
@@ -14,13 +14,16 @@
 
 #include "dbFldTypes.h"
 #include "dbCoreAPI.h"
+#include "link.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-DBCORE_API extern long (*dbFastGetConvertRoutine[DBF_DEVICE+1][DBR_ENUM+1])();
-DBCORE_API extern long (*dbFastPutConvertRoutine[DBR_ENUM+1][DBF_DEVICE+1])();
+/* typedef FASTCONVERTFUNC is now defined in link.h */
+
+DBCORE_API extern FASTCONVERTFUNC dbFastGetConvertRoutine[DBF_DEVICE+1][DBR_ENUM+1];
+DBCORE_API extern FASTCONVERTFUNC dbFastPutConvertRoutine[DBR_ENUM+1][DBF_DEVICE+1];
 
 #ifdef __cplusplus
 }
--- a/modules/database/src/ioc/db/dbConvertFast.h
+++ b/modules/database/src/ioc/db/dbConvertFast.h
@@ -7,7 +7,20 @@
 * EPICS Base is distributed subject to a Software License Agreement found
 * in file LICENSE that is included with this distribution.
 \*************************************************************************/
-/* dbConvertFast.h */
+/** @file dbConvertFast.h
+ *  @brief Data conversion for scalar values
+ *
+ * The typedef FASTCONVERTFUNC is defined in link.h as:
+ * @code
+ *   long convert(const void *from, void *to, const struct dbAddr *paddr);
+ * @endcode
+ *
+ * The arrays declared here provide pointers to the fast conversion
+ * routine where the first array index is the data type for the first
+ * "from" pointer arg, and the second array index is the data type for
+ * the second "to" pointer arg. The array index values are a subset of
+ * the DBF_ enum values defined in dbFldTypes.h
+ */
 
 #ifndef INCdbConvertFasth
 #define INCdbConvertFasth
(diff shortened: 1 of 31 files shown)


modules/database/test/ioc/db/dbStressLock.c
```diff
@@ -120,6 +120,9 @@ void doMulti(workerPriv *p)
     }
     dbScanUnlockMany(locker);
 
+    if (0)
+        testDiag("sum = %d", sum);
```
mdavidsaver, line 127: Debugging leftovers?

anjohnson: Core Group: Use the DEBUG macro in the `if ()`.

modules/database/src/ioc/db/dbConvertFast.h
```diff
@@ -19,8 +19,11 @@
 extern "C" {
 #endif
 
-DBCORE_API extern long (*dbFastGetConvertRoutine[DBF_DEVICE+1][DBR_ENUM+1])();
-DBCORE_API extern long (*dbFastPutConvertRoutine[DBR_ENUM+1][DBF_DEVICE+1])();
+struct dbAddr;
+typedef long (*FASTCONVERTFUNC)(const void *from, void *to, const struct dbAddr *paddr);
```
anjohnson: Core Group: Check build of external modules which are using this, if any. Search for the `dbFast{Get,Put}ConvertRoutine()` functions.

modules/database/src/ioc/dbStatic/link.h
```diff
@@ -77,8 +77,8 @@ struct macro_link {
     char *macroStr;
 };
 
-struct dbCommon;
-typedef long (*LINKCVT)();
+struct dbAddr;
+typedef long (*LINKCVT)(const void *, void *, const struct dbAddr *);
```
anjohnson: Is this one of the other typedef's that this PR already changes?

anjohnson (commented):
Look at these last points.

modules/database/src/ioc/dbStatic/link.h
anjohnson: Yes, `LINKCVT` is the same as `FASTCONVERTFUNC` from dbConvertFast.h shown in the previous point. The definition of that has to be moved here for the result to build. This change is now incorporated into 2nd commit after a rebase+squash.

modules/database/src/ioc/db/dbConvertFast.h
anjohnson: Looking in the epics-modules organization only `dbFastPutConvertRoutine` is found [here in autosave](https://github.com/epics-modules/autosave/blob/606903e177790c6431b277d4393700d9e5991b26/asApp/src/dbrestore.c#L80) and that's in a comment referring to code [added in 2004](https://github.com/epics-modules/autosave/commit/636baf20d6596d5cb84592b2ac04b04be052ae02) but [removed in 2017](https://github.com/epics-modules/autosave/commit/3a58565341d895f33d28ae0386697f2295542a82). Forks of autosave provide many references to this old code in a GitHub-wide search.

@klemenv's [PyDevice module](https://github.com/klemenv/PyDevice) is also using `dbFastPutConvertRoutine` from C++ code [here](https://github.com/klemenv/PyDevice/blob/ec1ef37a51f2ca3183c42baafa25c482e4822dc7/src/pycalcRecord.cpp#L210). I'm not sure if this change will break that code or not, but it looks like fixing it shouldn't be too hard with a version check for backwards compatibility.

Other uses of both dbFast{Get,Put}ConvertRoutine on GitHub all appear to be forks of Base or of the above.

modules/database/test/ioc/db/jlinkz.c
```diff
@@ -129,7 +129,7 @@ long z_putval(struct link *plink, short dbrType,
     if(nRequest==0) return 0;
 
     epicsMutexLock(priv->lock);
-    ret = (*pconv)(&priv->value, pbuffer, NULL);
+    ret = pconv(pbuffer, &priv->value, NULL);
```
anjohnson, line 129: Core group: Arg order change, is this right?

anjohnson: I believe my change here is correct, but incomplete. This routine doesn't actually get called by the `dbPutLinkTest` program at all, so we've never noticed that the code has a problem. The `lset::putValue()` method would be used by a `dbPutLink()` call in a record to send the `dbrType` value at `*pbuffer` through a `DBF_OUTLINK`, but we currently only test with input links and the `xRecord.c` doesn't call `dbPutLink()` anywhere.

For this link type `z_putval()` is trying to write the `*pbuffer` value to its own `epicsInt32 zpriv::value` member. However `dbFastPutConvertRoutine[DBF_LONG][dbrType]` looks up the function that converts from `DBF_LONG` to `dbrType` — both the Get and Put arrays use the order `[from][to]`, as do the functions arguments. The original code here looks to have been copied from the `z_getval()` method, but only the array used to look up the convert routine was changed.

New commits correct this routine, add tests, and provide some documentation.

anjohnson:
Core Group: The MSC-19 debug failure might be related to the code changes in this PR.

PR #457, opened 2024-01-19
Changes to start the EPICS diode instead of the IOC

ralphlange:
### !!! DO NOT MERGE !!!

This is only to show the changes that we had to do in iocInit.c to allow it to start the EPICS Diode engine instead of the regular IOC.

### DIFF
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -14,7 +14,7 @@
  *      Date:            06-01-91
  */
 
-
+
 #include <stddef.h>
 #include <stdlib.h>
 #include <stdarg.h>
@@ -98,17 +98,24 @@ static void iterateRecords(recIterFunc func, void *user);
 int dbThreadRealtimeLock = 1;
 epicsExportAddress(int, dbThreadRealtimeLock);
 
-enum iocStateEnum getIocState(void)
+enum iocStateEnum diode_getIocState(void)
 {
     return iocState;
 }
 
+int diode_iocBuild(void);
+int diode_iocRun(void);
+
+/* from devdiode.h */
+int diode_assign_fields();
+
 /*
  *  Initialize EPICS on the IOC.
  */
-int iocInit(void)
+int diode_iocInit(void)
 {
-    return iocBuild() || iocRun();
+    errlogPrintf("diode_iocInit: Using diode IOC initialization.\n");
+    return diode_iocBuild() || diode_assign_fields() || diode_iocRun();
 }
 
 static int iocBuild_1(void)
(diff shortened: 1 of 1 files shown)


modules/database/src/ioc/misc/iocInit.c
```diff
@@ -753,5 +766,61 @@ int iocShutdown(void)
...
+             "See more: iocBuild, iocPause"};
+static void iocRunCallFunc(const iocshArgBuf *args)
+{
+    iocshSetError(diode_iocRun());
+}
+
+/* iocPause */
+static const iocshFuncDef iocPauseFuncDef = {"diodeIocPause",0,NULL,
+             "Brings a running IOC to a quiescent state with all record processing frozen.\n"
+             "See more: iocBuild, iocRub, iocInit"};
```
anjohnson, line 804: What does rubbing an IOC do, is that like burping a baby? 😆🤓

anjohnson (commented):
I wonder whether we could combine these changes with the ones needed for unit testing by adding a bitmap to control whether specific subsystems get started & stopped by the routines in iocInit.c — I haven’t looked to see how much overlap there is there but I suspect there is some.

Alternatively, what about turning this into a C++ class with virtual functions which can be overloaded for specific stages? More work but might be cleaner, and now we have 3 examples of initialization to extract the requirements from.

It might be helpful if you could add an example of the diode’s st.cmd file to see how different it is.

ralphlange:
> It might be helpful if you could add an example of the diode’s st.cmd file to see how different it is.

Not very different, as far as I remember (basically calling `diodeIocInit` instead of `iocInit`), but I will ask for the file in our prototype.

modules/database/src/ioc/misc/iocInit.c
ralphlange: Nah.
Configuration of the Diode works like Aladdin's Magic Lamp: rub the IOC then say what you wish to be replicated.

ralphlange:
Nothing special, but there is diodeIocInit() call instead of iocInit():
```
cd ../../..

asSetFilename("cfg/readOnly.acf")

dbLoadDatabase("dbd/testDiode.dbd")
testDiode_registerRecordDeviceDriver(pdbbase)

# Trace = 0, Debug = 1, Config = 2, Info = 3 (default), Warning = 4, Error = 5 
diodeLogLevel(3)

# XPOZ side
dbLoadRecords("db/testDiodeX.db", "prefix=xpoz")

diodeReceiverStart("cfg/test_diode.json", 5080)

diodeIocInit()
```

Cheers,
Matej

ralphlange:
The idea is to start the "hollow" diode receiver IOC with the original record instances, same as they are on the inside. That will avoid having to send all the static record data through the one-way stream.

The list of `records.fields` to update is part of the diode configuration. All other fields will be left untouched.
Access security is configured for strict read-only access from the CA/PVA side.

PR #458, opened 2024-02-01
Accept should return SOCKET rather than int

FreddieAkeroyd:
Change return type and first parameter of `epicsSocketAccept()` from `int` to `SOCKET` to match its real data type - this removes a compiler warning on WIN32

### DIFF
--- a/modules/libcom/src/log/iocLogServer.c
+++ b/modules/libcom/src/log/iocLogServer.c
@@ -43,7 +43,7 @@ static char ioc_log_file_command[256];
 
 
 struct iocLogClient {
-    int insock;
+    SOCKET insock;
     struct ioc_log_server *pserver;
     size_t nChar;
     char recvbuf[1024];
--- a/modules/libcom/src/osi/os/WIN32/osdSock.c
+++ b/modules/libcom/src/osi/os/WIN32/osdSock.c
@@ -119,8 +119,8 @@ LIBCOM_API SOCKET epicsStdCall epicsSocketCreate (
     return socket ( domain, type, protocol );
 }
 
-LIBCOM_API int epicsStdCall epicsSocketAccept ( 
-    int sock, struct sockaddr * pAddr, osiSocklen_t * addrlen )
+LIBCOM_API SOCKET epicsStdCall epicsSocketAccept (
+    SOCKET sock, struct sockaddr * pAddr, osiSocklen_t * addrlen )
 {
     return accept ( sock, pAddr, addrlen );
 }
--- a/modules/libcom/src/osi/os/posix/osdSock.c
+++ b/modules/libcom/src/osi/os/posix/osdSock.c
@@ -124,8 +124,8 @@ LIBCOM_API SOCKET epicsStdCall epicsSocketCreate (
     return sock;
 }
 
-LIBCOM_API int epicsStdCall epicsSocketAccept ( 
-    int sock, struct sockaddr * pAddr, osiSocklen_t * addrlen )
+LIBCOM_API SOCKET epicsStdCall epicsSocketAccept (
+    SOCKET sock, struct sockaddr * pAddr, osiSocklen_t * addrlen )
 {
 #ifndef HAVE_SOCK_CLOEXEC
     int newSock = accept ( sock, pAddr, addrlen );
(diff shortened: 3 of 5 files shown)


anjohnson:
Core Group: In principle we approve these changes, but could they break any downstream modules such as Asyn on Windows? If that just generates warnings there that's okay. We're leaning towards merging, but would like to confirm that Asyn still builds.

dirk-zimoch:
Asyn uses `epicsSocketAccept()` twice and stores the result in an `int`. Need to check how it compiles when this PR is applied.

dirk-zimoch:
As expected, Linux builds of asyn  have no problem with this change because there `SOCKET` expands to `int`.
But also as expected, Windows now shows the warning:
```
asyn\drvAsynSerial\drvAsynIPServerPort.c(325): warning C4244: '=': conversion from 'SOCKET' to 'int', possible loss of data
```
The `epicsSocketAccept` call in asyn/vxi11/drvVxi11.c is not triggered because Windows gets its own (non-functional) implementation drvVxi11Win32.c.

The impact is probably more severe than a simple warning, because in Windows `SOCKET` is a 64 bit value on 64 bit Windows while an `int` has only 32 bits:
```
typedef unsigned __int64    UINT_PTR;
typedef UINT_PTR        SOCKET;
```
Thus, passing a `SOCKET` to an `int` may really corrupt the value! But this PR will not make the situation any worse as now the client code, such as asyn, can fix it. Before the socket value was already corrupted in `epicsSocketAccept()`, even if the client code had used `SOCKET`. Approved.

Did drvAsynIPServerPort ever work on Windows?

dirk-zimoch:
Looking for more warnings on Windows related to `SOCKET`, I found fdManager.cpp line 121, where a `SOCKET maxFD` is passed as first argument to `select(int nfds, ...)` (which is ignored by Windows, see https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select).

Also some code (e.g. iocLogServer.c but also drvAsynIPServerPort.c) attempts to print sockets in debug messages using `"%d"`.
This again causes a warning on Windows:
```
modules\libcom\src\log\ioclogserver.c(558): warning C4477: 'fprintf' : format string '%d' requires an argument of type 'int', but variadic argument 3 has type 'SOCKET'
```
I suggest to add a `PRISOCKET` macro (like the macros in inttypes.h) that expands to `"Iu"` (or `"Id"` because then `INVALID_SOCKET` == `~0` would show as `-1`?) on Windows and to `"d"` everywhere else. (E.g. the default in osiSock.h and the special case in os/WIN32/osdSock.h.)

This would allow code to print sockets like this:
`printf ("socket = %" PRISOCKET "\n", sock);`
and then make iocLogServer.c use it.

dirk-zimoch:
ca/src/client/access.cpp complains too about calling `CAFDHANDLER` with a socket. `CAFDHANDLER` is defined in cadef.h and should be changed to:
```
 typedef void CAFDHANDLER (void *parg, SOCKET fd, int opened);
```
(Needs to `#include "osiSock.h"` to do so.)

And then its use should to be fixed in acctst.c and fdmgrTest.c.
And the documentation in CAref.html.

dirk-zimoch:
See my additional changes here: https://github.com/FreddieAkeroyd/epics-base/pull/2

FreddieAkeroyd:
A `SOCKET` on windows can be used as a `HANDLE` https://learn.microsoft.com/en-us/windows/win32/winsock/socket-handles-2 and I believe only 32bits of a `HANDLE` are actually significant  https://stackoverflow.com/questions/18266626/what-is-the-range-of-a-windows-handle-on-a-64-bits-application so I think that then means that passing a `SOCKET` to an `int` will not (currently at least) cause corruption as no significant bits will be lost. I will check  `drvAsynIPServerPort.c` later and report back, thank you for your other changes @dirk-zimoch

dirk-zimoch:
> only 32bits of a HANDLE are actually significant

Good to know. In that case, casting to `int` seems safe and the warnings can be ignored. Also drvAsynIPServerPort should have no problem with either version. I tested with gcc that casting `~0` (== `-1`) == `INVALID_SOCKET` to `int` and back to an unsigned 64 bit integer (`SOCKET` in 64 bit Windows) gives the same value. But please test that cl is happy with that, too.

anjohnson:
Core group: Looks good, thanks for all the analysis.

PR #459, opened 2024-02-02
iocsh: allow history and history file size to be limited

JJL772:
Adds `EPICS_IOCSH_HISTSIZE` and `EPICS_IOCSH_HISTFILESIZE`

Figured this might be useful in some cases.

`EPICS_IOCSH_HISTSIZE` defines the number of history items to keep in memory, limited by stifle_history. Values of < 0 mean history is unlimited. This is not recommended for obvious reasons. Non-numeric values are simply ignored and the default is used instead.

`EPICS_IOCSH_HISTFILESIZE` defines the number of history items to store on disk. Values < 0 or non-numeric values remove any limit on this file's size. Values >= 0 truncate the history file to that many lines. Note that a value of 0 will simply generate an empty history file.

This behavior mirrors bash's HISTFILESIZE and HISTSIZE (but is not 1:1)

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -642,7 +642,12 @@ struct ReadlineContext {
                         fprintf(stderr, ERL_ERROR " %s (%d) loading '%s'\n",
                                 strerror(err), err, hist_file.c_str());
                 }
-                stifle_history(1024); // some limit...
+                // Limit history to 1024 entries, unless more (or less) is explicitly asked for via the env. values < 0 mean infinite history
+                epicsInt32 maxHistory = 1024;
+                if (const char* env = getenv("EPICS_IOCSH_HISTSIZE"))
+                    epicsParseInt32(env, &maxHistory, 10, NULL);
+                if (maxHistory >= 0)
+                    stifle_history(maxHistory); // some limit...
             }
         }
 #endif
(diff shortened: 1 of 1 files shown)


modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -653,10 +658,21 @@ struct ReadlineContext {
         if(context) {
 #ifdef USE_READLINE
             if(!hist_file.empty()) {
+                // Allow history file size to be limited. <0 or non-numeric value means infinite size, >=0 means truncate to that many lines
```
simon-ess, line 661: Just to be clear that I understand this: `write_history` overwrite the current history file.  So even if the history file is allowed to be infinite in size, we will only keep history from the previous IOC session. Is that correct? Moreover, is that the desired behaviour?

(That said, this is new functionality with 7.0.8, so it is sort of up to us collectively what "correct behaviour" is...)

simon-ess (commented):
One further comment: we now how two different variables that mean almost exactly the same thing, `EPICS_IOCSH_HISTSIZE` and `IOCSH_HISTSIZE`. Perhaps these should be joined...

(see https://github.com/epics-base/epics-base/blob/7.0/configure/CONFIG_SITE_ENV#L75 and its usage https://github.com/epics-base/epics-base/blob/7.0/modules/libcom/src/osi/os/default/gnuReadline.c#L50)

modules/libcom/src/iocsh/iocsh.cpp
mdavidsaver: > ... `write_history` overwrite the current history file ...

Yes.



> ... we will only keep history from the previous IOC session ...

... which would be initialized with the previous contents of that same history file.  (excepting `chdir()` as I think about it)

There is also an `append_history()` function.  I wonder if that would result in duplication?

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -653,10 +658,21 @@ struct ReadlineContext {
...
+                epicsInt32 maxLines = -1;
+                if (const char* env = getenv("EPICS_IOCSH_HISTFILESIZE"))
+                    epicsParseInt32(env, &maxLines, 10, NULL);
```
mdavidsaver, line 664: Need to handle parse errors.  eg. test with `EPICS_IOCSH_HISTFILESIZE=10o`, which should warn and use the default, not `10`.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -653,10 +658,21 @@ struct ReadlineContext {
...
+
                 if(int err = write_history(hist_file.c_str())) {
                     fprintf(stderr, ERL_ERROR " %s (%d) writing '%s'\n",
                             strerror(err), err, hist_file.c_str());
                 }
+                else if (maxLines >= 0) {
+                    if (int err = history_truncate_file(hist_file.c_str(), maxLines)) {
```
mdavidsaver, line 671: Why not `stifle_history(maxLines)` before `write_history()`?

mdavidsaver (changes requested):

anjohnson:
Core Group: Changes requested but not yet done; there is a hard-coded limit of 1024 lines for at least one of these. Converting to draft for now.

PR #460, opened 2024-02-11
WIN32: Move call to setThreadName()

FreddieAkeroyd:
The call to `setThreadName()` is moved to avoid a race condition that can happen with very short lived processes. If the process terminates very quickly e.g. is a google test runner or the `msi.exe` command called from a Makefile during a build, then very occasionally a crash can occur during process termination in `setThreadName()` when called from the newly created thread. This looks to be because the DLL it is trying to call gets unloaded between it getting a handle to the DLL and making the call. Moving the `setThreadName()` call to the creating thread avoids this problem, which probably stemmed from the delay between the thread being created and it subsequently running `setThreadName()`.  The issue was only ever seen with statically linked epics executables, I am unsure if the way a DLL based epics program unloads might avoid this, or just make it less likely but still possible. As mentioned above, the issue would only ever have occurred to threads that are created during process termination and so would not affect running IOCs

### DIFF
--- a/modules/libcom/src/osi/os/WIN32/osdThread.c
+++ b/modules/libcom/src/osi/os/WIN32/osdThread.c
@@ -500,8 +500,6 @@ static unsigned WINAPI epicsWin32ThreadEntry ( LPVOID lpParameter )
     BOOL success;
 
     if ( pGbl )  {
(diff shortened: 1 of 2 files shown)


anjohnson:
Core Group: Approved. Thanks for finding this weird corner-case!

PR #461, opened 2024-02-12
Fix issue where VSCode makefile extension can delete files

### LINKED ISSUE #456: make --always-make deletes some files
This issue was quite a bit fun to debug!

Turns out VSCode recommends installing the [Makefile tools](https://github.com/Microsoft/vscode-makefile-tools/) extension, which runs this command-line to get some info on the project:

```bash
make --dry-run --always-make --keep-going --print-directory
```

This command-line on an EPICS top removes all `CONFIGS` files during the installation phase. From my debugging, the files are removed by the `installEpics.pl`. Auditd says this command removed the `configure/CONFIG` file:

```bash
perl -CSD /nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/bin/linux-x86_64/installEpics.pl -d -m 444 ../CONFIG ../.
```

And `installEpics.pl` says:

```
installEpics.pl: Copy failed: No such file or directory
installEpics.pl:        ../CONFIG -> ../../configure/TEMP.CONFIG.1267659
```

<details>
<summary>Full make logs</summary>

```
make: Entering directory '/tmp/tmp.69OuGuGda6'
make -C ./configure install
make[1]: Entering directory '/tmp/tmp.69OuGuGda6/configure'
perl -CSD /nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/bin/linux-x86_64/makeMakefile.pl O.linux-x86_64 ../..
mkdir -p O.Common
make -C O.linux-x86_64 -f ../Makefile TOP=../.. \
    T_A=linux-x86_64 install
make[2]: Entering directory '/tmp/tmp.69OuGuGda6/configure/O.linux-x86_64'
Installing config file ../../configure/CONFIG
installEpics.pl: Copy failed: No such file or directory
installEpics.pl:        ../CONFIG -> ../../configure/TEMP.CONFIG.1267659
make[2]: *** [/nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/configure/RULES_BUILD:539: ../../configure/CONFIG] Error 2
Installing config file ../../configure/RELEASE
installEpics.pl: Copy failed: No such file or directory
installEpics.pl:        ../RELEASE -> ../../configure/TEMP.RELEASE.1267661
make[2]: *** [/nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/configure/RULES_BUILD:539: ../../configure/RELEASE] Error 2
Installing config file ../../configure/CONFIG_SITE
installEpics.pl: Copy failed: No such file or directory
installEpics.pl:        ../CONFIG_SITE -> ../../configure/TEMP.CONFIG_SITE.1267663
make[2]: *** [/nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/configure/RULES_BUILD:539: ../../configure/CONFIG_SITE] Error 2
Installing config file ../../configure/RULES
installEpics.pl: Copy failed: No such file or directory
installEpics.pl:        ../RULES -> ../../configure/TEMP.RULES.1267665
make[2]: *** [/nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/configure/RULES_BUILD:539: ../../configure/RULES] Error 2
../Makefile:3: Failed to remake makefile '../../configure/CONFIG'.
make[2]: Leaving directory '/tmp/tmp.69OuGuGda6/configure/O.linux-x86_64'
../../configure/CONFIG:9: Failed to remake makefile '../../configure/RELEASE'.
../../configure/CONFIG:33: Failed to remake makefile '../../configure/CONFIG_SITE'.
../Makefile:8: Failed to remake makefile '../../configure/RULES'.
make[1]: *** [/nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/configure/RULES_ARCHS:58: install.linux-x86_64] Error 2
make[1]: Leaving directory '/tmp/tmp.69OuGuGda6/configure'
make: *** [/nix/store/3jdcplbpdj2177csm7nbw3prjlaq1bdn-epics-base-7.0.7/configure/RULES_DIRS:85: configure.install] Error 2
make: Leaving directory '/tmp/tmp.69OuGuGda6'
```

</details>

simon-ess:
The problem is that VSCode's make extension, in order to determine some information about the project, runs
```
make --dry-run --always-make
```
which despite its name will actually try to remake the configure/* files. Running `installEpics.pl` on these will delete them first, then try copy them, resulting in an error.

Fixes #456

### DIFF
--- a/configure/RULES_BUILD
+++ b/configure/RULES_BUILD
@@ -532,11 +532,11 @@ endif # LOADABLE_SHRLIB_SUFFIX
 ifneq ($(INSTALL_CONFIGS),)
 $(INSTALL_CONFIG)/%: %
 	$(ECHO) "Installing config file $@"
-	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $< $(@D)
+	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $< $(@D))
 
 $(INSTALL_CONFIG)/%: ../%
 	$(ECHO) "Installing config file $@"
-	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $< $(@D)
+	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $< $(@D))
 endif
 
 $(INSTALL_INCLUDE)/%: $(COMMON_DIR)/%
--- a/src/tools/installEpics.pl
+++ b/src/tools/installEpics.pl
@@ -57,6 +57,9 @@ foreach my $source (@ARGV) {
     my $temp   = "$install_dir/TEMP.$name.$$";
     my $target = "$install_dir/$name";
 
+    # Don't try to install the file if it already exists
+    next if $source eq $target;
+
     if (-f $target) {
         next if -M $target < -M $source and -C $target < -C $source;
         # Remove old target, making sure it is deletable first


src/tools/installEpics.pl
```diff
@@ -57,19 +57,21 @@
     my $temp   = "$install_dir/TEMP.$name.$$";
     my $target = "$install_dir/$name";
 
-    if (-f $target) {
-        next if -M $target < -M $source and -C $target < -C $source;
-        # Remove old target, making sure it is deletable first
-        chmod 0777, $target;
-        unlink $target;
+    if ($source ne $target) {
```
anjohnson: I assume you’ve tested this and it resolves the original bug, but if `$target` is the absolute path version of `$source` they could still refer to the same file without the code knowing it. I think that could be triggered by setting `INSTALL_LOCATION` to `$(abspath $(TOP))` or the manual equivalent in the `CONFIG_SITE` file if you want to try it.

src/tools/installEpics.pl
```diff
@@ -57,19 +57,21 @@
...
 
-    # Using copy + rename fixes problems with parallel builds
-    copy($source, $temp) or die "$tool: Copy failed: $!\n" .
-        "$tool:\t$source -> $temp\n";
-    rename $temp, $target or die "$tool: Rename failed: $!\n" .
-        "$tool:\t$temp -> $target\n";
-
     # chmod 0555 <read-only> DOES work on Win32, but the above
     # chmod 0777 fails to install a newer version on top.
     chmod $mode, $target unless $^O eq 'MSWin32';
```
anjohnson: Should it be running this line if the source and target files are the same? That would probably make all the configure/ files read-only wouldn’t it?

anjohnson (commented):
I suggest just short-circuiting the `for ()` loop with
```Perl
next if $source eq $target;
```
but it might have to convert both paths to an absolute path before comparing them, see detailed comment. There are functions in the `EPICS::Path’ module for that (I think that’s its name).

src/tools/installEpics.pl
simon-ess: That's a fair point. I was mostly going off of @minijackson's comment (https://github.com/epics-base/epics-base/issues/456#issuecomment-1900765902), but I think in hindsight you are correct. This fits better with yoru simple short-circuit of the loop as well. I will update everything.

src/tools/installEpics.pl
simon-ess: I have switched to comparing absolute paths as you suggested. That said, it seemed that I need to directly use `abs_path` from `Cwd` instead of the EPICS one, which (as I understand) is installed by `installEpics.pl`.

I believe that this is fine; the limitations described in `EPICS::Path` about `abs_path` shouldn't apply here as the `source` should already exist, and the target directory is created just above.

simon-ess:
So unfortunately `abs_path` doesn't work as desired on older windows/perl combinations, see for example https://ci.appveyor.com/project/epics-base-7/epics-base/builds/49183965/job/wijpfyuy59158v8b.  Note that the failing ones all seem to have perl version 5.24 or less, while the passing ones that I have looked at have perl 5.32.

It seems that this behaviour might have been fixed in perl 5.28, see https://github.com/perl/perl5/issues/12693#issuecomment-586031602

I could switch to using `File::Spec->rel2abs`, but I am not sure that this is going to be the correct behaviour as this does not resolve all of the `../`s and `./`s. I guess I should give it a try...

anjohnson:
Another option might be to see whether the `$(abspath FILE)` function in GNU Make could be used in the build rules instead of doing the absoluting in the Perl script. It is described thusly:
```
'$(abspath NAMES...)'
     For each file name in NAMES return an absolute name that does not
     contain any '.' or '..' components, nor any repeated path
     separators ('/').  Note that, in contrast to 'realpath' function,
     'abspath' does not resolve symlinks and does not require the file
     names to refer to an existing file or directory.  Use the
     'wildcard' function to test for existence.
```

It doesn't completely solve the issue if someone is using soft-links, but maybe we could ignore that?

simon-ess:
That can be made to work, although it's a bit weird as make's `abspath` function has to be moved to the `RULES_*` files. The minimal change would be to do something like
```make
ifneq ($(INSTALL_CONFIGS),)
$(INSTALL_CONFIG)/%: %
	$(ECHO) "Installing config file $@"
	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $<) $(abspath $(@D))

$(INSTALL_CONFIG)/%: ../%
	$(ECHO) "Installing config file $@"
	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $<) $(abspath $(@D))
endif
```
in `RULES_BUILD` (i.e only modify the `INSTALL_CONFIGS` rules, since those are the affected ones. This feels _somewhat_ artificial, but it also feels weird to dump an `abspath` call into every build rule.

configure/RULES_BUILD
```diff
@@ -532,11 +532,11 @@ endif # LOADABLE_SHRLIB_SUFFIX
 ifneq ($(INSTALL_CONFIGS),)
 $(INSTALL_CONFIG)/%: %
 	$(ECHO) "Installing config file $@"
-	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $< $(@D)
+	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $<) $(abspath $(@D))
```
anjohnson: The `$(abspath ...)` function accepts multiple inputs, so the command could be shortened to just this:
```makefile
	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $< $(@D))
```

This doesn't support source or installation paths containing spaces, but I doubt that this is the only place where we'd need to add double-quotes around filenames if we wanted that.

configure/RULES_BUILD
```diff
@@ -532,11 +532,11 @@ endif # LOADABLE_SHRLIB_SUFFIX
...
 
 $(INSTALL_CONFIG)/%: ../%
 	$(ECHO) "Installing config file $@"
-	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $< $(@D)
+	@$(INSTALL) -d -m $(INSTALL_PERMISSIONS) $(abspath $<) $(abspath $(@D))
```
anjohnson: Ditto.

anjohnson (approved):
This is much cleaner now, with one suggestion for shortening it even further below.

configure/RULES_BUILD
simon-ess: Good catch. I was thinking of abspath implicitly working that way for all of the targets `$<` (although I believe in this case that will never be more than one), but I didn't think of putting the directory there too. Fixed.

anjohnson (approved):
Thanks!

anjohnson:
Core Group: Approved.

PR #464, opened 2024-02-21
Allow deleting a record at database creation.

hinxx:
Using a magical record type `"#"` will allow the user to delete previously created record  from the database.

I would find this particularly useful in cases where the initial db/template file that is loaded contains records that I do not want/need later on when additional db/template is loaded. It is kind of like the `"*"` magical record type that allows changing th record fields.

For example, if this is the `/tmp/db.db` file:
```
record(ai, "foo") { }

record(ao, "bar") { }
```

and this is the `/tmp/db2.db` file:
```
# this will delete the record foo from the DB
record("#", "foo") { }
```

then loading them would result in:
```
$ ./bin/linux-x86_64/softIoc 
epics> dbLoadRecords /tmp/db.db 
epics> dbl
foo
bar
epics> dbLoadRecords /tmp/db2.db 
WARNING: Record "foo" DELETED!
epics> dbl
bar
```

### DIFF
--- a/modules/libcom/src/osi/os/WIN32/osdThread.c
+++ b/modules/libcom/src/osi/os/WIN32/osdThread.c
@@ -500,8 +500,6 @@ static unsigned WINAPI epicsWin32ThreadEntry ( LPVOID lpParameter )
     BOOL success;
 
     if ( pGbl )  {
-        setThreadName ( pParm->id, pParm->pName );
-
         success = FlsSetValue ( pGbl->flsIndexThreadLibraryEPICS, pParm );
         if ( success ) {
             osdThreadHooksRun ( ( epicsThreadId ) pParm );
@@ -659,6 +657,7 @@ epicsThreadId epicsThreadCreateOpt (
         pParmWIN32->id = ( DWORD ) threadId ;
     }
 
+    setThreadName ( pParmWIN32->id, pParmWIN32->pName );
     osdPriority = epicsThreadGetOsdPriorityValue (opts->priority);
     bstat = SetThreadPriority ( pParmWIN32->handle, osdPriority );
     if (!bstat) {
--- a/modules/libcom/src/osi/os/WIN32/osdTime.cpp
+++ b/modules/libcom/src/osi/os/WIN32/osdTime.cpp
@@ -215,6 +215,7 @@ void currentTime :: startPLL ()
                 CREATE_SUSPENDED | STACK_SIZE_PARAM_IS_A_RESERVATION,
                 & this->threadId );
         assert ( this->threadHandle );
+        setThreadName ( this->threadId, "EPICS Time PLL" );
         BOOL bstat = SetThreadPriority (
             this->threadHandle, THREAD_PRIORITY_HIGHEST );
         assert ( bstat );
@@ -496,7 +497,6 @@ static unsigned __stdcall _pllThreadEntry ( void * pCurrentTimeIn )
 {
     currentTime * pCT =
         reinterpret_cast < currentTime * > ( pCurrentTimeIn );
-    setThreadName ( pCT->threadId, "EPICS Time PLL" );
     while ( ! pCT->threadShutdownCmd ) {
         Sleep ( currentTime :: pllDelay * 1000 /* mS */ );
         pCT->updatePLL ();


anjohnson:
Interesting idea, I like it!

Would the IOC need any kind of protection against this somehow getting run after iocInit? That isn’t legal, but I don’t remember what happens if you actually try that. I vaguely remember someone suggesting that it’s possible to load or define aliases after iocInit (was that @mdavidsaver)?

Another question, how does this removal interact with record aliases? 2 scenarios to consider: (1) The record name given is actually that of an alias, and (2) there are one or more aliases also defined for that record.

hinxx:
Thanks for the input @anjohnson ! Below are answers to your questions.

> Would the IOC need any kind of protection against this somehow getting run after iocInit? 

Seems that is already taken care of:
```

$ ./bin/linux-x86_64/softIoc 
epics> dbLoadRecords /tmp/db.db 
epics> iocInit 
epics> dbLoadRecords /tmp/db2.db 
ERROR failed to load '/tmp/db2.db'
    Records cannot be loaded after iocInit!
epics> dbl
foo
bar
```

In regards to aliases after `iocInit`, if `/tmp/db4.db` contains:

```
alias("foo", "duck")
```

trying to load it after `iocInit` results in similar error:

```
epics> dbLoadRecords /tmp/db4.db 
ERROR failed to load '/tmp/db4.db'
    Records cannot be loaded after iocInit!
```

> (1) The record name given is actually that of an alias

I added an alias to the original `/tmp/db.db` that now looks like this:
```
record(ai, "foo") { }

record(ao, "bar") { }

alias("foo", "dodo")
```

If we consider that `/tmp/db3.db` contains the following:
```
# this will delete the record alias dodo from the DB
record("#", "dodo") { }
```

then the result is as follows:

```
$ ./bin/linux-x86_64/softIoc 
epics> dbLoadRecords /tmp/db.db 
epics> dbl
foo
dodo
bar
epics> dbLoadRecords /tmp/db3.db 
WARNING: Record "dodo" DELETED!
epics> dbl
foo
bar
```

The real record remains, but the alias is gone. That seems expected and allows removing aliases without impacting the real record.

> (2) there are one or more aliases also defined for that record

Loading the same `/tmp/db.db` and then loading /tmp/db2.db` that deletes the real record that now has an alias looks like this:
```
$ ./bin/linux-x86_64/softIoc 
epics> dbLoadRecords /tmp/db.db 
epics> dbl
foo
dodo
bar
epics> dbLoadRecords /tmp/db2.db 
WARNING: Record "foo" DELETED!
epics> dbl
bar
```

Both, the read record and the alias, are gone. I believe this is what should happen to keep the DB sane.

anjohnson:
Core Group: We like this, it still needs an entry in the Release Notes, and a PR to the read-the-docs repo to document it there too, ask @ttkorhonen for help working out where.

hinxx:
Thanks @anjohnson !
Just want to make sure who needs to do what for this PR to be merged.
Am I supposed to write the Release Notes?
Am I supposed to create another PR for the read-the-docs repo (I'm not sure what that is)?

ericonr:
> read-the-docs repo (I'm not sure what that is)?

@hinxx I believe it's this one: https://github.com/epics-docs/epics-docs

hinxx:
> it still needs an entry in the Release Notes

@anjohnson should I edit the `documentation/RELEASE_NOTES.md` and add a line or two ? If so where?

Something like:

### Allow the user to delete previously created record from the database

**Previously defined record can be removed from the database using `"#"` in place of the record type in the record instance. This is useful when previously loaded records are not required or desired.**

anjohnson:
> Should I edit the `documentation/RELEASE_NOTES.md` and add a line or two ? If so where?

Yes please, new items go near the top, just under `## Changes made on the 7.0 branch since …`. Unfortunately this usually causes a git merge conflict since all new entries get bunched up together, so we have to perform those PR merges by hand. We haven't worked out a better way to handle this yet.

> Something like:

Here's my suggestion with a bit more explanation in it:

### Allow users to delete previously created records from the database

From this release, record instances and aliases that have already been loaded by an IOC can be removed from the database again before the call to `iocInit` by loading a second instance of the named records but using `"#"` in place of the record type. Values for the fields are not required or advised, just use an empty record body `{ }`. This is useful when a template defines records that are not wanted in some IOCs, without having to split or duplicate the original template.

ericonr:
> Yes please, new items go near the top, just under ## Changes made on the 7.0 branch since …. Unfortunately this usually causes a git merge conflict since all new entries get bunched up together, so we have to perform those PR merges by hand. We haven't worked out a better way to handle this yet.

I think the best way to handle this that I've seen is how CPython does it. Each PR should include a blurb in the relevant https://github.com/python/cpython/tree/main/Misc/NEWS.d/next directory, and when there is a release, the release notes are automatically generated by pooling all changes together.

anjohnson:
Too many cooks? This branch now needs rebasing against the latest 7.0.

@ericonr I was thinking about something along those lines. I'm adding a new issue to discuss this further, can't give you the number until I submit it though.

ttkorhonen:
For once, documentation would be ahead of the code releases ;-)

ralphlange:
> @ericonr I was thinking about something along those lines. I'm adding a new issue to discuss this further, [...]
+1

hinxx:
Apologies for closing this PR, but I was not able to make a clean rebase against the latest 7.0 branch code.

Please see #505 as a replacement PR.

Issue #429, opened 2023-10-03
`recGblRecordError` prints "ugly" messages


dirk-zimoch:
When called with `status=-1` as for example by `scanAdd()`, `recGblRecordError()` prints messages like this:
```
recGblRecordError: scanAdd: I/O Intr not valid Error (65535,65535) PV: <PVNAME>
```
The "ugly" `Error (65535,65535)` is caused by the fact that `recGblRecordError()` skips `errSymLookup()` only for `status==0` but not for `status==-1`.
The same is true for `recGblDbaddrError()` and `recGblRecSupError()` but at least nothing is base calls those with `-1` for the `status` argument .
`recGblRecordError()` is called with `-1` only from dbScan.c (at least in base).

mdavidsaver:
Addressed by 4c205188640502806c7747a9ebe7c0239ffcb468

PR #465, opened 2024-02-26
solving (recGblRecordError prints "ugly" messages) issue

seifalrahman:
modifying the condition from (status!=0) to (status>0) to skip the block in case the status variable ==-1


anjohnson:
Core group: This is a fix for #429. Triggering CI.

dirk-zimoch:
Maybe rebase to fix CI problems.

PR #467, opened 2024-03-13
Fix off by one error in constant link fetch

simon-ess:
For long string buffers, we currently write a null terminator one byte past the end of the buffer. This can be seen with a record of the type

```
record(aai, foo) {
  field(NELM, 1)
  field(FTVL, CHAR)
  field(INP, {const: "foo"})
}
```
where the buffer is only of size 1, but then we write at index 1 (aka past the end of the buffer).

### DIFF
--- a/modules/database/src/std/link/lnkConst.c
+++ b/modules/database/src/std/link/lnkConst.c
@@ -495,9 +495,11 @@ static long lnkConst_loadArray(struct link *plink, short dbrType, void *pbuffer,
         }
         else {
             /* Long string conversion */
-            strncpy(pbuffer, clink->value.scalar_string, *pnReq);
-            ((char *)pbuffer)[*pnReq] = 0;
-            nElems = strlen(pbuffer) + 1;
+            if (*pnReq > 0) {
+                strncpy(pbuffer, clink->value.scalar_string, *pnReq);
+                ((char *)pbuffer)[*pnReq - 1] = 0;
+                nElems = strlen(pbuffer) + 1;
+            }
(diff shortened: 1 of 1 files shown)


modules/database/src/std/link/lnkConst.c
```diff
@@ -496,7 +496,7 @@ static long lnkConst_loadArray(struct link *plink, short dbrType, void *pbuffer,
         else {
             /* Long string conversion */
             strncpy(pbuffer, clink->value.scalar_string, *pnReq);
-            ((char *)pbuffer)[*pnReq] = 0;
+            ((char *)pbuffer)[*pnReq - 1] = 0;
```
anjohnson: Core group: Check for the `*pnReq == 0` case?

simon-ess: So `pnReq` is the size of the buffer you are trying to write into. For something like an `aSub` record (the case where we saw this issue), the initialisation forces that the buffer has size at least one, so this cannot happen.

However, I see no reason why a custom record would have to behave that way; I could in theory have a custom record with an input link whose buffer size can be zero as well. What is the correct behaviour in that case? My inclination is that we should drop and out set `status` to 1, but I am not confident about that.

simon-ess:
As discussed, I have wrapped the block of code in a check to ensure that we do not have a buffer underflow error.

mdavidsaver (approved):
I think this looks correct now.

ralphlange (approved):
Agree.

PR #468, opened 2024-03-14
Fix issue with compress record

simon-ess:
The handling of N-to-M array compression was broken with the addition of the partial buffer option, which broke the bounds check that was being used.

Note that this also makes the partial buffer option more consistent; if, for example, you have
```
record(compress, foo) {
  field(ALG, "N to 1 Average")
  field(INP, "bar NPP")
  field(NSAM, 2)
  field(N, 2)
  field(PBUF, YES)
}
```
(with bar having, e.g., length 3) then this will now behave as expected on both of the samples.

### DIFF
--- a/modules/database/src/std/rec/compressRecord.c
+++ b/modules/database/src/std/rec/compressRecord.c
@@ -149,13 +149,16 @@ static int compare(const void *arg1, const void *arg2)
     else               return  1;
 }
 
+#define min(a, b) ((a) < (b) ? (a) : (b))
+
 static int compress_array(compressRecord *prec,
     double *psource, int no_elements)
 {
-    epicsInt32 i,j;
+    epicsInt32 j;
     epicsInt32 n, nnew;
     epicsInt32 nsam = prec->nsam;
-    double value;
+    epicsUInt32 samples_written = 0;
+    double value = 0.0;
 
     /* skip out of limit data */
     if (prec->ilil < prec->ihil) {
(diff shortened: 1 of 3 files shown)


modules/database/src/std/rec/compressRecord.c
```diff
@@ -149,13 +149,16 @@ static int compare(const void *arg1, const void *arg2)
     else               return  1;
 }
 
+#define min(a, b) ((a) < (b) ? (a) : (b))
```
anjohnson, line 152: Duplicate.

anjohnson (commented):
Core Group: Please add Release note too about the fix.

modules/database/src/std/rec/compressRecord.c
simon-ess: I actually can't find another definition anywhere for this. The only spot I can find similar definitions are:

https://github.com/epics-base/epics-base/blob/7.0/modules/libcom/src/flex/flexdef.h#L70

https://github.com/epics-base/epics-base/blob/7.0/modules/database/src/ioc/db/dbTest.c#L57

neither of which seems a good fit to try to include here.

ralphlange: Looking at https://stackoverflow.com/questions/3437404/min-and-max-in-c leaves me confused...

anjohnson: In epicsAlgorithm.h we defined `epicsMin()` (which is now deprecated in favor of `std::min()`), and Jeff Hill's even older tsMinMax.h defined `tsMin()`, but they are of course both for use from C++ only — sorry, my mistake there!

simon-ess:
@anjohnson - done.

PR #470, opened 2024-03-22
Memory leaks

simon-ess:
While running valgrind, we saw reports of a few memory leaks. One of these is real, the other ones are somewhat fictitious. However, the allocation and cleanup of tasks on threads that have no exit condition is somewhat inconsistent, so this is an attempt to make it more future-proof and consistent.

### DIFF
--- a/modules/database/src/ioc/as/asCa.c
+++ b/modules/database/src/ioc/as/asCa.c
@@ -225,6 +225,11 @@ static void asCaTask(void)
         if(asCaDebug) printf("asCaTask has cleared all channels\n");
         epicsEventSignal(asCaTaskWait);
     }
+
+    /* ATM never reached, just a placeholder */
+    cantProceed("Unreachable.  Perpetual thread.");
+
+    taskwdRemove(0);
 }
 
 void asCaStart(void)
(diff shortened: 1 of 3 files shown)


modules/database/src/ioc/as/asCa.c
```diff
@@ -225,6 +225,10 @@ static void asCaTask(void)
         if(asCaDebug) printf("asCaTask has cleared all channels\n");
         epicsEventSignal(asCaTaskWait);
     }
+
+    /* ATM never reached, just a placeholder */
+
+    taskwdRemove(0);
```
mdavidsaver, line 232: It might be clearer to add something like `cantProceed("Unreachable.  Perpetual thread.");` in addition.

modules/database/src/ioc/rsrv/caservertask.c
```diff
@@ -748,6 +752,7 @@ void rsrv_init (void)
         if(!havesometcp)
             cantProceed("CAS: No TCP server started\n");
     }
+    free(socks);
```
mdavidsaver, line 756: Yup. this is needed.

mdavidsaver (approved):

PR #478, opened 2024-04-25
Ignore .cache/ and compile_commands.json

JJL772:
* `.cache/` is generated by clangd
* `compile_commands.json` is a (usually) generated file consumed by various tools
* `.ccls-cache` is generated by [ccls](https://github.com/MaskRay/ccls) as noted by @exzombie


exzombie:
Please add also `.ccls-cache` which is created by [ccls](https://github.com/MaskRay/ccls)

mdavidsaver:
I typically [ignore](https://git-scm.com/docs/gitignore) editor/IDE related debris with entries in `.git/info/exclude` or more broadly `~/.config/git/ignore` since there is sooooo much variation.

JJL772:
Ah interesting, I didn't realize that was possible

anjohnson:
Maybe instead of adding those patterns, you could add a comment which links to the page @mdavidsaver pointed to, something like:

```
# Before adding patterns here, please read the gitignore
# documentation at https://git-scm.com/docs/gitignore
```

exzombie:
Oooh, that would be nice. Adding tooling-related entries is a pretty common practice, and it seems to me a comment like this would quite efficiently teach people about alternatives. Although it should probably provide more information on what exactly the reader should look for in the docs.

JJL772:
> Maybe instead of adding those patterns, you could add a comment which links to the page @mdavidsaver pointed to, something like:
> 
> ```
> # Before adding patterns here, please read the gitignore
> # documentation at https://git-scm.com/docs/gitignore
> ```

Done. I've left in the compile_commands.json line though, as that is a pretty standard generated file used by a variety of tools.

anjohnson:
I must be a programming dinosaur then, I've personally never seen a `compile_commands.json` file created inside any EPICS Base or support module repo's at APS. Do you know what tools generate it?

Anyone else want to weigh in here on what ignore patterns should or shouldn't be included?

exzombie:
The `compile_commands.json` file is used by language servers that I briefly mentioned yesterday. It's what tells them where all the include files are and such so that they can analyze the code properly and provide navigation and static analysis. It's editor-independent.

JJL772:
compile_commands.json is consumed by clang-based tooling such as clangd, clang-tidy, clang-doc and ccls. Editors like CLion and VSCode (w/Microsoft's Cpptools plugin) also consume it.

mdavidsaver:
> ... programming dinosaur ...

No comment ;)

> ... I've personally never seen a `compile_commands.json` file ...

```
$ ls -lh .qtc_clangd/compile_commands.json 
-rw-r--r-- 1 dev dev 5.9M Jun 14 16:45 .qtc_clangd/compile_commands.json
```

Note the size.  Clearly a generated file.

anjohnson:
Core group: We'd like to see the new comments added, but remove all the IDE files (should be in the Git personal area) and other local files like `*.orig` and do a general clean-up of these files. The template `.gitignore` files should get the same clean-up.

JJL772:
Updated. Let me know if I've understood the assignment here :)

PR #480, opened 2024-04-29
epicsThreadMustJoin() undefined behavior on POSIX.

ericonr:
Per [1], "The behavior is undefined if the value specified by the thread argument to pthread_join() refers to the calling thread.". The code relied, instead, on pthread_join() returning EDEADLK on such a situation.

This was caught when running with musl libc.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_join.html

---

I came across this when testing https://github.com/epics-modules/asyn/pull/171

I would have caught this earlier if I had run the test suite with musl. Can confirm it deadlocked in epicsThreadTest, and doesn't anymore after my fix.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -699,14 +699,22 @@ void epicsThreadMustJoin(epicsThreadId id)
         return;
     }
 
-    status = pthread_join(id->tid, &ret);
-    if(status == EDEADLK) {
-        /* Thread can't join itself (directly or indirectly)
-         * so we detach instead.
+    if(epicsThreadGetIdSelf()==id) {
+        /* Thread can't join itself, detach instead.
          */
         status = pthread_detach(id->tid);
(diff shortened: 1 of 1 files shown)


mdavidsaver:
> This was caught when running with musl libc.

How is this manifest?

Please quote the text of any warning/error message seen to help others find this ticket (and avoid duplicates).

ericonr:
@mdavidsaver there's no warning or error, pthread_join simply blocks forever due to the deadlock.

In asyn, this manifested as an event never being sent, but there wasn't a warning for that either.

Would it make sense to quote the failure in epicsThreadTest that this fixes? That's the only one I can think of.

mdavidsaver:
@ericonr  Could you test #496 ?

> ... no warning or error, pthread_join simply blocks forever ...

Noted.

> Would it make sense to quote the failure in epicsThreadTest that this fixes?

Not essential in this case, but maybe a good habit to develop.

anjohnson:
Core group: Closing in favor of #496, thanks!

Issue #485, opened 2024-05-17
IOC Segmentation fault when SIZV is larger than 32767


Insomnia1437:
Found on R7.0.8, and can repeat on latest commit d8b5616772df78c50e94fb642c4903d757232e38
Test db:
```
record(lsi, lsi){
  field(SIZV, "32767")
  field(INP,  {const: aaa})
  field(PINI, "YES")
}
record(lsi, lsi-bad){
  field(SIZV, "32768")
  field(INP,  {const: aaa})
  field(PINI, "YES")
}
```
When using  `caget lsi-bad`, ioc crashs.
```shell
# iocsh
$ softIocPVA -d bug.db
Starting iocInit
############################################################################
## EPICS R7.0.8
## Rev. 2024-03-27T10:23+0100
## Rev. Date build date/time:
############################################################################
iocRun: All initialization complete
epics> [1]    1485 segmentation fault  softIocPVA -d bug.db

# another terminal
$ caget lsi
lsi                            aaa
$ caget lsi-bad
CA.Client.Exception...............................................
    Warning: "Virtual circuit disconnect"
    Context: "172.29.165.14:5064"
    Source File: ../cac.cpp line 1238
    Current Time: Fri May 17 2024 22:15:48.322332782
..................................................................
CA.Client.Exception...............................................
    Warning: "Virtual circuit disconnect"
    Context: "op=0, channel=lsi-bad, type=DBR_TIME_STRING, count=1, ctx="172.29.165.14:5064""
    Source File: ../getCopy.cpp line 91
    Current Time: Fri May 17 2024 22:15:48.322409017
..................................................................
Read operation timed out: some PV data was not read.
lsi-bad
```
stack trace:
```
Thread 35 "CAS-client" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffaf6126c0 (LWP 1803)]
__strncpy_avx2 () at ../sysdeps/x86_64/multiarch/strcpy-avx2.S:845
845     ../sysdeps/x86_64/multiarch/strcpy-avx2.S: No such file or directory.
(gdb) bt
#0  __strncpy_avx2 () at ../sysdeps/x86_64/multiarch/strcpy-avx2.S:845
#1  0x00007ffff7e7beb8 in getStringString (paddr=0x7fff94005370, pto=<optimized out>, nRequest=<optimized out>, no_elements=1, offset=0)
    at ../db/dbConvert.c:146
[...]

PR #486, opened 2024-05-19
fix sizv for printf & fix doc 

Insomnia1437:
ses  #485


mdavidsaver:
Thank you for following up on this.  I had forgotten about printfRecord.

PR #490, opened 2024-06-09
Add makeBaseApp templateApp for instantiating site templates

anjohnson:
This adds a new makeBaseApp template which when instantiated gives you an application to which you can add your own (or your site's) EPICS application templates. It installs a Perl script `makeSiteApp` into its own `bin/$EPICS_HOST_ARCH` directory which calls the `makeBaseApp.pl` script, but offering its own templates for instantiation instead of those from Base itself. Thus after instatiating this template you can add your own `*App` and `*Boot` template directories under the `src/template/base/top` directory, and edit the `src/template/base/top/templateApp/Makefile` and list the files to be copied.

Don't try to use this particular application to try and understand how to create templates, it's likely to be confusing since it's fairly self-referential; look at the other templates which come with Base and can be found in the `$EPICS_BASE/templates/makeBaseApp/top` directory.

### DIFF
--- a/src/template/base/Makefile
+++ b/src/template/base/Makefile
@@ -15,6 +15,22 @@ TEMPLATES += top/configure/RULES.ioc
 TEMPLATES += top/configure/RULES_DIRS
 TEMPLATES += top/configure/RULES_TOP
 
+# This is a template for a module that provides local templates.
+# It needs to install the TEMPLATES files named above into the
+# templateApp/top/... tree.
+SUBTEMPL := $(TEMPLATES)
+T_APP = top/templateApp
+SUBFILES = $(addprefix $(T_APP)/,$(SUBTEMPL))
+
+TEMPLATES += $(T_APP)/Makefile
+TEMPLATES += $(T_APP)/makeSiteApp@
+TEMPLATES += $(SUBFILES)
+
 SCRIPTS_HOST += makeBaseApp.pl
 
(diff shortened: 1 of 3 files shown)


mdavidsaver:
Is another entry point script (`makeSiteApp`) necessary in Base?  Could `makeBaseApp` be given some kind of search path?

anjohnson:
There is no `makeSiteApp` program installed into Base by this, it's a template for creating template installers, the `makeSiteApp` program which the template builds provides a search path to the `makeBaseApp.pl` which already supports that. The point is to make it slightly easier for sites to make and use their own collection of IOC templates.

To use it at APS, I would instantiate this template once, possibly renaming the `makeSiteApp` script, then add our local templates to the app. After building it my users can run `makeSiteApp -l` to see what templates we have without also seeing the ones from Base which I might not want them to use at all.

anjohnson:
This needs documenting (as does `makeBaseApp.pl` and the whole "how do I create a template" process).

ralphlange:
https://docs.epics-controls.org/en/latest/appdevguide/gettingStarted.html#makebaseapp-pl is already pretty complete, isn't it?

anjohnson:
@ralphlange so it is, thanks! It seems such a long time since I looked at that document. It will need updating if/when we add this site template app to Base, but that is where to put detailed instructions on how to use this new template to create an app for a site's IOC templates to live in.

PR #491, opened 2024-06-09
Proposed R7.0.8.1

anjohnson:
**DO NOT MERGE THIS BRANCH,** it contains an empty commit where the proposed git tag belongs.

@ralphlange reminded me that at the last Core Developers meeting we discussed creating a release before the Codeathon to publish the fixes applied since December, and to provide ESS with the new release they're looking for. This is my proposal for R7.0.8.1, which includes the external submodules at the same commits as they were in R7.0.8 (there have been no substantive changes actually merged into the submodules, although a couple of them do have PRs waiting).

Note that this PR contains one commit _beyond_ the tag, so the diffs below don't show the changes to the release tag as they also include that extra commit.


mdavidsaver:
> ... creating a release before the Codeathon ...

I like this idea.

The change log does not mention that the return type of `epicsSocketAccept()` changed, and the potential effects on WIN32.

anjohnson:
I've been asked to merge and include epics-base/pvDatabaseCPP#82 in the 7.0.8.1 release as it's also a bug fix, which I'm willing to add if there are no objections to that.

This "make releases as a pull-request" seems like an interesting approach, it would allow for errors to be noticed and fixed, and lets others participate. Any comments on it from reviewers or others?

Some questions I have for discussion are:
* Should the PR be marked as a draft, or would that hide it too much?
* Whether my empty commit is good for marking the tag point; it's a pity Git doesn't let you add a tag to a branch as a normal commit.
* Should the work be done on a branch in an individual's fork, or a pre-release branch in the main repo?

ralphlange:
> * Whether my empty commit is good for marking the tag point; it's a pity Git doesn't let you add a tag to a branch as a normal commit.

What's your intent there?
Annotated tags are full git objects that are checksummed and can be GPG-signed. What are you missing?

anjohnson:
> > * Whether my empty commit is good for marking the tag point; it's a pity Git doesn't let you add a tag to a branch as a normal commit.
> 
> What's your intent there? Annotated tags are full git objects that are checksummed and can be GPG-signed. What are you missing?

I normally use annotated tags for releases anyway, but they don't appear in the branch history or the `git log` output as commits, they're stored under refs/tags. Both `git merge` and `git rebase` would do funny things if tags could be an inline commit though, so on reflection that idea doesn't work.

BTW @mdavidsaver I had typo'd the PR# when I submitted my earlier comment and then edited to correct it, I was actually asked to merge epics-base/pvDatabaseCPP#82 instead of 81 (but thanks for the comments there too, I will apply them to my similar clang-15 branches/PRs on the other modules).

ralphlange:
> I normally use annotated tags for releases anyway, but they don't appear in the branch history or the `git log` output as commits, they're stored under refs/tags. Both `git merge` and `git rebase` would do funny things if tags could be an inline commit though, so on reflection that idea doesn't work.

Try darcs. ;-)
"A tag is basically just an empty patch which depends on some other patches."
https://darcs.net/Using/Model

anjohnson:
@mdavidsaver commented
> > ... creating a release before the Codeathon ...
> 
> I think this idea.

I hope you meant "like" there instead of "think"…

Are the commits you added today intended to be included in R7.0.8.1, or should they remain unreleased to mature until the next release?

I will add the `epicsSocketAccept()` change to the Release Notes commit and force-push.

ralphlange:
Went through the changes and didn't see any unexpected things.

I was wondering: Should we include the fix or the workaround for the `-D_FORTIFY_SOURCE=3` issue (#517)?
Otherwise, given the popularity of Ubuntu, we would at least need a "Known Issues" entry for this.

ralphlange:
About the empty commit: You could still squash that and do a final force-push at the time you apply the tag.

Generally: Using a release branch is something very common (that we also did for EPICS Base in the past, IIRC). Takes the pressure out of the release process and doesn't interfere with development going on in parallel.

anjohnson:
I added the `_FORTIFY_SOURCE` limit as [08a242a](https://github.com/epics-base/epics-base/pull/491/commits/08a242ae823cf4c8acb582bc0e658b9aa224a6c0), do we have any GHA builds which will actually test this? I assume not since we haven't been getting failures in our CI builds (until @mdavidsaver's #517 which this build pre-dates. I don't have access to a recent Ubuntu system to do it manually.

Any ideas?

mdavidsaver:
> I added the `_FORTIFY_SOURCE` limit ...

```c
#if defined(_FORTIFY_SOURCE) && _FORTIFY_SOURCE>2
OP_SYS_CPPFLAGS += -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
#endif
```

This is what I had in mind.

ralphlange:
> I added the `_FORTIFY_SOURCE` limit as [08a242a](https://github.com/epics-base/epics-base/pull/491/commits/08a242ae823cf4c8acb582bc0e658b9aa224a6c0), do we have any GHA builds which will actually test this? I assume not since we haven't been getting failures in our CI builds (until @mdavidsaver's #517 which this build pre-dates. I don't have access to a recent Ubuntu system to do it manually.
> 
> Any ideas?

Can't you just pick the relevant change from #517 onto this branch for a test build? If it's after the "tag", it won't be part of the release.

mdavidsaver:
> ... I don't have access to a recent Ubuntu system to do it manually.

hmm...  This change does not have the desired effect.

```sh
podman pull ubuntu:latest
podman run --rm -ti ubuntu:latest
apt install --no-install-recommends build-essential git ca-certificates
git clone --depth 1 --branch Proposed-R7.0.8.1 https://github.com/anjohnson/epics-base
cd epics-base
make -j12
cat <<EOF > test.db
record(calc, "foo") {
    field(CALC, "VAL")
}
EOF
./bin/linux-x86_64/softIoc -d test.db
```

anjohnson:
I have a release note entry for this too now, but I won't push that rebase yet in the hope that @mdavidsaver might see what's wrong with the change. On my darwin-aarch64 laptop I forced the toolchain.c build to output those build flags, and now when I compile the libCom test programs I see the flags in compiler commands for both C and C++ sources.

mdavidsaver:
Right, the default for `_FORTIFY_SOURCE` is somehow set during `#include <features.h>`.

```
$ echo '#include <features.h>' | cpp -dM -x c|grep FORTI
#define __USE_FORTIFY_LEVEL 0
#define __bos(ptr) __builtin_object_size (ptr, __USE_FORTIFY_LEVEL > 1)
```

With Debian 12, adding `-O1` makes no difference.  With the latest Ubuntu, it does:

```
# echo '#include <features.h>' | cpp -dM -x c|grep FORTI
#define __REDIRECT_FORTIFY __REDIRECT
#define __USE_FORTIFY_LEVEL 0
#define __REDIRECT_FORTIFY_NTH __REDIRECT_NTH
#define __bos(ptr) __builtin_object_size (ptr, __USE_FORTIFY_LEVEL > 1)
```

becomes

```
# echo '#include <features.h>' | cpp -O1 -dM -x c|grep FORTI
#define __REDIRECT_FORTIFY __REDIRECT
#define _FORTIFY_SOURCE 3
#define __USE_FORTIFY_LEVEL 3
#define __REDIRECT_FORTIFY_NTH __REDIRECT_NTH
#define __bos(ptr) __builtin_object_size (ptr, __USE_FORTIFY_LEVEL > 1)
```

But where is this happening??

mdavidsaver:
> Right, the default for `_FORTIFY_SOURCE` is somehow set during `#include <features.h>`.

I am wrong.  It is a being injected as a pre-defined macro.  But only when `-O1` or greater is passed...  Currently `configure/toolchain.c` is processed with `$(CPPFLAGS)`, but not `$(CFLAGS)`.  So it doesn't get any `-O*` flag.

So I guess we would need to figure out a way to pass some/all `$(CFLAGS)` with `$(PREPROCESS.cpp)`?

This seems like it would be a good idea anyway to pick up `-f*` modifiers.  However, I am wondering what side-effects this would have...

```
$ cpp -dM -x c /dev/null|grep FORTI 
$ cpp -dM -O1 -x c /dev/null|grep FORTI
#define _FORTIFY_SOURCE 3
```

mdavidsaver:
With this additional change, the `_FORTIFY_SOURCE=3` pre-define is detected when processing `toolchain.c`.  I added the test of `$(GNU)`, and limited this all to `toolchain.c`, to avoid an 11th hour change effecting dependent modules.

```diff
diff --git a/configure/Makefile b/configure/Makefile
index ba77d50..70acd21 100644
--- a/configure/Makefile
+++ b/configure/Makefile
@@ -34,5 +34,8 @@ CFG += TOOLCHAIN.$(EPICS_HOST_ARCH).$(T_A)
 
 include $(TOP)/configure/RULES
 
+ifeq (YES,$(GNU))
+TOOLCHAIN.$(EPICS_HOST_ARCH).$(T_A): CPPFLAGS += $(CFLAGS)
+endif
 TOOLCHAIN.$(EPICS_HOST_ARCH).$(T_A): toolchain.c
        $(PREPROCESS.cpp)

```

```
$ tail cfg/TOOLCHAIN.linux-x86_64.linux-x86_64
...
OP_SYS_CPPFLAGS += -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
```

anjohnson:
Thanks, I'll push that, which will hopefully be the last change to this as a PR.

anjohnson:
Merged at 172bfce1f

PR #492, opened 2024-06-09
Improve error message for mingw64 architecture

FreddieAkeroyd:
As described in https://bugs.launchpad.net/epics-base/+bug/1873406 guessing the correct EPICS architecture for a MinGW system is not straightforward. Some suggestions are made there, it may also be possible to look at e.g. `gcc -v` to guess this based on its default target as to the best EPICS target. However in the meantime we will just print a better hint to the user as to how to resolve the situation.

### DIFF
--- a/src/tools/EpicsHostArch.pl
+++ b/src/tools/EpicsHostArch.pl
@@ -53,11 +53,11 @@ sub HostArch {
             die "$0: macOS CPU type '$cpu' not recognized\n";
         }
         # mingw64 has 32bit and 64bit build shells which give same arch result
-        if (m/^x86_64-msys/ or m/^i686-msys/) {
-            die "$0: Architecture '$arch' ambiguous.\n".
-                "Rerun after typing one of the following:\n".
-                "    export EPICS_HOST_ARCH=windows-x64-mingw   # for 64bit build\n".
-                "    export EPICS_HOST_ARCH=win32-x86-mingw     # for 32bit build\n";
+        if (m/^(x86_64|i686)-msys/) {
+            die "$0: Architecture '$arch' is unclear,\n".
+                "try again after explicitly setting the EPICS_HOST_ARCH environment variable.\n".
+                "Use  EPICS_HOST_ARCH=windows-x64-mingw  for a 64bit MinGW build, or\n".
+                "EPICS_HOST_ARCH=win32-x86-mingw  for a 32bit MinGW build.\n"
         }
         die "$0: Architecture '$arch' not recognized\n";
     }
--- a/src/tools/EpicsHostArch.pl
+++ b/src/tools/EpicsHostArch.pl
@@ -52,7 +52,13 @@ sub HostArch {
             }
             die "$0: macOS CPU type '$cpu' not recognized\n";
         }
-
+        # mingw64 has 32bit and 64bit build shells which give same arch result
+        if (m/^x86_64-msys/ or m/^i686-msys/) {
+            die "$0: Architecture '$arch' ambiguous.\n".
+                "Rerun after typing one of the following:\n".
+                "    export EPICS_HOST_ARCH=windows-x64-mingw   # for 64bit build\n".
+                "    export EPICS_HOST_ARCH=win32-x86-mingw     # for 32bit build\n";
+        }
         die "$0: Architecture '$arch' not recognized\n";
     }
 }


anjohnson:
Thanks @FreddieAkeroyd. What is the output from `make` in Base if you run this on an ambiguous system? How soon does the build die?

FreddieAkeroyd:
Hi Andrew,

It dies quite quickly
```
$ make
./src/tools/EpicsHostArch.pl: Architecture 'x86_64-msys-thread-multi' ambiguous.
Rerun after typing one of the following:
    export EPICS_HOST_ARCH=windows-x64-mingw   # for 64bit build
    export EPICS_HOST_ARCH=win32-x86-mingw     # for 32bit build
configure/CONFIG:67: configure/os/CONFIG..Common: No such file or directory
make: *** No rule to make target 'configure/os/CONFIG..Common'.  Stop.
```

anjohnson:
Thanks; sorry I didn't look at the message closely before — a minor nit, but that's why we do code reviews. There are shells that don't use the `export` syntax, which is a sh/bash/ksh/zsh-ism not supported by csh, tcsh or Command.exe and I don't know what syntax PowerShell uses at all. It's probably safer to say something like this in the case of ambiguity:
```
$0: This host architecture 'x86_64-msys-thread-multi' is ambiguous,
    try again after explicitly setting the EPICS_HOST_ARCH environment variable.
    Use  EPICS_HOST_ARCH=windows-x64-mingw  for a 64bit MinGW build, or
    EPICS_HOST_ARCH=win32-x86-mingw  for a 32bit MinGW build.
```
I note that in the LP bug comments back on 2020-04-23 @ralphlange wrote
> The environment settings in a "MinGW32 shell" or "MinGW64 shell" clearly indicate which toolchain is to be used. IMHO, the EpicsHostArch script should output the setting for the intended toolchain.

If those environment settings could be used to resolve the ambiguity we should make use of them, but I don't know if that's still the case. If not though we should defer like that.

Maybe:
```patch
- if (m/^x86_64-msys/ or m/^i686-msys/) {
+ if (m/^(x86_64|i686)-msys) {
```

FreddieAkeroyd:
Thank you @anjohnson - yes better to not make any assumptions about the environment/shell in the message

I think @ralphlange has looked at the toolchain issue in more details, but the issue seems to be that when you install MinGW64 it installs a 64bit common area, and then  /mingw64 and /mingw32 directories that contain compiler binaries that target that architecture plus some other binaries too. You then have windows shortcuts that start a "mingw 32bit" or "mingw 64bit" bash shell, these just differ by altering your path to search either /mingw64 or /mingw32 before the common area and setting an environment variable to indicate this initial choice. It seems a 32bit python is placed in the 32bit area, but not perl - so perl is always 64bit, hence the epics host arch test always thinks you are 64bit environment whatever mingw windows shortcut you start (you may be able to install a 32bit perl on mingw, but if it is not there in a default install it is probably not safe to rely on that). Also just because somebody started a mingw 64bit shell, they could still likely create a 32bit image by using something like `-m32` with gcc. So it is like an  x64 linux system that can have both 32bit and 64bit binaries and libraries present, and compile/use both architectures.     

So i think it is a question of what assumptions we want to make. Using the mingw environment variables would assume what we wanted to build was based entirely on the shortcut used to initially create the shell, however a user could of course later swap their PATH in a 64bit shell to point at the 32bit compiler and then using the envirionment variable would be incorrect. Another approach might be to query the gcc in the current PATH and see what its default target was at that time, and then guess the epics host target from this.  Changing the error message was just a just a way to help a new user of epics (as opposed to a developer) get things building quickly without making any assumptions as to what might be intended.  

I feel checking the environment variables could be a reasonable compromise, if not foolproof. It covers the most natural use case (start a 32bit shell to build 32bit, a 64bit shell to build 64bit). It does not cover the case of changing the compiler path etc in an existing shell, this could be judged as something an advanced user would be doing and leave it to them to sort out EPICS_HOST_ARCH. I am happy to look at checking the gcc path/target instead for the choice, or doing something like "check env variables, check current target of gcc in PATH, use if they agree or else error as ambiguous" if these seem more complete/useful solutions.

anjohnson:
Thanks Freddie, I'm fine with 'simpler is better'. I'm going to give this an "approved" label, assuming the Appveyor builds eventually pass (after retries of any unreliable tests as needed) that should get it merged. Our PR job queue there is very backed up after last week though, so it might take a while.

PR #493, opened 2024-06-10
centralize -g flag for gnu targets

dirk-zimoch:
Instead of adding `-g` to `OPT_CFLAGS_YES` and `OPT_CXXFLAGS_YES` for each architecture separately, it can be done once for all in CONFIG.gnuCommon.


mdavidsaver (approved):
Seems to me a straightforward and reasonably simplification.

anjohnson:
Rerunning the "Base / Ub-20 gcc-9 C++11, static" job, I don't think that would have failed due to these changes.

Looks fine to me if that passes.

PR #496, opened 2024-06-11
Avoid UB with self `pthread_join()`

mdavidsaver:
Attempt to address #480 by handling self-join in `epicsThreadMustJoin()` rather than depending on `pthread_join()`.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -687,34 +687,54 @@ static epicsThreadOSD *createImplicit(void)
 
 void epicsThreadMustJoin(epicsThreadId id)
 {
-    void *ret = NULL;
     int status;
+    int prev;
+    epicsThreadId self;
 
-    if(!id) {
+    if(!id)
         return;
-    } else if(epicsAtomicCmpAndSwapIntT(&id->joinable, 1, 0)!=1) {
-        if(epicsThreadGetIdSelf()==id) {
-            errlogPrintf("Warning: %s thread self-join of unjoinable\n", id->name);
+
+    prev = epicsAtomicCmpAndSwapIntT(&id->joinable, 1, 0);
+    self = epicsThreadGetIdSelf();
+
+    if(prev==0) {
+        /* join of unjoinable */
+        if(self==id) {
+            errlogPrintf(ERL_WARNING ": %s thread self-join of unjoinable\n", id->name);
+            return;
 
         } else {
             /* try to error nicely, however in all likelihood de-ref of
              * 'id' has already caused SIGSEGV as we are racing thread exit,
              * which free's 'id'.
              */
-            cantProceed("Error: %s thread not joinable.\n", id->name);
+            cantProceed(ERL_ERROR ": %s can't join unjoinable %s\n", self->name, id->name);
         }
-        return;
+
+    } else if(prev!=1) { /* also not 0, the only other allowed value. */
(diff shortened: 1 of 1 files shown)


modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -682,32 +682,45 @@ void epicsThreadMustJoin(epicsThreadId id)
...
-            errlogPrintf("Warning: %s thread self-join of unjoinable\n", id->name);
+
+    prev = epicsAtomicCmpAndSwapIntT(&id->joinable, 1, 0);
+    self = epicsThreadGetIdSelf();
+
+    if(prev==0) {
+        /* join of unjoinable */
+        if(self==id) {
+            errlogPrintf(ERL_WARNING ": %s thread self-join of unjoinable\n", id->name);
+            return;
```
exzombie, line 704: I think you are missing a call to `free_threadInfo()`

exzombie: Or is this meant to be a complete no-op? The comment at the end of the function says "always decrement refcnt"; the "always" is confusing to me, and there are no docs.

mdavidsaver: `prev==0` implies that `refcnt==1`.  For a self-join, this last reference is for the thread itself, and will be decremented when it actual exits.  This happens automatically through the use of a destructor function passed to `pthread_key_create()` when creating the thread local behind `epicsThreadGetIdSelf()`.

exzombie:
I was planning to open an issue against asyn because `asynPortDriver` causes a crash when EPICS is built with ThreadSanitizer ([this is the cause](https://github.com/epics-modules/asyn/blob/master/asyn/asynPortDriver/asynPortDriver.cpp#L934)). With this PR, the issue went away. It fixes more than just musl.

modules/libcom/src/osi/os/posix/osdThread.c
mdavidsaver: I changed the wording of this comment to "release joinable reference".

mdavidsaver:
I have added a fix for a memory leak I just noticed from `epicsThreadCreateOpt()` when `joinable` in the (I think) unlikely situation where initial SCHED_FIFO probing succeeded, but a later `pthread_create()` fails with `EPERM`.

modules/libcom/src/osi/os/posix/osdThread.c
```diff
@@ -689,32 +689,45 @@ void epicsThreadMustJoin(epicsThreadId id)
...
 
-    status = pthread_join(id->tid, &ret);
-    if(status == EDEADLK) {
+    } else if(prev!=1) { /* also not 0, the only other allowed value. */
+        cantProceed(ERL_ERROR ": %s joins corrupt thread handle\n", self->name);
+
+    } else if(self==id) {
         /* Thread can't join itself (directly or indirectly)
          * so we detach instead.
          */
```
ericonr: This comment needs to be updated, this block is for the thread joining itself directly.

ericonr (commented):
Doesn't this version lack the `pthread_detach` fallback when `pthread_join` fails with `EDEADLK` for valid calls?

mdavidsaver:
> Doesn't this version lack the `pthread_detach` fallback when `pthread_join` fails with `EDEADLK` for valid calls?

That it did.  Thank you for pointing this out.  Should now be correct.

anjohnson:
AI on @anjohnson : Examine closely, review.

ericonr (approved):
I can confirm that the test case now passed for me, on musl.

PR #497, opened 2024-06-11
Add `pushd`, `popd` and `dirs` commands for iocsh

### LINKED ISSUE #494: `pushd`, `popd` and `dirs` commands for iocsh
These commands tend to be quite useful in both interactive shell sessions and in scripts. Providing these for iocsh might be helpful.

This page describes their behavior in bash: https://www.gnu.org/software/bash/manual/html_node/Directory-Stack-Builtins.html

Of course, the iocsh equivalents of these don't need to support all of those features.

This might be a good quick codeathon task.

JJL772:
These are simplified versions of the bash equivalents detailed here: https://www.gnu.org/software/bash/manual/html_node/Directory-Stack-Builtins.html

`pushd` is used like `pushd path/to/my/directory`, and takes no other arguments or flags.

`popd` takes no arguments and removes the top item from the directory stack.

`dirs` shows the entire directory stack with the current working directory on top. Even if the stack is empty, `dirs` will still display the cwd.

I also used struct hack in the `dirStackEntry` structure. This is standard in C99, but was provided as a compiler extension prior to then. GCC as far back as 3.4.6 (at least) seems to [support this without issue](https://godbolt.org/z/94jEMTsd6). That being said, I don't know what support for these is like in Visual Studio. Everything that I can find seems to indicate that it's been supported since at least 2012. We'll see what Appveyor says :) 
According to the Microsoft docs, flexible arrays are available as long as Microsoft extensions are enabled (i.e. `/permissive-` and `/Za` are *not* set).

Closes #494

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -17,6 +17,8 @@
 #include <map>
 #include <string>
 #include <sstream>
+#include <list>
+#include <string>
 
 #include <stddef.h>
 #include <string.h>
@@ -38,6 +40,8 @@
 #include "registry.h"
 #include "epicsReadline.h"
 #include "cantProceed.h"
+#include "osiFileName.h"
+#include "osiUnistd.h"
 #include "iocsh.h"
 
 #include "epicsReadlinePvt.h"
@@ -988,9 +992,24 @@ struct iocshScope {
 
 // per thread executing iocshBody()
 struct iocshContext {
+    iocshContext() :
+        handle(NULL),
+        scope(NULL)
+    {
+    }
+
     MAC_HANDLE *handle;
     iocshScope *scope;
+    std::list<std::string> dirStack;
 };
+
+static iocshContext* iocshContextGet()
+{
+    iocshContext* ctxt = NULL;
+    if (iocshContextId)
+        ctxt = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
+    return ctxt;
+}
 
 int iocshSetError(int err)
 {
@@ -1084,10 +1103,10 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
     context = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
 
     if (!context) {
-        context = (iocshContext*)calloc(1, sizeof(*context));
+        context = new (std::nothrow) iocshContext;
         if (!context || macCreateHandle(&context->handle, pairs)) {
             errlogMessage("iocsh: macCreateHandle failed.");
-            free(context);
+            delete context;
             return -1;
         }
 
@@ -1294,7 +1313,7 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
 
     if (!scope.outer) {
         macDeleteHandle(handle);
-        free(context);
+        delete context;
         epicsThreadPrivateSet(iocshContextId, NULL);
     } else {
         context->scope = scope.outer;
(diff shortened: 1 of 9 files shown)


modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -28,6 +29,13 @@
 #include "freeList.h"
 #include "libComRegister.h"
 
+struct dirStackEntry {
+    ELLNODE node;
+	char path[];
+};
+
+static ELLLIST dirStack = ELLLIST_INIT;
```
mdavidsaver: IOC shell functions may run concurrently.  For basic user sanity, I don't think it is reasonably to have all threads share a common stack.

I think it would be better to add this stack to the per-thread `struct iocshContext`.

mdavidsaver: Also, since `iocsh.cpp` is C++, maybe `std::list<std::string>` would avoid some effort, and avoid memory leaks?

anjohnson: Do individual Posix, Windows and RTEMS-score threads each have their own record of the CWD, or do they share one CWD for the whole process? I'm pretty sure VxWorks only has one CWD, although I'm be okay ignoring that target if it's unique. If it isn't and concurrent iocsh instances are calling `cd`/`pushd`/`popd` there are going to be some major problems with relative paths. We tell people never to use relative paths for the access security configuration file because someone running a `cd` command would cause a DoS if/when that file gets reloaded (iocStats loads a PV for doing that).

mdavidsaver: @anjohnson I am not referring to `getcwd()`, but only to this iocsh directory stack.  I acknowledge that it is not possible to avoid races with concurrent `chdir()`.  I understand though that one of the goals of this change is to bring a measure of sanity.  I think that a per-thread stack will help with this.

modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -470,6 +478,65 @@ static void installLastResortEventProviderCallFunc(const iocshArgBuf *args)
...
+/* dirs */
+static const iocshFuncDef dirsFuncDef = {"dirs", 0, NULL,
+										 "Displays the contents of the directory stack, including current directory on top"};
+
+static void dirsFunc(const iocshArgBuf* args)
+{
+	struct dirStackEntry* e = NULL;
+
+	char cwd[PATH_MAX];
+	getcwd(cwd, sizeof(cwd));
```
mdavidsaver: You need to handle the possibility of failure here.  eg. if the current directory name is very long.  Also, `PATH_MAX` may not be available on all targets.

Actually, it looks like windows supports the posix-2001 behavior for `getcwd(NULL,0)` to return a `malloc()` buffer.  So maybe something like this untested code:

```c++
std::string cwd;
#ifdef PATH_MAX
char scratch[PATH_MAX];
if(getcwd(scratch, sizeof(scratch))) {
    cwd = scratch;
} else
#endif
if(char *alloc = getcwd(NULL,0)) {
    try {
        cwd = alloc;
        free(alloc);
    } catch(std::exception&) { // only bad_alloc possible
        free(alloc);
    }
}
if(!cwd.empty())
    context->dirstack.push(cwd);
else
    iocshSetErr(1);
```

JJL772: It might be worth investigating other uses of `getcwd` as well. It seems like `PATH_MAX` is defaulted to 100 in a couple of source files, even though Windows provides `MAX_PATH` instead of `PATH_MAX`. The `cd` command uses a seemingly arbitrary buffer size of 1024 with `getcwd`.

`PATH_MAX` is defined in `limits.h` (part of the C library), and it exists in both newlib and glibc. Windows has its own `MAX_PATH` and doesn't define `PATH_MAX`. Perhaps EPICS should provide `EPICS_PATH_MAX` to consolidate these definitions and to provide a default if it's missing?

modules/libcom/src/iocsh/libComRegister.c
JJL772: I've added this to `iocshContext` and switched to a `std::list` of `std::string`s

modules/libcom/src/iocsh/libComRegister.c
JJL772: I've added error handling to this particular call of `getcwd`, but haven't used the POSIX 2001 behavior for it. I'd like to get some feedback on the above comment before I go one way or the other.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1542,6 +1560,85 @@ static void onCallFunc(const iocshArgBuf *args)
...
+    }
+
+    char cwd[PATH_MAX+1];
+    if (!getcwd(cwd, sizeof(cwd))) {
+        iocshSetError(errno);
+        perror("getcwd failed");
+        return;
+    }
+
+    if (iocshSetError(chdir(args[0].sval)) != 0) {
```
mdavidsaver, line 1594: fyi. all iocsh arguments are optional.  So `args[0].sval` can be `NULL`.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1542,6 +1560,85 @@ static void onCallFunc(const iocshArgBuf *args)
...
+        return;
+    }
+
+    if (ctxt->dirStack.size() <= 0) {
+        fprintf(stderr, "Directory stack is empty.\n");
+        return;
+    }
+
+    if (iocshSetError(chdir(ctxt->dirStack.front().c_str())) != 0) {
+        fprintf(stderr, "Invalid directory path, ignored\n");
```
mdavidsaver: This error isn't be ignored.  (nor should it be)  Please correct the message.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1542,6 +1560,85 @@ static void onCallFunc(const iocshArgBuf *args)
...
+    if (!ctxt) {
+        iocshSetError(-EFAULT);
+        return;
+    }
+
+    char cwd[PATH_MAX+1];
+    if (!getcwd(cwd, sizeof(cwd))) {
+        iocshSetError(errno);
+        perror("getcwd failed");
+        return;
```
mdavidsaver, line 1645: Minor, but the dirStack could still be printed in this situation.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1084,10 +1102,10 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
     context = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
 
     if (!context) {
-        context = (iocshContext*)calloc(1, sizeof(*context));
+        context = new iocshContext;
```
mdavidsaver: Use [`std::nothrow`](https://en.cppreference.com/w/cpp/memory/new/nothrow) here.

```c++
context = new(std::nothrow) iocshContext;
```

mdavidsaver (changes requested):
Looking better.

modules/libcom/src/iocsh/libComRegister.c
mdavidsaver: > It might be worth investigating other uses of `getcwd` ...

I agree, though imo. this should be a separate project.

Since these commands are a new feature, I think I am ok with not handling the long CWD case right away.

JJL772:
I'd like #510 to be merged first so I can use `PATH_MAX`

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1545,6 +1563,94 @@ static void onCallFunc(const iocshArgBuf *args)
...
+static const iocshArg pushdArg0 = {"dir", iocshArgString};
+static const iocshArg* const pushdArgs[1] = {&pushdArg0};
+static const iocshFuncDef pushdFuncDef = {"pushd", 1, pushdArgs,
+                                         "Pushes the current directory to the directory stack and changes to the specified directory"};
+
+static void pushdCallFunc(const iocshArgBuf* args)
+{
+    if (!args[0].sval) {
+        iocshSetError(-1);
+        fprintf(stderr, "Invalid directory path\n");
```
mdavidsaver: Maybe a message would be: `"Missing required argument: dir"`?

mdavidsaver:
> I'd like #510 to be merged first so I can use `PATH_MAX`

@JJL772 I'm switching this PR to Draft pending development of an alternative to #510.

PR #500, opened 2024-06-11
Move asTrapWrite to dbCore

mdavidsaver:
As discussed here https://github.com/epics-base/epics-base/issues/474#issuecomment-2064728736 and originating with https://github.com/epics-modules/caPutLog/pull/23.  Move asTrapWrite code to dbCore to allow manipulation of `dbChannel`.  Type `asTrapWriteMessage::serverSpecific` to avoid pitfalls of `void*`.

I also had to move `asTrapWriteWithData()`, `asTrapWriteAfter()`, and their various macros from asLib.h.  Some macro trickery avoids breakages from code which included asLib.h neither asTrapWrite.h nor asDbLib.h.

Issue #474, opened 2024-04-15
Issues with asTrapWrite


mdavidsaver:
As mentioned in https://github.com/epics-modules/caPutLog/pull/23#issuecomment-2032499953 I think that the `asTrapWrite*()` API has a state corruption issue introduced when `dbAddr::pfield` became mutable.

Beyond this bug, there a couple of API design decisions which with hindsight would be done differently.

1. `asTrapWriteRegisterListener()` accepts a function pointer, but no context `void*` pointer.  This limits flexibility.

https://github.com/epics-base/epics-base/blob/cb1571783b4793cb9c545e5736659ee5ed7d03bb/modules/libcom/src/as/asTrapWrite.h#L82-L83

2.`asTrapWriteListener` function pointer accepts non-const `asTrapWriteMessage`, potentially allowing mutation.

https://github.com/epics-base/epics-base/blob/cb1571783b4793cb9c545e5736659ee5ed7d03bb/modules/libcom/src/as/asTrapWrite.h#L75

3. `asTrapWriteMessage::serverSpecific` is untyped.  This `void*` was originally a `dbAddr*`, then in >= 3.15 changed to `dbChannel*`.

https://github.com/epics-base/epics-base/blob/cb1571783b4793cb9c545e5736659ee5ed7d03bb/modules/libcom/src/as/asTrapWrite.h#L42

This was effectively an incompatible API change which slipped through.  The (Marty's?) original version of `dbChannel*` started with an embedded `dbAddr`, so casting as safe (68dfeb76c0c59b21e45f2f57819d412e8c6c9507).  However, this changed early (eee5a2898368005d9ce5c12a343cff5a0355c188).

So user code is forced to `#if`.

https://github.com/epics-modules/caPutLog/blob/3df5287558d4a7da4ac5ac4f333bf7861eecd39e/caPutLogApp/caPutLogAs.c#L115-L121

4. `asTrapWriteMessage::data` is non-const, potentially allowing mutation.

https://github.com/epics-base/epics-base/blob/cb1571783b4793cb9c545e5736659ee5ed7d03bb/modules/libcom/src/as/asTrapWrite.h#L53

(A note for anyone doing `git blame` on the early revision adding dbChannel.  27 April 2012 was the last day of a collaboration meeting.  I was furiously working to rebase the enormous `server-side-plugins` branch, and messed up the authorship of some of the revisions.  I believe that work originated with Marty at an earlier codeathon)

mdavidsaver:
[...]

PR #501, opened 2024-06-12
asTrap serverSpecific is dbChannel

mdavidsaver:
Originating with https://github.com/epics-modules/caPutLog/pull/23, and contrary to https://github.com/epics-base/epics-base/issues/474#issuecomment-2064728736 and #500.

Moving asTrapWrite from Com to dbCore presented too many difficulties for me.  Instead, keep asTrapWrite.c in Com, with special access to include `dbChannel.h` early.  This allows `asTrapWriteBeforeWithData()` and `asTrapWriteAfterWrite()` to save/restore dbAddr::pfield around callbacks to avoid corruption if a CB forgets to restore.

In order to peak at dbChannel.h during libCom build, generate dbCoreAPI.h early and add extra `-I` to source location when compiling dbTrapWrite.c .


anjohnson:
Core Group: Approved pending final Appveyor tests.

PR #503, opened 2024-06-12
caget now can bring values from multiple PVs even when some are disconnected

### LINKED ISSUE #489: catools enhancements
The `camonitor` program accepts multiple PV names and can be started before some of those PVs are online. Those that are accessible are monitored and their values shown immediately; the others connect if/when their IOC is booted.

The `caget` and `cainfo` programs also accept multiple PV names, but if any of the named PVs are offline they never act on the the PVs that are accessible. It would be nice to be able to give a list of PVs and for them to do their thing on the PVs that do successfully connect before the timeout period expires. That would let us run say `caget ioc{a,b,c,d,e}:status` and at least see the status values from those IOCs which are booted.

marciodo:
resolves #489 

This draft pull request is for caget only. I'll start the work on cainfo once I confirm that the changes in caget used the expected path. If I'm too far from the target, it is easier to change only caget, instead of both caget and cainfo.

The request from #489 asks for caget and cainfo to output data from a list PVs even when some of them are still waiting to timeout due to disconnection. I used one epicsThread per PV to achieve this behavior.
Here's a summary of all the changes in caget:
1. The caget() function is now called once per PV and its context is not in the main thread anymore. This is opposed to caget() being called once and operating in an array of PVs.
2. The original caget() function accepted an array of pv structures and a nPvs (number of PVs) argument. Due to item 1, now caget() receives one single PV on the argument and nPvs was removed.
3. A mutex is used to block when the thread is printing to the console. This is to avoid that output messages would be commingled with characters from different threads at the same time.
4. The threads are joinable, so the main thread can join all of them and wait for them to complete their tasks before quitting.
5. The output of the the caget binary indicates how many PVs are disconnected. So, if zero is returned, this was a successful run. An odd thing is that caget also returns 1 if someone calls it with wrong arguments. Maybe we could change it to return negative numbers to indicate this.

As I removed 2 if conditionals from caget(), I fixed the indentation and now diff shows that almost 100% of the function is changed.     :-(   If you want I can restore the original indentation so you can check where the real changes were.

Here's the result when running the command with some arguments:
```
└─>  ./caget -w 10 -a -# 10 EM1K0:GMD:HPS:BeamRate EM1K0:GMD:HPS:milliJoulesPerPulseHSTTH  dadsadsa hehgfsdgf
EM1K0:GMD:HPS:BeamRate         2024-05-22 11:44:20.569302 1 200
EM1K0:GMD:HPS:milliJoulesPerPulseHSTTH <undefined> 10 0 0 0 0 0 0 0 0 0 0 UDF INVALID
Channel connect timed out: 'dadsadsa' not found.
Channel connect timed out: 'hehgfsdgf' not found.

└─>  echo $?
2
```

### DIFF
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -37,8 +37,6 @@
 #include <epicsGetopt.h>
 #include "epicsVersion.h"
 #include "epicsMutex.h"
-#include "epicsExit.h"
-#include "epicsSpin.h"
 
 #include "tool_lib.h"
 
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -368,8 +368,7 @@ static int caget (pv pv_data, RequestT request, OutputT format,
     default :
         break;
     }
-    free(pvs_data.value);
-    }
+    free(pv_data.value);
 
     epicsMutexUnlock(printMutex);
 
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -24,6 +24,9 @@
  *     Added field separators
  *  2009/04/01 Ralph Lange (HZB/BESSY)
  *     Added support for long strings (array of char) and quoting of nonprintable characters
+ *  2026/02/26 Marcio Donadio (SLAC)
+ *     caget now operates wiht more threads and retrieves values or "PV not found" messages
+ *     for all PVs in the argument list instead of failing after the first PV is not found.
  *
  */
 
@@ -53,7 +56,8 @@ typedef enum { plain, terse, all, specifiedDbr } OutputT;
 /* Different request types */
 typedef enum { get, callback } RequestT;
 
-// Struct to hold data used by caget threads
+// Struct to hold data used by caget threads.
+// This is sent via the EPICS message queue mechanism.
 typedef struct
 {
     pv pv_data;
@@ -63,15 +67,24 @@ typedef struct
     unsigned long reqElems;
     epicsMutexId printMutex;
     epicsMutexId resultMutex;
-    epicsMessageQueueId queueId;
     int totalNumberOfPvs;
 } t_thread_data;
+
+// Struct to hold data used only during thread initialization
+typedef struct
+{
+    epicsMessageQueueId queueId;
+    struct ca_client_context* ca_context;
+
+} t_init_thread;
 
 static int nConn = 0;           /* Number of connected PVs */
 static int nRead = 0;           /* Number of channels that were read */
 static int floatAsString = 0;   /* Flag: fetch floats as string */
 static int pvs_with_trouble = 0; /* Counter for PVs with result not zero */
-static int completedPVs = 0;
+static int completedPVs = 0; /* Number of PVs already processed */
+static epicsMessageQueueId queueId; /* Transfers PV data from one producer to several consumer threads */
+static struct ca_client_context* ca_context; /* Unique CA context for all threads */
 
 static void usage (void)
 {
@@ -401,16 +414,27 @@ void thread_name (int index, char* th_name) {
 }
 
 // Function that runs inside each thread, calling the caget function
-void caget_caller (void* queueId_void)
+void caget_caller (void* init_thread_void)
 {
     int result;                 /* CA result */
     t_thread_data thread_data;
+    t_init_thread* init_thread = (t_init_thread*) init_thread_void;
 
-    epicsMessageQueueId queueId = (epicsMessageQueueId) queueId_void;
+    // Attach to the CA context created by the main thread
+    //struct ca_client_context* ca_context = malloc(sizeof(struct ca_client_context*)); 
+    //memcpy (ca_context, init_thread->ca_context, sizeof(struct ca_client_context));
+    //ca_context_status(ca_context, 1);
+    result = ca_attach_context(ca_context);
+    if (result != ECA_NORMAL) {
+        fprintf(stderr, "CA error %s occurred while trying "
+                "to attach to channel access context.\n", ca_message(result));
+        return;
+    }
+
+    //epicsMessageQueueId queueId = (epicsMessageQueueId) queueId_void;
 
     while(true) {
         if (epicsMessageQueueReceiveWithTimeout(queueId, &thread_data, sizeof(t_thread_data), 0.1) > 0) {
-            //t_thread_data* thread_data = (t_thread_data*)thread_data_void;
             result = connect_pvs(&thread_data.pv_data, 1);
 
                                         /* Read and print data */
@@ -418,21 +442,21 @@ void caget_caller (void* queueId_void)
                 result = caget(thread_data.pv_data, thread_data.request, thread_data.format, thread_data.dbrType, thread_data.reqElems, thread_data.printMutex);
 
             epicsMutexLock(thread_data.resultMutex);
-            pvs_with_trouble += result;
+            pvs_with_trouble += result ? 1 : 0;
             ++completedPVs;
             epicsMutexUnlock(thread_data.resultMutex);
         }
 
+        // All PVs processed; there's nothing else expected to arrive in the queue
         if (completedPVs >= thread_data.totalNumberOfPvs) {
+            //ca_detach_context();
             break;
         }
     }
-
-    //free(thread_data);
 }
 
 // Creates the threads that will read from the message queue
-void thread_builder (epicsMessageQueueId queueId, int index)
+void thread_builder (t_init_thread init_thread, int index)
 {
     char th_name[9];
     thread_name(index, th_name);
@@ -440,7 +464,7 @@ void thread_builder (epicsMessageQueueId queueId, int index)
     // Must be joinable so the main thread can wait for all the threads to
     // complete their job.
     thread_opts.joinable = 1;
-    epicsThreadCreateOpt(th_name, &caget_caller, queueId, &thread_opts);
+    epicsThreadCreateOpt(th_name, &caget_caller, &init_thread, &thread_opts);
 }
 
 
@@ -617,14 +641,16 @@ int main (int argc, char *argv[])
         fprintf(stderr, "No pv name specified. ('caget -h' for help.)\n");
         return 3;
     }
-                                /* Start up Channel Access */
+                                /* sTart up Channel Access */
 
-    result = ca_context_create(ca_disable_preemptive_callback);
+    result = ca_context_create(ca_enable_preemptive_callback);
     if (result != ECA_NORMAL) {
         fprintf(stderr, "CA error %s occurred while trying "
                 "to start channel access.\n", ca_message(result));
         return 3;
     }
+    // Global variable to be used by all threads
+    ca_context = ca_current_context();
                                 /* Allocate PV structure array */
 
     pvs = calloc (nPvs, sizeof(pv));
@@ -642,9 +668,10 @@ int main (int argc, char *argv[])
     epicsMutexId printMutex = epicsMutexCreate();
     // Mutex to access result counter
     epicsMutexId resultMutex = epicsMutexCreate();
+    // Global variable
+    queueId = epicsMessageQueueCreate(QUEUE_CAPACITY, sizeof(t_thread_data));
 
-    epicsMessageQueueId queueId = epicsMessageQueueCreate(QUEUE_CAPACITY, sizeof(t_thread_data));
-
+    // We use up to NUMBER_OF_THREADS threads
     int number_of_threads;
     if (nPvs > NUMBER_OF_THREADS) {
         number_of_threads = NUMBER_OF_THREADS;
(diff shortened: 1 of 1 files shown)


modules/ca/src/tools/caget.c
```diff
@@ -283,73 +356,85 @@ static int caget (pv *pvs, int nPvs, RequestT request, OutputT format,
...
+}
+/*
+void test(void* thread_data_void) {
+printf("called\n");
+    //t_thread_data* thread_data = (t_thread_data*)thread_data_void;
+    //epicsSpinLock(thread_data->spin);
+    //epicsMutexUnlock(thread_data->printMutex_lock_main);
+printf("End of test\n");
+}
+*/
```
anjohnson: Commented-out test code, marking it here to ensure it gets removed before merging. This is where you tried the spin-lock, but I think the mutex is better.

modules/ca/src/tools/caget.c
```diff
@@ -283,73 +356,85 @@ static int caget (pv *pvs, int nPvs, RequestT request, OutputT format,
...
+    char pv_num_str[3];
+    sprintf(pv_num_str, "%d", pv_num);
+    strcpy(th_name, "caget");
+    strcat(th_name, pv_num_str);
+}
+
+// Function that runs inside each thread, calling the caget function
+void caget_caller (void* thread_data_void)
+{
+    int result;                 /* CA result */
```
anjohnson, line 419: There is a call at the end of the `main()` function below to `ca_context_destroy()` so I assume there is also one to `ca_context_create()` earlier on. I would like to see all of these caget threads sharing the same CA context, as that means that all channels opened to the same IOC will share the same TCP socket, which uses fewer resources on the IOC. To do that your code needs to get and save the newly-created CA context pointer by calling [`ca_current_context()`](https://epics.anl.gov/base/R7-0/8-docs/CAref.html#ca_current_context), and then each thread should call [`ca_attach_context()`](https://epics.anl.gov/base/R7-0/8-docs/CAref.html#ca_attach_context) to use it. When the thread is finished it should detach from the context again.

modules/ca/src/tools/caget.c
```diff
@@ -549,14 +626,36 @@ int main (int argc, char *argv[])
...
+        epicsThreadMustJoin(thread_id);
+    }
 
-                                /* Shut down Channel Access */
     ca_context_destroy();
+    epicsMutexDestroy(printMutex);
+    epicsMutexDestroy(resultMutex);
 
-    return result;
+    return pvs_with_trouble;
```
anjohnson: I'm not convinced that this return status is a good idea. The C++ standards says this about `main()`:
> Execution of the return (or the implicit return upon reaching the end of main) is equivalent to first leaving the function normally (which destroys the objects with automatic storage duration) and then calling [std::exit](https://en.cppreference.com/w/cpp/utility/program/exit) with the same argument as the argument of the [return](https://en.cppreference.com/w/cpp/language/return) ([std::exit](https://en.cppreference.com/w/cpp/utility/program/exit) then destroys static objects and terminates the program).

and this abount the exit status value:
> If exit_code is 0 or [EXIT_SUCCESS](https://en.cppreference.com/w/cpp/utility/program/EXIT_status), an implementation-defined status indicating successful termination is returned. If exit_code is [EXIT_FAILURE](https://en.cppreference.com/w/cpp/utility/program/EXIT_status), an implementation-defined status indicating unsuccessful termination is returned. In other cases implementation-defined status value is returned.

modules/ca/src/tools/caget.c
```diff
@@ -36,6 +36,9 @@
 #include <cadef.h>
 #include <epicsGetopt.h>
 #include "epicsVersion.h"
+#include "epicsMutex.h"
+#include "epicsExit.h"
+#include "epicsSpin.h"
```
anjohnson: epicsExit.h and epicsSpin.h aren't needed by the current code AFAICS.

anjohnson (changes requested):
Assuming this works it seems an interesting and valid approach. Some changes requested here, but nothing major.

mdavidsaver:
> ... one epicsThread per PV ...

This raises my eyebrow.  [`pvget`](https://github.com/epics-base/pvAccessCPP/blob/f1268adb8ecbacbd74bb66c172d02d9d427bedfd/pvtoolsSrc/pvget.cpp#L205-L207) uses a single worker and a queue.  Something similar could be done with `epicsMessageQueue`.

marciodo:
> epicsMessageQueue

@mdavidsaver I'm thinking about your comment trying to guess what raised your eyebrow. My guess is that my change would allow for an unlimited number of threads to be opened by caget while, at the same time, flooding the network with simultaneous broadcast messages. Is that it?

The solution would be a queue, as you suggested, allowing a limited maximum number of threads to run at the same time.

I'm not familiar with epicsMessageQueue, so I'll start to Google around. In any case, I appreciate it if you have a reference for me to study. (I just found that the EPICS Application Developer's Guide has some information on the Message Queue functions).

anjohnson:
[epicsMessageQueue](https://epics.anl.gov/base/R7-0/8-docs/doxygen/epics_message_queue_8h.html) documentation

mdavidsaver:
> ... I'm thinking about your comment trying to guess what raised your eyebrow ...

The mention of one thread to handle each PV.  The design of CA and PVA protocol and libraries is geared specifically towards avoiding just this situation, and imo. contributes significantly to the scalability of EPICS systems.  Additionally taking one socket per PV would certainly compound this.

To give you a concrete example.  I sometimes issue `dbl > /tmp/pv.list` from IOC shell, then from a system shell do `cat /tmp/pv.list | xargs caget > snapshot.txt`.  A simple way to get a snapshot of the state of an IOC for offline troubleshooting.  Currently this works for large numbers of PVs.  I have used this with an IOC containing ~10,000 records.

I don't think we want to give up this scalability for (perhaps) shorter code.

modules/ca/src/tools/caget.c
marciodo: The original caget code calls ca_context_create() indirectly via:
ca/src/tools/tool_lib.c
connect_pvs -> create_pvs ->

ca/src/client/access.cpp
ca_create_channel -> fetchClientContext -> epicsThreadPrivateGet -> ca_context_create

Also, ca_context_destroy() was in the original caget code. Maybe @ralphlange can help us figure out the reason faster than I, because right now I'm reverse engineering.

My changes made each thread call connect_pvs, which is creating one channel per thread. Maybe each thread should call ca_clear_channel() after printing the PV data and ca_context_destroy() continues in the end of the main function?

anjohnson: Hi Márcio,

caget.c line 533 calls `ca_context_create(ca_disable_preemptive_callback)`, you were thinking of `ca_create_channel()`. The current version of the CA client library is multi-threaded internally, but the parameter passed to `ca_context_create()` controls whether the client application experiences the effects of that multi-threading or not; caget.c's setting `ca_disable_preemptive_callback` makes the API act like a single-threaded library in that the user callbacks will only be called while the main thread is executing inside the routines `ca_pend_io()`, `ca_poll()` or `ca_pend_event()`.

This mode has the advantage that you don't need to protect variables accessed in both the main thread and the callback routines with a mutex or atomic accesses. However a multi-threaded application can only share a context which was created using `ca_context_create(ca_enable_preemptive_callback)`. See the two sections of the CA reference manual starting [here](https://epics.anl.gov/base/R7-0/8-docs/CAref.html#Thread) for more information about the above.

I believe your code at the moment is creating a separate CA context for each thread. This is going to use significantly more resources, since each context creates multiple threads, internal data structures, and its own TCP and UDP sockets for doing name searches and connecting to IOCs. @mdavidsaver's raised eyebrow was related to this; I wasn't thinking about scaling to 10,000 channel names when I suggested that this approach might be valid.

You can still use the `ca_disable_preemptive_callback` mode to connect and fetch all the channel data if you re-work this. When using the callbacks instead of threads you don't actually need the mutexes that you added, the CA library takes care of the synchronization so that your code is never running multiple routines at the same time.

I hope you will be able to find some time to continue working on this.

Thanks,

- Andrew

mdavidsaver:
As a more concrete suggestion, how about exclusively using `ca_array_get_callback()`?

Then the `event_handler()` callback gets the index into the `pvs[]` array (add to `struct pv`), which it enqueues into an `epicsMessageQueueId`.  The main() thread then waits on that queue for notification of which PVs complete, in whatever order this occurs, and prints.  main() needs to keep track of how many PVs remain, and stop when this gets to zero (or timeout).  No extra threads needed.  (ok, maybe one special thread to sleep and then enqueue a special message `-1` to indicate timeout)

modules/ca/src/tools/caget.c
```diff
@@ -444,7 +444,9 @@ void thread_builder (t_thread_data* thread_data, int pv_num)
  *
  * Arg(s) Out:  none
  *
- * Return(s):   Return code = number of PVs that couldn't be reached
+ * Return(s):   Return code = number of PVs that couldn't be reached or
+ *              -1 if an error happened before we could start searching
+ *              for the PVs.
```
ericonr, line 485: What do you do for requests with more than 254 PVs? That's the maximum value that can be represented in a process return, because 255 isn't available, since -1 gets converted to 255, which is not very intuitive.

The exit status of the process feels like a somewhat limited place to express this information.

modules/ca/src/tools/caget.c
```diff
@@ -250,29 +274,76 @@ static int caget (pv *pvs, int nPvs, RequestT request, OutputT format,
...
+        break;
+    case all:
+        print_time_val_sts(&pv_data, reqElems);
+        break;
+    case specifiedDbr:
+        printf("%s\n", pv_data.name);
+        if (pv_data.status == ECA_DISCONN)
+            printf("    *** not connected\n");
+        else if (pv_data.status == ECA_NORDACCESS)
+            printf("    *** no read access\n");
```
ericonr, line 348: `print_time_val_sts` implements slightly different status checking:

https://github.com/epics-base/epics-base/blob/13d6ca598cca495b2e559e808392ed22265c57e3/modules/ca/src/tools/tool_lib.c#L517-L529

Maybe a previous commit should standardize the error handling. Possibly as a separate function encapsulating this logic?

ericonr (commented):
Shouldn't this be opt-in behavior? I have used `caget` and `cainfo` to check that a list of PVs was valid and all those PVs were actually available; or as @mdavidsaver said, to obtain a snapshot of a running IOC. If someone has built scripts relying on this behavior, their error handling might not work for these new conditions, and they would be dealing with invalid data.

In the case where PVs are disconnected, how can I programatically know which PVs are disconnected? If `-a` is the only machine parseable format, then it's easy to differentiate error strings from anything else; but if plain and `-t` formats should also be machine parseable, there's no way to differentiate `printf("*** not connected\n")` from `printf("%s\n", pv->string_value)"` where `pv->string_value` is `"*** not connected"`.

> The main() thread then waits on that queue for notification of which PVs complete, in whatever order this occurs, and prints.

Shouldn't order also be guaranteed? That's the current behavior of the tool, and it is helpful and consistent. It makes sense for something like `camonitor` to not enforce order for the first received values, since it would require additional logic separate from normal operation, but `caget` and `cainfo` are currently sequential, and it would be nice for their output to remain so.

modules/ca/src/tools/caget.c
marciodo: @ericonr, you are totally right. In issue #489, we had a discussion about this, and the definition we ended up agreeing on was:

- 0: success
- 1: some channels not found
- 2: all channels not found
- 3: other error

marciodo:
@ericonr 

> In the case where PVs are disconnected, how can I programatically know which PVs are disconnected? If `-a` is the only machine parseable format, then it's easy to differentiate error strings from anything else; but if plain and `-t` formats should also be machine parseable, there's no way to differentiate `printf("*** not connected\n")` from `printf("%s\n", pv->string_value)"` where `pv->string_value` is `"*** not connected"`.

Do you have a suggestion? How do you do this with the current version of caget?

> > The main() thread then waits on that queue for notification of which PVs complete, in whatever order this occurs, and prints.
> 
> Shouldn't order also be guaranteed? That's the current behavior of the tool, and it is helpful and consistent. It makes sense for something like `camonitor` to not enforce order for the first received values, since it would require additional logic separate from normal operation, but `caget` and `cainfo` are currently sequential, and it would be nice for their output to remain so.

I'll work on an implementation that will preserve the order, if possible.

marciodo:
@mdavidsaver and @ralphlange. @anjohnson wants your opinion about a thing. Implementing this with callbacks only, as Michael suggested almost 2 years ago, and Andrew just suggested to me, would:
1. Eliminate the need for a `-c` argument for caget.
2. Make caget print arrays only for the valid elements instead of adding padding zeroes for arrays not fully completed.

Mainly for point number 2, this is a change in behavior that might disrupt someone's expectations to receive back all the padding zeroes.

mdavidsaver:
> 1. Eliminate the need for a `-c` argument for caget.

Is this true?  Isn't "dynamic" array handling limited to subscription, and not GET?

> Mainly for point number 2, this is a change in behavior

Yes, printing the actual element count vs. the maximum would be a change, and would likely break someone's shell script.  Personally, I would see this as an improvement though.

anjohnson:
> Isn't "dynamic" array handling limited to subscription, and not GET?

@marciodo is right, the issue was that the basic `ca_array_get()` API provides no way for libca to tell the application how many array elements were valid, whereas the callback variant provides the array size in the callback args just like the subscription callback does.

We could always add an option to make something like `caget -z` print all the trailing zero elements, in case someone really needs that, but it would be nice to have the default just print the valid elements without having to give a `-c` flag.

PR #505, opened 2024-06-12
Allow deleting a record at database creation (take 2).

hinxx:
Using a magical record type "#" will allow the user to delete previously created record  from the database - see #464 for more details.

NOTE:

This is a second attempt at the PR after making a mess of my fork while trying to rebase aginst upstream code. This PR is on top of latest upstream 7.0 branch code and includes original code change from #464 along with the release notes.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1119,6 +1119,19 @@ static void dbRecordHead(char *recordType, char *name, int visible)
         return;
     }
 
+    if (recordType[0] == '#' && recordType[1] == 0) {
+        status = dbFindRecord(pdbentry, name);
+        if (status == 0) {
+            dbDeleteRecord(pdbentry);
+            fprintf(stderr, ERL_WARNING ": Record \"%s\" DELETED!\n", name);
+            return; /* done */
+        }
+        fprintf(stderr, ERL_ERROR ": Record \"%s\" not found\n", name);
+        yyerror(NULL);
+        duplicate = TRUE;
+        return;
+    }
+
     status = dbFindRecordType(pdbentry, recordType);
     if (status) {
         fprintf(stderr, "Record \"%s\" is of unknown type \"%s\"\n",
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1123,14 +1123,14 @@ static void dbRecordHead(char *recordType, char *name, int visible)
         status = dbFindRecord(pdbentry, name);
         if (status == 0) {
             dbDeleteRecord(pdbentry);
-            popFirstTemp();
-            dbFreeEntry(pdbentry);
-            duplicate = TRUE;
         } else {
             fprintf(stderr, ERL_WARNING ": Unable to delete record \"%s\".  Not found.\n"
                     "  at file %s line %d\n",
                     name, pinputFileNow->filename, pinputFileNow->line_num);
         }
+        popFirstTemp();
+        dbFreeEntry(pdbentry);
+        duplicate = TRUE;
         return;
     }
     status = dbFindRecordType(pdbentry, recordType);
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1123,15 +1123,12 @@ static void dbRecordHead(char *recordType, char *name, int visible)
         status = dbFindRecord(pdbentry, name);
         if (status == 0) {
             dbDeleteRecord(pdbentry);
-            fprintf(stderr, ERL_WARNING ": Record \"%s\" DELETED!\n", name);
-            return; /* done */
+            popFirstTemp();
+            dbFreeEntry(pdbentry);
+            duplicate = TRUE;
         }
-        fprintf(stderr, ERL_ERROR ": Record \"%s\" not found\n", name);
-        yyerror(NULL);
-        duplicate = TRUE;
         return;
     }
-
     status = dbFindRecordType(pdbentry, recordType);
     if (status) {
         fprintf(stderr, "Record \"%s\" is of unknown type \"%s\"\n",
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1126,6 +1126,8 @@ static void dbRecordHead(char *recordType, char *name, int visible)
             popFirstTemp();
             dbFreeEntry(pdbentry);
             duplicate = TRUE;
+        } else {
+            fprintf(stderr, ERL_WARNING ": Record \"%s\" not found\n", name);
         }
         return;
     }
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1127,7 +1127,9 @@ static void dbRecordHead(char *recordType, char *name, int visible)
             dbFreeEntry(pdbentry);
             duplicate = TRUE;
         } else {
-            fprintf(stderr, ERL_WARNING ": Record \"%s\" not found\n", name);
+            fprintf(stderr, ERL_WARNING ": Unable to delete record \"%s\".  Not found.\n"
+                    "  at file %s line %d\n",
+                    name, pinputFileNow->filename, pinputFileNow->line_num);
         }
         return;
     }
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -1519,7 +1519,6 @@ long dbDeleteRecord(DBENTRY *pdbentry)
     dbBase          *pdbbase = pdbentry->pdbbase;
     dbRecordType    *precordType = pdbentry->precordType;
     dbRecordNode    *precnode = pdbentry->precnode;
-    struct dbCommon *prec = pdbentry->precnode->precord;
     ELLLIST         *preclist;
     long            status;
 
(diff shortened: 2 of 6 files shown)


anjohnson:
The Core Group have already approved this once, thanks for the Release Notes. I just triggered the CI, once that passes I think this is ready to be merged — thanks!

mdavidsaver (changes requested):
I would like to see a unit test.

Since `dbDeleteRecord()` has ~never been used, I would be unsurprised to find some latent bug.

eg. I think this will leak the link string `epicsStrDup()` d in `dbPutString()`.  This would be far from the only leak during DB parsing, and doesn't need to be plugged by this PR.

The cases I can think of:

- add and remove a record.  Verify that it isn't found (by either iteration or hash lookup)
   - set INP/OUT link.  `valgrind` would turn up the leak.
   - add some `info()`
- add record and alias, remove alias.  verify that the alias is gone, but the record remains
- add record and alias, remove record.  verify both are gone

Since none of these cases should fail, I think they could all be in one `.db`.

hinxx:
I'm trying to add unit tests and I'm not sure how to proceed with regards to my fork and this PR branch being couple of commits behind the upstream 7.0. Should I rebase my fork and this PR branch before adding unit test commits or not?

Actually, I've tried rebasing and it looks like I would add commits that I currently lack from upstream 7.0 to this PR; I think this is not what you guys want.

mdavidsaver:
> ... I've tried rebasing and it looks like I would add commits that I currently lack from upstream 7.0 to this PR ...

What gives you this impression?

I rebase pull requests regularly, so I know it can be done successfully.  If you are in doubt though, then keep working where you are.  Or I can try to re-base for you.

hinxx:
> What gives you this impression?

Now that I rebased my PR includes 4 more commits that are not mine. This is what I did:
[git.txt](https://github.com/user-attachments/files/15880973/git.txt)

mdavidsaver:
@hinxx This is what I see (as shown by `gitk`):



Assuming that your repository has two `remote`s setup: `upstream` for this repository, and `origin` for your fork, then what you want to do is:

```
git stash save
# may say "No local changes to save"

git fetch upstream
git fetch origin
# create new local branch "temporary-name" to reflect current PR state
git checkout temporary-name origin/delete-record-magic
# should succeed, leaving only your one commit
git rebase upstream/7.0
```

At this point, stop and look at your local history.  `git diff upstream/7.0` at minimum.  You should see that `delete-record-magic` points to your one commit after `upstream/7.0`.

Now push the rewritten history to our feature branch.  The `+` expresses an intent to overwrite the remote `delete-record-magic` with the local branch `temporary-name`,

```
git push origin +temporary-name:delete-record-magic
```

hinxx:
Thanks @mdavidsaver for the instructions, my git kung-fu is not that strong. I was not doing the temp branch stuff and did not use `+` with git push. I followed what you wrote (I had to use `-b` with the `git checkout`) and now the log looks better to me.

hinxx:
BTW, I plan on changing the existing `dbStaticTest` and add the record deletion stuff; or would you prefer that I create a separate test?

mdavidsaver:
Adding to `dbStaticTest` is just fine.

hinxx:
@mdavidsaver, I think I've addressed all the cases you listed with the units tests.

I added a commit that solves the memory leak of link string `epicsStrDup()` in dbPutString(). I assume this does not solve the potential memory leaks for all link types. I'm willing to improve newly introduced `dbDeleteRecordLinks()` to cover more cases if required.

Apart from that I did not see any additional memory leaks when running `valgrind --leak-check=full  ./dbStaticTest`.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -1478,6 +1478,42 @@ long dbDeleteAliases(DBENTRY *pdbentry)
...
 
+static long dbDeleteRecordLinks(dbRecordType *rtyp, struct dbCommon *prec)
+{
+    short i;
+
+    for (i=0; i<rtyp->no_links; i++) {
+        dbFldDes *pflddes = rtyp->papFldDes[rtyp->link_ind[i]];
+        DBLINK *plink = (DBLINK *)(((char *)prec) + pflddes->offset);
+
+        // TODO: How to handle different link types?
```
mdavidsaver: I think the answer is:

https://github.com/epics-base/epics-base/blob/72d50ce2749bbc791907a32c1ed77af1333a21f5/modules/database/src/ioc/dbStatic/dbStaticLib.c#L107

hinxx: Addressed in https://github.com/epics-base/epics-base/pull/505/commits/3dc1ef2b73509346d86d3d9e3b4940161f13622d.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1119,6 +1119,19 @@ static void dbRecordHead(char *recordType, char *name, int visible)
...
 
+    if (recordType[0] == '#' && recordType[1] == 0) {
+        status = dbFindRecord(pdbentry, name);
+        if (status == 0) {
+            dbDeleteRecord(pdbentry);
+            fprintf(stderr, ERL_WARNING ": Record \"%s\" DELETED!\n", name);
+            return; /* done */
+        }
+        fprintf(stderr, ERL_ERROR ": Record \"%s\" not found\n", name);
+        yyerror(NULL);
```
mdavidsaver: Should attempting to delete a non-existent record be an error?  Was this discussed already?

I do think that the warning on successful deletion should be removed.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1119,6 +1119,19 @@ static void dbRecordHead(char *recordType, char *name, int visible)
         return;
     }
 
+    if (recordType[0] == '#' && recordType[1] == 0) {
+        status = dbFindRecord(pdbentry, name);
+        if (status == 0) {
+            dbDeleteRecord(pdbentry);
+            fprintf(stderr, ERL_WARNING ": Record \"%s\" DELETED!\n", name);
+            return; /* done */
```
mdavidsaver: Should the `duplicate` flag be set here as well?  What would (or should) happen with the following?

```
record("#", "foo") {
    field(VAL, "4")
}
```

mdavidsaver (commented):
A note on something which caught my attention, but which I think is __not__ an issue.

Looking at the earlier parts of `dbRecordHead()` I notice two allocations.

```c
    pdbentry = dbAllocEntry(savedPdbbase);
...
    allocTemp(pdbentry);
```

`dbDeleteRecord()` operates on `pdbentry`, but does not free() it.

Both should be matched in `dbRecordBody()`.  I think this will happen excepting `yyerror()`, when we currently accept there will be leaks.

```c
    pdbentry = (DBENTRY *)popFirstTemp();
...
    dbFreeEntry(pdbentry);
```

modules/database/src/ioc/dbStatic/dbLexRoutines.c
hinxx: > Should attempting to delete a non-existent record be an error? Was this discussed already?

I did not get any feedback on that one so far. I guess it could be silently ignored and not raise an error. I copied this code from the `"*"` handling code.

> I do think that the warning on successful deletion should be removed.

Will remove.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
hinxx: That would result in trouble.
Addressed in [37c1e8b](https://github.com/epics-base/epics-base/pull/505/commits/37c1e8b78ef02954d835e0a3e0cf2c63989bf573).

modules/database/src/ioc/dbStatic/dbLexRoutines.c
hinxx: Addressed in [37c1e8b](https://github.com/epics-base/epics-base/pull/505/commits/37c1e8b78ef02954d835e0a3e0cf2c63989bf573)

hinxx:
> A note on something which caught my attention, but which I think is **not** an issue.

I might have done something about this leak :

https://github.com/epics-base/epics-base/pull/505/commits/37c1e8b78ef02954d835e0a3e0cf2c63989bf573#diff-b2bf38a4674a0f1de1b667b0245f3e5a010fa89633d4caa913183454d60cee30R1126-R1127

anjohnson:
Core: Please add a warning to be printed if the record doesn't exist when deleted (but it's not an error, the parsing should continue). Users may be able to create the record immediately before deleting it to suppress that warning, but we'd prefer to push users to clean up their code.

mdavidsaver:
@hinxx I have added two commits to the PR.  The first adds file and line to the no-such-record warning.  The other adds a test to delete a non-existent record.  Unhappily, this cause `dbStaticTest` to crash.

```
$ ./dbStaticTest 
1..340
# Alloc jlinkz 0x55b636dfde90
# Free jlinkz 0x55b636dfde90
WARNING: Unable to delete record "no:such:record".  Not found.
  at file (null) line 37
WARNING: Unable to delete record "also:non:existant".  Not found.
  at file (null) line 38
Segmentation fault (core dumped)
```

hinxx:
@mdavidsaver good catch!

AFAICT, when this line is executed

https://github.com/epics-base/epics-base/blob/91e1d0ab8022d610bb12b5536a7a6a21644d47f7/modules/database/src/ioc/dbStatic/dbLexRoutines.c#L1170-L1171

the `dbGetRecordTypeName`  at https://github.com/epics-base/epics-base/blob/91e1d0ab8022d610bb12b5536a7a6a21644d47f7/modules/database/src/ioc/dbStatic/dbStaticLib.c#L1222 croaks because `pdbentry->precordType` is NULL.

We could make `dbGetRecordTypeName` more careful, just as `dbGetRecordName` is and return NULL instead; I tried this and the following was observed:

```
# Alloc jlinkz 0x555555591210
# Free jlinkz 0x555555591210
WARNING: Unable to delete record "no:such:record".  Not found.
  at file (null) line 37
WARNING: Unable to delete record "also:non:existant".  Not found.
  at file (null) line 38
(null) Record "(null)" does not have a field "FOO"
ERROR:  at or before ')' in standard input line 39
WARNING: Unable to delete record "testdelrec11".  Not found.
  at file (null) line 55
Bail out! Unable to load ..:dbStaticTestRemove.db
```

The test is aborted in a similar fashion as if one would try to load (not delete) an unknown record type; ie.

```
record(xx, "testdelrecxx") { }
```

Might be a bit far fetched, but would this outcome be acceptable also if someone tries to delete a non-existent record/field?

hinxx:
@mdavidsaver I've fixed the segfault you uncovered and the test suite should run as expected.

anjohnson:
Core Group: Approved.

Issue #487, opened 2024-05-29
Local variables in iocsh scripts


exzombie:
Being able to define local variables would make it easier to avoid repetition in iocsh scripts loaded by `iocshLoad`. Currently, the only way to define a variable is using `epicsEnvSet`, which creates a globally visible environment variable.

To be specific, I would like something like the following, with emphasis on the `$(PORT)`; the other stuff is there only to illustrate the use case (hopefully not obscuring the goal too much)

`script1.cmd`:
```
localSet("PORT", "A_COMPLICATED__AND_DYNAMIC_PORT_NAME")
drvAsynIPPortConfigure("$(PORT)", "$(ADDRESS)", 0, 0, 0)
dbLoadRecords("database.db", "PREFIX=$(DEVICE),PORT=$(PORT)")
...
iocshLoad("script2.cmd", "DEVICE=$(DEVICE),PARAM=SOMETHING")
...
```

`script2.cmd`:
```
localSet("PORT", "SOMETHING_COMPLETELY_DIFFERENT")
drvCustomDriver("$(PORT)", "ARG=$(PARAM)")
dbLoadRecords("other.db", "PREFIX=$(DEVICE),PORT=$(PORT)")
```

It would be nice if `script2.cmd`, which could come from another module, didn't need to know that it shouldn't use the name `PORT`. This cannot be achieved with `epicsEnvSet`. A possible workaround is to "namespace" variable names by prefixing them with something unique to each script, which is less than ideal.

Given that `iocshLoad` already creates a scope for the purpose of scoping its macro arguments, I'm wondering if this issue could be solved in a straightforward manner by exposing `macPutValue()` through an iocsh command.

anjohnson:
I like this, but would personally prefer the commands `set` and `show` — are these names better than `epicsSet` and `epicsShow` (to match `epicsEnvSet` etc.) or `iocshSet` and `iocshShow`? Implement with calls to `macPutValue(context->handle, name, value)` and `macReportMacros(context->handle)`. The implementation should go in iocsh.cpp.

@oksanagit has shown that using `macPutValue` with the iocsh `context->handle` does not put the name/value in the global environment, which is what @exzombie suggested.

This is related to #469; that issue would be modified to make `IOCSH_STARTUP_SCRIPT` a local variable; that would remove it from the environment and make it only visible to other iocsh commands, is that likely to be a problem?

mdavidsaver:
I like this idea.

[...]

Issue #469, opened 2024-03-16
iocshLoad sets IOCSH_STARTUP_SCRIPT but doesn't undo that afterwards


anjohnson:
The `iocshLoad` command takes the path to a script and a list of macro parameters. By pushing the macLib scope before setting them, those macro values only exist while that script is running, and get discarded afterwards.

However the `iocshLoad()` code currently calls `epicsEnvSet()` to write the script name to the `IOCSH_STARTUP_SCRIPT` variable before it passes the script name and macro list to `iocshBody()` to execute it. There is code in `epicsEnvSet()` to allow scripts that are run using `iocshLoad` to set variables in the global environment using the `epicsEnvSet` command, so no matter how many `iocshLoad` operations are in progress the top-level environment variable by that name will always be overwritten. This means that the variable doesn't get restored when the scope is popped, and at the end of the original st.cmd file the `IOCSH_STARTUP_SCRIPT` variable is left set to the last script that was executed by `iocshLoad` at any level.

One solution might be to move the code which sets the environment variable into `iocshBody` after the new macLib scope has been pushed. However the original startup script path should be left set after the top-level script has finished running, so this might not be the best answer.

mdavidsaver:
> ... at the end of the original st.cmd file the `IOCSH_STARTUP_SCRIPT` variable is left set to the last script ...

I find this a slightly surprising, I would probably have unset() afterwards.  What was your motivation?  (c11a31f5499f54bc471693a799f691f69b41da55)

mdavidsaver:
> ... macro parameters. By pushing the macLib scope ...

This idea may run into the same 256 character limit in macLib which has stalled https://github.com/epics-base/epics-base/pull/102#issuecomment-796958504 .

anjohnson:
> I find this a slightly surprising, I would probably have unset() afterwards.

Okay that bit is wrong, `IOCSH_STARTUP_SCRIPT` _should_ be unset after the script has finished executing.

However if one script calls another using an `iocshLoad` command, `IOCSH_STARTUP_SCRIPT` should be reverted back to the first script after that `iocshLoad` exits, which it doesn't now. That's the real bug.

anjohnson:
> 256 character limit in macLib

That sounds like another good Codeathon project. This comment appears in the `macLibNOTES` file in the source directory:

> Macro values can reference other macros in an arbitrarily complex way.
The only current restrictions are that a macro name or value cannot
exceed 256 characters in length. This restriction could fairly easily
[...]

PR #507, opened 2024-06-12
Add local variables to iocsh

oksanagit:
…Load() and now it is a local macro. #487: new fuctions are added set() and show(), allowing setting and showing local macros

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -1110,6 +1110,8 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
 
     macPushScope(handle);
     macInstallMacros(handle, defines);
+    if (pathname)
+        macPutValue(handle, "IOCSH_STARTUP_SCRIPT", pathname);
 
     wasOkToBlock = epicsThreadIsOkToBlock();
     epicsThreadSetOkToBlock(1);
@@ -1327,6 +1329,8 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
 int epicsStdCall
 iocsh (const char *pathname)
 {
+    if (pathname && !getenv("IOCSH_STARTUP_SCRIPT"))
+        osdEnvSet("IOCSH_STARTUP_SCRIPT", pathname);
     return iocshLoad(pathname, NULL);
 }
 
@@ -1339,8 +1343,6 @@ iocshCmd (const char *cmd)
 int epicsStdCall
 iocshLoad(const char *pathname, const char *macros)
 {
-    if (pathname && !getenv("IOCSH_STARTUP_SCRIPT"))
-        epicsEnvSet("IOCSH_STARTUP_SCRIPT", pathname);
     return iocshBody(pathname, NULL, macros);
 }
 
@@ -1502,6 +1504,64 @@ static void iocshLoadCallFunc(const iocshArgBuf *args)
     iocshSetError(iocshLoad(args[0].sval, args[1].sval));
 }
 
+/* set */
+struct setShowCtx {
+    char *match;
+    int none;
+};
+static long setShow(void *user, const char *name, const char *value)
+{
+    setShowCtx *shUsr = (setShowCtx *) user;
+
+    if (shUsr->match && !epicsStrGlobMatch(name, shUsr->match)) {
+        return 0;
+    }
+    shUsr->none = 0;
+    printf("  %s = \"%s\"\n", name, value);
+    return 0;
+}
+
+static const iocshArg iocshSetArg0 = { "name",iocshArgString};
+static const iocshArg iocshSetArg1 = { "value", iocshArgString};
+static const iocshArg *iocshSetArgs[2] = {&iocshSetArg0, &iocshSetArg1};
+static const iocshFuncDef iocshSetFuncDef = {"set",2,iocshSetArgs,
+    ANSI_BOLD("set") " -name\n\n"
+    ANSI_BOLD("set") " name/pattern\n\n"
+    ANSI_BOLD("set") "\n\n"
+    "Set, delete or show ioc shell variables\n"};
+static void iocshSetCallFunc(const iocshArgBuf *args)
+{
+    iocshContext *context;
+
+    if (!iocshContextId) return;
+    context = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
+
+    if (!context) return;
+
+    if (!args[0].sval) {
+        setShowCtx shCtx = {NULL, 1};
+
+        macIterateMacros(context->handle, setShow, &shCtx);
+        if (shCtx.none)
+            printf("No shell variables set.\n");
+    }
+    else if (!args[1].sval) {
+        if (args[0].sval[0] == '-') {
+            macPutValue(context->handle, args[0].sval + 1, NULL);
+        }
+        else {
+            setShowCtx shCtx = {args[0].sval, 1};
+
+            macIterateMacros(context->handle, setShow, &shCtx);
+            if (shCtx.none)
+                printf("No shell variables matched.\n");
+        }
+    }
+    else {
+        macPutValue(context->handle, args[0].sval, args[1].sval);
+    }
+}
+
 /* iocshRun */
 static const iocshArg iocshRunArg0 = { "command",iocshArgString};
 static const iocshArg iocshRunArg1 = { "macros", iocshArgString};
@@ -1610,6 +1670,7 @@ static void iocshOnce (void *)
     iocshRegisterImpl(&helpFuncDef,helpCallFunc);
     iocshRegisterImpl(&iocshCmdFuncDef,iocshCmdCallFunc);
     iocshRegisterImpl(&iocshLoadFuncDef,iocshLoadCallFunc);
+    iocshRegisterImpl(&iocshSetFuncDef,iocshSetCallFunc);
     iocshRegisterImpl(&iocshRunFuncDef,iocshRunCallFunc);
     iocshRegisterImpl(&onFuncDef, onCallFunc);
     iocshTableUnlock();
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -90,6 +90,10 @@ static epicsThreadPrivateId iocshContextId;
 
 static void iocshInit (void);
 
+/* These are defined in osi/os/.../osdEnv.c */
+extern void osdEnvSet(const char *name, const char *value);
+extern void osdEnvUnset (const char *name);
+
 /*
  * I/O redirection
  */
@@ -1349,19 +1353,13 @@ iocshRun(const char *cmd, const char *macros)
 }
 
 /*
- * Needed to work around the necessary limitations of macLib and
- * environment variables. In every other case of macro expansion
- * it is the expected outcome that defined macros override any
- * environment variables.
+ * Environment variables are global, so setting one should clear any
+ * iocsh variables (macros) with the same name. However we do allow
+ * local iocsh variables to hide environment variables in this context
+ * or child iocsh contexts created for scripts run using iocshLoad.
  *
- * iocshLoad/Run turn this on its head as it is very likely that
- * an epicsEnvSet command may be run within the context of their
- * calls. Thus, it would be expected that the new value would be
- * returned in any future macro expansion.
- *
- * To do so, the epicsEnvSet command needs to be able to access
- * and update the shared MAC_HANDLE that the iocsh uses. Which is
- * what this function is provided for.
+ * To implement that, epicsEnvSet and epicsEnvUnset clear local macros
+ * by calling iocshEnvClear.
  */
 void epicsStdCall
 iocshEnvClear(const char *name)
@@ -1376,6 +1374,21 @@ iocshEnvClear(const char *name)
         }
     }
 }
+
+void epicsStdCall
+epicsEnvSet (const char *name, const char *value)
+{
+    iocshEnvClear(name);
+    osdEnvSet(name, value);
+}
+
+void epicsStdCall
+epicsEnvUnset (const char *name)
+{
+    iocshEnvClear(name);
+    osdEnvUnset(name);
+}
+
 
 /*
  * Internal commands
--- a/modules/libcom/src/macLib/macCore.c
+++ b/modules/libcom/src/macLib/macCore.c
@@ -301,12 +301,9 @@ epicsStdCall macPutValue(
 long                            /* strlen(value), <0 if undefined */
 epicsStdCall macGetValue(
     MAC_HANDLE  *handle,        /* opaque handle */
-
     const char  *name,          /* macro name or reference */
-
     char        *value,         /* string to receive macro value or name */
                                 /* argument if macro is undefined */
-
     long        capacity )      /* capacity of destination buffer (value) */
 {
     MAC_ENTRY   *entry;         /* pointer to this macro's entry structure */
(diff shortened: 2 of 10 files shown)


modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1325,8 +1327,7 @@ iocshCmd (const char *cmd)
 int epicsStdCall
 iocshLoad(const char *pathname, const char *macros)
 {
-    if (pathname)
-        epicsEnvSet("IOCSH_STARTUP_SCRIPT", pathname);
+// IOCSH_STARTUP_SCRIPT is set inside iocshBody()
```
exzombie: I don't think this comment is needed. Documenting history of the code in code comments is important when the history affects the present. But this is not the case here, I doubt knowing that `IOCSH_STARTUP_SCRIPT` used to be set here once upon a time will help the future reader.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1367,6 +1368,35 @@ iocshEnvClear(const char *name)
     }
 }
 
+void epicsStdCall
+iocshSet(const char *name, const char * value)
```
exzombie: Please check for `NULL`, and if `name` is an empty string. Also, consider the return value of `macPutValue` and handle it using `iocshSetError()`.

I think `macPutValue` should check for `NULL` and empty name itself, but I checked and it doesn't. Well, somebody has to :)

exzombie: Speaking of `iocshSetError()`, I see now that this change removes a bunch of error handling calls that were added just yesterday. I believe you were making changes against an older version of the code. Please rebase so that only the desired changes are made part of this pull request.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1471,7 +1498,30 @@ static const iocshFuncDef iocshLoadFuncDef = {"iocshLoad",2,iocshLoadArgs,
...
+
+/* show */
+static const iocshArg iocshShowArg0 = { "pathname",iocshArgString};
+static const iocshArg iocshShowArg1 = { "macros", iocshArgString};
+static const iocshArg *iocshShowArgs[2] = {&iocshShowArg0, &iocshShowArg1};
+static const iocshFuncDef iocshShowFuncDef = {"show",2,iocshShowArgs,
+                                              "Shows all set local marcros\n"};
+static void iocshShowCallFunc(const iocshArgBuf *args)
+{
+    iocshShow();
```
exzombie: The metadata of these two commands is incomplete. The `show` command takes no arguments, and `iocshShowArgs` should reflect that. The `set` command *does* take two arguments, but the first one is no path, and the second one is not a list of macros. Please update this documentation.

I would avoid refering to "macros" for these user-facing descriptions. From the user's point of view, they are dealing with iocsh variables and don't care that the underlying implementation relies on the macro machinery.

exzombie:
@oksanagit, I see that this pull request is marked as a draft, but I chose to review it regardless, hoping to help you finish it. I'm sorry if I have commented on something you already know. But if I was too terse, ask away or stop by :)

modules/libcom/src/iocsh/iocsh.cpp
anjohnson: The new functions `iocshSet()` and `iocshShow()` should be marked `static` and not `epicsStdCall`; they will only be called from within this file. In fact unless they are called from elsewhere in this file (I don't think they are) it would be better to combine their functionality into the `iocshSetCallFunc()` and `iocshShowCallFunc()` routines directly; there's no need for them to even be separate routines.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1367,6 +1368,35 @@ iocshEnvClear(const char *name)
...
+iocshShow()
+{
+    iocshContext *context;
+
+    if (iocshContextId) {
+        context = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
+
+        if (context != NULL) {
+            macReportMacros(context->handle); // returns long, do we need it?
+            // takes macReportMacros((MAC_HANDLE *handle ), where to take it from?
```
anjohnson: We're going to need to do a bit more for this routine, since the output from `macReportMacros()` is more complex and shows internal information that would only confuse users. That may mean modifying macLib as well, but we can talk about that later.

anjohnson (changes requested):
I'm out of time for tonight, we can talk more tomorrow.

modules/libcom/src/iocsh/iocsh.cpp
mdavidsaver: > ... shows internal information that would only confuse users ...

I won't disagree, but does macLib currently provide any other report/dump/show functions?

Can we avoid scope creep into `macCore.c`?

modules/libcom/src/iocsh/iocsh.cpp
mdavidsaver: #507 may conflict with #508, #497, and perhaps #459.  @JJL772 fyi.

mdavidsaver: ... and yes, also 6ca716a77f24247e96952d1778bb1673600bb420.

ralphlange:
I'm not sure if I understand the commit message
'... IOCSH_STARTUP_SCRIP was moved from iocshLoad() to iocshLoad() ...'

oksanagit:
> I'm not sure if I understand the commit message
> '... IOCSH_STARTUP_SCRIP was moved from iocshLoad() to iocshLoad() ...'

This line should be ....IOCSH_STARTUP_SCRIP was moved from iocshLoad() to iocshBody() ... My bad

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1100,6 +1100,8 @@ iocshBody (const char *pathname, const char *commandLine, const char *macros)
 
     macPushScope(handle);
     macInstallMacros(handle, defines);
+     if (pathname)
```
mdavidsaver: Somehow the indentation is 5 spaces on this line.  Please change to 4.  (the next line has 8, which is correct)

oksanagit:
@mdavidsaver , could you point me to the example which might help me to write a unit test for the new set command? I was looking at iocshTest.cpp different tests.  And so far, with try and fail, I had 100% the later. You seem to be the author of this file. thx

modules/libcom/src/iocsh/iocsh.cpp
anjohnson: @mdavidsaver Unfortunately no, and looking at the output from `macReportMacros()` using @oksanagit's now-fixed branch we definitely need new code to generate something more like the output from `var` and `epicsEnvShow` instead of this:
```
Cox$ bin/darwin-aarch64/softIoc st.cmd 
# <<< st.cmd ---
echo "IOCSH_STARTUP_SCRIPT = st.cmd"
IOCSH_STARTUP_SCRIPT = st.cmd
show
e name             rawval           value
- ----             ------           -----
s ----             ------           -----
  IOCSH_STARTUP_SCRIPT st.cmd           st.cmd
#
iocshLoad l0.cmd "level=0,TOP_STARTUP_SCRIPT=st.cmd"
# <<< l0.cmd ---
show
e name             rawval           value
- ----             ------           -----
s ----             ------           -----
  IOCSH_STARTUP_SCRIPT st.cmd           st.cmd
s ----             ------           -----
  level            0                0
  TOP_STARTUP_SCRIPT st.cmd           st.cmd
  IOCSH_STARTUP_SCRIPT l0.cmd           l0.cmd
#
iocshLoad l1.cmd "level=1,PARENT=l0.cmd"
# <<< l1.cmd ---
echo "IOCSH_STARTUP_SCRIPT=l1.cmd, level=1"
IOCSH_STARTUP_SCRIPT=l1.cmd, level=1
show
e name             rawval           value
- ----             ------           -----
s ----             ------           -----
  IOCSH_STARTUP_SCRIPT st.cmd           st.cmd
s ----             ------           -----
  level            0                0
  TOP_STARTUP_SCRIPT st.cmd           st.cmd
  IOCSH_STARTUP_SCRIPT l0.cmd           l0.cmd
s ----             ------           -----
  level            1                1
  PARENT           l0.cmd           l0.cmd
  IOCSH_STARTUP_SCRIPT l1.cmd           l1.cmd
# --- l1.cmd >>>
#
show
e name             rawval           value
- ----             ------           -----
s ----             ------           -----
  IOCSH_STARTUP_SCRIPT st.cmd           st.cmd
s ----             ------           -----
  level            0                0
  TOP_STARTUP_SCRIPT st.cmd           st.cmd
  IOCSH_STARTUP_SCRIPT l0.cmd           l0.cmd
# --- l0.cmd >>>
#
echo "IOCSH_STARTUP_SCRIPT = st.cmd"
IOCSH_STARTUP_SCRIPT = st.cmd
show
e name             rawval           value
- ----             ------           -----
s ----             ------           -----
  IOCSH_STARTUP_SCRIPT st.cmd           st.cmd
# --- st.cmd >>>
epics> exit
Cox$ cat st.cmd l0.cmd l1.cmd 
# <<< st.cmd ---
echo "IOCSH_STARTUP_SCRIPT = $(IOCSH_STARTUP_SCRIPT)"
show
#
iocshLoad l0.cmd "level=0,TOP_STARTUP_SCRIPT=$(IOCSH_STARTUP_SCRIPT)"
#
echo "IOCSH_STARTUP_SCRIPT = $(IOCSH_STARTUP_SCRIPT)"
show
# --- st.cmd >>>
# <<< l0.cmd ---
show
#
iocshLoad l1.cmd "level=1,PARENT=$(IOCSH_STARTUP_SCRIPT)"
#
show
# --- l0.cmd >>>
# <<< l1.cmd ---
echo "IOCSH_STARTUP_SCRIPT=$(IOCSH_STARTUP_SCRIPT), level=$(level)"
show
# --- l1.cmd >>>
```
The separator lines labelled `s` in the show output LH column are where each new scope begins. Look at the `show` output from the l1.cmd file; that confuses even me at times.

oksanagit:
I hope all these failures are because new unit tests were added.

mdavidsaver:
I think the tests you want to preform would be something like the following.  Create two cmd files to test scoping, and using environment variables to communicate results out to iocshTest.c.

```c
testFile("iocshTestLocal1.cmd");
testOk1(getenv("innerA") && strcmp(getenv("innerA"), "A outer")==0);
testOk1(getenv("innerB") && strcmp(getenv("innerB"), "B inner")==0);
testOk1(getenv("outerB") && strcmp(getenv("outerB"), "B outer")==0);
```

In `iocshTestLocal1.cmd`

```
set "A" "A outer"
set "B" "B outer"
iocshLoad "iocshTestLocal2.cmd"
epicsEnvSet "outerB" "$(B)"
```

In `iocshTestLocal2.cmd`

```
set "B" "B inner"
epicsEnvSet "innerA" "$(A)"
epicsEnvSet "innerB" "$(B)"
```

oksanagit:
When the show() low level support is done I can write unit tests later in another PR @anjohnson

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1474,6 +1476,43 @@ static void iocshLoadCallFunc(const iocshArgBuf *args)
...
+                                              "Sets ioc shell variable to value\n"};
+static void iocshSetCallFunc(const iocshArgBuf *args)
+{
+    iocshContext *context;
+
+    if (iocshContextId) {
+        context = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
+
+        if (context != NULL) {
+            macPutValue(context->handle, args[0].sval, args[1].sval);
```
anjohnson: I discovered I can segfault the IOC by running the command `set` with no parameters, please add a check for `args[0].sval == NULL`. However it occurred to me that we could omit the `show` command completely, and just have `set` with no parameters display the list of variables, like we already do with `var`.

However, the following commands _should_ be allowed as the second `set` deletes the variable that was created by the first one.
```
set abcd 1234
set abcd
```

anjohnson (commented):
Hi @oksanagit, I worked on the macLib changes at LGA while waiting for my delayed flight on Friday and I have something that works, but it needs some additional modifications to untangle it from the `epicsEnvSet()` routines, and I have to add tests for the new functionality there.

If you add the check for `args[0].sval == NULL` that I discuss below and remove the `show` command completely we could merge this PR without waiting for my additions. I can adjust the NULL test and make that call the code in a later PR.

modules/libcom/src/iocsh/iocsh.cpp
anjohnson: I lied about the segfault above, I just found my notes and it required these:
```
set abcd 1234
set
```
The reason is still the same, `macPutValue()` doesn't like the parameter name to be NULL.

oksanagit: @anjohnson I am working on it, and will test. Will write here if have questions.

oksanagit:
@anjohnson please take a look, if this is what you meant having at the end. Also, your comment about crashing ioc, was helpful.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1474,6 +1476,33 @@ static void iocshLoadCallFunc(const iocshArgBuf *args)
...
+    if (iocshContextId) {
+        context = (iocshContext *) epicsThreadPrivateGet(iocshContextId);
+
+        if (context != NULL) {
+            if(args[0].sval !=0){
+                macPutValue(context->handle, args[0].sval, args[1].sval);
+            }
+        else {
+            macReportMacros(context->handle);
+        }
```
anjohnson: This `else` block needs to be indented by 4 more spaces.

However I don't think we should expose the output from `macReportMacros()` to our users, I'd rather you just do something like this for now:
```C
    else {
        printf("    'set' with no parameters is still being implemented\n");
    }
```

anjohnson (commented):
This looks good, just one more small change please, see below.

Thanks!

mdavidsaver:
Core Group: In trying to understand this PR and #551, why was the setting of `IOCSH_STARTUP_SCRIPT` moved from `iocshLoad()` to one of the calling functions `iocsh()`?

Is it worth trying to split out this part of the change as a separate PR?  eg. by amending #551

anjohnson:
The description in issue #469 answers the Core Group's question, this was the first bug @oksanagit was fixing — the first commit moved setting `IOCSH_STARTUP_SCRIPT` as a macLib macro into `iocshBody()` within the `macPushScope()` so at the end of an `ioshLoad`ed script (i.e. an `iocshLoad` command inside the main `st.cmd` file) the previous value of it and any other variables set in the command will be restored.

We realized later that we do have to call `epicsEnvSet()` at least once, because the startup script's name needs to remain in the environment after it ends; that explains the third commit, setting it as an environment variable in the `iocsh()` function.

I'm still working (off and on) on an iterate method for macLib...

anjohnson:
Release Notes for this updated version of the PR, not committed to avoid conflicts:

### New `set` and `show` commands for iocsh variables

Since EPICS 3.15.1 the IOC shell command `iocshLoad` has been available to run
an iocsh script and pass it `variable=value` settings for iocsh variables that
are limited in scope to that script. However there was no way to set variables
inside a script that would also disappear when it ends.

This release includes the new iocsh commands `set name value` to set an iocsh
variable in the current scope, and `show` to display all iocsh variables visible
in the current scope. Omitting `value` from the `set` command marks that name as
deleted in the current scope without affecting the parent scope.

Scripts that set variables which must be visible after the script ends should
continue to use `epicsEnvSet` or `epicsEnvUnset`, which both delete any iocsh
variables with the same name in all iocsh scopes.

Any iocsh variables given the same name as an environment variable will hide the
environment variable when iocsh expands `${name}` or `$(name)` in the current
and any child scopes; the original value will reappear when the iocsh variable
goes out of scope.

A new macLib routine was added to libCom to support the `show` command above,
`long macIterateMacros(handle, callback, user)` calls `callback()` for each
macro defined in `handle` passing it `user` followed by the macro's name and
fully-expanded value as strings. See macLib.h for details.

modules/libcom/src/macLib/macCore.c
```diff
@@ -465,14 +462,75 @@
...
+    for ( entry = first( handle ); entry != NULL; entry = next( entry ) ) {
+        const char *value;
+        MAC_ENTRY *final;
+
+        /* Ignore scope markers */
+        if ( entry->special ) continue;
+
+        /* There may be new values for this name in later scopes.
+         * Find the latest matching entry visible in any scope. */
+        final = lookup( handle, entry->name, FALSE );
```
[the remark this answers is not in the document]
anjohnson: The `lookup()` routine will return `entry` if no other macro is defined with the same name in later scopes. The above call should never be able to return `NULL`.

anjohnson (approved):

anjohnson:
@mdavidsaver I moved the `osdEnvSet()` and `osdEnvUnset()` declarations into iocsh.cpp since that's the only file which has calls to them, and removed all their decorations.

mdavidsaver:
> ... moved the `osdEnvSet()` and `osdEnvUnset()` declarations into iocsh.cpp ...

Can you move these into a private header included by both iocsh.cpp and the osdEnv.c files?

Otherwise the compiler can not detect if the function signatures are inconsistent.

How about `module/libcom/src/osi/osdEnvPvt.h`?

anjohnson:
Converting to draft for now, I now think this needs some tests for the macIterateMacros() API and maybe combining the new `show` command into `set` (which I will do).

Issue #498, opened 2024-06-11
missing iocshSetError()


mdavidsaver:
The iocsh `on error ...` construct depends on shell functions calling `iocshSetError()` to report errors.  This is so far not been done consistently.

#471 was due to `iocshLoadCallFunc()` not calling `iocshSetError()`.  Fixed by 6ca716a77f24247e96952d1778bb1673600bb420

mdavidsaver:
I think this would make a reasonable codeathon project.  Search the source for `CallFunc.*iocshArgBuf` and see that `iocshSetError()` is called if there is an error result available.

JJL772:
I'm going to start looking into this for the codeathon

PR #508, opened 2024-06-12
Add iocshSetError in a bunch of places

JJL772:
A quick attempt at addressing #498 

Many of the `dbxxx` and `asxxx` functions have a return type `int`, but always return 0. I wrapped them in `iocshSetError` regardless, just in case that changes in the future (as unlikely as that may be...)

### DIFF
--- a/modules/database/src/ioc/misc/dlload.c
+++ b/modules/database/src/ioc/misc/dlload.c
@@ -10,11 +10,13 @@
 #include "iocsh.h"
 #include "epicsExport.h"
 
-IOCSH_STATIC_FUNC void dlload(const char* name)
+IOCSH_STATIC_FUNC int dlload(const char* name)
 {
     if (!epicsLoadLibrary(name)) {
         printf("epicsLoadLibrary failed: %s\n", epicsLoadError());
+        return -1;
     }
+    return 0;
 }
 
 static const iocshArg dlloadArg0 = { "path/library.so", iocshArgStringPath};
@@ -28,7 +30,7 @@ static const iocshFuncDef dlloadFuncDef = {
 };
 static void dlloadCallFunc(const iocshArgBuf *args)
 {
-    dlload(args[0].sval);
+    iocshSetError(dlload(args[0].sval));
 }
 
 static void dlloadRegistar(void) {
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -510,6 +510,7 @@ static void rtshellCallFunc(const iocshArgBuf *args)
     
     if (!cmd) {
         fprintf(stderr, "ERR: No such command\n");
+        iocshSetError(-1);
     
     } else {
         fflush(stdout);
@@ -517,6 +518,7 @@ static void rtshellCallFunc(const iocshArgBuf *args)
         ret = (*cmd->command)(args[1].aval.ac,args[1].aval.av);
         fflush(stdout);
         fflush(stderr);
+        iocshSetError(ret);
         if(ret)
             fprintf(stderr, "ERR: %d\n",ret);
     }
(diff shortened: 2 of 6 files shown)


modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -611,20 +613,10 @@ static void nfsMountCallFunc(const iocshArgBuf *args)
         }
         *cp = '/';
     }
-    nfsMount(args[0].sval, args[1].sval, args[2].sval);
+    iocshSetError(nfsMount(args[0].sval, args[1].sval, args[2].sval));
 }
 #endif
 
-
-void zoneset(const char *zone)
```
mdavidsaver, line 619: Please reverse this change and keep `zoneset()` as a separate function.

There is method to this seemingly unnecessary verbosity.  RTEMS has alternate shells which directly lookup and execute functions by symbol names.

Changing the return type to `int` is OK.

modules/libcom/RTEMS/score/rtems_init.c
```diff
@@ -483,26 +483,21 @@ static void nfsMountCallFunc(const iocshArgBuf *args)
         }
         *cp = '/';
     }
-    nfsMount(args[0].sval, args[1].sval, args[2].sval);
+    iocshSetError(nfsMount(args[0].sval, args[1].sval, args[2].sval));
 }
 #endif
 
-
-void zoneset(const char *zone)
```
mdavidsaver, line 491: ditto

mdavidsaver (changes requested):
Two quick changes and I think this is ready to go.

mdavidsaver:
Noting the CI failure in the RTEMS specific code due to a mis-placed `;`.

modules/libcom/RTEMS/posix/rtems_init.c
JJL772: Done

JJL772:
Build error is due to newlib <1.18.0 declaring `unsetenv` as returning void. They fixed this in 1.18.

Wisely, they also made the `_NEWLIB_VERSION` macro a string, so you can't easily just check `#if _NEWLIB_VERSION_MINOR < 18`...
Newlib version is tied to RTEMS version though, so I just checked `__RTEMS_MINOR__`. RTEMS 4.10 upgraded to newlib 1.18.0

mdavidsaver:
> they also made the `_NEWLIB_VERSION` macro a string

Did you try testing `__NEWLIB__` and `__NEWLIB_MINOR__`?

https://github.com/cygwin/cygwin/blob/1ed15161362b2eb5649b049b7ab0aaf8097bd43a/newlib/_newlib_version.hin#L6-L8

You might make use of the `VERSION_INT` macro from `epicsVersion.h`.

```c
#include <stdlib.h>
#define NEWLIB_VERSION(__NEWLIB__, __NEWLIB_MINOR__, __NEWLIB_PATCHLEVEL__, 0)
...
# if NEWLIB_VERSION < VERSION_INT(1,2,3,0)
...
```

JJL772:
`__NEWLIB_MINOR__` was added in 2.2.0

JJL772:
As far as I can tell, the only version macro in newlib <2.2.0 is `_NEWLIB_VERSION`

mdavidsaver:
Ok, then just `iocshSetError()` with newlib >= 2.2 ?  I don't think we need to be super precise for historical OS versions.

modules/database/src/ioc/misc/dlload.c
```diff
@@ -13,6 +13,7 @@
 IOCSH_STATIC_FUNC void dlload(const char* name)
 {
     if (!epicsLoadLibrary(name)) {
+        iocshSetError(-1);
```
mdavidsaver: I think this call should be moved down into `dlloadCallFunc()`.  Which will mean changing the return type of `dlload()`.

mdavidsaver (approved):
Ok.  I think this change is ready to go.

anjohnson:
Core Group: Approved.

PR #509, opened 2024-06-13
makeBpt: fix memory leaks

daneos:
Free leaked memory allocated by `cmalloc` for filenames in makeBpt.


mdavidsaver (approved):
```diff
+    free(outFilename);
+    free(pname);
```

`pname` is initialized to `NULL`.  `outFilename` is not, but is currently set unconditionally later on.  So this looks like a fine improvment to me.

anjohnson:
Core Group: approved, thanks!

PR #510, opened 2024-06-13
Add epicsLimits.h to provide a portable definition of PATH_MAX

JJL772:
Leaving as a draft to collect feedback.

The goal of this PR is to provide a portable definition of `PATH_MAX`, mostly because Windows only defines `MAX_PATH`.

Updated the `UpdatePWD` function to use `PATH_MAX` instead of an arbitrary length limit of 1024.

As of Windows 10 v1607, path limits may be increased in the registry. Those changes are not affected by the `MAX_PATH` macro. I'm not sure of the best way to handle this edge case- or if it's worth handling it at all.

This is related to the work on #497

### DIFF
--- a/modules/libcom/RTEMS/epicsMemFs.c
+++ b/modules/libcom/RTEMS/epicsMemFs.c
@@ -14,11 +14,8 @@
 #include <sys/types.h>
 #include <fcntl.h>
 
+#include "epicsLimits.h"
 #include "epicsMemFs.h"
-
-#ifndef PATH_MAX
-#  define PATH_MAX 100
-#endif
 
 int epicsMemFsLoad(const epicsMemFS *fs)
 {
--- a/modules/libcom/src/iocsh/libComRegister.c
+++ b/modules/libcom/src/iocsh/libComRegister.c
@@ -18,6 +18,7 @@
 #include "epicsTime.h"
 #include "epicsThread.h"
 #include "epicsMutex.h"
+#include "epicsLimits.h"
 #include "envDefs.h"
 #include "osiUnistd.h"
 #include "logClient.h"
@@ -33,7 +34,7 @@
  */
 static void updatePWD() {
     static int lasterror;
-    char buf[1024];
+    char buf[PATH_MAX];
     char *pwd = getcwd(buf, sizeof(buf));
     if (pwd) {
         pwd[sizeof(buf) - 1] = '\0';
--- a/modules/libcom/src/osi/Makefile
+++ b/modules/libcom/src/osi/Makefile
@@ -57,6 +57,7 @@ INC += epicsStdioRedirect.h
 INC += epicsTempFile.h
 INC += epicsGetopt.h
 INC += epicsStackTrace.h
+INC += epicsLimits.h
 
 INC += devLib.h
 INC += devLibVME.h
--- a/modules/libcom/src/osi/os/Linux/osdgetexec.c
+++ b/modules/libcom/src/osi/os/Linux/osdgetexec.c
@@ -10,10 +10,7 @@
 #include <limits.h>
 
 #include <osiFileName.h>
-
-#ifndef PATH_MAX
-#  define PATH_MAX 100
-#endif
+#include <epicsLimits.h>
 
 char *epicsGetExecName(void)
 {
--- a/modules/libcom/src/osi/os/WIN32/epicsLimits.h
+++ b/modules/libcom/src/osi/os/WIN32/epicsLimits.h
@@ -1,0 +1,24 @@
+/*
+ * Copyright: Stanford University / SLAC National Laboratory.
+ *
+ * SPDX-License-Identifier: EPICS
+ * EPICS BASE is distributed subject to a Software License Agreement found
+ * in file LICENSE that is included with this distribution.
+ *
+ * Author: Jeremy Lorelli <lorelli@slac.stanford.edu>, 2024
+ */
+#ifndef _EPICS_LIMITS_H
+#define _EPICS_LIMITS_H
+
+#include <limits.h>
+#include <stdlib.h>
+
+#ifndef PATH_MAX
+#   define PATH_MAX _MAX_PATH
+#endif
+
+#ifndef NAME_MAX
+#   define NAME_MAX _MAX_FNAME
+#endif
+
+#endif /* _EPICS_LIMITS_H */
--- a/modules/libcom/src/osi/os/default/epicsLimits.h
+++ b/modules/libcom/src/osi/os/default/epicsLimits.h
@@ -1,0 +1,19 @@
+/*
+ * Copyright: Stanford University / SLAC National Laboratory.
+ *
+ * SPDX-License-Identifier: EPICS
+ * EPICS BASE is distributed subject to a Software License Agreement found
+ * in file LICENSE that is included with this distribution.
+ *
+ * Author: Jeremy Lorelli <lorelli@slac.stanford.edu>, 2024
+ */
+#ifndef _EPICS_LIMITS_H
+#define _EPICS_LIMITS_H
+
+#include <limits.h>
+
+#ifndef PATH_MAX
+#define PATH_MAX 255
+#endif
+
+#endif /* _EPICS_LIMITS_H */


JJL772:
I have also replaced `VC_EXTRALEAN` with `WIN32_LEAN_AND_MEAN`, and added `NOMINMAX` in `epicsAtomicOSD.h`. `VC_EXTRALEAN` doesn't seem to exist in my Windows SDK, nor in the MinGW headers.

mdavidsaver:
When I first read the PR title, I thought about adding this definition to `osiFileName.h`.  However, seeing the include of `windows.h` and the associated LEAN_AND_MEAN-ness gives me pause.

I think this would be the first time any header installed by epics-base includes `windows.h`.  I am not certain what pitfalls this might have.

@FreddieAkeroyd Any thoughts?

FreddieAkeroyd:
Base has generally kept windows.h and winsock2.h includes to stay within the osi source routines themselves and not leak out into code beyond that. One potential issue is that `winsock2.h` must be included before `windows.h` or else you get a compile error, so exposing it outside the osi tree may make some previously working user programs stop compiling.  Also i think we have wanted to avoid any random macros or functions in `windows.h` clashing with other outside code now or in the future. 

It looks like windows exposes the  MAX_PATH value  from its `stdlib.h` as `_MAX_PATH` 
https://learn.microsoft.com/en-us/cpp/c-runtime-library/path-field-limits?view=msvc-170 so this could be an alternative approach to including windows.h

JJL772:
OK, good find. I very much prefer including stdlib.h here.

FreddieAkeroyd:
I believe `VC_EXTRALEAN` only affects MFC projects, on my machine it is defined in  `afxv_w32.h` from `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\atlmfc\include` so using `WIN32_LEAN_AND_MEAN` would be of more potential use to EPICS

JJL772:
Thanks for checking that. I believe @exzombie was having an issue with `epicsAtomic.h` in asyn. `VC_EXTRALEAN` was probably the culprit there.

FreddieAkeroyd:
Did his problem go away when `VC_EXTRALEAN` was just removed, or was changed to  `WIN32_LEAN_AND_MEAN` ? If the latter, was it a socket issue or something else?

I think we should take a separate look at `epicsAtomicOSD.h` and its use of `windows.h` - it looks like there is a potential pathway  `epicsAtomic.h` -> `epicsAtomicCD.h` -> `epicsAtomicOSD.h` that can lead us to including `windows.h` in user code that may cause issues

exzombie:
> Thanks for checking that. I believe @exzombie was having an issue with `epicsAtomic.h` in asyn. `VC_EXTRALEAN` was probably the culprit there.

Indeed. `epicsAtomic.h` includes `windows.h` and using it in `asynPortDriver.cpp` caused conflicts with the C-linkage `connect()` function. I'm not sufficiently familiar with Windows header files to be able to say whether any of these leanness macros would help.

FreddieAkeroyd:
`WIN32_LEAN_AND_MEAN` would help. This would be an example of the case I described above that `winsock2.h` should be included before `windows.h` because otherwise `windows.h` includes winsock version 1 and then this conflicts with a later `winsock2.h`. I've always just included `winsock2.h` first myself, but what i hadn't realise until today was that one of the APIs that `WIN32_LEAN_AND_MEAN` excluded to make itself "lean and mean" was the socket library. So you can include `winsock2.h` after `windows.h` provided you have defined `WIN32_LEAN_AND_MEAN` first as that stops `windows.h` from including the conflicting library definitions.

JJL772:
> I think we should take a separate look at epicsAtomicOSD.h and its use of windows.h - it looks like there is a potential pathway epicsAtomic.h -> epicsAtomicCD.h -> epicsAtomicOSD.h that can lead us to including windows.h in user code that may cause issues

Definitely. There are some additional flags that we can set, i.e. `NOGDI`, to exclude even more from windows.h that isn't covered by `WIN32_LEAN_AND_MEAN`. I'm not sure if the atomic functions can be included directly; a lot of these Windows headers depend on architecture `#define`s at the top of windows.h... The Windows headers are really poorly designed, imo

FreddieAkeroyd:
It might be we can make it lean enough with the `NO*` , but ideally the `windows.h` would be kept inside of the epics  operating system specific area, but that is a bit more messy for define constants rather than functions. We could create global variables that contain the values from any `#define` rather than exposing the original header, but that needs a bit more looking at I feel. But this is beyond this ticket, the `windows.h` was there already, but using `WIN32_LEAN_AND_MEAN` is most likely an improvement, as was seen in the asyn case. However it is possible this could break somebody's code outside of base if they relied on the now missing definitions, but they could get round this by including `windows.h` themselves before `epicsAtomic.h`.

JJL772:
This is from the top of Windows.h:
```
/*  If defined, the following flags inhibit definition
 *     of the indicated items.
 *
 *  NOGDICAPMASKS     - CC_*, LC_*, PC_*, CP_*, TC_*, RC_
 *  NOVIRTUALKEYCODES - VK_*
 *  NOWINMESSAGES     - WM_*, EM_*, LB_*, CB_*
 *  NOWINSTYLES       - WS_*, CS_*, ES_*, LBS_*, SBS_*, CBS_*
 *  NOSYSMETRICS      - SM_*
 *  NOMENUS           - MF_*
 *  NOICONS           - IDI_*
 *  NOKEYSTATES       - MK_*
 *  NOSYSCOMMANDS     - SC_*
 *  NORASTEROPS       - Binary and Tertiary raster ops
 *  NOSHOWWINDOW      - SW_*
 *  OEMRESOURCE       - OEM Resource values
 *  NOATOM            - Atom Manager routines
 *  NOCLIPBOARD       - Clipboard routines
 *  NOCOLOR           - Screen colors
 *  NOCTLMGR          - Control and Dialog routines
 *  NODRAWTEXT        - DrawText() and DT_*
 *  NOGDI             - All GDI defines and routines
 *  NOKERNEL          - All KERNEL defines and routines
 *  NOUSER            - All USER defines and routines
 *  NONLS             - All NLS defines and routines
 *  NOMB              - MB_* and MessageBox()
 *  NOMEMMGR          - GMEM_*, LMEM_*, GHND, LHND, associated routines
 *  NOMETAFILE        - typedef METAFILEPICT
 *  NOMINMAX          - Macros min(a,b) and max(a,b)
 *  NOMSG             - typedef MSG and associated routines
 *  NOOPENFILE        - OpenFile(), OemToAnsi, AnsiToOem, and OF_*
 *  NOSCROLL          - SB_* and scrolling routines
 *  NOSERVICE         - All Service Controller routines, SERVICE_ equates, etc.
 *  NOSOUND           - Sound driver routines
 *  NOTEXTMETRIC      - typedef TEXTMETRIC and associated routines
 *  NOWH              - SetWindowsHook and WH_*
 *  NOWINOFFSETS      - GWL_*, GCL_*, associated routines
 *  NOCOMM            - COMM driver routines
 *  NOKANJI           - Kanji support stuff.
 *  NOHELP            - Help engine interface.
 *  NOPROFILER        - Profiler interface.
 *  NODEFERWINDOWPOS  - DeferWindowPos routines
 *  NOMCX             - Modem Configuration Extensions
 */
```

Macro pollution may still be an issue though (i.e. `PostMessage` being defined to `PostMessageA`)

exzombie:
I just double-checked. The problem with `asynPortDriver.cpp` and `epicsAtomics.h` was threefold:

- epics-base 3.14 doesn't have `epicsAtomics.h`
- mingw builds have the problem of including `socket.h`, which VS builds do not seem to have.
- RTEMS also has the `socket.h` issue.

Given the first point, I had to remove use of `epicsAtomics.h`. For the second point, I don't know how the headers differ from VS ones; do the leanness macros have any effect at all? And for the third point, I guess it may be best to open a separate issue.

FreddieAkeroyd:
which bit of code has the erroring `socket.h` include? It can't use `epicsSocket.h` instead?

exzombie:
On the contrary, you don't want sockets there. `asynPortDriver.cpp` defines it's own `connect()` function, that has nothing to do with sockets. It has C linkage and conflicts with the socket-related function. Sure, it could be renamed. But I think it's not nice that `epicsAtomics.h` pulls in socket functions.

FreddieAkeroyd:
@exzombie had a look at your appveyor failed build, the mingw windows.h and winsock.h may not behave exactly the same as on standard windows so needs more investigation. Potentially including winsock2.h before windows.h in the  epicsAtomicOSD.h  might resolve the later conflict, but i'm not keen on that as a long term solution. Removing leaking of winows.h definitions outside of epicsAtomicOSD.h into epicsAtomic.h would resolve the issue

FreddieAkeroyd:
sorry, didn't see your message before posting. defining `WIN32_LEAN_AND_MEAN` before including epicsAtomic.h woudl stop it including socket functions, you can test that on your own asyn branch

mdavidsaver:
Seeing the growing complexity, would we be better off instead encapsulating the `getcwd(NULL,0)` behavior?

```c
/** Return the current working directory
 *
 * On success, return an allocated buffer which must later be passed to @code free() @endcode .
 * On failure, return NULL and sets errno.
 */
char* epicsGetCWD(void);
```

The changes to `WIN32/epicsAtomicOSD.h` seem worthy of a separate PR.

JJL772:
> Seeing the growing complexity, would we be better off instead encapsulating the getcwd(NULL,0) behavior?

I don't think this is becoming too complex, we can rely on `_MAX_PATH` from `stdlib.h` on Windows, instead of `windows.h` (thankfully...)

As for `epicsGetCWD`, allocating during every call doesn't seem optimal. I do agree that it should be added, though.

> The changes to WIN32/epicsAtomicOSD.h seem worthy of a separate PR.

Sounds good, I'll split it out.

anjohnson:
This looks good to me now and has passed all CI builds, including Appveyor.

mdavidsaver:
> path limits may be increased in the registry. ... I'm not sure of the best way to handle this edge case

Historically, I could see a non-constant `PATH_MAX` as an edge case.  Today, I see it as the norm.

I (and [others](https://eklitzke.org/path-max-is-tricky)) see `PATH_MAX` as an artifact of API designs which omitted a buffer length.  cf. `getwd()` or `realpath()`.  I don't like the idea of encouraging new, "portable", code to be written around the idea that there is a maximum path length.

So I lean further towards the idea of a wrapper function which returns an allocated string.

Also, for fun.  On Debian 12, I am able to create a directory tree with a total name length of 4361 bytes.

mdavidsaver:
Something like this [untested code](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAZgCspBwAyeAyYAHLuAEaYxCAATBqkAA6oCoS2DM5uHt5%2Byak2AkEh4SxRMfEWmFYFDEIETMQEme6evpXV6XUNBEVhkdFxCQr1jc3ZbcPdvSVlgwCUFqiuxMjsHACksV7ByG5YANTrXo7D6LR4EQB0CEfY6xoAgpvbVFhU%2B8YA6gCSoV6x9yeW32%2Bx2e0wh2O%2BGImGs11ugOeVSUiOBoIYu1cByOjlcDDwp3hXjujyRDHwVFR2wx4MhjmixAYqCJJKepK2eCo5Mw72UDwAKgAJYwAWQeAA0qSC3sEIXyhaKJfsuLEABxUhycxGPZAIBoAKn2mESeGQCgA4pgmp8ACIQABuqDw6DmiIA7AAhKmc/YykLoCCmc0Bb4exymOaHN2OdbR3082UBr6/f6ux4gkEwgjLBj7YBW5AAdwDoTkAQCpH2GldXi9pPZXmRmEBGf2qQAXphjAR9hFXFR6Lmjjb9vLhWLJbWWxndQ19vqs5CR6Xy5X9X33sP9iuAkc62yHq3Cwg6JgIFxI7H963W7PiPOQqoe1uYWJaKhkBAN5WN4Oa9ebxBTkIDAMBHwIS9PWnQDWwZJkl32bBQgAeRFbART3aCYJBCJXwAa0w9MYNjG0sNbDcEPAwjDxg4D8wIIsA2/Xt%2Bz/SCAOwzMrQQjdqM4jNcMwJgCKnesaMAkijVoJRQSoCA4NQUDh2wAAlB5QnNbB2LImDBOEvjiLdUiiMA39GHnLcAVE8SbwAekNLNiAAT3nWysJI7UbOA0Cs20kybyoGEzw3f93KMrCsxzfYs2oskKXdYyHg4BZaE4HxeE8bheFQThHDbJYVghZ4eFIAhNGShY8JAHwElSjhJAy8rSByjheAUEAEjKjgtAWOBYCQNAWBNegyAoCBBuGgZkGAKQuD4OgCGidqvyaiJggaJzOBKtbmGc5CIm0WEupKwa2EEZCGFoTbut4LA%2B2ARw33arLSCwFhDGAcQbtevAYWsPB7UwZ6tH8VRYVcRatt4YJFrqkHzlwjbnCwJqCGIPAWCh0hAeICIUkwG1MHeoxziMcqFgHJhgAUAA1PBMELZDEkYLH%2BEEEQxHYKQZEERQVHUb7dDmgwydMYxzAR9rIAWVBEhqZ6AFpPjffZspx9GsCliAFksI70nsckxk8ObAllGYBjmvI0gEI29Ctmppn6GI5t1/6BC6UYXBaPRXZqD2ejNp2fZGJoveyF2Q8d0oLZ1grVgkFK0sa76Wv2VRVQANgVjPJDzZBkGVSRLi4fYIFwQgSEOLYL14Lqesq6ras4BrSEykGWrajrSvJxOOFiZP284Wue%2Bxpb9ckIA).

JJL772:
> Also, for fun. On Debian 12, I am able to create a directory tree with a total name length of 4361 bytes.

Oh wow! I didn't know that was possible-- I thought it was limited at 4096. In that case, yeah, the `epicsGetCWD` you proposed is the right way to go.

ericonr:
> Also, for fun. On Debian 12, I am able to create a directory tree with a total name length of 4361 bytes.

At least on Linux, `PATH_MAX` is mostly used to limit the size of the buffers that system calls can read from or write to. That includes things like `chdir`, `open`, `getcwd`.

If you use a combination of `chdir` and `open`, or `openat`, I don't know if there's an actual defined limit; but most utilities will break if you try to address these files and directories with their full paths, even if they don't hardcode `PATH_MAX` limits in their internal buffers.

At least on glibc, `getcwd` can work for paths longer than `PATH_MAX`, but it's due to their fallback implementation:

```
   C library/kernel differences
       On Linux, the kernel provides a getcwd() system call, which the functions described
       in this page will use if possible.  The system call takes the same arguments as the
       library  function  of  the  same name, but is limited to returning at most PATH_MAX
       bytes.  (Before Linux 3.12, the limit on the size of the returned pathname was  the
       system  page  size.   On  many architectures, PATH_MAX and the system page size are
       both 4096 bytes, but a few architectures have a larger page size.)  If  the  length
       of  the pathname of the current working directory exceeds this limit, then the sys‐
       tem call fails with the error ENAMETOOLONG.  In this case,  the  library  functions
       fall back to a (slower) alternative implementation that returns the full pathname.
```

mdavidsaver (changes requested):
To reiterate.  I think that current usage of `PATH_MAX` should be considered deprecated, and replaced.  So I do not want to begin exposing `PATH_MAX` as an OSI API.

Instead, my preference for an OSI API would be a wrapper which returns an allocated string.

Also, instead of a new header `epicsLimits.h`, maybe just add to the existing `osiFileName.h`?

anjohnson:
Core Group: Alternate approach agreed with @JJL772 .

PR #511, opened 2024-06-13
Add SPDX license tags to templates

exzombie:
The goal here is twofold:

- Provide clarification on the licensing of files that are "given over" to the user by `makeBaseApp.pl`. Until now, it was a bit unclear; one could only assume that the files are available under the EPICS license, but couldn't be sure.
- Allow using a linter (like the one provided by the [REUSE initiative](https://reuse.software/)) by tagging *every* file. In practice, dealing with exceptions and such is quite annoying, and tagging everything is the only reasonable way to deal with licensing.

Copyright information was gathered using `git log` with per-line history tracking, which allows following the history of a file through renames. There were a couple of cases (e.g.  caPerlApp) where this didn't work, so I traced the history manually.

Tagging every file means that plain-text READMEs are also tagged, which is a bit annoying. I put the tags at the bottom of the files so that they're not the first thing you read.

### DIFF
--- a/modules/ca/src/template/top/caClientApp/caExample.c
+++ b/modules/ca/src/template/top/caClientApp/caExample.c
@@ -1,6 +1,6 @@
-// SPDX-FileCopyrightText: 1998 Argonne National Laboratory
-//
-// SPDX-License-Identifier: EPICS
+/* SPDX-FileCopyrightText: 1998 Argonne National Laboratory */
+
+/* SPDX-License-Identifier: EPICS */
 
 /*caExample.c*/
 #include <stddef.h>
--- a/modules/ca/src/template/top/caClientApp/caExample.c
+++ b/modules/ca/src/template/top/caClientApp/caExample.c
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 1998 Argonne National Laboratory
+//
+// SPDX-License-Identifier: EPICS
+
 /*caExample.c*/
 #include <stddef.h>
 #include <stdlib.h>
--- a/modules/database/src/template/top/exampleApp/Db/circle.db
+++ b/modules/database/src/template/top/exampleApp/Db/circle.db
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2017 ITER Organization
+#
+# SPDX-License-Identifier: EPICS
+
 record(ao, "$(user):circle:step") {
   field(VAL , "1.0")
   field(DRVL, "0.0")
--- a/modules/database/src/template/top/exampleApp/src/sncExample.stt
+++ b/modules/database/src/template/top/exampleApp/src/sncExample.stt
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 1998 Argonne National Laboratory
+//
+// SPDX-License-Identifier: EPICS
+
 program sncExample
 double v;
 assign v to "{user}:aiExample";
--- a/modules/database/src/template/top/exampleApp/src/sncExample.stt
+++ b/modules/database/src/template/top/exampleApp/src/sncExample.stt
@@ -1,7 +1,3 @@
-// SPDX-FileCopyrightText: 1998 Argonne National Laboratory
-//
-// SPDX-License-Identifier: EPICS
-
 program sncExample
 double v;
 assign v to "{user}:aiExample";
@@ -24,3 +20,9 @@ ss ss1 {
         } state low
     }
 }
+
+/*
+SPDX-FileCopyrightText: 1998 Argonne National Laboratory
+
+SPDX-License-Identifier: EPICS
+*/
--- a/modules/database/src/template/top/exampleBoot/ioc/README@Common
+++ b/modules/database/src/template/top/exampleBoot/ioc/README@Common
@@ -10,6 +10,4 @@ st.cmd's dbLoadDatabase() command before starting the ioc.
 
 
 SPDX-FileCopyrightText: 2001 Argonne National Laboratory
-
 SPDX-License-Identifier: EPICS
-
--- a/modules/database/src/template/top/exampleBoot/ioc/README@Common
+++ b/modules/database/src/template/top/exampleBoot/ioc/README@Common
@@ -7,3 +7,9 @@ and check the executable name on the first line of the st.cmd file
 
 You may need to change the name of the .dbd file given in the
 st.cmd's dbLoadDatabase() command before starting the ioc.
+
+
+SPDX-FileCopyrightText: 2001 Argonne National Laboratory
+
+SPDX-License-Identifier: EPICS
+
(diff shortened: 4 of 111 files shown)


modules/ca/src/template/top/caClientApp/caExample.c
```diff
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 1998 Argonne National Laboratory
+//
+// SPDX-License-Identifier: EPICS
+
```
anjohnson: Please don't use `//` comments in .c files. These changes won't fail our CI tests because we don't attempt to instantiate or build the templates, but one or two of our oldest compilers don't accept C++-style comments.

modules/database/src/template/top/exampleApp/src/sncExample.stt
```diff
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 1998 Argonne National Laboratory
+//
+// SPDX-License-Identifier: EPICS
+
```
anjohnson: This might cause a problem with older versions of the sequencer, although hopefully it's been fixed in newer ones. The old `snc` that I remember fighting with wouldn't accept _any_ text including comments before the `program` statement. If that was fixed before the first seq-2.1 release then I think this is acceptable, but please check.

modules/database/src/template/top/exampleBoot/ioc/README@Common
```diff
@@ -7,3 +7,9 @@ and check the executable name on the first line of the st.cmd file
 
 You may need to change the name of the .dbd file given in the
 st.cmd's dbLoadDatabase() command before starting the ioc.
+
+
+SPDX-FileCopyrightText: 2001 Argonne National Laboratory
+
+SPDX-License-Identifier: EPICS
+
```
anjohnson: If you can cuddle these up they'd be less prominent, meaning no blank lines between or after them?

modules/database/src/template/top/exampleBoot/ioc/README@vxWorks
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2001 Argonne National Laboratory
+
+SPDX-License-Identifier: EPICS
```
anjohnson: I suspect this file has to be completely empty, which tells `makeBaseApp.pl` to not instantiate it at all. It has to be there though to prevent the `README@Common` file from being installed, but being empty there can be no copyright or license to it. If your tool complains that's a bug in the tool IMHO…

src/template/ext/top/configure/RULES_JAVA
```diff
@@ -1,13 +1,8 @@
...
 # Copyright (c) 2002 The Regents of the University of California, as
-#     Operator of Los Alamos National Laboratory.
+# Copyright (c) 2002 The University of Chicago, as Operator of Argonne
+# SPDX-FileCopyrightText: 2000 Argonne National Laboratory
+#
 # SPDX-License-Identifier: EPICS
-# EPICS Base is distributed subject to a Software License Agreement found
-# in file LICENSE that is included with this distribution. 
-#*************************************************************************
-
```
anjohnson: I should probably delete this file, but until I do I think your tool mangled the existing copyright header a bit, please fix.

anjohnson (changes requested):
Some adjustments please; I didn't mark every file that needs changing, just one of each type.

modules/ca/src/template/top/caClientApp/caExample.c
exzombie: Done

modules/database/src/template/top/exampleBoot/ioc/README@Common
exzombie: Done

modules/database/src/template/top/exampleBoot/ioc/README@vxWorks
exzombie: If it's not instantiated, then it's not a problem. What's important to me is to have a clear copyright status of files that *are* instantiated.

src/template/ext/top/configure/RULES_JAVA
exzombie: Uh, I totally missed this. The tool I used tries to preserve the original copyright notices, but apparently can't handle multiline notices. Are you ok with just replacing all the notices with SPDX tags? They may not have been correct in the first place. I will try to trace the history of the file again, but it seems to be older than the notices state, I had previously traced its origins to the year 2000.

exzombie: Ah, I found the reason. Janet added a bunch of copyright notices in 2002. The file was originally added by Marty with commit 65d9c939bc0bb5e03cf040e71927ae83e44061c6. So, 2000 appears to be the correct date.

modules/database/src/template/top/exampleApp/Db/circle.db
```diff
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2017 ITER Organization
+#
+# SPDX-License-Identifier: EPICS
+
```
anjohnson, line 4: @ralphlange Would this be the correct copyright attribution for this file? You added it to Git in 1c03e615f16ab9a1279456846b414858a4e94274 but I suspect the original might have had a different source.

modules/database/src/template/top/exampleBoot/ioc/README@RTEMS
```diff
@@ -4,3 +4,7 @@ contents to
 
 Then load the executable into the IOC (floppy disk, network boot, debugger,
 etc.) and start it.
+
+
+SPDX-FileCopyrightText: 2001 Lawrence Berkeley National Laboratory
```
anjohnson: In 2001 Eric Norum was working for APS at Argonne, but we converted the EPICS CVS repository to Bazaar and then to Git after he had moved to LBL. The repo conversion put his newer email address on all his commits — CVS only records the local username of the committer, so the repo conversion tools have to be told how to convert them.

anjohnson (changes requested):
Hi Jure,

Please adjust the 2001 copyright from LBL => ANL as noted below, I only saw that in one file.
@ralphlange should be able to resolve the Iter question.

Thanks!

modules/database/src/template/top/exampleApp/Db/circle.db
ralphlange: Yes, that's fine.

modules/database/src/template/top/exampleBoot/ioc/README@RTEMS
exzombie: Done. Thank you so much for reviewing these things!

anjohnson:
Core Group: Approved, thanks!

anjohnson (approved):

Issue #513, opened 2024-06-14
Investigate uses of Windows.h


JJL772:
`epicsAtomicOSD.h` for Windows includes `windows.h` for the `InterlockedXXX` declarations. This leads to some pollution due to the pile of other headers `windows.h` includes. User code including `epicsAtomic.h` will eventually include `epicsAtomicOSD.h` and thus `windows.h` on Windows.

#510 improves on this behavior somewhat by replacing `VC_EXTRALEAN` with `WIN32_LEAN_AND_MEAN` and adding `NOMINMAX`. `VC_EXTRALEAN` is only relevant to MFC sources.

More cleanup is required though. There are some options for tweaking what is included or declared, as seen at the top of windows.h:
```
/*  If defined, the following flags inhibit definition
 *     of the indicated items.
 *
 *  NOGDICAPMASKS     - CC_*, LC_*, PC_*, CP_*, TC_*, RC_
 *  NOVIRTUALKEYCODES - VK_*
 *  NOWINMESSAGES     - WM_*, EM_*, LB_*, CB_*
 *  NOWINSTYLES       - WS_*, CS_*, ES_*, LBS_*, SBS_*, CBS_*
 *  NOSYSMETRICS      - SM_*
 *  NOMENUS           - MF_*
 *  NOICONS           - IDI_*
 *  NOKEYSTATES       - MK_*
 *  NOSYSCOMMANDS     - SC_*
 *  NORASTEROPS       - Binary and Tertiary raster ops
 *  NOSHOWWINDOW      - SW_*
 *  OEMRESOURCE       - OEM Resource values
 *  NOATOM            - Atom Manager routines
 *  NOCLIPBOARD       - Clipboard routines
 *  NOCOLOR           - Screen colors
 *  NOCTLMGR          - Control and Dialog routines
 *  NODRAWTEXT        - DrawText() and DT_*
 *  NOGDI             - All GDI defines and routines
 *  NOKERNEL          - All KERNEL defines and routines
 *  NOUSER            - All USER defines and routines
 *  NONLS             - All NLS defines and routines
 *  NOMB              - MB_* and MessageBox()
 *  NOMEMMGR          - GMEM_*, LMEM_*, GHND, LHND, associated routines
 *  NOMETAFILE        - typedef METAFILEPICT
 *  NOMINMAX          - Macros min(a,b) and max(a,b)
 *  NOMSG             - typedef MSG and associated routines
[...]

PR #518, opened 2024-06-15
Fix conflicts from `windows.h` in `epicsAtomicOSD.h`

JJL772:
In short:
* `VC_EXTRALEAN` is only relevant to the ATL/MFC headers
* `WIN32_LEAN_AND_MEAN` excludes a bunch of random declarations that come from `windows.h` (i.e. winsock1)
* `NOMINMAX` disables the min/max macros defined by `windows.h`. These tend to cause funny issues in C++ STL headers such as `<algorithm>`

Part of the work for #513 split from #510

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -18,6 +18,14 @@ __This version of EPICS has not been released yet.__
 
 __Add new items below here__
 
+### Reduce symbol and macro pollution from epicsAtomic.h on WIN32
+
+`epicsAtomic.h` no longer pulls in as many unneeded declarations and macros from
+`windows.h`. Prior to this change, including `epicsAtomic.h` at the wrong time
+could result in unexpected compiler errors. Due to the nature of `windows.h`,
+some unneeded declarations are still pulled in, however the number is greatly reduced.
+Code that needs these declarations should explicitly include `windows.h` before `epicsAtomic.h`.
+
 ### epicsExport simplifications
 
 `epicsExportAddress()`, `epicsExportRegistrar()` and `epicsRegisterFunction()`
--- a/modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
+++ b/modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
@@ -19,27 +19,50 @@
 
 #define EPICS_ATOMIC_OS_NAME "WIN32"
 
-#ifdef VC_EXTRALEAN
-#   define VC_EXTRALEAN_DETECTED_epicsAtomicOSD_h
-#else
-#   define VC_EXTRALEAN
-#endif
+/* Disable extra declarations that we don't need here (i.e. winsock1, rpc, etc.) */
+#pragma push_macro("WIN32_LEAN_AND_MEAN")
+#undef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
 
-#ifdef STRICT
-#   define STRICT_DETECTED_epicsAtomicOSD_h
-#else
-#   define STRICT
-#endif
+#pragma push_macro("STRICT")
+#undef STRICT
+#define STRICT
+
+/* Disable min/max macros from windows.h. These macros can cause issues with headers such as <algorithm> that declare or use std::min/max */
+#pragma push_macro("NOMINMAX")
+#undef NOMINMAX
+#define NOMINMAX
+
+/* Disable 'service controller' includes */
+#pragma push_macro("NOSERVICE")
+#undef NOSERVICE
+#define NOSERVICE
+
+/* Disable 'input management engine' includes */
+#pragma push_macro("NOIME")
+#undef NOIME
+#define NOIME
+
+/* Disable 'modem configuration extensions' includes */
+#pragma push_macro("NOMCX")
+#undef NOMCX
+#define NOMCX
+
+/* Disable GDI includes */
+#pragma push_macro("NOGDI")
+#undef NOGDI
+#define NOGDI
 
 #include "windows.h"
 
-#ifndef VC_EXTRALEAN_DETECTED_epicsAtomicOSD_h
-#   undef VC_EXTRALEAN
-#endif
-
-#ifndef STRICT_DETECTED_epicsAtomicOSD_h
-#   undef STRICT
-#endif
+/* Restore previous macro values */
+#pragma pop_macro("WIN32_LEAN_AND_MEAN")
+#pragma pop_macro("STRICT")
+#pragma pop_macro("NOMINMAX")
+#pragma pop_macro("NOSERVICE")
+#pragma pop_macro("NOIME")
+#pragma pop_macro("NOMCX")
+#pragma pop_macro("NOGDI")
 
 #if defined ( _WIN64 )
 #    define MS_ATOMIC_64
--- a/modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
+++ b/modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
@@ -53,6 +53,21 @@
 #undef NOGDI
 #define NOGDI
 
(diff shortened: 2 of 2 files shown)


modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
```diff
@@ -19,10 +19,10 @@
...
 #define EPICS_ATOMIC_OS_NAME "WIN32"
 
-#ifdef VC_EXTRALEAN
-#   define VC_EXTRALEAN_DETECTED_epicsAtomicOSD_h
+#ifdef WIN32_LEAN_AND_MEAN
+#   define WIN32_LEAN_AND_MEAN_DETECTED_epicsAtomicOSD_h
 #else
-#   define VC_EXTRALEAN
+#   define WIN32_LEAN_AND_MEAN
 #endif
```
anjohnson: It would be nice to include a brief description of what all this macro magic achieves as a comment towards the top of the file, for future maintainers to read without having to research for themselves.

I'm wondering whether all the `_DETECTED_epicsAtomicOSD_h` should be undefined at the end or not; this header does have a `#include` guard, so later includes of the same file won't ever see them and nothing else in Base is using them AFAICT. Not a big point, just a minor clean-up if you're making the above change.

JJL772: I've added a couple comments about `WIN32_LEAN_AND_MEAN` and `NOMINMAX`

Just want to note that [push_macro](https://gcc.gnu.org/onlinedocs/gcc/Push_002fPop-Macro-Pragmas.html) may be a less-verbose solution to saving macro definitions in the future. Minimum GCC version for this is 4.2.0, Clang has supported it since 2010 and MSVC added it before GCC. But I think the vxWorks compilers are still too old for this feature.

mdavidsaver: > ... vxWorks compilers are still too old for this feature.

Which is relevant in a `os/WIN32/` header how?  ;)

@FreddieAkeroyd Any objections to usage of `push_macro()` ?

JJL772: > Which is relevant in a os/WIN32/ header how? ;)

Eventually I'll learn to have my morning coffee *before* making comments on my pull requests :laughing:

FreddieAkeroyd (approved):
looks good to me

modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
FreddieAkeroyd: `push_macro` seems a good approach to improve readability

JJL772: Ok, I've switched this to use `push_macro`/`pop_macro`. I also added some additional `NOXXX` defines to strip out as much as possible. Seems like `WIN32_LEAN_AND_MEAN` covers most.

mdavidsaver:
@MarkRivers fyi. if you want to weight in.

modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
```diff
@@ -19,27 +19,50 @@
...
+#include "windows.h"
+
+/* Restore previous macro values */
+#pragma push_macro("WIN32_LEAN_AND_MEAN")
+#pragma push_macro("STRICT")
+#pragma push_macro("NOMINMAX")
+#pragma push_macro("NOSERVICE")
+#pragma push_macro("NOIME")
+#pragma push_macro("NOMCX")
+#pragma push_macro("NOGDI")
```
FreddieAkeroyd: I think you meant `pop_macro` here

JJL772: That's embarrassing. Fixed.

mdavidsaver:
Given the tricky-ness of `windows.h`, I think some kind of release notes entry is needed.

documentation/RELEASE_NOTES.md
```diff
@@ -28,6 +28,13 @@ In Base 7.0.8, an update to the compress record was added to allow for certain
 algorithms to use partially filled buffers in their computations. Unfortunately,
 this broke the behaviour of the records in certain cases. This has been fixed.
 
+### Reduce symbol and macro pollution from epicsAtomic.h on WIN32
+
+`epicsAtomic.h` no longer pulls in as many unneeded declarations and macros from
+`windows.h`. Prior to this change, including `epicsAtomic.h` at the wrong time
+could result in unexpected compiler errors. Due to the nature of `windows.h`,
+some unneeded declarations are still pulled in, however the number is greatly reduced.
+
```
mdavidsaver: ... Code which needs these declarations and macros should explicitly include `windows.h` before `epicsAtomic.h`.

mdavidsaver (commented):
> It would be nice to include a brief description of what all this macro magic achieves as a comment towards the top of the > file, for future maintainers to read without having to research for themselves.

@anjohnson This may be difficult since the only description I can (easily) find of the `NO*` macros [says](https://learn.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers)

```
* Define one or more of the NOapi symbols to exclude the API. For example, NOCOMM
  excludes the serial communication API. For a list of support NOapi symbols, see Windows.h.

#define NOCOMM
```

It isn't clear to me if `see Windows.h` means read the source, or some other document.  I can't find a list of all (current?) `NO*` macros.

JJL772:
> It isn't clear to me if see Windows.h means read the source, or some other document. I can't find a list of all (current?) NO* macros.

It means read the source. The documentation on this stuff is quite lacking. I pasted the comment from the top of Windows.h in #513

documentation/RELEASE_NOTES.md
JJL772: Done.

JJL772:
Is there anything else that needs to be done for this PR?

mdavidsaver:
@FreddieAkeroyd @anjohnson Does this change look OK in its present state?

anjohnson:
Core Group: Building some other modules that include `windows.h` is probably worth doing before merge: Asyn, AreaDetector and iocStats.

modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
```diff
@@ -19,27 +19,50 @@
...
+#define WIN32_LEAN_AND_MEAN
 
-#ifdef STRICT
-#   define STRICT_DETECTED_epicsAtomicOSD_h
-#else
-#   define STRICT
-#endif
+#pragma push_macro("STRICT")
+#undef STRICT
+#define STRICT
```
FreddieAkeroyd, line 29: To be sure of `STRICT` I think we also need to push and undef `NO_STRICT`  https://learn.microsoft.com/en-us/windows/win32/winprog/disabling-strict

FreddieAkeroyd:
@JJL772  does  it still work if you define  `NOCRYPT` as well?

modules/libcom/src/osi/os/WIN32/epicsAtomicOSD.h
FreddieAkeroyd: Actually, on reflection if somebody has explicitly defined `NO_STRICT` then we should leave as is for include - so ignore this suggestion

JJL772:
@FreddieAkeroyd Yeah base, asyn, modbus, etc. still build with NOCRYPT and the other definitions I added.

FYI, I'm testing with mingw here.

mdavidsaver:
@anjohnson @ralphlange Immediately post 7.0.9 seems like the best time to merge this PR.

anjohnson:
> Immediately post 7.0.9

Ack, added this to that milestone. I'm happy to merge any of those as soon as they're ready and approved.

PR #519, opened 2024-06-18
caget: fix memory leaks

daneos:
Free memory leaked by PVs array.

### DIFF
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -343,6 +343,7 @@ static int caget (pv *pvs, int nPvs, RequestT request, OutputT format,
         default :
             break;
         }
+        free(pvs[n].value);
     }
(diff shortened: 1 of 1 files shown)


modules/ca/src/tools/caget.c
```diff
@@ -388,7 +388,7 @@ int main (int argc, char *argv[])
     int digits = 0;             /* getopt() no. of float digits */
 
     int nPvs;                   /* Number of PVs */
-    pv* pvs;                    /* Array of PV structures */
+    pv* pvs = NULL;             /* Array of PV structures */
```
ralphlange, line 392: pvs is used with `calloc()`, so this is not strictly necessary, but it's better coding style, agreed.

anjohnson:
Core Group: Thanks, approved!

PR #520, opened 2024-06-18
readline: only keep history for interactive session

mdavidsaver:
Only `add_history()`, and persist history, for interactive readline session.  Otherwise entries in iocsh scripts end up in `.iocsh_history` repeatedly.

This change is to some extent a matter of opinion, and mine is changing.  I initially through that having the start script history would be useful, but so far have only found it gets in the way of my most common task, re-running the same commands before and after an IOC restart.


anjohnson:
+1 — agreed.

ralphlange:
+1

In real life, the only thing that I am not annoyed by is having the top level `st.cmd` in the immediate history - but that is always recent and not coming from the history file.

PR #523, opened 2024-06-22
Show buffered line along with DB parse error

mdavidsaver:
Yet again I find myself looking at a parse error in a generated .db file with the usual sigh.  Wondering what I mistyped, and where.  Then it occurs to me that the .db file parse input is buffered by line.  So why not show the whole line with the syntax error?

```
ERROR: field(bi, "LOCALHOST:NET:lo:CR") {
ERROR: syntax error
 at or before 'field' in path "/home/dev/epics/linStat/iocBoot/iocdemo/../../db"  file "linStatNIC.db" line 192
ERROR failed to load 'linStatNIC.db'
```

Now I see both what and where I typo'd!

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbYacc.y
+++ b/modules/database/src/ioc/dbStatic/dbYacc.y
@@ -378,6 +378,7 @@ static int yyerror(char *str)
         dbIncludePrint();
         yyFailed = TRUE;
     }
+    fprintf(stderr, "\n %d | %s\n", pinputFileNow->line_num, my_buffer);
     return(0);
 }
 static long pvt_yy_parse(void)


anjohnson:
I approve of the intention but can you please distinguish better between the full line that contains the error and the part which the parser didn’t actually understand. The extra line doesn’t really clarify the problem for me in your example (I assume the actual error there was that an earlier line missed something like a ‘}’ character).

ralphlange:
`field` instead of `record`.

mdavidsaver:
> ... distinguish better between the full line that contains the error and the part which the parser didn’t actually understand ...

Do any of the following strike your fancy?

```
>>> field(bi, "LOCALHOST:NET:lo:CR") {
ERROR: syntax error
 at or before 'field' in path "/home/dev/epics/linStat/iocBoot/iocdemo/../../db"  file "linStatNIC.db" line 192
ERROR failed to load 'linStatNIC.db'
```

```
====
field(bi, "LOCALHOST:NET:lo:CR") {
====
ERROR: syntax error
 at or before 'field' in path "/home/dev/epics/linStat/iocBoot/iocdemo/../../db"  file "linStatNIC.db" line 192
ERROR failed to load 'linStatNIC.db'
```

```
 192 | field(bi, "LOCALHOST:NET:lo:CR") {
ERROR: syntax error
 at or before 'field' in path "/home/dev/epics/linStat/iocBoot/iocdemo/../../db"  file "linStatNIC.db" line 192
ERROR failed to load 'linStatNIC.db'
```

anjohnson:
The words `ERROR: Syntax error` need to come first; that's like a title and sets the context for what comes next. My brain will usually ignore anything that appears _above_ that line when I'm examining an error.

The `at or before 'field'` is really part of the error, so lets have it on the same line. The quoted str isn't likely to be very long, but if it is it can just wrap. Then a newline:

`ERROR: Syntax error at or before 'field'`

Going from the general to the specific, the file path & name should come next. They're currently output by `dbIncludePrint()` which is a static function in dbLexRoutines.c but this is the only call to it so changing it won't affect anything else. It would be nice to combine the path and filename into a single file-path instead of separating them as it currently does, but that would be a bit more work. We're going to see the filename by itself on the `ERROR failed to load '…'` line which comes from `dbLoadRecords()` in dbAccess.c, so combining them here would make it easy to copy the whole file-path and paste it into an editor command-line or File::Open dialog.

` in file "/home/dev/epics/linStat/iocBoot/iocdemo/../../db/linStatNIC.db"`

I like the line number immediately before the text, which I think is fairly common usage although it might not be universally understood:

` 192: field(bi, "LOCALHOST:NET:lo:CR") {`

Either of these look good to me:
```
ERROR: Syntax error at or before 'field'
 in file "/home/dev/epics/linStat/iocBoot/iocdemo/../../db/linStatNIC.db"
 192: field(bi, "LOCALHOST:NET:lo:CR") {
ERROR failed to load 'linStatNIC.db'
```

```
ERROR: Syntax error at or before 'field' at line 192 of file
 /home/dev/epics/linStat/iocBoot/iocdemo/../../db/linStatNIC.db

 field(bi, "LOCALHOST:NET:lo:CR") {

ERROR failed to load 'linStatNIC.db'
```

(you did ask...)

mdavidsaver:
> (you did ask...)

So you can get it out of your system ;)

For this PR I am thinking about an ~small incremental improvement.  That is, not changing the lines which are currently printed.  Just adding more.

> I like the line number immediately before the text

I was trying to follow the pattern of error messages from GCC, which I see as a something of a de-facto standard.  Qt Creator at least tries to parse them.  Making Base parse error messages more uniform might be a good codeathon project (or a summer student?).

ralphlange:
> I was trying to follow the pattern of error messages from GCC, which I see as a something of a de-facto standard. Qt Creator at least tries to parse them. Making Base parse error messages more uniform might be a good codeathon project (or a summer student?).

Running the IOC in my IDE (which I do when developing), and having the editor jump to the right place when the IOC prints errors would be _awesome_.

mdavidsaver:
Updated to look like:

typo

anjohnson:
Core Group: Approved.

PR #527, opened 2024-07-17
freebsd: Add support for x86 and amd64 builds

kiwichris:
Use the OS compiler, ie cc and c++.

The version overrides the GNU CC support.

### DIFF
--- a/configure/os/CONFIG.freebsd-x86_64.freebsd-x86_64
+++ b/configure/os/CONFIG.freebsd-x86_64.freebsd-x86_64
@@ -2,13 +2,23 @@
 # Definitions for freebsd-x86_64 host - freebsd-x86_64 target builds
 # Sites may override these definitions in CONFIG_SITE.freebsd-x86_64.freebsd-x86_64
 #-------------------------------------------------------
-GNU_DIR=/usr/local
+GNU_DIR=/usr
 
 # Include common gnu compiler definitions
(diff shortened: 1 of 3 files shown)


anjohnson:
@tboegi you have submitted PRs for FreeBSD to epics-base in the past, do you have any comments on this PR?

As a more general matter, given the following quote from the [FreeBSD 14.0 Release Notes](https://www.freebsd.org/releases/14.0R/relnotes/#future-releases) does anyone have any objections if removed the freebsd-x86 configure/ files in the EPICS 7.0.9 release?
> FreeBSD 15.0 is not expected to include support for 32-bit platforms other than armv7. The armv6, i386, and powerpc platforms are deprecated and will be removed.

tboegi:
@anjohnson @kiwichris 

My previous patch was based on the (wrong) assumption, that gcc is the
compiler to be used (as in other OS).
However, Unix and BSD do use cc as the C-compiler.

This PR looks good, has now been tested and works like a charm:
please merge.

kiwichris:
Ping?

kiwichris:
Can this PR please be merged? I use FreeBSD as my development machine and having it merged would make things simpler and quicker for me when working on EPICS changes for RTEMS. Thanks

anjohnson:
Core Group: Merging. We'd like to remove the freebsd-x86 target files sometime, but this is fine for now.

PR #528, opened 2024-07-18
Add RTEMS ntpd support

kiwichris:
With RTEMS 6 this changes require `libntp` and that is available in the `rtems-net-services` package. The repository is available on https://gitlab.rtems.org and it builds for the legacy and `libbsd` networking stacks. 

RTEMS 5 builds the default POSIX support as RTEMS 5 does not have the kernel NTP support.

Configure ntpd with the following env variables:

- `EPICS_TS_NTP_INET` : NTP server
- `EPICS_TS_NTP_CONF_FILE` : ntpd configuration file

or use the IOC shell commands:

- NTPTime_Start()
- NTPTime_SyncWait()
- NTPTime_Update()

The `NTPTime_Start()` command accepts `ip` of `config` as the first argument and an IP address if `ip` or the path to an `ntpd` configuration file if `config`.

The `NTPTime_SyncWait()` command lets you wait in the IOC shell until the time is synchronized before you run `iocInit` and start to initialize the system. This can be important if an accurate time is needed. 

The `NTPTime_Update()` command accepts the IP address of an NTP server time is fetched from.

A configuration file is used if both a configuration file and server IP address are set.

This PR depends on https://github.com/epics-base/epics-base/pull/375. I do not know how to make a merge train on GitHub.

### DIFF
--- a/configure/os/CONFIG.Common.RTEMS
+++ b/configure/os/CONFIG.Common.RTEMS
@@ -140,7 +140,12 @@ OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_NETWORKING))
 POSIX_CPPFLAGS_posix = -D_GNU_SOURCE -D_DEFAULT_SOURCE
 POSIX_CPPFLAGS = $(POSIX_CPPFLAGS_$(OS_API))
 
+ifneq ($(RTEMS_VERSION),5)
+LDLIBS_posix_NET_ntp = -lntp
+endif
+
 OP_SYS_LDLIBS_posix_NET_yes = -ltftpfs -lz -ltelnetd
+OP_SYS_LDLIBS_posix_NET_yes += $(LDLIBS_posix_NET_ntp)
 OP_SYS_LDLIBS_posix_NET_yes += $(RTEMS_NET_LIB_$(RTEMS_NETWORKING))
 OP_SYS_LDLIBS_posix_NET_no = -ltftpfs -lz
 OP_SYS_LDLIBS_score_NET_yes = -lnfs
--- a/modules/libcom/src/osi/Makefile
+++ b/modules/libcom/src/osi/Makefile
@@ -82,7 +82,7 @@ Com_SRCS += epicsGeneralTime.c
 # Time providers
 Com_SRCS += osiClockTime.c
 Com_SRCS_vxWorks += osiNTPTime.c tz2timezone.c
-Com_SRCS_RTEMS += osiNTPTime.c
+Com_SRCS_RTEMS += osiNTPTime.c epicsSimpleNtp.c
 
 ifeq ($(OS_CLASS),vxWorks)
 osdTime.o: USR_CXXFLAGS += -DBUILD_TIME=$(shell perl -e 'print time')
--- a/modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.c
@@ -1,0 +1,90 @@
+/*************************************************************************\
+* (C) 2014 David Lettier.
+* http://www.lettier.com/
+* SPDX-License-Identifier: BSD-3-Clause
+*
+* use of UDP funktions and implement timeout, HPJ 15.12.2025
+\*************************************************************************/
+
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netdb.h>
+#include <arpa/inet.h>
+#include <rtems.h>
+#include "epicsSimpleNtp.h"
+
+int epicsSimpleNtpGetTime(char *ntpIp, struct timespec *now)
+{
+  int sockfd, n; // Socket file descriptor and the n return result from writing/reading from the socket.
+  int portno = 123; // NTP UDP port number.
+
+  // Create and zero out the packet. All 48 bytes worth.
+  ntp_packet packet = { 0 };
+  // Set the first byte's bits to 00,011,011 for li = 0, vn = 3, and mode = 3. The rest will be left set to zero.
+  *( ( char * ) &packet + 0 ) = 0x1b; // Represents 27 in base 10 or 00011011 in base 2.
+
+  /* Create an UDP socket, use IP address of given nameserver, set the port number
+     struct sockaddr_in serv_addr; // Server address data structure.
+     add timout of 2 seconds.
+   */
+  struct sockaddr_in serv_addr = {
+    .sin_family = AF_INET,
+    .sin_addr.s_addr = inet_addr(ntpIp),
+    .sin_port = htons( portno ),
+  };
+
+  sockfd = socket( AF_INET, SOCK_DGRAM, 0 );
+  if ( sockfd < 0 ) {
+    perror( "epicsNtpGetTime creating socket" );
+    return -1;
+  }
+
+  /* Set receive timeout to 2 seconds */
+  struct timeval tv = { .tv_sec = 2, .tv_usec = 0 };
+  if (setsockopt(sockfd, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv)) < 0) {
+    perror("setsockopt");
+    return -1;
+  }
+
+  // Send it the NTP packet it wants. If n == -1, it failed.
+  n = sendto(sockfd, ( char* ) &packet, sizeof( ntp_packet ), 0, (struct sockaddr *)&serv_addr, sizeof(serv_addr));
+  if ( n < 0 ) {
+    perror( "epicsNtpGetTime sending NTP request" );
+    close( sockfd );
+    return -1;
+  }
+
+  // Wait and receive the packet back from the server. If n == -1, it failed.
+  n = recvfrom(sockfd, ( char* ) &packet, sizeof( ntp_packet ), 0, NULL, NULL);
+  if (n < 0) {
+    if (errno == EAGAIN || errno == EWOULDBLOCK) {
+      printf("Timeout - no response\n");
+    } else {
+      perror( "epicsNtpGetTime reading NTP reply" );
+    }
+    close( sockfd );
+    return -1;
+  }
+  close( sockfd );
+
+  // These two fields contain the time-stamp seconds as the packet left the NTP server.
+  // The number of seconds correspond to the seconds passed since 1900.
+  // ntohl() converts the bit/byte order from the network's to host's "endianness".
+  packet.txTm_s = ntohl( packet.txTm_s ); // Time-stamp seconds.
+  packet.txTm_f = ntohl( packet.txTm_f ); // Time-stamp fraction of a second.
+
+  // Extract the 32 bits that represent the time-stamp seconds (since NTP epoch) from when the packet left the server.
+  // Subtract 70 years worth of seconds from the seconds since 1900.
+  // This leaves the seconds since the UNIX epoch of 1970.
+  // (1900)------------------(1970)**************************************(Time Packet Left the Server)
+  time_t txTm = ( time_t ) ( packet.txTm_s - NTP_TIMESTAMP_DELTA );
+  now->tv_sec = txTm;
+  return 0;
+}
--- a/modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.h
+++ b/modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.h
@@ -1,0 +1,50 @@
+/*************************************************************************\
+* (C) 2014 David Lettier.
+* http://www.lettier.com/
+* SPDX-License-Identifier: BSD-3-Clause
+\*************************************************************************/
+#ifndef EPICSSIMPLENTP_H
+#define EPICSSIMPLENTP_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define NTP_TIMESTAMP_DELTA 2208988800ull
+
+int epicsSimpleNtpGetTime(char *ntpIp, struct timespec *now);
+
+typedef struct
+{
+
+  uint8_t li_vn_mode __attribute__((unused));      // Eight bits. li, vn, and mode.
+                                                   // li.   Two bits.   Leap indicator.
+                                                   // vn.   Three bits. Version number of the protocol.
+                                                   // mode. Three bits. Client will pick mode 3 for client.
+
+  uint8_t stratum __attribute__((unused));         // Eight bits. Stratum level of the local clock.
+  uint8_t poll __attribute__((unused));            // Eight bits. Maximum interval between successive messages.
+  uint8_t precision __attribute__((unused));       // Eight bits. Precision of the local clock.
+
+  uint32_t rootDelay __attribute__((unused));      // 32 bits. Total round trip delay time.
+  uint32_t rootDispersion __attribute__((unused)); // 32 bits. Max error aloud from primary clock source.
+  uint32_t refId __attribute__((unused));          // 32 bits. Reference clock identifier.
+
+  uint32_t refTm_s __attribute__((unused));        // 32 bits. Reference time-stamp seconds.
+  uint32_t refTm_f __attribute__((unused));        // 32 bits. Reference time-stamp fraction of a second.
+
+  uint32_t origTm_s __attribute__((unused));       // 32 bits. Originate time-stamp seconds.
+  uint32_t origTm_f __attribute__((unused));       // 32 bits. Originate time-stamp fraction of a second.
+
+  uint32_t rxTm_s __attribute__((unused));         // 32 bits. Received time-stamp seconds.
+  uint32_t rxTm_f __attribute__((unused));         // 32 bits. Received time-stamp fraction of a second.
+
+  uint32_t txTm_s __attribute__((unused));         // 32 bits and the most important field the client cares about. Transmit time-stamp seconds.
+  uint32_t txTm_f __attribute__((unused));         // 32 bits. Transmit time-stamp fraction of a second.
+
+} ntp_packet;              // Total: 384 bits or 48 bytes.
+
+#ifdef __cplusplus
+}
+#endif
+#endif // EPICSSIMPLENTP_H
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp
@@ -1,0 +1,749 @@
+/*************************************************************************\
+* Copyright (c) 2002 The University of Saskatchewan
+* Copyright (c) 2008 UChicago Argonne LLC, as Operator of Argonne
+*     National Laboratory.
+* Copyright (c) 2023 Chris Johns <chris@contemporary.software>
+* SPDX-License-Identifier: EPICS
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+/*
+ * Author: Chris Johns
+ */
+#include <errno.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+
+#define EPICS_EXPOSE_LIBCOM_MONOTONIC_PRIVATE
+#include <epicsStdio.h>
+#include "epicsExit.h"
+#include "epicsTime.h"
+#include "errlog.h"
+#include "osdTime.h"
+#include "osiNTPTime.h"
+#include "osiClockTime.h"
+#include "generalTimeSup.h"
+#include "iocsh.h"
+#include "taskwd.h"
+
+#include "epicsSimpleNtp.h"
+
+#include <iocsh/initHooks.h>
+
+#include <rtems.h>
+
+/*
+ * This file is built automatically by EPICS's build system
+ */
+#if __RTEMS_MAJOR__ < 6
+
+/* Include the code from posix/osdTime.cpp. Better solutions welcome */
+
+#include "../posix/osdTime.cpp"
+
+#else /* __RTEMS_MAJOR__ < 6 */
+
+#include <string.h>
+
+#include <rtems/ntpd.h>
+#include <rtems/ntpq.h>
+
+extern "C" {
+
+/*
+ * NTPD is configured by a file /etc/ntp.conf. The RTEMS /etc/ directory
+ * does not survive a reset so ntp.conf is created each boot.
+ *
+ * Users can configure ntpd by:
+ *
+ *  None:
+ *     ntpd is not started
+ *
+ *  BOOTP/DHCP/Init:
+ *     The BOOTP server IP address is used if there is a valid BOOTP
+ *     record and variable rtemsInit_NTP_server_ip is empty and
+ *     rtems_bsdnet_config.ntp_server is set. If the variable
+ *     rtems_bsdnet_config.ntp_server is empty and rtemsInit_NTP_server_ip
+ *     is set it is used.
+ *
+ *  NTP Server IP:
+ *     The environment variable EPICS_TS_NTP_INET is used as the
+ *     address if no configuration file is set. This option takes
+ *     priority over BOOTP/DHCP/Init.
+ *
+ *  Configuration File:
+ *     The environment variable EPICS_TS_NTP_CONF_FILE is a path to a
+ *     configuration file that is used as is. This allows site specifc
+ *     and even board specific configuration support at run time. It
+ *     is recommended you provide a site specific configuration file
+ *     if you need site specific control. This option takes
+ *     priority over NTP Server IP.
+ *
+ *  The Configuration File option provides a system level means to
+ *  configure a site specific ntpd configuration that can be loaded
+ *  from a NFS file system when the system starts. This option
+ *  provides the ability to implement a site specific configuration of
+ *  ntpd.
+ *
+ *  Notes:
+ *    1. ntpd can take a while to synchronize the realtime clock
+ *    2. Changes in the environment variables not do reconfigure ntpd
+ */
+
+#define NTPD_CONFIG_NONE   0
+#define NTPD_CONFIG_BOOTP  1
+#define NTPD_CONFIG_NTP_IP 2
+#define NTPD_CONFIG_FILE   3
+
+#define osdNTP_NTPD_Fails  4
+
+/*
+ * NTPD Runner Interval
+ *
+ * The period in seconds the ntpd thread checks for a valid
+ * configuration.
+ */
+#define EPICS_NTPD_RUNNER_INTERVAL 1.0
+
+/*
+ * NTPD Monitor Intervals
+ *
+ * The periods are in seconds. The first is the period used when
+ * NtpD is not running. The second is the period when running.
+ * configuration.
+ */
+#define EPICS_NTPD_MONITOR_IDLE_INTERVAL   1.0
+#define EPICS_NTPD_MONITOR_ACTIVE_INTERVAL 10.0
+
+/*
+ * Environment variables
+ */
+#define EPICS_TS_NTP_INET      "EPICS_TS_NTP_INET"
+#define EPICS_TS_NTP_CONF_FILE "EPICS_TS_NTP_CONF_FILE"
+
+typedef ntp_sys_var_data ntpq_rl_data;
+
+/*
+ * Declared in rtems_init.c. No header.
+ */
+extern char rtemsInit_NTP_server_ip[16];
+
+/*
+ * For rtems_bsdnet_config.
+ */
+#ifdef RTEMS_LEGACY_STACK
+#include <rtems/rtems_bsdnet.h>
+extern struct rtems_bsdnet_config rtems_bsdnet_config;
+#endif /* RTEMS_LEGACY_STACK */
+
+ /*
+  * NTPD Private data
+  */
+static struct {
+    int          config_source;
+    char         config_arg[256]; /* could be a long path */
+    char         config_ip[32];
+    int          active;
+    int          synchronized;
+    int          ntpd_active;
+    int          ntpd_active_secs;
+    int          ntpd_finished;
+    float        runner_interval;
+    float        monitor_interval;
+    epicsEventId monitorEvent;
+    epicsEventId runnerEvent;
+    epicsMutexId lock;
+    int          ntpq_open;
+    ntpq_rl_data rl;
+    char         last_updated[64];
+    char         ntpq_output[2048];
+} osdNTPPvt;
+
+static epicsThreadOnceId onceId = EPICS_THREAD_ONCE_INIT;
+
+static void osdNTP_Configure(void);
+static void osdNTP_Runner(void *dummy);
+static void osdNTP_Monitor(void *dummy);
+
+void osdNTP_Shutdown(void *dummy);
+
+static const char etc_services[] =
+    "ntp                123/tcp      # Network Time Protocol  [Dave_Mills] [RFC5905]\n"
+    "ntp                123/udp      # Network Time Protocol  [Dave_Mills] [RFC5905]\n";
+
+static void set_rl_last_updated()
+{
+    struct timespec ts;
+    int r = clock_gettime(CLOCK_REALTIME, &ts);
+    if (r == 0) {
+        struct tm tm;
+        localtime_r(&ts.tv_sec, &tm);
+        strftime(
+            osdNTPPvt.last_updated, sizeof(osdNTPPvt.last_updated),
+            "%Y-%m-%d %H:%M:%S", &tm);
+    } else {
+        strlcpy(osdNTPPvt.last_updated, "unknown", sizeof(osdNTPPvt.last_updated));
+    }
+}
+
+static void ntpq_open(void)
+{
+    const char *argv[] = {
+        "host",
+        "127.0.0.1",
+        NULL
+    };
+    const int argc = ((sizeof(argv) / sizeof(argv[0])) - 1);
+    rtems_ntpq_create(1024);
+    int r = rtems_ntpq_query(
+        argc, argv, osdNTPPvt.ntpq_output, sizeof(osdNTPPvt.ntpq_output));
+    if (r != 0) {
+        errlogPrintf("osdNTP: ntpq host 127.0.0.1 failed\n");
+    }
+}
+
+static bool osdNTP_Get_Active(void)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    bool active = osdNTPPvt.ntpd_active;
+    epicsMutexUnlock(osdNTPPvt.lock);
+    return active;
+}
+
+static void osdNTP_Set_Active(bool active)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    osdNTPPvt.ntpd_active = active ? 1 : 0;
+    epicsMutexUnlock(osdNTPPvt.lock);
+}
+
+static void osdNTP_Set_ConfigSource(int config_source)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    osdNTPPvt.config_source = config_source;
+    epicsMutexUnlock(osdNTPPvt.lock);
+}
+
+static int osdNTP_Get_ConfigSource(void)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    int config_source = osdNTPPvt.config_source;
+    epicsMutexUnlock(osdNTPPvt.lock);
+    return config_source;
+}
+
+static bool osdNTP_Configured(void)
+{
+    return osdNTP_Get_ConfigSource() != NTPD_CONFIG_NONE;
+}
+
+static bool osdNTP_Synchronized(void)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    int synchronized = osdNTPPvt.synchronized;
+    epicsMutexUnlock(osdNTPPvt.lock);
+    return synchronized != 0;
+}
+
+static void osdNTP_Set_Synchronized(int synchronized)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    osdNTPPvt.synchronized = synchronized;
+    epicsMutexUnlock(osdNTPPvt.lock);
+}
+
+static bool osdNTP_Update_ActiveSecs(int secs)
+{
+    const int active_before_ntpq_open = 15;
+    epicsMutexMustLock(osdNTPPvt.lock);
+    if (osdNTPPvt.active) {
+        int active_secs = osdNTPPvt.ntpd_active_secs;
+        osdNTPPvt.ntpd_active_secs += secs;
+        /*
+         * Has the active time been long enough to check the ntpd
+         * port?  The ntpq host command checks.
+         */
+        if (active_secs <= active_before_ntpq_open
+            && osdNTPPvt.ntpd_active_secs > active_before_ntpq_open) {
+            epicsMutexUnlock(osdNTPPvt.lock);
+            ntpq_open();
+            epicsMutexMustLock(osdNTPPvt.lock);
+            osdNTPPvt.ntpq_open = 1;
+        }
+    }
+    bool ntpq_is_open = osdNTPPvt.ntpq_open != 0;
+    epicsMutexUnlock(osdNTPPvt.lock);
+    return ntpq_is_open;
+}
+
+static void osdNTP_InitOnce(void *not_used)
+{
+    (void) not_used;
+
+    osdNTPPvt = { 0 };
+    osdNTPPvt.config_source    = NTPD_CONFIG_NONE;
+    osdNTPPvt.active           = 1;
+    osdNTPPvt.synchronized     = 0;
+    osdNTPPvt.ntpd_active      = 1;
+    osdNTPPvt.ntpd_active_secs = 0;
+    osdNTPPvt.ntpd_finished    = 0;
+    osdNTPPvt.runner_interval  = EPICS_NTPD_RUNNER_INTERVAL;
+    osdNTPPvt.monitor_interval = EPICS_NTPD_MONITOR_IDLE_INTERVAL;
+    osdNTPPvt.monitorEvent     = epicsEventMustCreate(epicsEventEmpty);
+    osdNTPPvt.runnerEvent      = epicsEventMustCreate(epicsEventEmpty);
+    osdNTPPvt.lock             = epicsMutexCreate();
+    osdNTPPvt.ntpq_open        = 0;
+
+    epicsThreadCreate("osdNTP_Runner", epicsThreadPriorityHigh,
+        epicsThreadGetStackSize(epicsThreadStackMedium),
+        osdNTP_Runner, nullptr);
+    epicsThreadCreate("osdNTP_Monitor", epicsThreadPriorityMedium,
+        epicsThreadGetStackSize(epicsThreadStackSmall),
+        osdNTP_Monitor, nullptr);
+    epicsAtExit(osdNTP_Shutdown, nullptr);
+}
+
+static void osdNTP_Configure(void)
+{
+    int config_source = NTPD_CONFIG_NONE;
+    const char* ev;
+    int r;
+
+    /*
+     * Add NTP ports to /etc/services
+     */
+    r = rtems_ntpd_add_etc_services();
+    if (r == 0) {
+        /*
+         * Configuration file? Set the leapfile field in the config file.
+         */
+        ev = getenv(EPICS_TS_NTP_CONF_FILE);
+        if (ev != nullptr) {
+            strlcpy(osdNTPPvt.config_arg, "--configfile=", sizeof(osdNTPPvt.config_arg));
+            strlcat(osdNTPPvt.config_arg, ev, sizeof(osdNTPPvt.config_arg));
+            config_source = NTPD_CONFIG_FILE;
+        }
+
+        /*
+         * Server IP address?
+         */
+        if (config_source == NTPD_CONFIG_NONE) {
+            ev = getenv(EPICS_TS_NTP_INET);
+            if (ev != nullptr) {
+                strlcpy(
+                    osdNTPPvt.config_ip, "EPICS_TS_NTP_INET: ",
+                    sizeof(osdNTPPvt.config_ip));
+                strlcpy(osdNTPPvt.config_ip, ev, sizeof(osdNTPPvt.config_ip));
+                r = rtems_ntpd_client_pool_config(ev);
+                if (r == 0) {
+                    config_source = NTPD_CONFIG_NTP_IP;
+                }
+            }
+        }
+
+        /*
+         * BOOTP/DHCP/Init?
+         */
+        if (config_source == NTPD_CONFIG_NONE) {
+            ev = nullptr;
+#ifdef RTEMS_LEGACY_STACK
+            if (rtems_bsdnet_config.ntp_server[0] != nullptr) {
+                ev = rtems_bsdnet_config.ntp_server[0];
+                strlcpy(
+                    osdNTPPvt.config_ip, "BSDNET_CONFIG: ",
+                    sizeof(osdNTPPvt.config_ip));
+            }
+#endif /* RTEMS_LEGACY_STACK */
+            if (ev == nullptr && rtemsInit_NTP_server_ip[0] != '\0') {
+                ev = &rtemsInit_NTP_server_ip[0];
+                strlcpy(
+                    osdNTPPvt.config_ip, "RTEMS_INIT: ",
+                    sizeof(osdNTPPvt.config_ip));
+            }
+            if (ev != nullptr) {
+                strlcat(osdNTPPvt.config_ip, ev, sizeof(osdNTPPvt.config_ip));
+                r = rtems_ntpd_client_pool_config(ev);
+                if (r == 0) {
+                    config_source = NTPD_CONFIG_BOOTP;
+                }
+            }
+        }
+    }
+
+    osdNTP_Set_ConfigSource(config_source);
+}
+
+void osdNTP_Shutdown(void *dummy)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    osdNTPPvt.config_source = NTPD_CONFIG_NONE;
+    osdNTPPvt.active = 0;
+    osdNTPPvt.synchronized = 0;
+    osdNTPPvt.ntpq_open = 0;
+    epicsMutexUnlock(osdNTPPvt.lock);
+    rtems_ntpd_stop();
+    rtems_ntpq_destroy();
+    epicsEventSignal(osdNTPPvt.runnerEvent);
+    epicsEventSignal(osdNTPPvt.monitorEvent);
+}
+
+static int osdNTP_RunCommand(const char* activity, int *fails, int argc, char** argv)
+{
+    osdNTP_Set_Active(true);
+    int r = rtems_ntpd_run(argc, argv);
+    osdNTP_Set_Active(false);
+    if (r != 0) {
+        *fails += 1;
+        errlogPrintf("osdNTP: ntpd error (fails=%d): %d: %s\n", *fails, r, activity);
+    } else {
+        *fails = 0;
+    }
+    return r;
+}
+
+static int osdNTP_Run(int *fails)
+{
+    const char *argv[] = {
+        "ntpd",
+        "-g",
+#if NTP_DEBUG
+        "--set-debug-level=2",
+#endif
+        osdNTPPvt.config_arg,
+        nullptr
+    };
+    const int argc = ((sizeof(argv) / sizeof(argv[0])) - 1);
+    return osdNTP_RunCommand("daemon", fails, argc, (char**) argv);
+}
+
+static void osdNTP_Runner(void *dummy)
+{
+    int ntpd_fails = 0;
+
+    taskwdInsert(0, nullptr, nullptr);
+
+    /*
+     * Wait for a valid configuration mode to be set. Restart ntpd if
+     * it exits limiting the number of attempts if there is a persistent
+     * error.
+     */
+    for (epicsEventWaitWithTimeout(osdNTPPvt.runnerEvent, osdNTPPvt.runner_interval);
+         osdNTPPvt.active && ntpd_fails < osdNTP_NTPD_Fails;
+         epicsEventWaitWithTimeout(osdNTPPvt.runnerEvent, osdNTPPvt.runner_interval)) {
+
+        if (osdNTP_Configured()) {
+            int r = osdNTP_Run(&ntpd_fails);
+            if (r == 0) {
+                osdNTPPvt.monitor_interval = EPICS_NTPD_MONITOR_ACTIVE_INTERVAL;
+            }
+        }
+    }
+
+    epicsMutexMustLock(osdNTPPvt.lock);
+    osdNTPPvt.ntpd_finished = 1;
+    epicsMutexUnlock(osdNTPPvt.lock);
+
+    taskwdRemove(0);
+}
+
+static void osdNTP_Monitor(void *dummy)
+{
+    taskwdInsert(0, nullptr, nullptr);
+
+    for (epicsEventWaitWithTimeout(osdNTPPvt.monitorEvent, osdNTPPvt.monitor_interval);
+         osdNTPPvt.active;
+         epicsEventWaitWithTimeout(osdNTPPvt.monitorEvent, osdNTPPvt.monitor_interval)) {
+
+        if (osdNTP_Configured()
+            && osdNTP_Get_Active()
+            && osdNTP_Update_ActiveSecs(osdNTPPvt.monitor_interval)) {
+            epicsMutexMustLock(osdNTPPvt.lock);
+            rtems_ntpd_get_sys_vars(&osdNTPPvt.rl);
+            set_rl_last_updated();
+            epicsMutexUnlock(osdNTPPvt.lock);
+        }
+
+        osdNTP_Set_Synchronized(rtems_ntpd_is_synchronized(&osdNTPPvt.rl));
+
+        if (!osdNTP_Configured()) {
+            osdNTP_Configure();
+        }
+    }
+
+    osdNTP_Set_Synchronized(0);
+
+    taskwdRemove(0);
+}
+
+void osdTimeRegister(void)
+{
+    /* Init NTP first so it can be used to sync ClockTime */
+    NTPTime_Init(100);
+    ClockTime_Init(CLOCKTIME_SYNC);
+
+    osdMonotonicInit();
+}
+
+int osdNTPGet(struct timespec *ts)
+{
+    int r;
+    if (osdNTP_Synchronized()) {
+        r = clock_gettime(CLOCK_REALTIME, ts);
+    } else {
+        r = -1;
+    }
+    return r;
+}
+
+void osdNTPInit(void)
+{
+    printf("ntpd: RTEMS service init\n");
+    epicsThreadOnce(&onceId, osdNTP_InitOnce, nullptr);
+}
+
+void osdNTPReport(void)
+{
+    epicsMutexMustLock(osdNTPPvt.lock);
+    int config_source = osdNTPPvt.config_source;
+    int synchronized = osdNTPPvt.synchronized;
+    bool ntpd_active = osdNTPPvt.ntpd_active == 1;
+    int ntpd_active_secs = osdNTPPvt.ntpd_active_secs;
+    bool ntpq_open = osdNTPPvt.ntpq_open == 1;
+    epicsMutexUnlock(osdNTPPvt.lock);
+    printf("RTEMS NTP Status:\n");
+    printf(" config type = ");
+    if (config_source <= NTPD_CONFIG_FILE) {
+        static const char* config_str[4] = {
+            "NTPD_CONFIG_NONE",
+            "NTPD_CONFIG_BOOTP",
+            "NTPD_CONFIG_NTP_IP",
+            "NTPD_CONFIG_FILE"
+        };
+        printf("%s\n", config_str[config_source]);
+
+    } else {
+        printf("INVALID (%i)\n", config_source);
+    }
+    if (osdNTPPvt.config_arg[0] != '\0') {
+        printf(" config arg = %s\n", osdNTPPvt.config_arg);
+    }
+    if (osdNTPPvt.config_ip[0] != '\0') {
+        printf(" config ip = %s\n", osdNTPPvt.config_ip);
+    }
+    printf(" ntpd active = %s\n", ntpd_active ? "yes" : "no");
+    printf(" ntpd active sec = %d\n", ntpd_active_secs);
+    printf(" ntpq open = %s\n", ntpq_open ? "yes" : "no");
+    printf(" status last updated = %s\n", osdNTPPvt.last_updated);
+    printf(" synchronized = %s\n", synchronized ? "yes" : "no");
+    printf(" status = (%04x) %s\n", osdNTPPvt.rl.status, osdNTPPvt.rl.status_str);
+    printf(" version = %s\n", osdNTPPvt.rl.version);
+    printf(" processor = %s\n", osdNTPPvt.rl.processor);
+    printf(" system = %s\n", osdNTPPvt.rl.system);
+    printf(" leap = %d\n", (int) osdNTPPvt.rl.leap);
+    printf(" stratum = %d\n", (int) osdNTPPvt.rl.stratum);
+    printf(" precision = %i\n", (int) osdNTPPvt.rl.precision);
+    printf(" rootdelay = %f\n", osdNTPPvt.rl.rootdelay);
+    printf(" rootdisp = %f\n", osdNTPPvt.rl.rootdisp);
+    printf(" refid = %s\n", osdNTPPvt.rl.refid);
+    printf(" reftime = %" PRIu64 ".%" PRIu64"\n",
+           osdNTPPvt.rl.reftime_sec, osdNTPPvt.rl.reftime_nsec);
+    printf(" clock = %" PRIu64 ".%" PRIu64 "\n",
+           osdNTPPvt.rl.clock_sec, osdNTPPvt.rl.clock_nsec);
+    printf(" peer = %d\n", osdNTPPvt.rl.peer);
+    printf(" tc = %d\n", osdNTPPvt.rl.tc);
+    printf(" mintc = %d\n", osdNTPPvt.rl.mintc);
+    printf(" offset = %f\n", osdNTPPvt.rl.offset);
+    printf(" frequency = %f\n", osdNTPPvt.rl.frequency);
+    printf(" sys_jitter = %f\n", osdNTPPvt.rl.sys_jitter);
+    printf(" clk_jitter = %f\n", osdNTPPvt.rl.clk_jitter);
+    printf(" clk_wander = %f\n", osdNTPPvt.rl.clk_wander);
+    printf(" tai = %d\n", osdNTPPvt.rl.tai);
+    printf(" leapsec = %" PRIu64 "\n", osdNTPPvt.rl.leapsec);
+    printf(" expire = %" PRIu64 "\n", osdNTPPvt.rl.expire);
+    /*
+     * Query the peers. We need to be able to see what ntpd is using
+     * as a time source.
+     */
+    printf("Peers:\n");
+    if (ntpq_open) {
+        const char *argv[] = {
+            "peers",
+            nullptr
+        };
+        const int argc = ((sizeof(argv) / sizeof(argv[0])) - 1);
+        int r = rtems_ntpq_query(
+            argc, argv, osdNTPPvt.ntpq_output,
+            sizeof(osdNTPPvt.ntpq_output));
+        if (r == 0) {
+            printf(osdNTPPvt.ntpq_output);
+        } else {
+            printf("%s\n", rtems_ntpq_error_text());
+        }
+    } else {
+        printf("ntpq: not open\n");
+    }
+}
+
+int osdTickGet(void)
+{
+    return rtems_clock_get_ticks_since_boot();
+}
+
+int osdTickRateGet(void)
+{
+    return rtems_clock_get_ticks_per_second();
+}
+
+/*
+ * Use reentrant versions of time access
+ */
+int epicsTime_gmtime ( const time_t *pAnsiTime, struct tm *pTM )
+{
+    struct tm * pRet = gmtime_r ( pAnsiTime, pTM );
+    if ( pRet ) {
+        return epicsTimeOK;
+    }
+    else {
+        return errno;
+    }
+}
+
+int epicsTime_localtime ( const time_t *clock, struct tm *result )
+{
+    struct tm * pRet = localtime_r ( clock, result );
+    if ( pRet ) {
+        return epicsTimeOK;
+    }
+    else {
+        return errno;
+    }
+}
+
+} // extern "C"
+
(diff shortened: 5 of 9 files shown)


anjohnson:
Core Group comment, not a review/request: This PR contains an NTP client for use by RTEMS-5 targets. We might not need to support RTEMS-5 after RTEMS-6 has been released and integrated into Base. There seems to be an issue with the current osdTime code on RTEMS-posix that could be fixed to allow RTEMS-5 to use the EPICS-provided NTP time provider (less accurate than a proper NTP client), but that might be sufficient for any remaining RTEMS-5 users.

kiwichris:
I have reviewed the warnings the static checker has raised and they seem like noise to me.

mdavidsaver:
Firstly, Is this PR relevant now that RTEMS 6.1 is released?

> I have reviewed the warnings the static checker has raised and they seem like noise to me.

Is the added `RTEMS-posix/osdTime.cpp` new code for this PR, or taken from some other source?

If it is new, and would be unique to epics-base, then I think it would be be a worthwhile exercise to follow up on these warnings.  I agree that the texts of these Codacy warnings are unhelpful, but I think several are actionable in that what is being attempted could be accomplished in simpler ways.

There is a lot of C string manipulation being added with this PR, with zero test coverage.  This makes me nervous.

kiwichris:
> Firstly, Is this PR relevant now that RTEMS 6.1 is released?
> 
> > I have reviewed the warnings the static checker has raised and they seem like noise to me.
> 
> Is the added `RTEMS-posix/osdTime.cpp` new code for this PR, or taken from some other source?
> 

New and I have reviewed each case again. My comments are below.

> If it is new, and would be unique to epics-base, then I think it would be be a worthwhile exercise to follow up on these warnings. I agree that the texts of these Codacy warnings are unhelpful, but I think several are actionable in that what is being attempted could be accomplished in simpler ways.

Case 1:

The suggestion is the clear memory implies a needs to be clear after use. The memory is a local variable on a calling stack and goes out of scope when the called function returns. The call is `static` and the compiler will deal with this call in the module wide optimization so doubt a real call actually exists. I can add a clear at the end of the block that uses the code to see if this help but it implies the checker is weak in some areas.

Case 2:

The buffer pointed to by `buf` is initialized by calls to `strlcpy` and `strlcat` to make sure there is no buffer overrun. The `strlen` call is not protected because the other calls are considered OK. It is shame the analyzer cannot see that. Maybe using `strnlen` will avoid the warning.

Case 3:

I will add a length to the call using `sizeof(a_string)`.

Case 4:

Looking at this again I think the `write` call should use the size returned from the `read` call `s` if positive.

Any suggestions on how to check the changes without pushing to this PR?

> There is a lot of C string manipulation being added with this PR, with zero test coverage. This makes me nervous.

How would test coverage to this code be added to EPICS?

Chris

mdavidsaver:
Looking more closely... do I understand that the `struct ntpq_rl_data` parsing and printing is entirely local to osdTime.cpp?  Is this reformatting really worth the significant extra code complexity?  Why not just print the output of `ntpq` verbatim?

mdavidsaver:
For that matter, could the whole `ntpd` setup be externalized through the RTEMS shell?  (exposed via `rtems ...` from the IOC shell)

I don't like seeing configuration file contents hard-coded into epics-base as it prevents, or at least complicates site/instance configuration.  eg. your `osdNTP_Monitor` thread looking for changes to environment variables, and overwriting `/etc/...` files.

mdavidsaver:
> How would test coverage to this code be added to EPICS?

epics-base has a framework for unit-testing, and many example in the tree.  eg. `modules/libcom/test/`.

There is so far no `modules/libcom/RTEMS/test/`, but that would be one obvious place to put test of `librtemsCom.a`.

For code like this string parsing, which is not really RTEMS specific, I might structure this to build and run also as a host executable.  (with the happy side of being much easier to develop and debug...)

kiwichris:
> For that matter, could the whole `ntpd` setup be externalized through the RTEMS shell? (exposed via `rtems ...` from the IOC shell)
> 
> I don't like seeing configuration file contents hard-coded into epics-base as it prevents, or at least complicates site/instance configuration. eg. your `osdNTP_Monitor` thread looking for changes to environment variables, and overwriting `/etc/...` files.

Maybe the comment at the start of the file may help:
```
 * NTPD is configured by a file /etc/ntp.conf. The RTEMS /etc/ directory
 * does not survive a reset so ntp.conf is created each boot.
 *
 * Users can configure ntpd by:
 *
 *  None:
 *     ntpd is not started
 *
 *  BOOTP/DHCP:
 *     The BOOTP server IP address is used if there is a valid BOOTP
 *     record. This is currently not supported because it is not
 *     implemented with the libbsd stack.
 *
 *  NTP Server IP:
 *     The environment variable EPICS_TS_NTP_INET is used as the
 *     address if no configuration file is set.
 *
 *  Configuration File:
 *     The environment variable EPICS_TS_NTP_CONF_FILE is a path to a
 *     configuration file that is used as is. This allows site specifc
 *     and even board specific configuration support at run time. It
 *     is recommended you provide a site specific configuration file
 *     if you need site specific control.
 *
 *  The Configuration File option provides a system level means to
 *  configure a site specific ntpd configuration that can be loaded
 *  from a NFS file system when the system starts. This option
 *  provides the ability to implement a site specific configuration of
 *  ntpd.
```

The config file fragment is present is to provide backward compatibility for existing users who want a simple NTP set up similar to the simple client Eric and other provided years ago. I suggest you provide external configuration files fit for you system including the leapyear file.

If you want EPCIS to break backward compatibility just say but I did not think EPICS did that sort of thing so I went to the trouble of finding a solution.

kiwichris:
> Looking more closely... do I understand that the `struct ntpq_rl_data` parsing and printing is entirely local to osdTime.cpp?

Yes. 

[ Also I have attempted to deal with the checker but it seems there is a limit to how far it can see. The buffer is being cleared and the `read` length is calculated the line before based on the buffer size and the `write` uses the read length (which is checked against the length passed to `read`). ]

> Is this reformatting really worth the significant extra code complexity? 

The NTP client is replacing what existed. If you do not want this monitoring and reporting of the NTP client state it can be removed or disabled by default. Users can be pointed to the the `ntpd` documentation for solutions.

> Why not just print the output of `ntpq` verbatim?

There is an RTEMS shell command available in the net-services package however it is not monitoring the state. It should not be hard for someone who knows that area of EPICS to add it.

You could also provide a suitable config file to provide external access and do the same thing with any `ntpq` command based on `ntpd`.

mdavidsaver:
> If you do not want this monitoring and reporting of the NTP client state it can be removed or disabled by default.

I am not asking for removal.  I am asking whether this (imo. expert debugging) information can simply be presented in a "raw" form with less effort?

Your comment makes it look like the "raw" form is space separated key=value pairs.  Potentially with more keys than would be parsed into `struct ntpq_rl_data`.  To my eye, this looks clear enough for an expert to parse.

```c
 * The variable list as returned by the `rl` ntpq command. The output is:
 *
 * associd=0 status=0615 leap_none, sync_ntp, 1 event, clock_sync,
```

anjohnson:
The commit 144f9756eac4e800f18d3d9a35f0d42a712614fd from the iocshSetError additions that the Core Group merged earlier today has caused a conflict in this branch, sorry @kiwichris, please rebase and force-push.

Note that the Appveyor build failure was a single "Failed in 1 hr" false positive job; I was trying to re-run that job but it aborted with "Pull request #528 is non-mergeable."

kiwichris:
> > If you do not want this monitoring and reporting of the NTP client state it can be removed or disabled by default.
> 
> I am not asking for removal. 

I am sorry, I misunderstood. This is a great discussion and well worth having.

> I am asking whether this (imo. expert debugging) information can simply be presented in a "raw" form with less effort?
> 
> Your comment makes it look like the "raw" form is space separated key=value pairs. Potentially with more keys than would be parsed into `struct ntpq_rl_data`. 

This is correct. The [ntpq](https://www.ntp.org/documentation/4.2.8-series/ntpq/) documentation provides a complete list. On Linux, FreeBSD etc `ntpq` uses the same protocol to query a server. There is no direct access to the internals of `ntpd`. I briefly looked into direct access and found a rabbit hole of potential problems and stepped back from that path. 

> To my eye, this looks clear enough for an expert to parse.
> 
> ```c
>  * The variable list as returned by the `rl` ntpq command. The output is:
>  *
>  * associd=0 status=0615 leap_none, sync_ntp, 1 event, clock_sync,
> ```

Yes this would remove the need for that code however you also lose the ability to know the synchronized state and if the clock source is usable:
```
        if (strcmp("sync_ntp", rl.sync_status) == 0) {
            osdNTP_Set_Synchronized(1);
        } else {
            osdNTP_Set_Synchronized(0);
        }
```
If you have more than one clock source there could be issues if this one cannot cleanly signal its state. This code currently does not propagate this state. I was waiting for this change and this discussion to explore what this means. I feel on RTEMS and in an embedded system this may be important. Currently the code makes the same assumption about the clock source Linux does.

EPICS on Linux or Windows would use the OS system for time unless there is a specialized time source. RTEMS being an embedded RTOS does not come with external networked "clock management" as a standard feature. You need to embed it and that means you need to manage it. I suppose this thread of discussion ends up at the EPICS system level and what the core dev see as being needed? Does NTP clock management on RTEMS:

1. Rely on external management interfaces built into NTP (remote `ntpq`)? I do not favor this approach because it could be a DOS vector complicating the ability to lock down a system at the network level.
2. Rely on an EPICS shell command mapped to the `ntpq` command? Does not help multiple clock management reporting and status.
3. Add NTP monitoring to `iocStats` using this parsed set of data?

Items 1 and 2 means the parsing code can be removed and item 3 means it remains. Note, the monitoring being discussed uses the `ntpq` networked protocol however it can be restrict to the `127.0.0.1` interface in the configuration file.

kiwichris:
> The commit [144f975](https://github.com/epics-base/epics-base/commit/144f9756eac4e800f18d3d9a35f0d42a712614fd) from the iocshSetError additions that the Core Group merged earlier today has caused a conflict in this branch, sorry @kiwichris, please rebase and force-push.

All good and thanks for the heads up.

kiwichris:
I have rebased the PR against 7.0 of of today. The Codacy issues raised have me confused so I am not sure how to address them or how you report things like this to Codacy.

Is this PR able to be merged?

anjohnson:
Thanks, we are getting confused by Codacy issues at the moment too. There's an error in the Codacy image logs for this job which shows me this, @ralphlange any ideas?

image

I cancelled the Appveyor build for this PR, the queue there is slow and it doesn't build the RTEMS code anyway.

I don't think we'll let those weird Codacy issues stop us from merging this PR once we're happy with it. I haven't reviewed the latest changes here myself (just answering this for now) but I am putting RTEMS on our meeting Agenda for Wednesday.

ralphlange:
When logged into Coday, I can see more reasoning:
image

mdavidsaver:
> Items 1 and 2 means the parsing code can be removed and item 3 means it remains. Note, the monitoring being discussed uses the `ntpq` networked protocol however it can be restrict to the `127.0.0.1` interface in the configuration file.

Then I am in favor of approaches 1 or 2 as I would prefer not to see this complex print-and-reparse logic in Base.

Personally, I have no problem with allowing a read-only subset of NTP mode 6.  imo. EPICS IOCs are by their nature always network attached, and thus always subject to possible DoS.

> ... you also lose the ability to know the synchronized state and if the clock source is usable: ...

How about a simpler way to parse out this one key/value pair?  Of course a regexp.  Perhaps also `sscanf()`?

mdavidsaver (changes requested):
> Is this PR able to be merged?

imo. no.

I'm sorry but I do not like the design being proposed.  I would like to avoid our growing a custom, difficult to test, mostly hard coded, init system for RTEMS.  This PR seems like a significant step in that direction.

#595 has my thinking about what an alternative design(s) could look like.

wrt. this PR specifically, I would want to see `osdNTP_Monitor()` removed.  I would also want to see some explanation of why `osdNTP_Runner()` is needed.  What all is the `osdNTPPvt.config_source` state machine meant to handle?

In general, I would like to see a PR which starts with the absolute minimum needed to setup and call `rtems_ntpd_run()` (assuming it can't already be called from the RTEMS shell).   Then add only the simplest diagnostics necessary for a human to verify that the NTP client is running and synced.  No new `$EPICS_*` environment variables.  No hardcoded `/etc` file contents.

kiwichris:
> > Is this PR able to be merged?
> 
> imo. no.

Thanks for the review and feedback.
 
> I'm sorry but I do not like the design being proposed.

Thanks but there is no need to be sorry. It is good we work towards a suitable design and solution. My response here is to try and present the reasons for the choices made.

> I would like to avoid our growing a custom, difficult to test, mostly hard coded, init system for RTEMS. This PR seems like a significant step in that direction.

From your ending comment I assume the hard coding reference is the configuration and leap second file fragments? The config files are in the [NTP client configuration](https://www.ntp.org/documentation/4.2.8-series/ntp.conf/) format. The file fragment is taken from the [ntp01](https://gitlab.rtems.org/rtems/pkg/rtems-net-services/-/blob/main/testsuites/ntp01/test_main.c) test code in RTEMS. It is the way this stuff works on RTEMS. The code is only in the RTEMS POSIX directory. 

Does EPICS testing for RTEMS support networking protocols like NTP? If so I would love the details so I can take a look. Did the NTP client for the legacy stack Eric wrote have a set of tests?

In the simplest configuration mode set the environment variable `EPICS_TS_NTP_INET` in a boot script with the IP address of the NTP server and the NTP client will use that server when started. I believe that meets a key requirement of your #595 approach. 

The single env var `EPICS_TS_NTP_INET` does not handle some important things a fully functional NTP client may need. Examples are multiple NTP servers or updating the leap year seconds file when a new version is released (without rebuilding the software)? The `EPICS_TS_NTP_CONF_FILE` is an environment variable you set with a path to a site specific configuration file. The approach is simple and meets the #595 goals plus it means EPICS and RTEMS have no extra code between users and the NTP client code. The NTP standard configuration file format with it documentation can be used as a reference. These configuration files are no different to the files the systems integrators would use if running on Unix.

> #595 has my thinking about what an alternative design(s) could look like.

I have responded in that issue and I agree with the approach. I am happy to consider changes however I would prefer I understand and we agree on what is acceptable first. The current approach attempts to:

1. Follow what I saw in EPICS and Andrew explained which is backwards compatibility to older version. This is achieved with the `EPICS_TS_NTP_INET` environment variable.
2. Provide a synchronized status for the NTP clock source. It is my understanding from reviewing the EPICS clock source code it is important to indicate if the clock source is synchronized. If I have misunderstood how that works or the fact NTP on EPICS systems is assumed to work then please let me know?

> wrt. this PR specifically, I would want to see `osdNTP_Monitor()` removed.

How do you monitor the NTP client to determine the clock source is synchronized and the clock source is stable and usable?

On an RTEMS system there is no ability to run some other process or use some other tool to monitor the state of NTP.  I felt it important for systems engineers to know the state of the NTP client on RTEMS to aid integration as well as know the state. The monitoring and the EPICS command has proved to be valuable in the systems running this code in production. I added all I could because it was not easy to see what could be removed and yet useful. The data was put there by the NTP designers so I respected their decisions and provided that data to the EPICS command. If EPICS has suitable hooks I can use then yes lets move to using them.

NTP running on RTEMS is a black box. I did look into inspecting the internal state of the client before adding the monitoring via a loopback socket however the NTP client code is not thread safe so there would have been race conditions.

> I would also want to see some explanation of why `osdNTP_Runner()` is needed. What all is the `osdNTPPvt.config_source` state machine meant to handle?

The `osdNTP_Runner()` and `osdNTPPvt.config_source` manage the way a user can configure the NTP client.

RTEMS now supports a complete NTP client and like an instance running on Unix it is configured with a couple of files, the `/etc/ntp.conf` and `/etc/leapseconds` file. Without valid configurations the client will not run. RTEMS has moved away from providing back doors and unsupportable global variables and calls to using the standard means a package supports. The approach will bring long term stability to EPICS and RTEMS integration because it is an agreed public facing interface for the related package.

The monitoring and triggering of configuration states lets a user add an environment variable independent of the state of the NTP client and the client will start. There is no `services` or `systemd` to manage that on RTEMS. I was not sure how EPICS on RTEMS universally initialized across all the supported BSPs and possible configuration methods. When is a network interface available to use or a DHCP lease received? The running and the configuration state lets the NTP configuration and start happen asynchronously to the initialization and environment variable set up.

In respect to #595 I did mention the timing of the network interface initialization and you could view the solution here as a self-contained instance of the same problem that code would face. If #595 handled initialization events and some sort of dispatch for RTEMS specific services the running code here could be removed.

> In general, I would like to see a PR which starts with the absolute minimum needed to setup and call `rtems_ntpd_run()` (assuming it can't already be called from the RTEMS shell). 

I am not sure what that achieves as this PR provides that and then everything else that is needed? This code has been running in production systems no-stop from July 2024. As stated I added all the NTP fields to the command to avoid any extra overheads and complexity for EPICS users of this NTP client. They do not need a remote `ntpq` command or a specific `ntp.conf` file to provide access.

> Then add only the simplest diagnostics necessary for a human to verify that the NTP client is running and synced.

Could you please indicate using the `ntpq` data what that is?

> No new `$EPICS_*` environment variables. No hardcoded `/etc` file contents.

How would the NTP client work without the configuration files it needs?

I think a "hardcoded" restriction that bans publicly documented file formats will make RTEMS on EPICS difficult. Maybe the EPICS core devs should discuss what this means and then maybe reach out to the RTEMS community? This PR and #595 is aimed at cleaning up years of EPICS peeking under the hood at RTEMS and then breaking when RTEMS changes. Part of the solution where appropriate is to move to file formats the underlying software provides. Nothing here or in #595 has been created by RTEMS and in the case of `ntp.conf` that format is over 30 years old. I hope you reconsider this position and appreciate these are not formats created by RTEMS for EPICS. We consider integration and use carefully in RTEMS and the impact of users.

There are parts of this code that could be moved to RTEMS net-services and test added if this helps EPICS and EPICS integration such as the default configuration for a single IP address and the monitoring? The monitoring thread would be best handled in EPICS given it can manage priorities. I think the ability to allow a user to provide a configuration file is a good addition for EPICS and should be kept.

hjunkes:
Good morning Chris,
Today on my Beagleboneblack with RTEMS 6.1 and EPICS 7 with the NTP-PR I have this in the console:
```
beaglebone>
beaglebone> NTPTimeSync: Sync recovered at <undefined>
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (1 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (18 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (13 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (13 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (12 occurrences)
NTPTimeSync: NTP requests failed since <undefined> - Success
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (13 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (13 occurrences)
NTPTimeSync: Sync recovered at 2025-03-11 10:31:25.576713
cpsw0: ipv4_addroute: mcpsw0: ipv4_addroute: m_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (12 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (2 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (2 occurrences)

beaglebone> _bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (4 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (2 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (2 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (2 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (2 occurrences)
_bsd_sonewconn: pcb 0x8e694148: Listen queue overflow: 7 already in queue awaiting acceptance (3 occurrences)
pthread_create error No more processes
epicsThread: Unexpected C++ exception "unable to create thread" with type "N11epicsThread20unableToCreateThreadE" in thread "TCP-acceptor" at Tue Mar 11 2025 10:48:12.375603187

*** EXIT STATUS NOT ZERO ***
exit source: 5 (RTEMS_FATAL_SOURCE_EXIT)
exit code: 1 (0x00000001)
RTEMS version: 6.0.0.b4149b2282ee3543455894ae7981703de1d719ac-modified
RTEMS tools: 13.3.0 20240521 (RTEMS 6, RSB b1aec32059aa0e86385ff75ec01daf93713fa382-modified, Newlib 1b3dcfd)
executing thread ID: 0x0b010021
executing thread name: TCP-acceptor
U-Boot SPL 2018.09-00002-g0b54a51eee (Sep 10 2018 - 19:41:39 -0500)
```
I have not yet investigated where this suddenly comes from. I haven't seen it before.
Heinz

kiwichris:
> Good morning Chris, Today on my Beagleboneblack with RTEMS 6.1 and EPICS 7 with the NTP-PR I have this in the console:
> 
> [snip]
<
> I have not yet investigated where this suddenly comes from. I haven't seen it before. Heinz

Thanks for the report. It looks resources or network related. NTP has kindly reporting its state to the console. I do not think it is related to this PR.

I suggest moving this to the RTEMS Discourse site at https://users.rtems.org under [Applications / EPICS](https://users.rtems.org/c/apps/epics/26) category?

mdavidsaver:
@kiwichris You write quite an essay ;)  I will try to pick a couple of essential points to address.  Please feel free to call me out if I skip over something which you fell is essential.

I'll start with the notion of configuration files.

> From your ending comment I assume the hard coding reference is the configuration and leap second file fragments?
> ... It is the way this stuff works on RTEMS. ...

I am referring these these configuration files.  While this may be the way things are normally done with RTEMS in general, and I'm sure there are good reasons for this.  eg. your general case is likely very general.  However, for the much less general case of EPICS IOCs, I think there is a better way.

> ... These configuration files are no different to the files the systems integrators would use if running on Unix.

My objection is not to the content, but rather to the packaging.  imo. these stop being practical configuration files when the content is embedded in C code!

An analogy which might help you to understand what I see.  This PR seems to me like a Linux distributor providing a single pre-built initramfs image, expecting each installation to unpack that image, manually edit some files, then pack it back up again.  What I have in mind is analogous to the automation which avoids requiring end users to even know about initramfs.  Now I don't know if we will be able to get all the way to that point, but I think it is necessary to try.

I did not attempt to expand on these ideas in detail above because it will require more knowledge than I think is reasonable to ask of you ( @kiwichris ) about the EPICS build system, how the wider EPICS ecosystem is organized, and what our user's expectations are.

Touching on a couple of points:

1. Config files should appear in the source tree as individual files.
2. Where possible, epics-base should provide reasonable defaults.  BSP specific if necessary.
3. It should be possible for a support module or IOC application to replace some of these default files **without recompiling epics-base**

To give you an idea of where I think this design starts, I have encouraged @anjohnson (our Make maker) to think about how target filesystem fragments (including config files) should be placed into the installed tree of each EPICS module.

mdavidsaver:
> The monitoring and triggering of configuration states lets a user add an environment variable independent of the state of the NTP client and the client will start.

In the EPICS world, this change action should be handled with an IOC shell command.

Remember that **every IOC instance will have an IOC shell script.**  This is how an instance is "personalized".  So it is normal for an engineer setting up, or changing, an IOC to do so by issuing IOC shell commands.

So while RTEMS shell is mostly used (I guess) for diagnostics, the IOC shell is first and foremost **the central method for configuring an IOC instance.**  The many diagnostic tools are icing on that cake.  Understand this, and all that it implies, is one of the big mental hurdles everyone has to clear in understanding the way EPICS IOCs work in practice.  (so you are in good company if this seems foreign to you now)


> There is no `services` or `systemd` to manage that on RTEMS.

Clearly.  Otherwise you would not be trying to add a lite version to epics-base ;)

> I was not sure how EPICS on RTEMS universally initialized across all the supported BSPs and possible configuration methods.

There isn't a way.  The situation so far is imo. a mess of organic growth over many years.  I am looking to exploit the disruption which is already being caused by the RTEMS 4 -> 6 transition to sweep some of this mess away, or at least isolate it along with the "legacy" network stack code.

> When is a network interface available ...

Right, exactly my concern.  There will never be one answer to a question like this.  So my "answer" is to make it scriptable, and allow this scripting to be easily overridden/extended at various levels.

kiwichris:
> @kiwichris You write quite an essay ;) 

Yes it was a lot but I felt it easier to lay out things in the hope it saves time.

> I will try to pick a couple of essential points to address. Please feel free to call me out if I skip over something which you fell is essential.

I welcome your feedback because it makes me rethink how RTEMS integrates with users. The work in the PR is what is needed and that was goal. Your feedback highlights real issues and that is a good thing and it shows the work present needs to be broken up. I am happy to take this PR as a first pass base to work from.

> I'll start with the notion of configuration files.
> 
> > From your ending comment I assume the hard coding reference is the configuration and leap second file fragments?
> > ... It is the way this stuff works on RTEMS. ...
> 
> I am referring these these configuration files. While this may be the way things are normally done with RTEMS in general, and I'm sure there are good reasons for this. eg. your general case is likely very general. However, for the much less general case of EPICS IOCs, I think there is a better way.

Makes sense, thanks.

> > ... These configuration files are no different to the files the systems integrators would use if running on Unix.
> 
> My objection is not to the content, but rather to the packaging. imo. these stop being practical configuration files when the content is embedded in C code!

We could package them into EPCIS as you say but we need to preprocess the file to add the user's IP address from the boot environment when coping it to `/etc`. Does EPICS have support to take a file and expand env vars?

Another possibility is moving that code into RTEMS's net-services where the NTP code lives and handled there via an agreed interface work? This offers the benefit that each RTEMS release's config file format has to work and EPICS does not care.

> An analogy which might help you to understand what I see. This PR seems to me like a Linux distributor providing a single pre-built initramfs image, expecting each installation to unpack that image, manually edit some files, then pack it back up again. What I have in mind is analogous to the automation which avoids requiring end users to even know about initramfs. Now I don't know if we will be able to get all the way to that point, but I think it is necessary to try.

The analogy works. If the file is in the EPICS configuration data do you see a user updating a `make` variable to point to a user specific configuration file to override the default behavior?

> I did not attempt to expand on these ideas in detail above because it will require more knowledge than I think is reasonable to ask of you ( @kiwichris ) about the EPICS build system, how the wider EPICS ecosystem is organized, and what our user's expectations are.

Fair. I am learning and welcome any guidance or insights so please call out any issues you see.

> Touching on a couple of points:
> 
>     1. Config files should appear in the source tree as individual files.

Agreed.

>     2. Where possible, epics-base should provide reasonable defaults.  BSP specific if necessary.

OK

>     3. It should be possible for a support module or IOC application to replace some of these default files **without recompiling epics-base**

Agreed and nice.

> To give you an idea of where I think this design starts, I have encouraged @anjohnson (our Make maker) to think about how target filesystem fragments (including config files) should be placed into the installed tree of each EPICS module.

Sure and thanks.

kiwichris:
> > The monitoring and triggering of configuration states lets a user add an environment variable independent of the state of the NTP client and the client will start.
> 
> In the EPICS world, this change action should be handled with an IOC shell command.

Thanks and that makes sense. There is a complication here. The RTEMS net-services wrapper around the NTP code provides the ability to stop and restart the service. How well it works depends on the options provided to the client. The NTP code is old and its clean up code when closing down is questionable. It appears to rely on a process exiting to clean up. We have looked into dealing with this and in some cases it works but you enter a deep hole with no bottom when attempting changes.

As a result should we protect EPICS users and not allow runtime configuration changes? A change requires a reboot? Doing so removes the code.

> Remember that **every IOC instance will have an IOC shell script.** This is how an instance is "personalized". So it is normal for an engineer setting up, or changing, an IOC to do so by issuing IOC shell commands.

You see my embedded background conflicting here. This is a nice feature.

> So while RTEMS shell is mostly used (I guess) for diagnostics

The RTEMS shell is not a good example of a shell and yes it can run commands which is basically a hard coded text name to function caller. Having access via the EPICS `rt` command is an important diagnostic tool.

> the IOC shell is first and foremost **the central method for configuring an IOC instance.** The many diagnostic tools are icing on that cake. Understand this, and all that it implies, is one of the big mental hurdles everyone has to clear in understanding the way EPICS IOCs work in practice. (so you are in good company if this seems foreign to you now)

I am seeing it and starting to understand but I do not think I am there. I only use EPICS enough to deal with RTEMS related issues and it shows.

> > There is no `services` or `systemd` to manage that on RTEMS.
> 
> Clearly. Otherwise you would not be trying to add a lite version to epics-base ;)

Haha yeah lets not.

> > I was not sure how EPICS on RTEMS universally initialized across all the supported BSPs and possible configuration methods.
> 
> There isn't a way. The situation so far is imo. a mess of organic growth over many years. 

Agreed

> I am looking to exploit the disruption which is already being caused by the RTEMS 4 -> 6 transition to sweep some of this mess away, or at least isolate it along with the "legacy" network stack code.

Excellent. I am onboard with this.

> > When is a network interface available ...
> 
> Right, exactly my concern. There will never be one answer to a question like this. So my "answer" is to make it scriptable, and allow this scripting to be easily overridden/extended at various levels.

We saw the issue from the RTEMS perspective and added the net-services package as a place to smooth over the difference in the legacy, libbsd and even lwip stacks. I am happy to provide interfaces that can be used in a consistent way in EPICS. For example a call that probes an interface to see if it has an IP address. This is off topic for this PR and removal of the configuration probing code will be actioned.

One item remaining. I can move the monitoring code to RTEMS net-services and then EPCIS can selectively decide on monitoring NTP for synchronized time. This would let EPICS control the synchronized state of the clock source. Does this work for you?

mdavidsaver:
> ... Does EPICS have support to take a file and expand env vars?

Oh yes.  Several ways at build and runtime.

Most relevantly, the IOC shell expands environment variables as each line is executed.  This is why the notion of eg. pre-populating the environment from eeprom is a natural one.  Quite a lot can then be done with these values.

Here is one [intimidatingly complex](https://github.com/areaDetector/ADCore/blob/master/iocBoot/EXAMPLE_commonPlugins.cmd) example of a reusable IOC shell fragment.

kiwichris:
I have created an issue in RTEMS to address the status query and the C string parsing code : https://gitlab.rtems.org/rtems/pkg/rtems-net-services/-/issues/16

kiwichris:
MR [rtems-net-services/!28](https://gitlab.rtems.org/rtems/pkg/rtems-net-services/-/merge_requests/28) has been posted for review. The MR is on the development branch (RTEMS 7). Once merged an MR for the RTEMS 6 branch will be made and that will end up in RTEMS 6.2.

The change accesses the `ntpd` data directly while holding the `ntpd` lock. Direct access avoids the overhead of a UPD query packet to the loopback interface and the need for any C string parsing code. A call `rtems_ntpd_is_synchronized()` has been added and it is all an NTP client on RTEMS needs to let the EPICS clock server know the clock source is synchronized.

I will update this change to:

- Use the new RTEMS NTP call
- Support configuration files the EPICS way

kiwichris:
@mdavidsaver and @anjohnson Can I assume the user will handle the NTP client configuration files in their application? That is they will make a suitable `/etc/ntpd.conf` and `/etc/leapsecond` file using the IOC shell and what ever is in the app?

If there is nothing I can print an error?

There should be some default files in EPICS base and I can help but where to put them and how to handle them is beyond my knowledge.

I am still learning this area.

kiwichris:
I have completed the changes for RTEMS NTP for EPICS. They are on the [rtems-ntpd-7_0-with-5-and-6-v2](https://github.com/kiwichris/epics-base/tree/rtems-ntpd-7_0-with-5-and-6-v2) branch. I am waiting for the [RTEMS Net Services MR !28](https://gitlab.rtems.org/rtems/pkg/rtems-net-services/-/merge_requests/28) changes to be merged so a back port for RTEMS 6 can be made. Once made I bring the v2 patch here so this PR's history is maintained.

mrippa:
I have built and tested the latest branch https://github.com/kiwichris/epics-base/tree/rtems-ntpd-7_0-with-5-and-6-v2 on the mvme2700 at Gemini. This is built on [RTEMS 6.2-rc1](https://ftp.rtems.org/pub/rtems/releases/6/rc/6.2-rc1/).

This looks good and is locked with our TimeServer.

image

mrippa:
At Gemini we're looking forward to the seeing RTEMS 6.2 formally accepted in a future EPICS 7.x release. Is this PR ready for inclusion in the next EPICS Release?

hjunkes:
I have built and tested the latest branch https://github.com/kiwichris/epics-base/tree/rtems-ntpd-7_0-with-5-and-6-v2 on a beagleboneblack board with libbsd (including DHCP, NFSv4). 

To setup the rtems base was a little bit tricky:
rtems source builder and kernel from release 6.2-rc1
```
https://ftp.rtems.org/pub/rtems/releases/6/rc/6.2-rc1/sources/rtems-source-builder-6.2-rc1.tar.xz
https://ftp.rtems.org/pub/rtems/releases/6/rc/6.2-rc1/sources/rtems-6.2-rc1.tar.xz
```
libbsd 6-freebsd-12
```
git clone https://gitlab.rtems.org/rtems/pkg/rtems-libbsd
cd rtems-libbsd
git checkout 6-freebsd-12
```
and net services 6
```
git clone https://gitlab.rtems.org/rtems/pkg/rtems-net-services
cd rtems-net-services
git checkout 6
```
 
Works out and  is locked with our TimeServer: 
```
beagle-debug> NTPTime_Report
NTP driver is synchronized with server
Syncronization interval = 60.0 seconds
Last synchronized at 2025-09-09 12:14:55.438605
Current OS tick rate = 10000 Hz
Measured tick rate = 8696.435 Hz
RTEMS NTP Status:
 config type = NTPD_CONFIG_NTP_IP
 config ip = 141.14.138.238
 ntpd active = yes
 ntpd active sec = 633
 ntpq open = yes
 status last updated = 2025-09-09 12:15:34
 synchronized = yes
 status = (c618) leap_alarm, sync_ntp, 1 event, no_sys_peer
 version = ntpd 4.2.8p12
 processor = ARM/ARMv4
 system = RTEMS/6.2.0
 leap = 3
 stratum = 4
 precision = -19
 rootdelay = 8.043111
 rootdisp = 954.880043
 refid = 141.14.138.238
 reftime = 3966408609.2855234189
 clock = 3966408934.2321755104
 peer = 1833
 tc = 6
 mintc = 3
 offset = 0.000000
 frequency = 0.000000
 sys_jitter = 0.000000
 clk_jitter = 5.823318
 clk_wander = 0.000000
 tai = 0
 leapsec = 3692217600
 expire = 3896899200
Peers:
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
 141.14.138.238  .POOL.          16 p    -   64    0    0.000    0.000   0.002
*141.14.138.238  d484611a         3 u  326   64    1    0.276  -16.471  12.338
```
```
beagle-debug> ClockTime_Report
IOC is synchronizing OS Clock to a priority=100 provider
Initial sync was at 2025-09-09 12:10:19.465853
Last successful sync was at 2025-09-09 12:25:16.442594
IOC synchronization interval = 60 seconds
beagle-debug> date
2025/09/09 12:26:39.812920
```
```
beagle-debug> rt cpuuse
-------------------------------------------------------------------------------
                              CPU USAGE BY THREAD
------------+----------------------------------------+---------------+---------
 ID         | NAME                                   | SECONDS       | PERCENT
------------+----------------------------------------+---------------+---------
 0x09010001 | IDLE                                   |    663.086109 |  44.953
 0x0a010001 | TIME                                   |     11.934815 |   0.809
 0x0a010002 | IRQS                                   |      1.651950 |   0.111
 0x0a010003 | swi6: Giant tas                        |      0.000006 |   0.000
 0x0a010004 | config_0                               |      0.000012 |   0.000
 0x0a010005 | kqueue_ctx task                        |      0.000008 |   0.000
 0x0a010006 | swi6: task queu                        |      0.006575 |   0.000
 0x0a010007 | swi5: fast task                        |      0.000003 |   0.000
 0x0a010008 | thread taskq                           |      0.000005 |   0.000
 0x0a010009 | swi1: netisr 0                         |      0.009910 |   0.000
 0x0a01000a | usbus0                                 |      0.000007 |   0.000
 0x0a01000b | usbus0                                 |      0.000004 |   0.000
 0x0a01000c | usbus0                                 |      0.000004 |   0.000
 0x0a01000d | usbus0                                 |      0.030921 |   0.002
 0x0a01000e | usbus0                                 |      0.000009 |   0.000
 0x0a01000f | usbus1                                 |      0.000006 |   0.000
 0x0a010010 | usbus1                                 |      0.000004 |   0.000
 0x0a010011 | usbus1                                 |      0.000005 |   0.000
 0x0a010012 | usbus1                                 |      0.024098 |   0.001
 0x0a010013 | usbus1                                 |      0.000005 |   0.000
 0x0a010014 | bufdaemon                              |      0.076781 |   0.005
 0x0a010015 | vnlru                                  |      0.011266 |   0.000
 0x0a010016 | syncer                                 |      0.013352 |   0.000
 0x0a010017 | softirq_0                              |      0.000008 |   0.000
 0x0a010018 | DHCP                                   |      0.080277 |   0.005
 0x0a010019 | bufspacedaemon-                        |      0.009756 |   0.000
 0x0a01001a | nfscl                                  |      0.023497 |   0.001
 0x0a01001b | TNTa                                   |      0.000051 |   0.000
 0x0a01001c | TNTb                                   |      0.000032 |   0.000
 0x0a01001d | TNTc                                   |      0.000026 |   0.000
 0x0a01001e | TNTd                                   |      0.000027 |   0.000
 0x0a01001f | TNTe                                   |      0.000028 |   0.000
 0x0a010020 | TNTD                                   |      0.000028 |   0.000
 0x0b010001 | _main_                                 |      2.522759 |   0.171
 0x0b010002 | PVAS timers                            |      0.001644 |   0.000
 0x0b010003 | errlog                                 |      0.025618 |   0.001
 0x0b010004 | osdNTP_Runner                          |      0.012471 |   0.000
 0x0b010005 | osdNTP_Monitor                         |      0.101306 |   0.006
 0x0b010006 | NTPTimeSync                            |      0.001100 |   0.000
 0x0b010007 | ClockTimeSync                          |      0.002776 |   0.000
 0x0b010008 | taskwd                                 |      0.001869 |   0.000
 0x0b010009 | timerQueue                             |      0.037397 |   0.002
 0x0b01000a | cbLow                                  |      0.030809 |   0.002
 0x0b01000b | cbMedium                               |      0.000031 |   0.000
 0x0b01000c | cbHigh                                 |      0.000032 |   0.000
 0x0b01000d | dbCaLink                               |      0.212245 |   0.014
 0x0b01000e | PVAL                                   |      0.000023 |   0.000
 0x0b01000f | PDB-event                              |      0.000016 |   0.000
 0x0b010010 | pvAccess-client                        |    793.521242 |  53.783
 0x0b010011 | UDP-rx 0.0.0.0:                        |      0.000049 |   0.000
 0x0b010012 | UDP-rx 141.14.1                        |      0.000021 |   0.000
 0x0b010013 | UDP-rx 141.14.1                        |      0.052960 |   0.003
 0x0b010014 | UDP-rx 224.0.0.                        |      0.000018 |   0.000
 0x0b010015 | scanOnce                               |      0.000051 |   0.000
 0x0b010016 | scan-10                                |      0.001549 |   0.000
 0x0b010017 | scan-5                                 |      0.015400 |   0.001
 0x0b010018 | scan-2                                 |      0.038373 |   0.002
 0x0b010019 | scan-1                                 |      0.109191 |   0.007
 0x0b01001a | scan-0.5                               |      0.021854 |   0.001
 0x0b01001b | scan-0.2                               |      0.052750 |   0.003
 0x0b01001c | scan-0.1                               |      0.115112 |   0.007
 0x0b01001d | CAS-TCP                                |      0.003935 |   0.000
 0x0b01001e | CAS-UDP                                |      0.070716 |   0.004
 0x0b01001f | CAS-beacon                             |      0.005269 |   0.000
 0x0b010020 | CAC-event                              |      0.057317 |   0.003
 0x0b010021 | TCP-acceptor                           |      0.000069 |   0.000
 0x0b010022 | UDP-rx 0.0.0.0:                        |      0.000032 |   0.000
 0x0b010023 | UDP-rx 141.14.1                        |      0.000016 |   0.000
 0x0b010024 | UDP-rx 141.14.1                        |      1.253523 |   0.084
 0x0b010025 | UDP-rx 224.0.0.                        |      0.000017 |   0.000
 0x0b010026 |                                        |      0.034040 |   0.002
 0x0b010027 | CAS-event                              |      0.019673 |   0.001
 0x0b010028 | CAS-client                             |      0.004499 |   0.000
------------+----------------------------------------+---------------+---------
 TIME SINCE LAST CPU USAGE RESET IN SECONDS:                       1475.559511
-------------------------------------------------------------------------------
```

hjunkes:
Build on RTEMS 6.1 with libbsd for MVME2500 'qoriq_e500'. Works well, looks ok ...
```
brutus> NTPTime_Report
NTP driver is synchronized with server
Syncronization interval = 60.0 seconds
Last synchronized at 2025-09-12 15:37:29.660201
Current OS tick rate = 10000 Hz
Measured tick rate = 10000.001 Hz
RTEMS NTP Status:
 config type = NTPD_CONFIG_NTP_IP
 config ip = 141.14.138.238
 ntpd active = yes
 ntpd active sec = 151
 ntpq open = yes
 status last updated = 2025-09-12 15:38:01
 synchronized = yes
 status = (c618) leap_alarm, sync_ntp, 1 event, no_sys_peer
 version = ntpd 4.2.8p12
 processor = PowerPC/Generic (no FPU)
 system = RTEMS/6.0.0
 leap = 3
 stratum = 4
 precision = -18
 rootdelay = 8.256660
 rootdisp = 193.897003
 refid = 141.14.138.238
 reftime = 3966680188.2895003608
 clock = 3966680281.2849329016
 peer = 10687
 tc = 6
 mintc = 3
 offset = 0.000000
 frequency = 0.000000
 sys_jitter = 0.000000
 clk_jitter = 1.417666
 clk_wander = 0.000000
 tai = 0
 leapsec = 3692217600
 expire = 3896899200
Peers:
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
 141.14.138.238  .POOL.          16 p    -   64    0    0.000    0.000   0.004
*141.14.138.238  d484611a         3 u   93   64    1    0.581   -3.889   3.458
```
```
brutus> ClockTime_Report
IOC is synchronizing OS Clock to a priority=100 provider
Initial sync was at 2025-09-12 15:36:29.662206
Last successful sync was at 2025-09-12 15:38:29.662198
IOC synchronization interval = 60 seconds
```
```
brutus> rt cpuuse
-------------------------------------------------------------------------------
                              CPU USAGE BY THREAD
------------+----------------------------------------+---------------+---------
 ID         | NAME                                   | SECONDS       | PERCENT
------------+----------------------------------------+---------------+---------
 0x09010001 | IDLE                                   |     34.900560 |  12.780
 0x0a010001 | TIME                                   |      0.606558 |   0.222
 0x0a010002 | IRQS                                   |      0.096446 |   0.035
 0x0a010003 | kqueue_ctx task                        |      0.000009 |   0.000
 0x0a010004 | swi6: Giant tas                        |      0.000003 |   0.000
 0x0a010005 | config_0                               |      0.000004 |   0.000
 0x0a010006 | swi6: task queu                        |      0.000228 |   0.000
 0x0a010007 | swi5: fast task                        |      0.000002 |   0.000
 0x0a010008 | thread taskq                           |      0.000003 |   0.000
 0x0a010009 | swi1: netisr 0                         |      0.003165 |   0.001
 0x0a01000a | bufdaemon                              |      0.013516 |   0.004
 0x0a01000b | vnlru                                  |      0.001913 |   0.000
 0x0a01000c | syncer                                 |      0.001278 |   0.000
 0x0a01000d | softirq_0                              |      0.000002 |   0.000
 0x0a01000e | bufspacedaemon-                        |      0.000904 |   0.000
 0x0a01000f | DHCP                                   |      0.018529 |   0.006
 0x0a010010 | nfscl                                  |      0.002271 |   0.000
 0x0a010011 | TNTa                                   |      0.000030 |   0.000
 0x0a010012 | TNTb                                   |      0.000021 |   0.000
 0x0a010013 | TNTc                                   |      0.000022 |   0.000
 0x0a010014 | TNTd                                   |      0.000023 |   0.000
 0x0a010015 | TNTe                                   |      0.000022 |   0.000
 0x0a010016 | TNTD                                   |      0.000011 |   0.000
 0x0b010001 | _main_                                 |      0.272770 |   0.099
 0x0b010002 | PVAS timers                            |      0.000100 |   0.000
 0x0b010003 | errlog                                 |      0.000190 |   0.000
 0x0b010004 | osdNTP_Runner                          |      0.016979 |   0.006
 0x0b010005 | osdNTP_Monitor                         |      0.010215 |   0.003
 0x0b010006 | NTPTimeSync                            |      0.000129 |   0.000
 0x0b010007 | ClockTimeSync                          |      0.000413 |   0.000
 0x0b010008 | taskwd                                 |      0.000053 |   0.000
 0x0b010009 | timerQueue                             |      0.005591 |   0.002
 0x0b01000a | cbLow                                  |      0.003539 |   0.001
 0x0b01000b | cbMedium                               |      0.000018 |   0.000
 0x0b01000c | cbHigh                                 |      0.000019 |   0.000
 0x0b01000d | dbCaLink                               |      0.027510 |   0.010
 0x0b01000e | PVAL                                   |      0.000010 |   0.000
 0x0b01000f | PDB-event                              |      0.000009 |   0.000
 0x0b010010 | pvAccess-client                        |    236.860874 |  86.653
 0x0b010011 | UDP-rx 0.0.0.0:                        |      0.000020 |   0.000
 0x0b010012 | UDP-rx 141.14.1                        |      0.000011 |   0.000
 0x0b010013 | UDP-rx 141.14.1                        |      0.006493 |   0.002
 0x0b010014 | UDP-rx 224.0.0.                        |      0.000010 |   0.000
 0x0b010015 | scanOnce                               |      0.000025 |   0.000
 0x0b010016 | scan-10                                |      0.000299 |   0.000
 0x0b010017 | scan-5                                 |      0.001845 |   0.000
 0x0b010018 | scan-2                                 |      0.004135 |   0.001
 0x0b010019 | scan-1                                 |      0.013327 |   0.004
 0x0b01001a | scan-0.5                               |      0.004055 |   0.001
 0x0b01001b | scan-0.2                               |      0.011365 |   0.004
 0x0b01001c | scan-0.1                               |      0.023344 |   0.008
 0x0b01001d | CAS-TCP                                |      0.000042 |   0.000
 0x0b01001e | CAS-UDP                                |      0.002660 |   0.000
 0x0b01001f | CAS-beacon                             |      0.000441 |   0.000
 0x0b010020 | CAC-event                              |      0.005050 |   0.001
 0x0b010021 | TCP-acceptor                           |      0.000038 |   0.000
 0x0b010022 | UDP-rx 0.0.0.0:                        |      0.000014 |   0.000
 0x0b010023 | UDP-rx 141.14.1                        |      0.000010 |   0.000
 0x0b010024 | UDP-rx 141.14.1                        |      0.158843 |   0.058
 0x0b010025 | UDP-rx 224.0.0.                        |      0.000008 |   0.000
 0x0b010026 |                                        |      0.007417 |   0.002
------------+----------------------------------------+---------------+---------
 TIME SINCE LAST CPU USAGE RESET IN SECONDS:                        273.499060
-------------------------------------------------------------------------------
```

hjunkes:
Hi Chris,
I’m currently in the process of getting the BeagleBone Black up and running with RTEMS 7 (I had to tweak the network settings) and have come across NTP again.

One issue is that NTP synchronisation is enabled via a hook after IOCInit.

For example, when I use IOCstats, the PV 'IOC:M_BootTime' is set to the wrong time during IOCInit.

In other words, I need to set the time before calling `IOCInit`. I’ll do that now using a one-shot NTP read:

in rtems_init.c:

```
    fixup_hosts();

    /* - NTP time synchronisation before filesystem and iocInit -*/

    /* Step 1: one-shot SNTP to set OS clock immediately after DHCP */

    //struct timespec now;
    if (rtemsInit_NTP_server_ip[0] != '\0') {
        printf("\n***** One-shot NTP time set from %s *****\n",
               rtemsInit_NTP_server_ip);
        if (epicsNtpGetTime(rtemsInit_NTP_server_ip, &now) == 0) {
            if (clock_settime(CLOCK_REALTIME, &now) == 0) {
                char tbuf[32];
                strftime(tbuf, sizeof(tbuf), "%Y/%m/%d %H:%M:%S",
                         gmtime(&now.tv_sec));
                printf("***** Clock set to %s UTC *****\n", tbuf);
            } else {
                printf("WARNING: clock_settime failed: %s\n", strerror(errno));
            }
        } else {
            printf("WARNING: epicsNtpGetTime failed, clock not set\n");
        }
    }

    /* Step 2: initialise EPICS time provider now, with valid OS clock.
    NTPTime_Init uses epicsThreadOnce so the initHookAtBeginning call
    inside iocInit becomes a no-op. synchronized=1 from the start. */

    printf(" Initialising EPICS time provider...\n");
    osdTimeRegister();
    /* end NTP sync */

    printf("\n***** Setting up file system *****\n");
```

 I’ve also fixed a few issues in osiNTPTime.c (cosmetic?):
  
[osiNTPTime.patch](https://github.com/user-attachments/files/28479779/osiNTPTime.patch)

kiwichris:
> Hi Chris, I’m currently in the process of getting the BeagleBone Black up and running with RTEMS 7 (I had to tweak the network settings) and have come across NTP again.

Thanks for the testing and comments.

> One issue is that NTP synchronisation is enabled via a hook after IOCInit.

What do you mean via sychronisation is enabled via a hook?

> For example, when I use IOCstats, the PV 'IOC:M_BootTime' is set to the wrong time during IOCInit.

Yes this is something I see on my MVME5500 beatnik board. The start up date has a year of 2055 or something like that. Given RTEMS's initial time defaults to Jan 1 1988 something else is happening. I am wondering if an RTC driver is setting the time?

This raises a key question of when is the RTEMS arena considered operational enough for EPICS to run? I have not seen any documentation on this topic?

> In other words, I need to set the time before calling `IOCInit`. I’ll do that now using a one-shot NTP read:
> 
> in rtems_init.c:
> 
> ```
>     fixup_hosts();
> 
>     /* - NTP time synchronisation before filesystem and iocInit -*/
> 
>     /* Step 1: one-shot SNTP to set OS clock immediately after DHCP */
> 
>  [snip]
> 
>     printf("\n***** Setting up file system *****\n");
> ```

I think NTP has an update option which would be nice to have work but the NTP code when embedded in RTEMS has a [restart/rerun](https://gitlab.rtems.org/rtems/pkg/rtems-net-services/-/issues/14) issue we have not resolved. I think an SNTP solution is a good alternative.

It however again raises the question of how accurate the time needs to be when EPICS starts? NTP lets you get accurate time but it does take a bit of time to be synchronized.

> 
> I’ve also fixed a few issues in osiNTPTime.c (cosmetic?):
> 
> [osiNTPTime.patch](https://github.com/user-attachments/files/28479779/osiNTPTime.patch)

Thanks. I will be looking at this soon.

anjohnson:
> how accurate the time needs to be when EPICS starts?

Before the `iocInit` command in the startup script the IOC code doesn't really care much about the wall-clock time; during and after that the IOC uses time providers registered with the generalTime system to get the current time, and that prevents the IOC from seeing time going backwards (but doesn't stop forward jumps). Attempts to step time backwards will result in timers and periodic scans not triggering until reality catches up, and the timestamps given to records triggered by other means will similarly appear to be frozen.

Applications that need the OS time to have synchronized could register an initHook routine to delay the IOC initialization during iocInit, or can register an iocsh command to do the same thing that could be run earlier. The startup script can normally load DBD and DB files and initialize hardware and I/O before the clock is accurate.

Does NFS require the OS clock to be accurate for read-access to files? IOCs rarely write files before iocInit (some IOC applications might) so that's less likely to be a problem.

hjunkes:
In the current implementation, in modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp,
osdTimeRegister is launched when initHookAtBeginning is announced. So during IocInit?

```
static void ntpInitHook(initHookState state)
{
    switch (state) {
        case initHookAtBeginning:
            osdTimeRegister();
            break;
        default:
            break;
    }
}
```
NTP synchronisation will then also start, and this usually takes 60–80 seconds in the background until the clock is synchronised. In the meantime, the db files have been loaded and the PVs have been initialised and processed.
If I then use IocStats, for example, IOC:M_BootTime is set and then shows, say, 1 January 1990!? And this PV will not be corrected later once the clock has been synchronised.

kiwichris:
> In the current implementation, in modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp, osdTimeRegister is launched when initHookAtBeginning is announced. So during IocInit?

This was the previous behavior. It can be removed and command added?

> NTP synchronisation will then also start, and this usually takes 60–80 seconds in the background until the clock is synchronised. In the meantime, the db files have been loaded and the PVs have been initialised and processed. If I then use IocStats, for example, IOC:M_BootTime is set and then shows, say, 1 January 1990!? And this PV will not be corrected later once the clock has been synchronised.

I will add a IOC shell command to wait until synchronized with a timeout.

kiwichris:
I have added IOC shell commands:

1. `NTPTime_Start`
2. `NTPTime_SyncWait`

The `NTPTime_Start` command lets you start with an IP address or a configuration file. The `NTPTime_SyncWait` takes the number of seconds to wait and returns once NTP is synchronized.

This is on a MVME5500 NTP test IOC I have:

```
Amnesiac > NTPTime_Start config /opt/src/ntp/ntp-test/iocBoot/iocntp/ntp.conf
NTP: setenv /opt/src/ntp/ntp-test/iocBoot/iocntp/ntp.conf -> EPICS_TS_NTP_CONF_FILE
Amnesiac > NTPTime_SyncWait 20
RTEMS NTP: waiting 20 seconds for sync
NTP is synchronized
```

Note, by default the boot server IP address is used to be compatible so you need to make sure the environment variables `EPICS_TS_NTP_INET` and `EPICS_TS_NTP_CONF_FILE` are not set.

kiwichris:
I have changed the IOC shell register position to before `iocInit()` so you can have in an `st.cmd` that initializes `ntpd` with a site specific configuration then waits for the server to synchronize.  On my network this takes around 20 seconds:
```shell
#- SPDX-FileCopyrightText: 2001 Argonne National Laboratory
#-
#- SPDX-License-Identifier: EPICS

#- Example RTEMS ntpd startup script

< envPaths

NTPTime_Start("config", "/opt/src/noirlab/gemini/ntp/ntp-test/iocBoot/iocntp/ntp.conf")
NTPTime_SyncWait(30)

Debug_Start()

cd /opt/src/noirlab/gemini/ntp/ntp-test

## Register support components
dbLoadDatabase("dbd/ntp.dbd")
ntp_registerRecordDeviceDriver(pdbbase)

## Load record instances
#dbLoadTemplate("db/user.substitutions")
#dbLoadRecords("db/ntpVersion.db", "user=chris")
#dbLoadRecords("db/dbSubExample.db", "user=chris")

#- Set this to see messages from mySub
#-var mySubDebug 1

#- Run this to trace the stages of iocInit
#-traceIocInit

iocInit

## Start any sequence programs
#seq(sncExample, "user=chris")
```
The console log is:
```
EPICS RTEMS rtems-7.0.0 (PowerPC/Generic (classic FPU)/beatnik)
  Tools: 15.2.0 20250808 (RTEMS 7, Build 2026.06.12, RSB f2eb74759a8d5a313428459e31ca2a429f89c8ca, Newlib 7d4336cf)
  Cores: 1

] Init (15,1/11,enabled): test.ntp
] Init (200,2/11,enabled): system.log.reset
Startup
] Init (210,3/11,enabled): system.filesys.root
File System: root set up
] Init (250,4/11,enabled): system.filesys.local
] Init (500,5/11,enabled): system.net
nexus0: <RTEMS Nexus device>
pcib0 pcibus 0
pci0: <PCI bus> on pcib0
pci0: <memory> at device 0.0 (no driver attached)
pci0: <memory> at device 0.1 (no driver attached)
pci0: <memory> at device 0.2 (no driver attached)
pci0: <memory> at device 0.3 (no driver attached)
pci0: <memory> at device 0.4 (no driver attached)
pci0: <memory> at device 0.5 (no driver attached)
pci0: <memory> at device 0.6 (no driver attached)
pci0: <memory> at device 0.7 (no driver attached)
pcib1: <PCI-PCI bridge> at device 10.0 on pci0
pci1: <PCI bus> on pcib1
pci1: <bridge> at device 0.0 (no driver attached)
pcib2 pcibus 2
pci2: <PCI bus> on pcib2
pci2: <memory> at device 0.0 (no driver attached)
pci2: <memory> at device 0.1 (no driver attached)
pci2: <memory> at device 0.2 (no driver attached)
pci2: <memory> at device 0.3 (no driver attached)
pci2: <memory> at device 0.4 (no driver attached)
pci2: <memory> at device 0.5 (no driver attached)
pci2: <memory> at device 0.6 (no driver attached)
pci2: <memory> at device 0.7 (no driver attached)
em0: <Intel(R) Legacy PRO/1000 XT 82544EI (Copper)> port 0xf0810000-0xf081001f mem 0xe1100000-0xe111ffff,0xe1120000-0xe113ffff irq 84 at device 10.0 on pci2
em0: Using 1024 TX descriptors and 1024 RX descriptors
info: em0: Ethernet address: 00:01:af:31:a8:5d
cpu0: <legacy CPU>
Setting hostname: Amnesiac.
Starting network:  lo0 info: lo0: link state changed to UP
em0.
 em0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=48525bb<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,TSO4,LRO,WOL_MAGIC,VLAN_HWFILTER,VLAN_HWTSO,NOMAP>
        ether 00:01:af:31:a8:5d
        inet 10.10.5.114 netmask 0xffffff00 broadcast 10.10.5.255
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
        media: Ethernet autoselect
        status: no carrier
        inet6 ::1 prefixlen 128
        inet 127.0.0.1 netmask 0xff000000
        groups: lo
add net default: gateway 10.10.5.1
em0: waiting for link (timeout 20s)... info: em0: link state changed to UP
up
] Init (600,6/11,enabled): system.nfs
mount: nfs: 10.10.5.4:/opt/src -> /opt/src
nfs: mount: nfs -> 10.10.5.4:/opt/src (nfsv4,minorversion=1)
nfs: mount args: 8
  16 addr=10 02 08 01 0a 0a 05 04 00 00 00 00 00 00 00 00
   0 nfsv4
   9 dirpath=/opt/src
   4 fstype=nfs
   7 fspath=/src-0
  19 hostname=10.10.5.4:/opt/src
   0 rw
   2 minorversion=1
nfs: mount: (0) Success
] Init (840,7/11,enabled): system.debugger
] Init (850,8/11,disabled): system.ntp.ip
] Init (900,9/11,enabled): system.cmds
RTEMS Commands registered
] Init (999,10/11,enabled): system.log.err.flush
] Init (1000,11/11,enabled): system.ioc
IOC work directory: "/opt/src/noirlab/gemini/ntp/ntp-test/iocBoot/iocntp/"

IOC Command line: /rtems/app.boot /opt/src/noirlab/gemini/ntp/ntp-test/iocBoot/iocntp/st.cmd

< envPaths
epicsEnvSet("IOC","iocntp")
epicsEnvSet("TOP","/opt/src/noirlab/gemini/ntp/ntp-test")
epicsEnvSet("EPICS_BASE","/opt/src/noirlab/gemini/epics-base/7-libbsd")
#epicsEnvSet("EPICS_TS_NTP_CONF_FILE", "/opt/src/noirlab/gemini/ntp/ntp-test/iocBoot/iocntp/ntp.conf");
NTPTime_Start("config", "/opt/src/noirlab/gemini/ntp/ntp-test/iocBoot/iocntp/ntp.conf")
NTP: setenv /opt/src/noirlab/gemini/ntp/ntp-test/iocBoot/iocntp/ntp.conf -> EPICS_TS_NTP_CONF_FILE
NTPTime_SyncWait(30)
RTEMS NTP: waiting 30 seconds for sync
NTP is sync[CPU:0] rtems-db: remote running
[CPU:0] rtems-db: tcp remote: listing on port: 1122
hronized
Debug_Start()
Starting debugger: tcp:1122
cd /opt/src/noirlab/gemini/ntp/ntp-test
## Register support components
dbLoadDatabase("dbd/ntp.dbd")
ntp_registerRecordDeviceDriver(pdbbase)
## Load record instances
#dbLoadTemplate("db/user.substitutions")
#dbLoadRecords("db/ntpVersion.db", "user=chris")
#dbLoadRecords("db/dbSubExample.db", "user=chris")
iocInit
Starting iocInit
## EPICS R7.0.10.1-DEV
## Rev. 71751648d10571c6c91d-dirty
## Rev. Date Git: 2026-07-09 18:32:00 +1000
## Start any sequence programs
#seq(sncExample, "user=chris")
Amnesiac >
```
I am testing this with the RTEMS C++ initialization support.

hjunkes:
I’d actually tried that approach before. That didn’t go down well with the people around us ;-)
The HW-IOCs are expected to reboot immediately, with no delay. Above all, they should not be dependent on a network service such as NTP. I think the approach of querying the NTP server directly (without a delay) just once at the start is more suitable in our environment.

anjohnson:
Hi @kiwichris 
1. Is the `NTPTime_SyncWait` command optional in the st.cmd file, and
2. Does the call to `epicsNtpGetTime(rtemsInit_NTP_server_ip, &now)` in `rtems_init.c::POSIX_Init()` still jump-set the clock, assuming the NTP server responds to the request?

If so, it sounds like @hjunkes just needs to omit the `NTPTime_SyncWait` command on his systems, which should start with a rough time sync and get more accurate once ntpd has its engine up and running.

kiwichris:
>     1. Is the `NTPTime_SyncWait` command optional in the st.cmd file, and

Yes it is optional. You need to add the command to have it run. It is a system design decision to use. The command takes the number of seconds to wait if you use it so it does not stop booting if the time cannot be synchronized within your set timeout period.

>     2. Does the call to `epicsNtpGetTime(rtemsInit_NTP_server_ip, &now)` in `rtems_init.c::POSIX_Init()` still jump-set the clock, assuming the NTP server responds to the request?

It should. I have not touched that code. I did not see it as part of this work as this PR is about adding `ntpd` and management of it. It would be nice  to use `ntpd` and the NTP update command to do this however the NTP options management code does not like to be reinitialized. We have an issue in our GitLab for it. I do not have any time allocated to fixing this.

I see this type of functionality as an initialization function and I can look at a command to the RTEMS C++ init work if you want to have a command to set the time once by a SNTP type function. I saw the C++ init PR as being a framework we can add these types of base modules to. I also see some systems using RTC hardware if available.

> If so, it sounds like @hjunkes just needs to omit the `NTPTime_SyncWait` command on his systems, which should start with a rough time sync and get more accurate once ntpd has its engine up and running.

Correct.

hjunkes:
ok, tested:
osdTimeRegister() alone (no one-shot pre-sync) does step the clock roughly correct on its own, then afterward NTPTimeSync()
  (osiNTPTime.c:157, polling every 60s) keeps refining/disciplining it.
  
  Update : https://github.com/hjunkes/RTEMS_BBB/commit/9e4e817c2b748b92e73759fc433100666c7e5930

kiwichris:
Added `NTPTime_Update` command. The following line in an `st.cmd` script:
```
NTPTime_Update("${RTEMS_NET_NTP_IP}")
```
results in:
```
NTPTime_Update("10.10.5.2")
NTP update: time is 08/19/26 01:59:00.011927552 UTC
```
This command can be used before `NTPUpdate_Start` to update the RTEMS clock to the NTP time while the started `ntpd` works to synchronize the clock. For example:
```
NTPTime_Update("${RTEMS_NET_NTP_IP}")
NTPTime_Start("ip", "${RTEMS_NET_NTP_IP}")
```

kiwichris:
Added `528.md` release note.

configure/os/CONFIG.Common.RTEMS
```diff
@@ -140,7 +140,12 @@ OP_SYS_CFLAGS += $(OP_SYS_CFLAGS_NET_$(RTEMS_NETWORKING))
 POSIX_CPPFLAGS_posix = -D_GNU_SOURCE -D_DEFAULT_SOURCE
 POSIX_CPPFLAGS = $(POSIX_CPPFLAGS_$(OS_API))
 
+ifneq ($(RTEMS_VERSION),5)
+LDLIBS_posix_NET_ntp = -lntp
```
thelunatic, line 144: As we discussed, it would be preferable if this is a switchable option and `-lntp` is not a mandatory dependency for every build

modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.c
```diff
@@ -0,0 +1,90 @@
...
+  packet.txTm_f = ntohl( packet.txTm_f ); // Time-stamp fraction of a second.
+
+  // Extract the 32 bits that represent the time-stamp seconds (since NTP epoch) from when the packet left the server.
+  // Subtract 70 years worth of seconds from the seconds since 1900.
+  // This leaves the seconds since the UNIX epoch of 1970.
+  // (1900)------------------(1970)**************************************(Time Packet Left the Server)
+  time_t txTm = ( time_t ) ( packet.txTm_s - NTP_TIMESTAMP_DELTA );
+  now->tv_sec = txTm;
+  return 0;
+}
```
thelunatic, line 90: This file is mostly similar to the epicsNtp.c file already in the tree. However, I see that your version fixes a few pre-existing bugs with that file. I believe the cleanest way forward is to replace the epicsNtp.c with this version of the epicsSimpleNtp.c in the common path of `modules/libcom/RTEMS`.

The name change from `epicsNtpGetTime` to `epicsSimpleNtpGetTime` is also a nice change and will help differentiate between the actual NTP vs simple NTP time.

modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.h
```diff
@@ -0,0 +1,50 @@
...
+#endif
+
+#define NTP_TIMESTAMP_DELTA 2208988800ull
+
+int epicsSimpleNtpGetTime(char *ntpIp, struct timespec *now);
+
+typedef struct
+{
+
+  uint8_t li_vn_mode __attribute__((unused));      // Eight bits. li, vn, and mode.
```
thelunatic, line 20: Are all elements of this struct unused? This one is likely added to make codacy happy. I wonder if all should be marked unused 🤔

modules/libcom/src/osi/os/RTEMS-posix/epicsSimpleNtp.h
```diff
@@ -0,0 +1,50 @@
...
+
+  uint32_t txTm_s __attribute__((unused));         // 32 bits and the most important field the client cares about. Transmit time-stamp seconds.
+  uint32_t txTm_f __attribute__((unused));         // 32 bits. Transmit time-stamp fraction of a second.
+
+} ntp_packet;              // Total: 384 bits or 48 bytes.
+
+#ifdef __cplusplus
+}
+#endif
+#endif // EPICSSIMPLENTP_H
```
thelunatic, line 50: Same comment as the `epicsSimpleNtp.c` file. Can this replace the common epicsNtp.h, which is basically the simple NTP?

modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp
```diff
@@ -0,0 +1,749 @@
...
+#include "epicsSimpleNtp.h"
+
+#include <iocsh/initHooks.h>
+
+#include <rtems.h>
+
+/*
+ * This file is built automatically by EPICS's build system
+ */
+#if __RTEMS_MAJOR__ < 6
```
thelunatic, line 42: Somewhere around here, maybe we can use the ntpd switch to allow users to decide what they want. Something like `-DHAVE_NTPD` or `-DHAVE_SIMPLE_NTP` might be a cleaner approach

modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp
```diff
@@ -0,0 +1,749 @@
...
+
+    osdNTP_Set_Synchronized(0);
+
+    taskwdRemove(0);
+}
+
+void osdTimeRegister(void)
+{
+    /* Init NTP first so it can be used to sync ClockTime */
+    NTPTime_Init(100);
```
thelunatic, line 485: This one initializes the coarse 60 second update clock. I think if a user is opting for NTPD, this one should be excluded (or modified) for real ntp. In general, it would be better if the ntpd and simple ntp branches are completely separate.

modules/libcom/src/osi/os/RTEMS-posix/osdTime.cpp
```diff
@@ -0,0 +1,749 @@
...
+    ClockTime_Init(CLOCKTIME_SYNC);
+
+    osdMonotonicInit();
+}
+
+int osdNTPGet(struct timespec *ts)
+{
+    int r;
+    if (osdNTP_Synchronized()) {
+        r = clock_gettime(CLOCK_REALTIME, ts);
```
thelunatic, line 495: Is it going to update the OS clock every 60 seconds? I think this one is coming from the prior existing simpleNTP coarse clock in `RTEMS-score/osdTime.cpp`. Perhaps this needs some updating too, to further separate the simple NTP and NTP. As we discussed before, a user should have NTP or simple NTP explicitly mentioned so they know what they are getting, instead of assuming that RTEMS (and hence EPICS + RTEMS) does not have actual NTP support.

Also, as line `44` of this file says. Better ideas welcome :)

modules/libcom/src/osi/Makefile
```diff
@@ -82,7 +82,7 @@ Com_SRCS += epicsGeneralTime.c
 # Time providers
 Com_SRCS += osiClockTime.c
 Com_SRCS_vxWorks += osiNTPTime.c tz2timezone.c
-Com_SRCS_RTEMS += osiNTPTime.c
+Com_SRCS_RTEMS += osiNTPTime.c epicsSimpleNtp.c
```
thelunatic, line 85: This will break the build for `RTEMS-score`. I am not sure what a better solution would be here, if we want to remove the rtems-score branch either way. But as of now, this would break some builds and a easier approach might be to tie this include to the same switch that allows user to pick NTP

thelunatic (commented):
Hi @kiwichris. I added a few comments, mainly relating to adding clear distinction between the simple ntp and actual ntp to avoid any future confusions.

anjohnson:
I have a suggestion that I think will help clean up this code and separate it from the built-in last-resort time provider osiClockTime.c: I would like to see NTPD for RTEMS as a separate stand-alone GeneralTime current time provider with its own `ntpdTimeGetCurrent()` routine, using say priority 90 (smaller numbers are higher priority). Its routine may be identical to `osdTimeGetCurrent()` from the posix/osdTime.cpp code if that's appropriate, but it would be registered separately with its own provider name and priority. The code can remain inside EPICS Base, but I do want to separate it from the other providers.

Currently this NTPD code is using the osiClockTime provider which registers itself at the `LAST_RESORT_PRIORITY` (999), and that makes it impossible to add any other providers that might be used as a fall-back from NTPD if the NTP servers all died (unlikely I know, but that's how generalTime was designed to work).

Disabling the call to `NTPTime_Init()` as @thelunatic says would ensure that the old time-sync system doesn't get started when NTPD is being used, but when NTPD isn't wanted (say for someone who's quickly trying out EPICS on RTEMS and doesn't want to have to build libntp) I'd like it to be available (although it might not be started automatically).

Vijay tells me your time availability is limited, so I'm hoping we can help with this restructuring to preserve your time as much as possible.

PR #530, opened 2024-08-06
use -nologo with rc like with many other MSVC commands

dirk-zimoch:
Just reduce the build noise a bit.


mdavidsaver:
This change seems innocuous.  I'm inclined to merge as soon as appveyor (successfully) completes.

ralphlange:
Agree. Approve.

anjohnson:
Agreed, as long all the older versions of the rc tool support it.

PR #535, opened 2024-08-12
fix string to epicsUInt32 conversion via double

dirk-zimoch:
Some compilers (e.g. clang 17) show warnings like this on 64 bit architectures:
```
../db/dbConstLink.c:97:21: warning: implicit conversion from 'unsigned long' to 'double' changes value from 18446744073709551615 to 18446744073709551616 [-Wimplicit-const-int-float-conversion]
   97 |             dval <= ULONG_MAX)
      |                  ~~ ^~~~~~~~~
/usr/bin/../lib/clang/17/include/limits.h:61:37: note: expanded from macro 'ULONG_MAX'
   61 | #define ULONG_MAX (__LONG_MAX__ *2UL+1UL)
      |                    ~~~~~~~~~~~~~~~~~^~~~

```
The reason is that on most 64 bit architectures (except Windows), `unsigned long` and thus `ULONG_MAX` has 64  bits  but a `double` has less that 64 bits mantissa. Thus, converting `unsigned long` to `double` for the comparison may lead to unexpected results when the value uses too many bits (as `ULONG_MAX` does).
But I think the use of `ULONG_MAX` is wrong in this context anyway because, despite some of their names, the affected functions want to convert a string to an `epicsUInt32` and not to an `unsigned long` and `epicsUInt32` is defined as `unsigned int` in epicsTypes.h. Thus, the value should be tested against `UINT_MAX` instead (which does not cause compiler warnings).

BTW: Any reason that only the conversions to `epicsUInt32` have a code path that uses `epicsParseFloat64` and thus accept exponential notation but none of the conversion to other integer types like `epicsInt32`, `epicsUInt16`, `epicsUInt64` etc?

### DIFF
--- a/modules/database/src/ioc/db/dbConstLink.c
+++ b/modules/database/src/ioc/db/dbConstLink.c
@@ -94,7 +94,7 @@ static long cvt_st_UInt32(const char *from, void *pfield, const dbAddr *paddr)
         status = epicsParseFloat64(from, &dval, &end);
         if (!status &&
             dval >=0 &&
-            dval <= ULONG_MAX)
+            dval <= UINT_MAX)
             *to = dval;
     }
     return status;
--- a/modules/database/src/ioc/db/dbConvert.c
+++ b/modules/database/src/ioc/db/dbConvert.c
@@ -302,7 +302,7 @@ static long getStringUlong(const dbAddr *paddr,
                 epicsFloat64 dval;
 
                 status = epicsParseFloat64(psrc, &dval, &end);
-                if (!status && 0 <= dval && dval <= ULONG_MAX)
+                if (!status && 0 <= dval && dval <= UINT_MAX)
                     *pdst = dval;
             }
             if (status)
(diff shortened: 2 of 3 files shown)


mdavidsaver (approved):
I think usage of `UINT_MAX` is correct for `epicsUInt32`.

https://github.com/epics-base/epics-base/blob/86cdfc596f402657401b07c3a38493159022d023/modules/libcom/src/misc/epicsTypes.h#L49

anjohnson:
I agree with the `ULONG_MAX` to `UINT_MAX` changes, they appear to have been a thinko on my part.

As to why `DBF_ULONG` fields can accept exponential notation, an old comment from Marty explained:
```C
  /*Problem was old database access said to get unsigned long as double*/
```
which I apparently edited to the current:
```C
        /*
         * Convert via double so numbers like 1.0e3 convert properly.
         * db_access pretends unsigned long is double.
         */
```
That refers to the way that RSRV presents a `DBF_ULONG` field as a double since CA doesn't have an unsigned 32-bit integer type and a double is the only way for it to correctly return integer values larger then `2**31-1`.

My guess is that when a display manager text editor widget fetched `DBF_ULONG` fields it would get an exponent in the string representation of numbers above 99999:
```
% perl -e 'printf "%g, %g\n",999999, 1000000'
999999, 1e+06
```
That string couldn't be written directly back to a `DBF_ULONG`, so the user would have to retype the number in their text input box without the exponent to submit a new value.

I don't think we want to allow that for the smaller integer types. Maybe there's a case for the int64 types though, since CA also represents them as a double.

anjohnson:
Core group approved.

PR #536, opened 2024-08-13
Aliases

dirk-zimoch:
Allow to redefine the same alias for the same record.

Normally, loading the same record (db file) twice is allowed (and is a no-op), but that fails if the db file contains aliases. This patch allows it even with aliases.

Also if creating the alias fails because the name already exists (as a record or an alias for a different record), the error message is more specific about this.

My use case: I have a template file for a multi channel device with many records per channel and a few per device. When I load the template multiple times with different macro values per channel but for the same device, the per-device records are re-loaded multiple times -- with the same values, which is no problem. But having per-device aliases fails in this case.

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -22,6 +22,13 @@ should also be read to understand what has changed since earlier releases:
 
 ## Changes made on the 7.0 branch since 7.0.8.1
 
+### Allow to load the same alias multiple times
+
+Aliases can now be defined multiple times as long as they still refer to the
+same record, unless the shell variable dbRecordsOnceOnly is set.
+This allows to load database files multiple times, even if they contain
+alias definitions.
+
 ### DBE_PROPERTY event rate changed
 
 Updating property fields now only post DBE_PROPERTY events if the
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1263,24 +1263,50 @@ static void dbRecordInfo(char *name, char *value)
     }
 }
 
+static long createAlias(DBENTRY *pdbentry, const char *alias)
+{
+    DBENTRY tempEntry;
+    long status;
+    dbRecordNode *precnode = pdbentry->precnode;
+
+    if (precnode->aliasedRecnode) precnode = precnode->aliasedRecnode;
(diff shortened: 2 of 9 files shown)


anjohnson:
I load per-device records from a different .db file and then set fields in the device records and create aliases for them in the per-channel .db file(s) loaded later. However your approach is fine too.

So this looks fine to me in principle, but would you please add a couple of unit tests to `modules/database/test/ioc/dbStaticTest.c` where we already have some alias testing going on. I'll also want a Release Note entry before merging (so in the future users can easily see when doing this became legal).

Thanks!

mdavidsaver:
> Normally, loading the same record (db file) twice is allowed ...

Note that setting

```
var dbRecordsOnceOnly 1
```

Changes the default to disallow record redefinition.

Should this be relevant to aliases?

dirk-zimoch:
Applying `dbRecordsOnceOnly` to aliases too makes sense to me.

dirk-zimoch:
How can I access a shell variable (here: `dbRecordsOnceOnly`) in the tests? Simply declaring it `extern` and using it as a C variable fails on all Windows builds.

dirk-zimoch:
> How can I access a shell variable (here: dbRecordsOnceOnly) in the tests? 

I use `iocshFindVariable` now.

anjohnson:
Re `dbRecordsOnceOnly` welcome to the wonders of using Windows DLLs! The definition in dbLexRoutines.c would have to be marked with `DBCORE_API` to export it from the DLL, and your test program would need a similar declaration marking it as to be imported from a DLL. Using `iocshFindVariable` is probably a simpler solution.

mdavidsaver:
> I use `iocshFindVariable` now.

Probably simplest to call through the shell.

```c
iocshCmd("var dbRecordsOnceOnly 1");
```

dirk-zimoch:
> Probably simplest to call through the shell.
> `iocshCmd("var dbRecordsOnceOnly 1");`

Works. I have changed it.

anjohnson (approved):
Seems fine to me, thanks!

anjohnson:
Core Group: Needs rebase to merge Release Notes, then can be merged.

dirk-zimoch:
rebased to resolve conflict in RELEASE_NOTES.md

mdavidsaver:
Merged as 72026a27a04ec914cb70c4c3caff0224492a96b4

PR #539, opened 2024-08-26
genVersionHeader: work with git submodules and worktrees.

ericonr:
When the epics-base directory is used as a git submodule, or checked out in multiple places as a git worktree, ".git" is not a directory, but a file which points to the location of the git directory. In that case, the git tools can still be used to determine the version information.

Without this change, the version string is determined from the build date/time, which has multiple issues:

- it doesn't include the full information about the build environment
- it is not reproducible
- it leads to unnecessary rebuilds, since a lot of files depend on "epicsVCS.h"

### DIFF
--- a/src/tools/genVersionHeader.pl
+++ b/src/tools/genVersionHeader.pl
@@ -86,7 +86,7 @@ elsif (-d '.hg') { # Mercurial
     }
     $cv = `hg log -l1 --template '{date|isodate}'`
 }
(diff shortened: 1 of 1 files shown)


anjohnson (approved):
Approved, the script was written before git worktrees were a thing. Some CI checks are failing for reasons unconnected to this PR; I'll try to re-run them after the Win32 job stops.

PR #540, opened 2024-08-27
decorate functions that do not return

dirk-zimoch:
I have decorated most of the functions that gcc found with the option `-Wsuggest-attribute=noreturn`.
Exceptions: `cast_server()` and `rsrv_online_notify_task()`.

### DIFF
--- a/modules/ca/src/client/acctst.c
+++ b/modules/ca/src/client/acctst.c
@@ -3396,7 +3396,7 @@ void verifyContextRundownChanStillExist (
     showProgressEnd ( interestLevel );
 }
 
-int acctst ( const char * pName, unsigned interestLevel, unsigned channelCount,
+void acctst ( const char * pName, unsigned interestLevel, unsigned channelCount,
             unsigned repetitionCount, enum ca_preemptive_callback_select select )
 {
     chid chan;
@@ -3549,8 +3549,6 @@ int acctst ( const char * pName, unsigned interestLevel, unsigned channelCount,
     printf ( "\nTest Complete\n" );
 
     epicsExit ( EXIT_SUCCESS );
-
-    return 0;
 }
 
 
--- a/modules/libcom/src/flex/flexdef.h
+++ b/modules/libcom/src/flex/flexdef.h
@@ -696,16 +696,16 @@ extern void dataend (void);
 extern void flexerror (char[]) NORETURN;
 
 /* report a fatal error message and terminate */
-extern void flexfatal (char[]);
+extern void flexfatal (char[]) NORETURN;
 
 /* return current time */
 extern char *flex_gettime();
 
 /* report an error message formatted with one integer argument */
-extern void lerrif (char[], int);
+extern void lerrif (char[], int) NORETURN;
 
 /* report an error message formatted with one string argument */
-extern void lerrsf (char[], char[]);
+extern void lerrsf (char[], char[]) NORETURN;
 
 /* spit out a "# line" statement */
 extern void line_directive_out (FILE*);
--- a/modules/libcom/src/osi/epicsThread.h
+++ b/modules/libcom/src/osi/epicsThread.h
@@ -437,6 +437,7 @@ private:
     epicsThread ( const epicsThread & );
     epicsThread & operator = ( const epicsThread & );
     friend void epicsThreadCallEntryPoint ( void * );
+    EPICS_NORETURN
     void printLastChanceExceptionMessage (
         const char * pExceptionTypeName,
         const char * pExceptionContext );
--- a/modules/libcom/src/yacc/defs.h
+++ b/modules/libcom/src/yacc/defs.h
@@ -297,7 +297,7 @@ extern void tokenized_start(char *s) NORETURN;
 extern void retyped_warning(char *s);
 extern void reprec_warning(char *s);
 extern void revalued_warning(char *s);
-extern void terminal_start(char *s);
+extern void terminal_start(char *s) NORETURN;
 extern void restarted_warning(void);
 extern void no_grammar(void) NORETURN;
 extern void terminal_lhs(int s_lineno) NORETURN;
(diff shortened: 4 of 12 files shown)


modules/ca/src/client/acctst.c
```diff
@@ -3549,8 +3549,6 @@ int acctst ( const char * pName, unsigned interestLevel, unsigned channelCount,
     printf ( "\nTest Complete\n" );
 
     epicsExit ( EXIT_SUCCESS );
-
-    return 0;
```
dirk-zimoch, line 3553: The compiler complains about returning once `acctst()` is maked "noreturn".

modules/libcom/src/flex/flexdef.h
```diff
@@ -696,16 +696,16 @@ extern void dataend (void);
 extern void flexerror (char[]) NORETURN;
```
dirk-zimoch, line 696: This one already had `NORETURN`. So I used it for the other functions in this file as well. Replace with `EPICS_NORETURN` to add the hints for Windows too?

modules/libcom/src/yacc/defs.h
```diff
@@ -297,7 +297,7 @@ extern void tokenized_start(char *s) NORETURN;
 extern void retyped_warning(char *s);
 extern void reprec_warning(char *s);
 extern void revalued_warning(char *s);
-extern void terminal_start(char *s);
+extern void terminal_start(char *s) NORETURN;
```
dirk-zimoch, line 300: Replace with `EPICS_NORETURN` here and for the other functions in this file?

dirk-zimoch:
Rebased and added `epicsThread :: printLastChanceExceptionMessage()` which has recently changed.

mdavidsaver:
Can the function type `EPICS_NORETURN(...)` be avoided?  I find declaration wrappers harder to read.

Testing with https://godbolt.org/ the following compiles on gcc/clang/msvc.  It seems to produce the desired effect with gcc/clang as the second print is omitted, even at `-O0`.

```c
#include <stdio.h>
#include <string>

#ifdef _MSC_VER
int __declspec(noreturn) blah();
#else
int __attribute__((noreturn)) blah();
#endif

void foo(void)
{
    printf("xxx\n");
    blah();
    printf("yyy\n");
}
```

mdavidsaver:
... it looks like gcc also accepts

```c
__attribute__((noreturn)) int blah();
```

with the same effect.

This led me to the [gcc manual](https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html), which I find unclear.  I guess we are seeing the effects of `... laxity is allowed in the placing of attributes.  ...`?

dirk-zimoch:
If both, cl and gcc can work with prefixes, that will make it much easier. I will re-write it.

anjohnson (approved):
Looks much cleaner, thanks. I don't think replacing the `NORETURN` macros in flex and antelope is worth it.

dirk-zimoch:
We have the choice to put `EPICS_NORETURN` before or after the return type of the function. Both works. I did it consistently before now (like boost, btw).

modules/ca/src/client/acctst.c
mdavidsaver: The last MSVC 2019 build warns about `function declared with 'noreturn' has non-void return type`.

```
2025-02-28T16:10:29.0630831Z D:\a\epics-base\epics-base\modules\ca\src\client\acctst.c(3401): warning C4646: function declared with 'noreturn' has non-void return type
```

One goes down, another pops up...

modules/libcom/src/osi/epicsThread.h
```diff
@@ -437,7 +437,7 @@ class LIBCOM_API epicsThread {
     epicsThread ( const epicsThread & );
     epicsThread & operator = ( const epicsThread & );
     friend void epicsThreadCallEntryPoint ( void * );
-    void printLastChanceExceptionMessage (
+    EPICS_NORETURN void printLastChanceExceptionMessage (
```
mdavidsaver: I have a slight preference for adding decorators macros to the previous line.  Partly for (imo.) readability, and partly for `git-blame`.

```c
EPICS_NORETURN
void printLastChanceExceptionMessage (
```

mdavidsaver (approved):
This PR meets my "good enough for now" criteria.

modules/libcom/src/osi/epicsThread.h
anjohnson: I second that preference. Should it be mentioned in a comment in modules/libcom/src/osi/compilerDependencies.h? That file looks like it has Doxygen comments even though in reality only the file header is one so the file does get processed. We could start them with this macro…

modules/ca/src/client/acctst.c
dirk-zimoch: The function acctst() ends with epicsExit(0); Does that even make sense? Why not let it return normally? The main() function in acctstMain() calls acctst() and then returns 0 without checking the return value of int acctst().
The iocsh wrapper acctstCallFunc() in acctstRegister.cpp calls acctst() but that would then terminate any ioc which runs it. Is that desired? However that file is only linked to VxWorks in the Makefile:
OBJS_vxWorks = […] acctstRegister
„Exiting“ from VxWorks does not make much sense.

I can change the function to return void. Probably nobody will ever notice. I can also change the function to not call epicsExit but return 0. Will that change anything we care of?

modules/libcom/src/osi/epicsThread.h
dirk-zimoch: What with 
LIBCOM_API EPICS_NORETURN void epicsExit ?
would you prefer

LIBCOM_API
EPICS_NORETURN
void epicsExit

or

LIBCOM_API EPICS_NORETURN
void epicsExit

or

LIBCOM_API
EPICS_NORETURN
void
epicsExit 

or

EPICS_NORETURN
LIBCOM_API void epicsExit ?

anjohnson: The second one looks good to me:
```C
LIBCOM_API EPICS_NORETURN
void epicsExit(int status);
```

dirk-zimoch: Done.

Issue #541, opened 2024-08-27
Compile error with isnan/isinf and Visual Studio 2022 17.11.1


FreddieAkeroyd:
This occurred after a recent Visual Studio 2022 update (previous 2022 versions were OK). When compiling e.g. epicsThread.cpp on Windows x64 
```
 C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2588: '::!
_finite': illegal global finalizer
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2143: synt
ax error: missing ')' before '!'
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2143: synt
ax error: missing ';' before '!'
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2270: '_fi
nite': modifiers not allowed on nonmember functions
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2588: '::!
_isnan': illegal global finalizer
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2059: synt
ax error: ')'
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C4430: miss
ing type specifier - int assumed. Note: C++ does not support default-int
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(832): error C2556: 'boo
l _isnan(signed char) noexcept': overloaded function differs only by return type from 'int _isnan(signed char) noexcept'

C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): note: see declara
tion of '_isnan'
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(832): error C2371: '_is
nan': redefinition; different basic types
[...]

PR #542, opened 2024-08-29
protect isnan/isinf define on c++

FreddieAkeroyd:
Protect  isnan/isinf defines from being called under c++ as these now create a compile failure on a recent VS2022 update 

See #541


FreddieAkeroyd:
The PR also removes the redefines for any G++ compile via MinGW, this appears to work OK but maybe we want to align a bit more with  https://github.com/epics-base/epics-base/blob/7.0/modules/libcom/src/osi/os/posix/epicsMath.h ?

anjohnson:
Core Group: Realignment doesn't have to happen now, we're okay with this for now, thanks!

PR #543, opened 2024-08-29
libCom: don't use std::unexpected in epicsThread.

### LINKED ISSUE #343: `std::unexpected()` deprecated
The function  `epicsThread :: printLastChanceExceptionMessage` in modules/libcom/src/osi/epicsThread.cpp calls `std::unexpected ()`.
https://github.com/epics-base/epics-base/blob/fa00572780a00bea2f57dbff69653b14bd792be0/modules/libcom/src/osi/epicsThread.cpp#L84

The gcc-12 compiler claims that `std::unexpected ()` is deprecated:
```
../osi/epicsThread.cpp: In member function ‘void epicsThread::printLastChanceExceptionMessage(const char*, const char*)’:
../osi/epicsThread.cpp:84:21: warning: ‘void std::unexpected()’ is deprecated [-Wdeprecated-declarations]
   84 |     std::unexpected ();
      |     ~~~~~~~~~~~~~~~~^~
In file included from ../osi/epicsThread.cpp:14:
/opt/rh/gcc-toolset-12/root/usr/include/c++/12/exception:109:8: note: declared here
  109 |   void unexpected() __attribute__ ((__noreturn__));
      |        ^~~~~~~~~~
```
When using `-std=c++23`, compilation fails:
```
../osi/epicsThread.cpp: In member function ‘void epicsThread::printLastChanceExceptionMessage(const char*, const char*)’:
../osi/epicsThread.cpp:84:10: error: ‘unexpected’ is not a member of ‘std’
   84 |     std::unexpected ();
      |          ^~~~~~~~~~
```
See also: https://en.cppreference.com/w/cpp/error/unexpected

ericonr:
Besides being deprecated in C++11 and removed in C++17, the intended use case for this was always wrong, since std::unexpected() is called by the C++ runtime when a function throws an exception that was not specified in its "dynamic exception specification", which is different from an exception thrown by user code which wasn't caught [1,2].

Allowing the program to proceed instead of terminating it (via std::terminate(), which is the default unexpected handler) also seems to be more correct, since the thread which threw the exception has already exited anyway.

[1] https://github.com/epics-base/epics-base/issues/343
[2] https://en.cppreference.com/w/cpp/error/unexpected

---

Fixes #343

### DIFF
--- a/modules/libcom/src/osi/epicsThread.cpp
+++ b/modules/libcom/src/osi/epicsThread.cpp
@@ -16,6 +16,7 @@
 
 #include <stdio.h>
 #include <stddef.h>
+#include <stdlib.h>
 #include <float.h>
 #include <string.h>
 
@@ -78,10 +79,8 @@ void epicsThread :: printLastChanceExceptionMessage (
         "with type \"%s\" in thread \"%s\" at %s\n",
         pExceptionContext, pExceptionTypeName, name, date );
     errlogFlush ();
-    // This behavior matches the C++ implementation when an exception
-    // isn't handled by the thread code. Users can install their own
-    // application-specific unexpected handler if preferred.
-    std::unexpected ();
+
+    abort();
 }
 
 extern "C" void epicsThreadCallEntryPoint ( void * pPvt )


anjohnson:
Core Group: Suggest calling `std::abort()` rather than just letting the thread die, we want this to cause the exception to not be ignored.

mdavidsaver:
> Core Group: Suggest calling `std::abort()` ...

While strictly correct, it may simplify backwards compatibility to use the unqualified name, as in...

https://github.com/epics-base/epics-base/blob/2e4113b63bb60e4952eb0fcc97adfb8349b2be3e/modules/ca/src/client/ca_client_context.cpp#L415

ericonr:
Both suggestions have been implemented :)

PR #544, opened 2024-08-30
send DBE_PROPERTY events only if property field actually changed

dirk-zimoch:
The `dbPut()` function now compares (scalar) property fields to their old values and posts a DBE_PROPERTY event only if the value changed.

This change allows IOCs to write to property fields regularly (e.g. an IOC receiving values and alarm limits from a PLC periodically) without generating a large number of DBE_PROPERTY events. In the case I have seen, the IOC updated LOLO, LOW, HIGH and HIHI each time it updated VAL, resulting in a DBE_PROPERTY event rate 4 time as high as the DBE_VALUE event rate, even though the limits rarely ever changed. This undermines the idea of DBE_PROPERTY monitors to update properties only when needed.

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -1325,6 +1325,7 @@ long dbPut(DBADDR *paddr, short dbrType,
     long status = 0;
     dbFldDes *pfldDes;
     int isValueField;
+    int propertyUpdate = paddr->pfldDes->prop && precord->mlis.count;
 
     if (special == SPC_ATTRIBUTE)
         return S_db_noMod;
@@ -1369,8 +1370,22 @@ long dbPut(DBADDR *paddr, short dbrType,
         if (nRequest < 1) {
             recGblSetSevr(precord, LINK_ALARM, INVALID_ALARM);
         } else {
-            status = dbFastPutConvertRoutine[dbrType][field_type](pbuffer,
-                paddr->pfield, paddr);
+            if (propertyUpdate && paddr->field_size <= MAX_STRING_SIZE) {
+                char propBuffer[MAX_STRING_SIZE];
+                status = dbFastPutConvertRoutine[dbrType][field_type](pbuffer,
+                    &propBuffer, paddr);
+                if (!status) {
+                    if (memcmp(paddr->pfield, &propBuffer, paddr->field_size) != 0) {
+                        memcpy(paddr->pfield, &propBuffer, paddr->field_size);
+                    } else {
+                        /* suppress DBE_PROPERTY event if property did not change */
+                        propertyUpdate = 0;
+                    }
+                }
+            } else {
+                status = dbFastPutConvertRoutine[dbrType][field_type](pbuffer,
+                    paddr->pfield, paddr);
+            }
             nRequest = 1;
         }
     }
@@ -1391,11 +1406,7 @@ long dbPut(DBADDR *paddr, short dbrType,
     if (precord->mlis.count &&
         !(isValueField && pfldDes->process_passive))
         db_post_events(precord, pfieldsave, DBE_VALUE | DBE_LOG);
-    /* If this field is a property (metadata) field,
-     * then post a property change event (even if the field
-     * didn't change).
-     */
-    if (precord->mlis.count && pfldDes->prop)
+    if (propertyUpdate)
         db_post_events(precord, NULL, DBE_PROPERTY);
 done:
     paddr->pfield = pfieldsave;
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -1390,6 +1390,11 @@ long dbPut(DBADDR *paddr, short dbrType,
         }
     }
 
+    /* Post property updates before second dbPutSpecial */
+    /* which may post DBE_VALUE and/or DBE_LOG events */
+    if (propertyUpdate && !status)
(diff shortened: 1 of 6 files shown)


dirk-zimoch:
The special code path is only taken if monitors are connected, the property field is a scalar and `field_size <= sizeof(epicsAny)`, i.e. 40 bytes or less. This protects against possible future extensions with properties that are arrays or any hypothetical larger data type. In that case `dbPut` cannot compare and falls back to the previous behavior of always posting a DBE_PROPERTY event. When no monitor is connected, the comparison is skipped too as it would be a pointless waste of time and the value is simply written to the field directly.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -1369,8 +1370,22 @@ long dbPut(DBADDR *paddr, short dbrType,
         if (nRequest < 1) {
             recGblSetSevr(precord, LINK_ALARM, INVALID_ALARM);
         } else {
-            status = dbFastPutConvertRoutine[dbrType][field_type](pbuffer,
-                paddr->pfield, paddr);
+            if (propertyUpdate && paddr->field_size <= sizeof(epicsAny)) {
+                epicsAny propBuffer;
```
anjohnson: I don't think you should use `epicsAny` here — the `string` union member is _not_ a char array; `sizeof(epicsAny)` is 8 bytes on a 32-bit system and 16 bytes on a 64-bit system because this how that string member type is defined:
```C
/**
 * \brief !! Don't use this - it may vanish in the future !!
 */
typedef struct {
    unsigned    length;
    char        *pString;
} epicsString;
```

There is only one place in EPICS Base that actually uses the `epicsAny` type, and nothing that references `epicsAny.string` that I know of (I haven't checked external modules though). We should discuss removing either the `string` member or the whole `epicsAny` type at some point.

Looking through the DBD files at the fields marked `prop(YES)` I suggest using `char propBuffer[MAX_ENUM_STRING_SIZE]` which should be large enough for any of the fields that supply metadata properties.

anjohnson (changes requested):
I might have some more comments/questions but I don't have time to research and write them right now, so here's the important one to get started with.

modules/database/src/ioc/db/dbAccess.c
dirk-zimoch: Oops, I did not check what `epicsString` actually is and assumed 40 chars (confusing it with `epicsOldString`). It only worked "coincidentally" when I tested it with .EGU because I used a 64 bit machine.

dirk-zimoch: Small problem with `MAX_ENUM_STRING_SIZE`: It is part of the CA definitions (db_access.h) which notoriously disagree with the rest of EPICS (e.g. compare `MAX_UNITS_SIZE` 8 with EGU size of 16 for records in base). We also cannot be sure that some record outside base would never use longer string property fields. Thus, I prefer to use `MAX_STRING_SIZE`(i.e. 40 bytes) here. I don't think the additional 14 bytes of stack use will hurt us.

dirk-zimoch: I just noticed that mbbi/o records handle `DBE_PROPERTY` in their `special()` function and need to be fixed as well (unlike bi/o records).

dirk-zimoch:
I just noticed that mbbi/o records handle `DBE_PROPERTY` in their special() function and need to be fixed as well (unlike bi/o records).

dirk-zimoch:
I have now removed the `DBE_PROPERTY` event handling from the `special()` function of mbbi/o records and instead made the `xxST` fields `prop(YES)` like `xNAM` in bi/bo records.
However, this changes the timing of events: Before, `DBE_PROPERTY` and potentially `DBE_VALUE | DBE_LOG` were sent together. Now, the `DBE_PROPERTY` event is sent after the other two, because `dbPutSpecial(paddr, 1)` is called before `dbPost_events(precord, NULL, DBE_PROPERTY)`. Not ideal. I think I will change that order.

modules/database/src/ioc/db/dbAccess.c
anjohnson: Ack, sorry my bad; `MAX_STRING_SIZE` is the correct value for the "new" dbAccess routines, as shown here in dbAccessDefs.h:
```C
#define DBRenumStrs \
        epicsUInt32     no_str;         /* number of strings*/\
        epicsInt32      padenumStrs;    /*padding to force 8 byte align*/\
        char            strs[DB_MAX_CHOICES][MAX_STRING_SIZE];  /* string values    */
```

mdavidsaver:
Following up on the discussion during yesterday's core meeting.  Talking about "before" and "after" in the context of subscription updates resulting from calls to `db_post_events()` is complicated the way dbEvent buffering is currently implemented.

In brief.  Multiple subscriptions (`struct evSubscrip`) __may__ share the same buffer and mutex (`struct event_que`).  Each event thread (`struct event_user`) steps through its list of buffers (`event_user::firstque`) in allocation order.  So the relative order of updates between two subscriptions is maintained only iff 1) both updates result from the same `db_post_events()` call, 2) both subscriptions were created through the same `event_user`, and 3) happen to be assigned to the same `event_que`.  User code has no control over condition 3.

As a consequence, I think that:

```c
db_post_events(..., DBE_VALUE | DBE_LOG | DBE_ALARM | DBE_PROPERTY);
```

Does __not__ guarantee ordering between any two subscriptions.

anjohnson:
Core Group: Approved.

PR #546, opened 2024-08-31
Support for riscv64

Gui-Yue:
I noticed this package has not supported the arch riscv64.Maybe the modification I commit can support the arch riscv64.Please let me know wheather the solution can be accepted.


anjohnson:
Core developers group: We're reluctant to add architectures that we can't test or build without CI. Is there a tool-chain for cross-builds or even a docker container available for GitHub Actions for this architecture?

Gui-Yue:
> Core developers group: We're reluctant to add architectures that we can't test or build without CI. Is there a tool-chain for cross-builds or even a docker container available for GitHub Actions for this architecture?

I may add riscv arch auto-test ci for this pr.

mdavidsaver:
I think that additional riscv64 CI jobs should follow the pattern of the ARM jobs with `cross: linux-riscv64@riscv64-linux-gnu`.

https://github.com/epics-base/epics-base/blob/2e4113b63bb60e4952eb0fcc97adfb8349b2be3e/.github/workflows/ci-scripts-build.yml#L188-L198

The `@riscv64-linux-gnu` should cause [`cue.py`](https://github.com/epics-base/ci-scripts/blob/0e93b708551cb7bb212fda7029eccdf872dabc81/cue.py#L1090) to run `apt-get install g++-riscv64-linux-gnu`, which I think is available to the [ubuntu-latest](https://packages.ubuntu.com/search?keywords=g%2B%2B-riscv64-linux-gnu) image.

Gui-Yue:
how can i trigger the workflow to build Base?

ralphlange:
Done.

Gui-Yue:
I dont know why 'Ub-20 gcc-9 unsigned char' test failed here.It passed in my forked repo ci.

simon-ess:
I have re-run that job; it looks like a flaky failure.

Gui-Yue:
> I have re-run that job; it looks like a flaky failure.

right! it now successfully passes the test

mdavidsaver:
> right! it now successfully passes the test

The build succeeds, which is the important point for this PR.  Our unit tests only run for select cross targets.  This should be possible with QEMU support for riscv64 as it is for arm-linux-gnueabihf, but I don't see this as an immediate requirement since this could be quite tricky.

Are you able to manually build and run unittests on a riscv64 host?

eg. `make runtests` then __attach__ the output as a text file.

Gui-Yue:
> > right! it now successfully passes the test
> 
> 
> 
> The build succeeds, which is the important point for this PR.  Our unit tests only run for select cross targets.  This should be possible with QEMU support for riscv64 as it is for arm-linux-gnueabihf, but I don't see this as an immediate requirement since this could be quite tricky.
> 
> 
> 
> Are you able to manually build and run unittests on a riscv64 host?
> 
> 
> 
> eg. `make runtests` then __attach__ the output as a text file.
> 
> 

I'll do that

configure/os/CONFIG.Common.linux-riscv64
```diff
@@ -0,0 +1,21 @@
+# CONFIG.Common.linux-x86_64
+#
+# Definitions for linux-x86_64 target builds
+# Sites may override these definitions in CONFIG_SITE.Common.linux-x86_64
```
mdavidsaver: Please update the leading comments to match the filenames.

configure/os/CONFIG.Common.linux-riscv64
```diff
@@ -0,0 +1,21 @@
...
+
+ARCH_DEP_CFLAGS = $(GNU_TUNE_CFLAGS)
+ARCH_DEP_CPPFLAGS += -D_RISCV64_
+
+OP_SYS_CFLAGS +=
+OP_SYS_LDFLAGS +=
+
+# If your crosscompiler name has a GNU target prefix like <gnutarget>-gcc,
+# e.g. x86_64-redhat-linux-gcc, put a GNU_TARGET definition in
+# CONFIG_SITE.<host>.linux-x86_64 file, e.g. GNU_TARGET=x86_64-redhat-linux
```
mdavidsaver: `x86_64` -> `riscv64`

configure/os/CONFIG.linux-riscv64.Common
```diff
@@ -0,0 +1,9 @@
+# CONFIG.linux-x86_64.Common
+#
+# Definitions for linux-x86_64 host builds
+# Sites may override these definitions in CONFIG_SITE.linux-x86_64.Common
```
mdavidsaver: Also here...

configure/os/CONFIG.linux-riscv64.linux-riscv64
```diff
@@ -0,0 +1,14 @@
+#  CONFIG.linux-ppc64.linux-ppc64
+#
+# Definitions for linux-ppc64 host - linux-ppc64 target builds
+# Sites may override these definitions in CONFIG_SITE.linux-ppc64.linux-ppc64
```
mdavidsaver: Also here...

configure/os/CONFIG.Common.linux-riscv64
```diff
@@ -0,0 +1,21 @@
...
+# Sites may override these definitions in CONFIG_SITE.Common.linux-x86_64
+#-------------------------------------------------------
+
+# Include definitions common to all linux targets
+include $(CONFIG)/os/CONFIG.Common.linuxCommon
+
+ARCH_CLASS = riscv64
+
+ARCH_DEP_CFLAGS = $(GNU_TUNE_CFLAGS)
+ARCH_DEP_CPPFLAGS += -D_RISCV64_
```
mdavidsaver: Please remove the `-D_RISCV64_` definition.  Historically, Base provided its own ID macros for some targets.  However, these days we prefer to depend on compiler pre-defined macros.

eg. with GCC, query pre-defined macros by running `riscv64-linux-gnu-cpp -dM -x c /dev/null`.

Comparing the outputs for riscv 32 vs. 64 from [godbolt.org](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,source:'//+Type+your+code+here,+or+load+an+example.%0Aint+square(int+num)+%7B%0A++++return+num+*+num%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:rv64-gcc1420,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-dM+-x+c',overrides:!(),selection:(endColumn:19,endLineNumber:15,positionColumn:19,positionLineNumber:15,selectionStartColumn:19,selectionStartLineNumber:15,startColumn:19,startLineNumber:15),source:1),l:'5',n:'0',o:'+RISC-V+(64-bits)+gcc+14.2.0+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:pp,i:(clang-format:'1',editorid:1,filter-headers:'0',fontScale:14,fontUsePx:'0',j:1,treeid:0),l:'5',n:'0',o:'Preprocessor+Output+RISC-V+(64-bits)+gcc+14.2.0+(Editor+%231,+Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) leads me towards ID macros `#define __riscv` and `#define __SIZEOF_POINTER__`.

eg. testing the most specific case:

```c
#if defined(__linux__) && defined(__GNUC__) && defined(__riscv) && __SIZEOF_POINTER__==8
/* GCC compiling for Linux running on RISC-V 64 */
#endif
```

mdavidsaver (changes requested):

configure/os/CONFIG.Common.linux-riscv64
Gui-Yue: done

configure/os/CONFIG.linux-riscv64.Common
Gui-Yue: done

configure/os/CONFIG.linux-riscv64.linux-riscv64
Gui-Yue: done

configure/os/CONFIG.Common.linux-riscv64
Gui-Yue: done

Gui-Yue:
I dont know why it test failed , could you give me some suggestions?
[epics_log.log](https://github.com/user-attachments/files/17004294/epics_log.log)

mdavidsaver:
The MSI test fails, without useful output.

```
/home/debian/epics-base/epics-base/modules/database/test/ioc/dbtemplate/O.linux-riscv64
msi.t .. 
not ok 1
not ok 2
not ok 3
not ok 4
not ok 5
not ok 6
Dubious, test returned 2 (wstat 512, 0x200)
Failed 12/12 subtests 
```

anjohnson:
Please run the following commands in your Base build area so we can see what might be wrong with your msi output, and either paste that output here or attach a log file containing that output. We hope the 6 tests are all failing for the same reason, and unfortunately our test output doesn't show how the text output differs from what we're expecting.

```shell
cd modules/database/test/ioc/dbtemplate/O.linux-riscv64
../../../../../../bin/linux-riscv64/msi -I .. ../t1-template.txt
```

mdavidsaver:
@Gui-Yue MSI is a text processing tool, so this test failure could be a locale issue unrelated to this PR.  If you have convenient access to a similarly configured linux-x86_64 host, please also run the epics-base tests there.

Gui-Yue:
> Please run the following commands in your Base build area so we can see what might be wrong with your msi output, and either paste that output here or attach a log file containing that output. We hope the 6 tests are all failing for the same reason, and unfortunately our test output doesn't show how the text output differs from what we're expecting.
> 
> ```shell
> cd modules/database/test/ioc/dbtemplate/O.linux-riscv64
> ../../../../../../bin/linux-riscv64/msi -I .. ../t1-template.txt
> ```

I'll do that

anjohnson:
Core Group: @Gui-Yue We would still be interested in your manual results from the previous comments.

Gui-Yue:
> Core Group: @Gui-Yue We would still be interested in your manual results from the previous comments.

I will be working on this soon, sorry for forgetting about this pr in the past while.

PR #547, opened 2024-09-03
Use Perl for filtering MAKEFLAGS properly

### LINKED ISSUE #545: Build messages with `$(ECHO)` are suppressed on Windows due to spurious detection of `make-s` flag
Build messages that should be output using `$(ECHO) <message>` are suppressed.

The file `configure/CONFIG_COMMON` sets a value for `ECHO` based on the result of single-letter flag detection. The actual value of `makeflags` is being set to `T_A=windows-x64` instead of an expected word containing all single-letter flags, and the subsequent use in `$(findstring $1,$(makeflags))` results in a spurious detection of the `make-s` due to the presence of "s" within "T_A=windows-x64".

I added debug output to CONFIG_COMMON, which shows:
```
DEBUG $(MAKEFLAGS) =  -- T_A=windows-x64 TOP=../..
DEBUG $(makeflags) = T_A=windows-x64
DEBUG $(make-s) = s
DEBUG $(ECHO) = @:
```

See the attached [545_BuildLog.txt](https://github.com/user-attachments/files/16818829/545_BuildLog.txt)

## Notes

- GNU Make 4.4.41, x86_64-w64-mingw32

anjohnson:
Requires moving the checkflags code into CONFIG_BASE as that's where FIND_TOOL gets defined.

Fixes #545

### DIFF
--- a/configure/CONFIG_BASE
+++ b/configure/CONFIG_BASE
@@ -49,6 +49,7 @@ PODTOHTML_pl     = $(TOOLS)/podToHtml.pl
 PODTOHTML_dep    = $(PODTOHTML_pl) $(call FIND_PM,EPICS/PodHtml.pm)
 PODTOHTML        = $(PERL) $(PODTOHTML_pl)
 CONVERTRELEASE   = $(PERL) $(call FIND_TOOL,convertRelease.pl)
+FILTERMAKEFLAGS  = $(PERL) $(call FIND_TOOL,filterMakeflags.pl)
 FULLPATHNAME     = $(PERL) $(TOOLS)/fullPathName.pl
 GENVERSIONHEADER = $(PERL) $(TOOLS)/genVersionHeader.pl $(QUIET_FLAG) $(QUESTION_FLAG)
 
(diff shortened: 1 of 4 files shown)


simon-ess:
I can confirm that this seems to behave as expected on make 3.81.

anjohnson:
Tested on RHEL-7 which has Make 3.82, and on RHEL-8 (4.2.1).

The previous code worked on RHEL because there's a `-w` flag which seems to be included in `MAKEFLAGS` when no flags are given on the command-line:
```makefile
MAKEFLAGS = w -- T_A=linux-x86_64 TOP=../..
makeflags = w
```
That isn't something Red Hat added since I built my own copy of 4.2.1 from source on RHEL-7 and the `w` appears with that too. It doesn't appear on macOS though, using either Apple's 3.81 or 4.4.1 from Homebrew, and from @levismith's report in #545 it isn't set in the Windows builds either.

Apple 3.81:
```makefile
MAKEFLAGS = T_A=darwin-x86 TOP=../..
makeflags = 
```

Homebrew 4.4.1
```makefile
MAKEFLAGS =  -- T_A=darwin-x86 TOP=../..
makeflags = T_A=darwin-x86
```

A reminder from re-reading the stack-overflow post that I linked to in #395, the value of `$(MAKEFLAGS)` can be different when used inside a recipe vs. outside a recipe. The variable assignments weren't included when used outside the recipe with some Make versions. The Perl script stops parsing when it gets to either `--` or an assignment.

mdavidsaver:
Would this incantation be sufficient to "clean" the flags?

```make
BASICMAKEFLAGS=$(filter-out %=%,$(filter-out --%,$(firstword $(MAKEFLAGS))))
```

Dissecting...

1. `$(firstword ...` starts by taking only the first argument.  Which is one of: short args, a long arg, or a variable definition.  (I think the only possibilities)
2. `$(filter-out --%` strips out a long arg or literal `--`.
3. `$(filter-out %=%` strips out a variable definition.

I think this would leave only the short args or an empty string.

anjohnson:
@mdavidsaver The GNU Make documentation says that there can only be one wildcard in the pattern:
```makefile
all:
	@echo filter: $(filter-out %=%,x T_A=os-arch y)
```
```bash
$ make
filter: x T_A=os-arch y
```

mdavidsaver:
> ... can only be one wildcard in the pattern:

Well, that is less useful...

mdavidsaver:
Core Meeting.  Seems like the least bad solution ;)

Adding another `$(shell ...)` will have some effect on run-time.  If this turns out to be significant, it may be worth investigating `$(MAKE_VERSION)` as a condition to detect older Make.

PR #550, opened 2024-09-19
Drop execute bits from startup/win32.bat

jlmuir:
The file `startup/win32.bat` shouldn't be executable, so drop the execute bits from its file mode.


anjohnson:
Hi Lewis, I believe you but could you please briefly explain why — should the file be edited by the end-user first to set the right paths in it, or is it that Windows doesn't look at those bits for a .bat file anyway? Thanks.

jlmuir:
Hi, Andrew!  Sure, it's because Windows doesn't do anything with the execute bits, but on UNIX-like OSs with a POSIX file system, the execute bits are respected.  Since `win32.bat` is not intended to be executed, it doesn't make sense for it to have its execute bits set.

It's also inconsistent with the execute bits of the other files in the `startup` directory:

```
$ ls -Fglo
total 56
-rwxr-xr-x  1    965 Jul 21  2021 EpicsHostArch*
-rw-r--r--  1   3875 Jul 21  2021 unix.csh
-rw-r--r--  1   3872 Jul 21  2021 unix.sh
-rwxr-xr-x  1   4229 Sep 19 20:53 win32.bat*
-rw-r--r--  1   4231 Jul 21  2021 windows.bat
```

As you can see, the file mode for `win32.bat` doesn't match `windows.bat`, yet they're exactly the same except for two lines; `win32.bat` defaults to 32-bit x86, and `windows.bat` defaults to 64-bit x64.  (Aside: I don't even know why there are both `win32.bat` and `windows.bat`; it seems like one would have been fine.  Maybe for backward compatibility?)  And the `unix.{csh,sh}` files are intended to be sourced, not executed, so they don't have their execute bits set either.

The only file that should be executable in that directory is `EpicsHostArch` because it's intended to be executed, and it has a shebang line making that possible.  (Aside: It seems that macOS Sonoma and RHEL 7 [and probably others] will assume a shebang line of `#!/bin/sh` if the file is executable and has no shebang line.  I don't like it that they do that, but that's what I've noticed.)

jlmuir:
Hi, Andrew!  Is my explanation OK?  Also, the build check failed for this PR, but I couldn't imagine why it would, so I suspected a transient failure, and I force-pushed to try to make the checks run again, and this time it seems the build check passed.

anjohnson:
Hi Lewis, yes thanks for the extra info, that was a helpful reminder for me. As you've seen Appveyor often fails random tests, sometimes due to their VMs not being able to install some necessary package, sometimes for other reasons that we don't fully understand. When that happens we may ask it to re-run the tests that failed, and often (but not always) that is sufficient to turn the board green (which isn't always a hard requirement for us to merge PRs, but it gives us more confidence if it is). Your force-push triggered a rebuild during which one job failed, and I hit the rerun button earlier which finally passed that job.

We usually merge straightforward PRs like this one on alternate Wednesdays when the Core developers group meet. I'm pretty sure we'll get to this one tomorrow — thanks!

PR #552, opened 2024-09-20
Check if test was killed by signal in makeTestFile.

ericonr:
Tests killed by a signal should cause the .t script to exit with an error code, otherwise its exit code will be 0, and the test harness won't be aware of the unsuccessful exit.

### DIFF
--- a/src/tools/makeTestfile.pl
+++ b/src/tools/makeTestfile.pl
@@ -140,6 +140,8 @@ else {
     ######################################## Code for Unix run-hosts
     print $OUT <<__UNIX__;
 
+use POSIX qw(WIFEXITED WIFSIGNALED WEXITSTATUS);
+
 my \$pid = fork();
 die "\$tool: Can't fork for '$error': \$!\\n"
     unless defined \$pid;
@@ -154,9 +156,19 @@ if (\$pid) {
     };
 
     alarm \$timeout;
-    waitpid \$pid, 0;
-    alarm 0;
-    exit \$? >> 8;
+    while (1) {
+        waitpid \$pid, 0;
+        if (WIFEXITED(\$?)) {
+            # normal exit
+            alarm 0;
+            exit WEXITSTATUS(\$?);
+        } elsif (WIFSIGNALED(\$?)) {
+            # terminated by signal
+            alarm 0;
+            die "\$tool: Test was terminated by signal '\$?'\\n";
+        }
+        # non-terminal change of status, continue waiting
+    }
 }
 else {
     # Child process


anjohnson:
Is the intention here to exit the test collection early? There are 2 different ways that the tests are run, by Perl's `App::Prove` module in our `epicsProve.pl` script from the `make runtests` target, or by having Perl run the `.t` script directly from the `make tapfiles` target and capturing the output into a `.tap` file. Does this change affect both methods similarly, or do they behave differently?

I remember having some difficulty getting this to do exactly what I wanted, I may have given up trying to get the best solution.

ericonr:
I wasn't aware of the `runtests` target, was always doing `tapfiles` and then `test-results`. I will take a look at its behavior!

My intention wasn't to exit the test collection early, but to actually detect and report more errors. I had a test that was segfaulting (annoyingly, only in CI), but `make tapfiles` didn't report anything, the only clue was an incomplete `.tap` file. In my case `test-results` managed to catch it, but it wouldn't be the case, for example, if destructors trigger segfaults/assertions after all test results have already been printed.

mdavidsaver:
Core Group: `make tapfiles` is intended to be used with `make test-results`.  So this PR is unnecessary.

@ericonr Do you have any thoughts on how this pairing could be made more apparent?

eg. would you have noticed if `make tapfiles` ends by printing something like "To see results run: make test-results"?

ericonr:
@mdavidsaver I don't know if I agree with closing this.

The pairing between tapfiles and test-results was known to me (I learned of it by looking at the CI definitions from epics-base), but it doesn't cover the possibility of clean up code (e.g. C++ destructors) segfaulting/causing assertions, which will be missed if the cleanup happens after all tests have been ran.

Maybe we can find a better solution that doesn't cause the `tapfiles` target to error out, though.

mdavidsaver:
I see now that the mentions of `make tapfiles` and `make test-results` were a bit of a distraction from the change being proposed.  This PR is addressing an incomplete handling of the return value from `waitpid`.  So reopen.

mdavidsaver:
wrt. handling `waitpid`, rather than using magic numbers like `127`, it would be preferable to use the `W*()` macros, or notating the equivalence with the [`waitpid()`](https://man7.org/linux/man-pages/man2/wait.2.html) system call.

src/tools/makeTestfile.pl
```diff
@@ -156,6 +156,9 @@ BEGIN
     alarm \$timeout;
     waitpid \$pid, 0;
     alarm 0;
+    if ($? & 127) {
```
anjohnson: Core Developers: The `$?` here needs to be `\$?` (as below), since this code is inside a heredoc. We don't like the magic number and would prefer to see one the appropriate WIF macro from the Perl POSIX module.

mdavidsaver:
I did a little digging into what the status communicated `waitpid()` actually encodes, which turns out to be complicated with a lot of incomplete examples.

Starting with the example loop from the `waitpid()` manpage the first point is that needs to be a loop when non-terminal status changes can be reported.  eg. when any of the `WNOHANG | WUNTRACED | WCONTINUED` options is passed, which isn't the case in `makeTestfile.pl`.  However, the linux manpage wording for `WIFSTOPPED()` says `this is possible only if the call was done using WUNTRACED or when the child is being traced`.  As I read it, that "or" adds ambiguity.

I think a more robust handling would look something like the following:

```perl
    alarm \$timeout;
    while(1) {
        waitpid \$pid, 0;
        if(WIFEXITED(\$?)) { # normal exit with code
            alarm 0;
            exit WEXITSTATUS(\$?);
        } elif(WIFSIGNALED(\$?)) { # terminated by signal
            alarm 0;
            exit 1;
        }
        # non-terminal change of status, continue waiting
    }
```

ericonr:
> However, the linux manpage wording for WIFSTOPPED() says this is possible only if the call was done using WUNTRACED or when the child is being traced. As I read it, that "or" adds ambiguity.

The paragraph for `WUNTRACED` makes this less ambiguous, I think:

>  Status for traced children which have stopped is provided even if this option is not specified.

It's possible someone attaches gdb/strace to a running test, so it makes sense to be prepared for that; I think your suggestion is correct, I will include it here.

~Btw, on a related topic, I think procServ doesn't handle waitpid() returns caused by children being stopped by tracing them; I have attached GDB to IOCs under procServ and that caused it to launch a new IOC. I will take a look at the code there and make a PR with a similar implementation, if relevant.~ my mistake, it does things correctly. I was restarting the process with `run` after attaching; that does kill the original process and start another one

ericonr:
Btw, do you have any suggestions for better wording the commit message, to make my intention clearer?

src/tools/makeTestfile.pl
```diff
@@ -154,9 +154,19 @@ BEGIN
...
-    waitpid \$pid, 0;
-    alarm 0;
-    exit \$? >> 8;
+    while (1) {
+        waitpid \$pid, 0;
+        if (WIFEXITED(\$?)) {
+            # normal exit
+            alarm 0;
+            exit WEXITSTATUS(\$?);
+        } elif (WIFSIGNALED(\$?)) {
```
anjohnson: You need to load Perl's POSIX module to be able to use those macros, add the line shown below just above the call to `fork()` in line 143 above.
```perl
use POSIX qw(WEXITSTATUS WIFEXITED WIFSIGNALED);
```
I'm pretty sure this will fix all the CI failures.

ericonr: I was also missing elif -> elsif

ericonr:
There are two workflows for tests, right? `make tapfiles` + `make test-results` or `make runtests`.

##  `make tapfiles` and `make test-results`:

At least in epics-base, I can't get this to actually cause the error suite to error out if `.t` *only* exits with an error value; it can only work if there are issues with the tap file, from what I understand. I believe it worked with a test in the module I was developing, so will try to understand that; maybe it has to do with the fact that tests in epics-base using `.t` files are inside `modules`, so `RULES_MODULES` are in effect as well?

Generating the .tap file won't error out, because even though we add that test to `TAPS_FAILED_LOG`, the `TAPS_FAILED_LOG` files will only be read by `testFailures.pl` if toplevel `.tests-failed.log` exists, and that file is only created by the `PROVE_FAILURE` command, which is only run if there are issues with tapfiles (it's a circular situation).

https://github.com/epics-base/epics-base/blob/98a30e8029d1dcb0b650c9e2ae5c3cef2621fe86/configure/RULES_BUILD#L410-L415

https://github.com/epics-base/epics-base/blob/98a30e8029d1dcb0b650c9e2ae5c3cef2621fe86/configure/RULES_BUILD#L390-L396

Would it be okay to change `TAPFILE_FAILURE` to exit with an error? I think it would be enough to fix this. Before 7beb32e209715424607dc263b7f3244cecdf7844, an error in generating the tapfile would have errored out, but that commit added `|| $(TAPFILE_FAILURE)`, which might have had this unintended consequence?

## `make runtests`

`make runtests` doesn't error out immediately while running tests (though it does comment on strangeness with `Dubious, test returned 1 (wstat 256, 0x100)`), but it populates `.tests-failed.log`, which causes `TESTFAILURES` to exit with 1, and therefore causes make to exit with an error code as well.

mdavidsaver:
> Generating the .tap file won't error out ...

The decision to not terminate on the first failure in a run comes from:

1. Allowing `make -jXXX tapfiles` to run tests in parallel which speed up execution
2. The first test which fails may not be the most precise/clear indication of the root cause(s).

The two recipes should both ultimately communicate failure (or its a bug).  As you found, `make runtests` runs all tests, then exits non-zero if any failed.  However, `make tapfiles` explicitly always returns zero regardless, depending on a later call to `make test-results`, which should return non-zero if any tests failed.

As for why the two part recipe exists.  One complaint about `make runtests` is that the success/failure messages are spread out and easier to miss when scrolling.  The output of `make test-results` is supposed to be more concise.

mdavidsaver (approved):
> Btw, do you have any suggestions for better wording the commit message, to make my intention clearer?

I think your commit message is reasonable, as is the change.

ericonr:
> The two recipes should both ultimately communicate failure (or its a bug). .... later call to `make test-results`, which should return non-zero if any tests failed.

So we have a bug, because `make test-results` returns 0, since `testFailures.log` doesn't ever look at the `.taps-failed.log` files when there isn't a `.tests-failed.log`.

mdavidsaver (approved):
> This PR is addressing an incomplete handling of the return value from `waitpid`.

I think the PR now accomplishes this.

mdavidsaver:
> So we have a bug, because `make test-results` returns 0, since `testFailures.log` doesn't ever look at the `.taps-failed.log` files when there isn't a `.tests-failed.log`.

I am afraid that I don't follow.  @anjohnson ?

anjohnson:
Core Group: This change improves the behavior so we're merging it, but the other bug discussed above isn't handled here and we would have to look at it in more detail to work out if and how it could be changed.

Please open a separate issue describing the change that we can look at separately if you think this is still worth doing.

PR #558, opened 2024-10-21
feat: afterIocRunning (aka: atInit, afterInit) IOC shell command integrated into epics-base

jerzyjamroz:
1) I revised the ```afterInit``` module that is used by some parts of the epics community and integrated it into the epics-base (if it should go to another location then advise).
1) The module allows you to write better-structured ioc shell files to include in your startup scripts without tracking where ```iocInit``` is located (and how IOC is deployed) e.g.:
    * to achieve the best maintainability,
    * to improve writing boot sequences.
1) The module itself is simple but the idea behind is what gives it value.
1) It requires the config variable: ```afterInit=YES``` to activate, therefore the user decides to include it or not. Moreover, if included, it has zero effect on IOC that never uses that.
1) @dirk-zimoch , @jeonghanlee , @ttkorhonen , @gabrielfedel - please review it as well and share your opinion.

Feel free to modify the PR according to your preference.


dirk-zimoch:
Hi Jerzy

Actually, I have a newer version that allows to run scripts at any initHook stage. Maybe you want to use that one. I had not pushed that one yet to github (since end of 2019 or so).
It is pushed now.

https://github.com/paulscherrerinstitute/iocsh_utilities/blob/master/afterInit.c

Dirk

anjohnson:
Is this a more sophisticated rewrite of the `doAfterIocInit` command from the [synApps std module](https://github.com/epics-modules/std/blob/master/stdApp/src/delayCmd.cpp)?

Requesting a review from @timmooney and/or @keenanlang for their experience with that.

If we merge this into the Base softIoc we'd probably also want it in softIocPVA and softIocPVX but let's do the functional review here first since those are both external repo's and projects.

mdavidsaver:
I have vague memory that there is at least one other variation on this idea out there, but I can't remember a name just now.

My first thoughts.

How to manage/avoid iocsh command name conflicts with existing external module(s)?

This command should be available unconditionally.  (forgetting to include `system.dbd` is annoying enough)

I am uncertain about allowing callbacks at all hook points.  Is there a known use case for a hook other than `initHookAfterIocRunning`?

The `doAfterIocInit` implementation looks closer to what I have in mind.  eg. using ellLib.  I see a further simplification by making `cmdNode` a variable sized struct with a single allocation.

```c
struct cmdNode {
	ELLNODE node;
	char cmd[1];
};
```

And on the pathological side.  Disallow adding of additional commands after `iocInit()` starts.

simon-ess:
The version we use at ESS (essentially the same one pushed by @jerzyjamroz ) is derived from and simpler than the one that @dirk-zimoch has posted; the main difference is that we only allow the usage
```
afterInit "an entire command with args"
```
instead of
```
afterInit an entire command with args
```
I agree with @mdavidsaver that using elllib would be an improvement. We have not needed yet to use any hooks other than `initHookAfterIocRunning`, but who can say what others might need...

dirk-zimoch:
> Is this a more sophisticated rewrite of the `doAfterIocInit` command from the [synApps std module](https://github.com/epics-modules/std/blob/master/stdApp/src/delayCmd.cpp)?

When I wrote `afterInit` back in 2016, I was not aware of `doAfterIocInit`. It seems the same requirement came up at two different labs with two very similar solutions. I had probably used `doAfterIocInit` had I known of it.

> We have not needed yet to use any hooks other than `initHookAfterIocRunning`

I had been asked for something similar "before the records initialize", i.e. at `initHookAtBeginning`. Thus I generalized it to support running commands at any initHook.

> we only allow the usage `afterInit "an entire command with args"`

I like `Init an entire command with args` more.

>  using elllib would be an improvement

I have never seen the advantage of ellLib. It uses a lot of unsafe type casting. A `next` pointer with the correct type is simple enough and obvious. I would consider C++ `std::list` as it is at least type safe.

> variable sized struct with [...] `char cmd[1];`

Doable. My version uses `char cmd[256]`, a limit implied by iocsh. `doAfterInit` uses `char *` which is worse in memory locality and in allocation performance. On the other hand, this stuff runs only once and thus performance is not an issue. I also do not expect so many installed commands that `char[256]` would become a memory size problem.

> This command should be available unconditionally.

Agreed. The user is always free to not use it. No need to keep it unavailable.

ttkorhonen:
> 
> > we only allow the usage `afterInit "an entire command with args"`
> 
> I like `Init an entire command with args` more.
> 

Something like "I prefer \<option\> because \<argument\>" from both sides would help to resolve the issue...
(edit: I forgot escaping  \< \> )

dirk-zimoch:
I wrote it as I like it. Anyone else may write it as they like it... Because if saves typing quotes and allows for quotes arguments that contain spaces.

jerzyjamroz:
1) @dirk-zimoch , thank you for sharing, do you have an example with ```afterInit``` where the usage of ```iocshArgString``` would not be enough (compare our cmditem vs yours)? Basically, ```afterInit "any complex command"``` works fine. e.g.```$(EXTPPS="#")afterInit "dbpf $(P)PpsInp-MbbiDir_.TPRO 1"```
1) @anjohnson - thanks for the reference.
1) @mdavidsaver 
    * I use only ```afterInit``` hook but I didn't think about the benefits of others.
    * I can do the upgrade with your ELLNODE proposal once there is a green light for this PR.

jerzyjamroz:
@mdavidsaver , we have the use case of another hook, check the ```missing``` from the chart below:
```mermaid
graph LR;
    A["iocInit"]-->B["afterInitEarly (missing)"];
    B-->C["autosave"];
    C-->D["afterInit (this PR)"];
```
It would avoid overloading the existing ```.db``` records.

jerzyjamroz:
> I had been asked for something similar "before the records initialize", i.e. at `initHookAtBeginning`. Thus I generalized it to support running commands at any initHook.

@dirk-zimoch , I think this ```initHookAtBeginning``` solves the concern from my chart above. The question is if this hook is the best choice for that use case?

dirk-zimoch:
The thing I had been asked for is your `afterInitEarly` case. I had called it `atInit`. However, I cannot find the example right now. Not everything people request is actually being used later.
I also cannot give an example where using a single command string is actually more difficult than argv. I simply liked it without the extra quotes.

dirk-zimoch:
I think the request was "before autosave restores the values".

simon-ess:
As I recall, our reason for using the syntax
```
afterInit "foo bar baz"
```
was simply to provide a single consistent syntax for every usage. Most of the ones we have seen are just a
```
afterInit "dbpf $(P)$(R)PvName value"
```
and we thought this was the simpler of the two choices.

That said, if this is merged into EPICS base, that would be one less bit of custom code we need to maintain ourselves 😄

ralphlange:
Could anyone please show me a use case for this mechanism?

In 25 years of EPICS, I never heard of this module and never had a requirement in that direction.
Are all my `st.cmd` files badly structured, then?

simon-ess:
For us it's due to the inclusion of module-specific snippets of startup scripts. Those don't get to run iocInit, but they might want to specify something that happens afterwards.

(This might be a thing that is mostly more necessary for ESS or PSI as the local require users...)

ttkorhonen:
One of the use cases mentioned `dbpf`. I personally would like to get rid of all dbpf's in startup scripts. There are a few cases where they _may_ make sense but in general they risk undoing autosave... Most often they are used to override defaults from community templates. I think we have found a better way to do that, so at least that use case is quite weak IMO.

I am not against integrating this functionality in Base, even if I do not see a pressing need either. At least we should merge the different efforts if we decide to do it.

ralphlange:
Ahh, ESS and PSI, I should have noticed.

Assemblies (creating `st.cmd` files where modules can insert lines/commands) don't work for you as everything is found out at boot time, including commands you would want to run further down the script you're already executing.

Is the requirement to do that at run-time common enough to include the mechanism in Base?
Are there users outside of ESS and PSI?

dirk-zimoch:
> Ahh, ESS and PSI, I should have noticed.

The usual suspects. Our typical use case is to include a script of some "module" which wants to arrange for something after iocInit. E.g. starting sequencer threads.

ttkorhonen:
Well, somebody at APS had written the doAfterIocInit command, in [synApps std module](https://github.com/epics-modules/std/blob/master/stdApp/src/delayCmd.cpp), as Andrew mentioned. So I assume that there has been at least one use case outside PSI and ESS.

dirk-zimoch:
With my latest version, you can also prepare commands to run at `iocPause` and `iocRun` or at exit. Isn't that wonderful?
(And the commands are even executed in reverse order at the appropriate initHook stages.)

ttkorhonen:
Now, that would be something! 

> With my latest version, you can also prepare commands to run at `iocPause` and `iocRun` or at exit. Isn't that wonderful?

Indeed, that would be something interesting. The initHooks are there anyway, why not make them more user-friendly?

dirk-zimoch:
So let's add it to initHooks.c and call it a day.

mdavidsaver:
> > Ahh, ESS and PSI, I should have noticed.
> 
> The usual suspects

For ALS-U @jeonghanlee is promoting a pattern of using `iocshLoad()` of common snippets, some featuring `doAfterIocInit`.  eg. one line for autosave...

```
iocshLoad("$(IOCSH_TOP)/autosave.iocsh", "AS_TOP=$(TOP),IOCNAME=$(IOCNAME),DATABASE_TOP=$(DB_TOP),SEQ_PERIOD=60")
```

dirk-zimoch:
> For ALS-U @jeonghanlee is promoting [...]

He was spoiled by ESS as well. ;-)

jeonghanlee:
> > For ALS-U @jeonghanlee is promoting [...]
> 
> He was spoiled by ESS as well. ;-)

Haha, @dirk-zimoch Nice try, but that is not true. 

I introduced `iocshLoad` to ESS inspired by mostly synApps, especially `std` `doAfterIncInit` first. And I developed the documentation. 

Then, I found `require` had its own "similar" function, so I switched to `require` functions to let ESS users know about them.

The `std` module inspires me first, and I saw that function can reduce and hide complexity to users within ESS EPICS Environment. 

So, I spoiled ESS, not reversely. 

@jerzyjamroz now you know more than I know about the EPICS base. So, please don't call me to subjects which I don't know. ;)  I am pretty busy to do the intense troubleshooting for the Phoebus :)

dirk-zimoch:
> So, I spoiled ESS, not reversely.

Aha. Thanks anyway. ;-)

keenanlang:
>Requesting a review from @timmooney and/or @keenanlang for their experience with that.

Looks like a slightly more robust version of doAfterIocInit. Looks good to me. If it gets included in base, it looks to be capable of being a drop-in replacement with no issues.


>Could anyone please show me a use case for this mechanism?
>
>Is the requirement to do that at run-time common enough to include the mechanism in Base?
>Are there users outside of ESS and PSI?

The beamline controls for the APS have been shifting to using it extensively. Like others have put forward, we also use iocshLoad to call iocsh scripts that package the initialization of devices. Sometimes there are sequencer programs that are associated with said device support and it is useful to have the iocsh script to trigger that sequencer program (as the script has already been given all relevant information).

This has the benefit of making it very difficult to accidentally make a mistake in setting support up. You can't accidentally forget the sequencer program due to copying the database load but missing that there was a line at the end of the st.cmd file. Nor can you accidentally break the support because you changed a macro for the database but forgot to change the same one for the sequencer program. 

So, for example, we have a set of filter bank support that has a iocsh script implemented here (https://github.com/epics-modules/optics/blob/master/opticsApp/iocsh/pf4filters.iocsh). That can be included into an IOC with the line:

> iocshLoad("$(OPTICS)/iocsh/pf4filters.iocsh", "PREFIX=$(PREFIX), INSTANCE=pf4")

And removed from an IOC with a single comment.

jeonghanlee:
> Could anyone please show me a use case for this mechanism?
> 
> In 25 years of EPICS, I never heard of this module and never had a requirement in that direction. Are all my `st.cmd` files badly structured, then?

@ralphlange my example is the exact same as what @keenanlang shared since his approach (actually codes iocsh) inspired me to follow that direction around 2017(?), which is a major method to use the iocsh (cmd) snippet to the ESS I believe.

It would be really nice to have them into the EPICS base from my view. I have to use `std` in order to use `doAfterIocInit`. Still I am good not to have `require` 😄 

We added the following line in `st.cmd`

```
iocshLoad("$(IOCSH_TOP)/autosave.iocsh", "AS_TOP=$(TOP),IOCNAME=$(IOCNAME),DATABASE_TOP=$(DB_TOP),SEQ_PERIOD=60")
```
And we can disable and enable the entire `autosave` configuration with one "#", and it removes the chance for  users to think and rewrite codes multiple times to implement in each IOC which we need to do. And it also give us the consistent `autosave` strategies across overall IOCs. I saw a lot of bad examples in multiple IOCs across the Europe and US.

And here is the complete `iocsh` file for `autosave` which I used. At this moment, for ALS-U, I don't decide yet to use `doAfterIocInit` since it requires the additional module which I don't need it such as `std`. I am okay to split the single iocsh file into two iocsh files such as `autosave_bf_iocinit.iocsh` and `autosave_af_iocinit.iocsh`. 

  



```
##############################################################################
####                         AUTOSAVE iocsh                                ###
##############################################################################
#- IOCNAME        - AUTOSAVE IOCNAME (IOCNAME should not have :)
#- AS_TOP         - Autosave TOP path
#- DATABASE_TOP       - database TOP path
#-#############################################################################
#-
#- INCOMPLETE      - Optional: Ok to save/restore save sets with missing values?
#-                   Default: 1
#- CA_RECONNECT    - Optional: Retry connecting to PVs whose initial connection
#-                             attempt failed?,  Default: 1
#- DATED_BACKUP    - Optional: Save dated backup files?
#-                   Default: 1
#- CA_RECONNECT    - Optional: Retry connecting to PVs whose initial connection
#-                             attempt failed?, Default: 1
#- NUM_SEQ         - Optional: Number of sequenced backup files to write
#-                   Default: 3
#- SEQ_PERIOD      - Optional: Time interval in seconds between sequenced backups
#-                 - Default: 300
#- VALUES_FILES_PASS0    - Optional: Filename, Default: values_pass0
#- VALUES_FILES_PASS1    - Optional: Filename, Default: values_pass1
#- SETTINGS_FILES  - Optional: Filename, Default : settings
#- WAVEFORM_FILES  - Optional: Filename, Default : waveform
#- VALUES_PERIOD_PASS0   - Optional, Default 5
#- VALUES_PERIOD_PASS1   - Optional, Default 10
#- SETTINGS_PERIOD  - Optional, Default 5
#- WAVEFORM_PERIOD - Optional, Default 30
#-
set_savefile_path    ("$(AS_TOP)/$(IOCNAME)", "/save")
set_requestfile_path ("$(AS_TOP)/$(IOCNAME)", "/req")
#
#- If there is no path, create it
system("install -d $(AS_TOP)/$(IOCNAME)/save")
system("install -d $(AS_TOP)/$(IOCNAME)/req")
#-
#- iocBoot/iocas/save_restore.cmd
#-
save_restoreSet_status_prefix("$(IOCNAME)-as:")
#-
#- Debug-output level
save_restoreSet_Debug(0)
#-
#- Ok to save/restore save sets with missing values (no CA connection to PV)?
save_restoreSet_IncompleteSetsOk($(INCOMPLETE=1))
#-
#- Tell save_restore to writed dated backup files. 
save_restoreSet_DatedBackupFiles($(DATED_BACKUP=1))
#-
#- Tell save_restore to save sequence files.
#- save_restore to to maintain three copies of each *.sav file
save_restoreSet_NumSeqFiles($(NUM_SEQ=3))
#-
#- at 5 minute intervals
save_restoreSet_SeqPeriodInSeconds($(SEQ_PERIOD=300))
#-
#- should periodically retry connecting to PVs whose initial connection attempt failed
#-
save_restoreSet_CAReconnect($(CA_RECONNECT=1))

#-
#- Time interval in seconds between forced save-file writes.  (-1 means forever).
#- This is intended to get save files written even if the normal trigger mechanism is broken.
save_restoreSet_CallbackTimeout(-1)
#- 
#- pass0 : save files are to be restored before record initialization
set_pass0_restoreFile("$(SETTINGS_FILES=settings).sav")
set_pass0_restoreFile("$(VALUES_FILES_PASS0=values_pass0).sav")
#- pass1 : save files are to be restored after record initialization
set_pass1_restoreFile("$(SETTINGS_FILES=settings).sav")
set_pass1_restoreFile("$(VALUES_FILES_PASS1=values_pass1).sav")
#-
#-
dbLoadRecords("$(DATABASE_TOP)/save_restoreStatus.db", "P=$(IOCNAME)-as:, DEAD_SECONDS=5")

#- epics-modules/std has the doAfterIocInit function

doAfterIocInit("makeAutosaveFileFromDbInfo('$(AS_TOP)/$(IOCNAME)/req/$(SETTINGS_FILES=settings).req','autosaveFields')")
doAfterIocInit("makeAutosaveFileFromDbInfo('$(AS_TOP)/$(IOCNAME)/req/$(VALUES_FILES_PASS0=values_pass0).req','autosaveFields_pass0')")
doAfterIocInit("makeAutosaveFileFromDbInfo('$(AS_TOP)/$(IOCNAME)/req/$(VALUES_FILES_PASS1=values_pass1).req','autosaveFields_pass1')")
#-
#- We don't need PREFIX, because generated req files has the hard-code PV names from DBInfo : info tag
doAfterIocInit("create_monitor_set('$(SETTINGS_FILES=settings).req','$(SETTINGS_PERIOD=5)')")
doAfterIocInit("create_monitor_set('$(VALUES_FILES_PASS0=values_pass0).req','$(VALUES_PASS0_PERIOD=5)')")
doAfterIocInit("create_monitor_set('$(VALUES_FILES_PASS1=values_pass1).req','$(VALUES_PASS1_PERIOD=10)')")

```

jeonghanlee:
> > > Ahh, ESS and PSI, I should have noticed.
> > 
> > 
> > The usual suspects
> 
> For ALS-U @jeonghanlee is promoting a pattern of using `iocshLoad()` of common snippets, some featuring `doAfterIocInit`. eg. one line for autosave...
> 
> ```
> iocshLoad("$(IOCSH_TOP)/autosave.iocsh", "AS_TOP=$(TOP),IOCNAME=$(IOCNAME),DATABASE_TOP=$(DB_TOP),SEQ_PERIOD=60")
> ```

Thank you for invitation, @mdavidsaver. I didn't want to join the discussion since I am terribly busy for the project and Phoebus, but I am welcome to do so.

simon-ess:
initHooks.c makes sense to me as well.

For funsies, @dirk-zimoch , with your setup allowing to run commands at each of the init hooks, can you do something like
```
atInitState IocBuild "atInitStage Beginning \"atInitStage AfterCallbackInit echo foo\""
```
Not that I would _suggest_ to do this, but it would be pretty funny.

jerzyjamroz:
> or at exit

@dirk-zimoch - I realized that it can be useful to ensure the orderly/controlled termination of certain hardware processes.

anderslindho:
We have at ESS often been asked to ensure adequate tests for this particular feature, so I think it would be nice if tests could be added to any version accepted into EPICS base.

anjohnson:
Core Group: Agreement in principle that we should have this functionality in Base, but please look at moving the implementation to initHooks.c, does that simplify or shorten the code significantly?

Question about whether we want to support more than one initHook, don't want users to have to remember the names of the hooks. Suggest 2 commands `atInit "command"` and `atInitHook <hook> "command"`

mdavidsaver:
To avoid needing to interact with .dbd files, ~iocsh~ libCom has a special means of registration.

https://github.com/epics-base/epics-base/blob/9f8a8b9c1fc96dcbb18a35b627530428f74aaa62/modules/libcom/src/iocsh/libComRegister.c#L479-L482

anjohnson:
Core Group: We aren't convinced about the need for the more general (any hook) case, that could cause problems in some cases. Please move the implementation to libCom and just provide the `atInit "command"` case.

jerzyjamroz:
I) If we resign from the generic hook implementation e.g. ```atInitHook <hook> “command”```, then the data structure will be affected, e.g. ([cmditem](https://github.com/jerzyjamroz/epics-base/blob/c844b92fa3e00a4daab93b0e6c4b79a70ffa9136/modules/libcom/src/iocsh/initHooks.c#L190C1-L194C3))
```
struct cmditem {
   struct cmditem *next;
   char           *cmd;
   initHookState   hook;
};
```
So something to consider:
1) To implement it generic for all hooks and then “link” ```atInit``` to ```initHookAfterIocRunning``` (so to expose that in IOC shell, then repeat that for any hook one would like to share in the IOC shell. E.g. At the ```initHookAtBeginning``` and ```initHookAtShutdown```.
2) To implement it straight so: one implementation per one hook that one wants to expose in the IOC shell
```
struct cmditem {
   struct cmditem *next;
   char           *cmd;
};
```
This is what we agreed with ```atInit``` during the epics meeting.

---
II) Another consideration is the ```initHookAtShutdown```
Useful, when one wants to assure a controlled behaviour of the shutdown process, and even EVG can be an example in here:
```
initHookAtShutdown "dbpf $(P)SoftSeq0Disable-Cmd 1”
initHookAtShutdown "dbpf $(P)SoftSeq1Disable-Cmd 1"
```
So If one uses two sequencers, it is good to assure that the EVG does not send any events while IOC gets switched off, so to avoid any hazardous situtations.

mdavidsaver:
> II) Another consideration is the `initHookAtShutdown`

To my mind, the example you cite is a good one of shutdown behavior which should be done at the driver level.  (yes, I am aware that this specific omission from mrfioc2 is my doing)

Certainly any required shutdown should be performed at driver level.  eg. the way mrfioc2 disables VME/PCI interrupts.

> ... so to avoid any hazardous situtations.

I am somewhat concerned about users adding eg. `atExit dbpf SOME:ASYN:PARAM 0` without understanding that this creates a race condition.  Still, I guess situations like this are why we describe EPICS as a toolkit.

So I would accept allowing deferred commands for `initHookAfterIocBuilt` and `initHookAtShutdown`.

modules/database/src/ioc/misc/afterInit.c
```diff
@@ -0,0 +1,126 @@
...
+  "  afterInit \"dbpf <PV> <VAL>\"\n"
+  "  afterInit \"date\"\n";
+
+struct cmditem
+{
+  struct cmditem *next;
+  char *cmd;
+};
+
+static struct cmditem *cmdlist, **cmdlast = &cmdlist;
```
mdavidsaver: How about switching to `ellLib.h` when changing `cmditem` to be a single allocation?  Also, I think this one list will become several.  One for each permitted `initHook*`.

modules/database/src/ioc/misc/afterInit.c
```diff
@@ -0,0 +1,126 @@
...
+
+static void afterInitHook(initHookState state)
+{
+  if(state != initHookAfterIocRunning)
+    return;
+
+  struct cmditem *item = cmdlist;
+  struct cmditem *next = NULL;
+  while(item)
```
mdavidsaver: fyi. with `ellLib.h` this could become

```c
struct cmditem *item;
while((item = ellGet(theList)) {
```

Also, an AtShutdown hook should iterate in the opposite order.

modules/database/src/ioc/misc/afterInit.c
```diff
@@ -0,0 +1,126 @@
...
+  helpMessage};
+
+static void afterInitFunc(const iocshArgBuf *args)
+{
+  static int first_time = 1;
+  char *cmd = args[0].sval;
+
+  if(interruptAccept)
+  {
+    errlogPrintf("WARNING afterInit can only be used before iocInit (check help)\n");
```
mdavidsaver: IOCsh functions should print messages with the `epicsStdout.h` wrappers.  These are synchronous, it is clear which function resulting in which messages.  Also, these wrappers handle iocsh redirection to file (eg. `dbl > records.list`).

modules/database/src/ioc/misc/afterInit.c
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  struct cmditem *item = newItem(cmd);
+
+  if(!item)
+    errlogPrintf("ERROR afterInit failed to add the command '%s' %s\n", cmd, strerror(errno));
+}
+
+static void afterInitRegister(void)
```
mdavidsaver: Move this code into libCom and handle registration from `libComRegister()`.

https://github.com/epics-base/epics-base/pull/558#issuecomment-2536374560

modules/database/src/ioc/misc/afterInit.c
```diff
@@ -0,0 +1,126 @@
...
+}
+
+static struct cmditem *newItem(char *cmd)
+{
+  struct cmditem *item = malloc(sizeof(struct cmditem));
+
+  if(item == NULL)
+  {
+    errno = ENOMEM;
+    return NULL;
```
mdavidsaver: I don't think there is much point to setting `errno` here.  Since this code can only run before `iocInit()`, I think it would be acceptable to use `mallocMustSucceed()`.

modules/database/src/ioc/misc/afterInit.c
```diff
@@ -0,0 +1,126 @@
...
+  item->next = NULL;
+  *cmdlast = item;
+  cmdlast = &item->next;
+  return item;
+}
+
+static const iocshFuncDef afterInitDef = {
+  "afterInit",
+  1,
+  (const iocshArg *[]){&(iocshArg){"command (before iocInit)", iocshArgString}},
```
mdavidsaver: I had to look at this for awhile...  Is this valid C89?  I guess clang, GCC, and MSVC all  accepts it, which is the practical criteria.

mdavidsaver (changes requested):

modules/database/src/ioc/misc/afterInit.c
anjohnson: I don't know why, but I don't see any output from GHA builds of this PR at all. Appveyor appears to have successfully built the branch with all the compilers, but there are no tests in or of this code so it hasn't been exercised.

dirk-zimoch: Older compiler versions may have a problem with that. Do we care? All the compilers I use are fine with that.

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: Done

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: Done.

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: Done.

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: Done.

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: I mark it as solved.

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: Ongoing, once the actual code of ```atInit``` is confirmed, I will proceed with this request.

jerzyjamroz:
@mdavidsaver , thank you for the review. I applied all your comments except the movement to the ```libCom``` - I will do that right after the green light to this ELLLIST implementation.

anjohnson:
@jerzyjamroz The build-time conditionals `ifeq ($(atInit),YES)` in your Makefile changes here mean that the code you've added doesn't get compiled by our CI jobs at all. This PR is effectively a no-op to them. Was there a specific reason for adding them?

GitHub is asking me to approve it building 3 GHA workflows, but I don't plan to do that having found this. I just cancelled the remaining Appveyor builds after searching through one job's complete build log for "afterInit." and got no results.

modules/database/src/ioc/misc/atInit.c
```diff
@@ -0,0 +1,119 @@
...
+    free(item->cmd);
+    free(item);
+  }
+}
+
+static struct cmditem* newItem(char* cmd)
+{
+  struct cmditem* item = mallocMustSucceed(sizeof(struct cmditem), "ERROR Failed to allocate memory for cmditem");
+
+  item->cmd = epicsStrDup(cmd);
```
mdavidsaver: Please change to use a single allocation.  eg.

```c
struct cmditem* item = mallocMustSucceed(sizeof(struct cmditem)+strlen(cmd)+1, "ERROR ...");
item->cmd = (char*)(item+1);
strcpy(item->cmd, cmd);
```

modules/database/src/ioc/misc/atInit.c
```diff
@@ -0,0 +1,119 @@
...
+    return;
+
+  struct cmditem* item = NULL;
+
+  while(item = (struct cmditem*)ellGet(&cmdlist))
+  {
+    epicsStdoutPrintf("%s\n", item->cmd);
+
+    if(iocshCmd(item->cmd))
+      epicsStdoutPrintf("ERROR atInit command '%s' failed to run\n", item->cmd);
```
mdavidsaver: For consistency and clarify, prepend the preceding printf with `"INFO atInit: "`.

I also think that use of the colored `ERL_ERROR` from errlog.h would give a better chance of this error being noticed.

mdavidsaver (commented):
> ... green light to this ELLLIST implementation.

You are using ellLib as is intended, and the code is becoming pleasingly short.  With the changes noted made, including registration, I think this PR will be ready.

jerzyjamroz:
> Was there a specific reason for adding them?

@anjohnson The reason was to have the selection mechanism by a user with ```atInit = YES``` in CONFIG_SITE. I removed that feature (I guess it is your intention).

modules/database/src/ioc/misc/atInit.c
jerzyjamroz: Done.

modules/database/src/ioc/misc/atInit.c
jerzyjamroz: As I understood previously ```errlogPrintf``` does not do that so I added a local macro ```__AT_INIT_LOG``` managing all ```ERL_ERROR```-s (and ```ERL_WARNING```-s) according to your comment.

modules/database/src/ioc/misc/atInit.c
```diff
@@ -0,0 +1,123 @@
...
+#include <initHooks.h>
+#include <iocsh.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define __AT_INIT_LOG(svr) svr " atInit: "
+
+// Version within the message
+static const char helpMessage[] =
+  "atInit version 2.0.1\n"
```
simon-ess: I'm a bit surprised at adding the _version_ here. Is there any other help command that includes a version?

jerzyjamroz: @simon-ess , I had the same concern.

modules/database/src/ioc/misc/afterInit.c
jerzyjamroz: @mdavidsaver , The last commit presents the code movement to ```libCom```. There was an issue that ```#include <dbAccess.h>``` could not be used so ```interruptAccept``` also could not be used. I modified the idea with ```static int s_initendflag = 0;``` but if you have a better proposal, all is welcome.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,124 @@
...
+#include <iocsh.h>
+#include <string.h>
+
+#include "atInit.h"
+
+#define __AT_INIT_LOG(svr) svr " atInit: "
+
+// Version within the message
+static const char helpMessage[] =
+  "atInit version 2.1.1\n"
```
anjohnson: I don't think we want a version number here, it's just going to be noise and won't get updated so please remove it.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,124 @@
...
+  "Allows you to define commands to be run after the iocInit\n"
+  "Example commands:\n"
+  "  atInit \"dbpf <PV> <VAL>\"\n"
+  "  atInit \"date\"\n";
+
+struct cmditem
+{
+  ELLNODE node;
+  char* cmd;
```
anjohnson: The modern standard for EPICS is to indent by 4 spaces, I would much prefer to see that for new code.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,124 @@
...
+static void atInitHook(initHookState state)
+{
+  if(state != initHookAfterIocRunning)
+    return;
+
+  struct cmditem* item = NULL;
+
+  while(item = (struct cmditem*)ellGet(&s_cmdlist))
+  {
+    epicsStdoutPrintf("%s\n", item->cmd);
```
anjohnson: The `epicsStdoutPrintf()` routine is explicitly needed by some internal libCom calls, but most code should be calling `printf()` to allow the stream redirection code in iocsh to work as documented.

@dirk-zimoch — did these `epicsStdoutPrintf()` calls come from your original, and if so was there a functional reason that you used them instead?

modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -27,7 +27,7 @@
 #include "epicsGeneralTime.h"
 #include "freeList.h"
 #include "libComRegister.h"
-
+#include "atInit.h"
```
anjohnson: Please leave a blank line after all the `#include` lines, it seems to have been lost here.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,124 @@
+/* Copyright (C) 2020 Dirk Zimoch */
+/* Copyright (C) 2020-2025 European Spallation Source, ERIC
+ * Maintainer: Jerzy Jamroz
+ */
```
anjohnson: Please use the standard EPICS license header format for this block, and include the SPDX License ID tag. In this case it should probably be:

```C
/*************************************************************************\
* Copyright (C) 2020 Dirk Zimoch
* Copyright (C) 2020-2025 European Spallation Source, ERIC
* SPDX-License-Identifier: EPICS
* EPICS BASE is distributed subject to a Software License Agreement found
```

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,124 @@
...
+                                           __AT_INIT_LOG(ERL_ERROR) "failed to allocate memory for cmditem");
+  item->cmd = (char*)(item + 1);
+  memcpy(item->cmd, cmd, strlen(cmd) + 1);
+
+  if(item->cmd == NULL)
+  {
+    free(item);
+    errno = ENOMEM;
+    return NULL;
+  }
```
anjohnson: This pointer check has a code smell. The `mallocMustSucceed()` call is guaranteed to return a non-NULL value that points to a buffer large enough to hold the requested size. That means `item->cmd` can never be NULL.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,124 @@
...
+    return;
+  }
+
+  if(first_time)
+  {
+    first_time = 0;
+    if(initHookRegister(atInitHook) < 0)
+    {
+      errno = ENOMEM;
+      epicsStdoutPrintf(__AT_INIT_LOG(ERL_ERROR) "initHookRegister memory allocation failure %s\n", strerror(errno));
```
anjohnson: How do we know this was a memory allocation failure and not some other kind (maybe in a future version of `initHookRegister()`)?

Setting `errno` ourselves also has a code smell, the C library should have set it if that was why `initHookRegister()` gave an error.

anjohnson (changes requested):
These comments are mostly about non-functional issues that affect human readability and style, but there are a couple that I spotted when investigating the code that sets `errno` (which sometimes can be a code smell).

modules/libcom/src/iocsh/atInit.c
mdavidsaver: @anjohnson I might have suggested the explicit call, which I am coming to prefer vs. the `#define printf ...`.

modules/libcom/src/iocsh/atInit.c
mdavidsaver: Usage of `mallocMustSucceed()` makes this `NULL` check unnecessary, so it can simply be removed.

modules/libcom/src/iocsh/atInit.c
mdavidsaver: fyi. `initHookRegister()` already logs this failure, so no need to print again.

https://github.com/epics-base/epics-base/blob/c8eccfcb8ffcba80c7464982c3e6abbc9716d029/modules/libcom/src/iocsh/initHooks.c#L60-L63

modules/libcom/src/iocsh/atInit.c
jerzyjamroz: Done.

modules/libcom/src/iocsh/atInit.c
jerzyjamroz: Ok, ```printf``` implemented (the ```epicsStdoutPrintf``` was used with the portability intentions).
```#define``` was used to reuse the same format. I will change it to the explicit call.

File: modules/libcom/src/iocsh/libComRegister.c (hunk 13 above)
Comment (jerzyjamroz) [reply modules/libcom/src/iocsh/libComRegister.c]: Done.

File: modules/libcom/src/iocsh/atInit.c (hunk 14 above)
Comment (jerzyjamroz) [reply modules/libcom/src/iocsh/atInit.c]: Done.

File: modules/libcom/src/iocsh/atInit.c (hunk 15 above)
Comment (jerzyjamroz) [reply modules/libcom/src/iocsh/atInit.c]: Done.

File: modules/libcom/src/iocsh/atInit.c (hunk 16 above)
Comment (jerzyjamroz) [reply modules/libcom/src/iocsh/atInit.c]: @mdavidsaver , the only gain is that you get an extra info ```"ERROR atInit: ..."``` whereas in your case, you only know that ```Cannot malloc a new initHookLink```.

Comment (jerzyjamroz) [reply modules/libcom/src/iocsh/atInit.c]: I have removed ```errno``` and the printf message.

File: modules/libcom/src/iocsh/atInit.c [hunk 17]
```diff
@@ -0,0 +1,104 @@
...
+
+    if (!cmd || !cmd[0]) {
+        printf(ERL_WARNING " atInit: "
+                           "received an empty argument (check help)\n");
+        return;
+    }
+
+    if (first_time) {
+        first_time = 0;
+        initHookRegister(atInitHook);
```
mdavidsaver: Move this `initHookRegister()` into `atInitRegister()`.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,104 @@
...
+    s_initendflag = 1;
+}
+
+static struct cmditem* newItem(char* cmd)
+{
+    struct cmditem* item = mallocMustSucceed(sizeof(struct cmditem) + strlen(cmd) + 1,
+                                             ERL_ERROR " atInit: "
+                                                       "failed to allocate memory for cmditem");
+    item->cmd = (char*)(item + 1);
+    strlcpy(item->cmd, cmd, strlen(cmd) + 1);
```
mdavidsaver: I don't think `strlcpy()` is portable.  Also, it is unnecessary since the allocation directly above is assured to be sufficient.

mdavidsaver (changes requested):

modules/libcom/src/iocsh/atInit.c
jerzyjamroz: I have not changed that before as my guess was that if one does not use ```atInit``` in his IOC, then ```initHookRegister(atInitHook)``` is never provisioned (I guess this intention of @dirk-zimoch ).
Anyway, I have applied this modification.

modules/libcom/src/iocsh/atInit.c
jerzyjamroz: Ok, I changed it to ```memcpy()``` (unless you prefer simply ```strcpy```?).

ralphlange: I guess `strcpy()` will be marked as unsafe. I think `strncpy()` works best. (While still creating some warnings.)

jerzyjamroz: Ok, I have changed to ```strncpy()```.

jerzyjamroz: It might be a good idea to add e.g. ```epicsStrlcpy()``` function to the EPICS toolkit so all operations are unified.

ralphlange: That thought had crossed my mind a few times already.
Speaking of: We could also think about adapting an existing string handling library adding the UTF-8 support that Base is still missing. That would really add value.

dirk-zimoch: I think utf-8 support should go on the EPICS 7.1 wish list. It is out of scope here.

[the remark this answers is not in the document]
dirk-zimoch: Any standard  xxxcpy() funtion call that uses the source length instead of the target buffer size will be flagged as warnings. EPICS is full of such code.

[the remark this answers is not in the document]
ralphlange: And the QA tool is right in flagging this!
In the above case: if the source buffer happens to miss the null termination,`strlen()` will happily read until the end of the universe. This code is not entirely wrong, but definitely not robust, either. Exactly the stuff you want a QA tool to point out.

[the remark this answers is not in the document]
dirk-zimoch: Exactly what I mean. Unfortunately many people seem to think replacing strcpy(&dest, &src) with strncpy(&dest, &src, sizeof(src)) or strncpy(&dest, &src, strlen(&src)) would fix anything and are then surprized that they still get warnings. EPICS has several of such "fixes" (which should be cleaned up). Please do not add more.
The compiler warnings are sometimes a bit misleading in just saying: strcpy() is unsafe, use strncpy()

anjohnson: We have talked about adding `strlcat()` and `strlcpy()` before, e.g. [the 2020 core-talk thread ending here](https://epics.anl.gov/core-talk/2020/msg01335.php), but I think we verbally decided at the time we didn't want to add yet another set of string routines, since these don't solve all of the problems either.

I agree with @dirk-zimoch that proper UTF-8 support belongs in 7.1. See the new [Road-map Wiki page](/epics-base/epics-base/wiki/Road%E2%80%90map-for-EPICS-7.1) that I just created.

ralphlange: > I think utf-8 support should go on the EPICS 7.1 wish list. It is out of scope here.

Obviously.
I meant it as a general remark: Instead of adding just another stoopid thin boilerplate layer, let's do something actually useful. (In 7.1, of course.)

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,99 @@
...
+    s_initendflag = 1;
+}
+
+static struct cmditem* newItem(char* cmd)
+{
+    struct cmditem* item = mallocMustSucceed(sizeof(struct cmditem) + strlen(cmd) + 1,
+                                             ERL_ERROR " atInit: "
+                                                       "failed to allocate memory for cmditem\n");
+    item->cmd = (char*)(item + 1);
+    strncpy(item->cmd, cmd, strlen(cmd) + 1);
```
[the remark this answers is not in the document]
jerzyjamroz: Fixed in the last commit.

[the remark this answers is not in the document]
mdavidsaver: @jerzyjamroz I would be happy to ignore this and similar codacy complaints as it is easy for me to see that there is no issue since the allocation size is apparent only a couple of lines above.

modules/libcom/src/iocsh/atInit.c
```diff
@@ -0,0 +1,101 @@
...
+
+static struct cmditem* newItem(char* cmd)
+{
+    size_t cmd_len = strlen(cmd) + 1;
+
+    struct cmditem* item = mallocMustSucceed(sizeof(struct cmditem) + cmd_len,
+                                             ERL_ERROR " atInit: "
+                                                       "failed to allocate memory for cmditem\n");
+    item->cmd = (char*)(item + 1);
+    strncpy(item->cmd, cmd, cmd_len);
```
mdavidsaver: I think this present code is fine.

To be optimally optimal, since the sizes are all known after `strlen()`, `memcpy()` is best so long as the trailing nil is accounted for.

```c
size_t cmd_len = strlen(cmd) + 1;
struct cmditem* item = mallocMustSucceed(sizeof(struct cmditem) + cmd_len, "atInitHook");
item->cmd = (char*)(item + 1);
memcpy(item->cmd, cmd, cmd_len);
```

Also, no error formatting should be used with `mallocMustSucceed()`, the second argument is meant to be just an identifier.  Inserting `ERL_ERROR` into the message printed from `mallocMustSucceed()` would be reasonable, but a separate change.

https://github.com/epics-base/epics-base/blob/144f9756eac4e800f18d3d9a35f0d42a712614fd/modules/libcom/src/misc/cantProceed.c#L43-L47

mdavidsaver: > Inserting `ERL_ERROR` into the message printed from `mallocMustSucceed()`

1d19ba4cc2e8c9ab17288b684022a0682e398bde, which also updates the doc commends for `*MustSucceed()`.

anjohnson:
Have you looked at the errors from the older MS compilers on our Appveyor Windows builds? I suspect there's some syntax you're using that they might not like.

modules/database/src/ioc/misc/afterInit.c
mdavidsaver: Unresolving this sub-thread due to MSVC vs2010 and vs2012 build failures.

```
c:\projects\epics-base\modules\libcom\src\iocsh\atinit.c(70) : error C2059: syntax error : '{'
c:\projects\epics-base\modules\libcom\src\iocsh\atinit.c(70) : error C2059: syntax error : ','
```

@anjohnson @ralphlange  @FreddieAkeroyd

Do we in fact care about these decade+ old MSVC versions enough to ask for a change?

jerzyjamroz:
> Have you looked at the errors from the older MS compilers on our Appveyor Windows builds?

I can check that only the next day as after the commit it is always in ```Queued``` for all jobs.

modules/database/src/ioc/misc/afterInit.c
FreddieAkeroyd: I feel they could probably be dropped now, I recall Mark Rivers needing to keep VS2010 compatibility in the past but i don't know for what particular environment that was and if it is still the case

anjohnson: They would have been for some areaDetector cameras or other drivers which call libraries from the manufacturer built with those compilers. Dropping them seems reasonable at this point, I just wrote a "last call" email to tech-talk about them.

jerzyjamroz:
> ❌ [Build EPICS Base 7 base-7.0-1489 failed](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/51529991) (commit [3d3af2e052](https://github.com/epics-base/epics-base/commit/3d3af2e052) by @jerzyjamroz)

It looks like an issue with AppVeyor.

ralphlange:
One of the two failures - the build hanging and timing out after 1hr - is known to be happening and we haven't found the underlying issue.

The other failure (first a compilation error [here](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/51529991/job/m7gyipb3yd2sfhj8#L12449) then - as a consequence - a missing library [there](https://ci.appveyor.com/project/epics-base-7/epics-base/builds/51529991/job/m7gyipb3yd2sfhj8#L12895)) looks different.
Not related to your changes, but not really like a typical AppVeyor problem, either.

mdavidsaver:
> The other failure ... looks different.

Something of an understatement.  I think this is `cl` hitting the windows equivalent of `SIGSEGV`.

```
HostX64/x64/cl: Bad address
make[4]: *** [C:/projects/epics-base/configure/RULES_BUILD:268: pdb.obj] Error -1073741819
```

I have re-started the failed jobs.  We will see what happens...

jerzyjamroz:
If you prefer some changes, then all comments are welcome.
I think this CodeQl improvement may raise a concern:
```
const size_t cmd_len = strnlen(cmd, 32768 - 1) + 1;
```
I selected 32768 as the maximum limit for IOC shell command arguments unless there is an EPICS macro that defines those limits.

mdavidsaver:
> const size_t cmd_len = strnlen(cmd, 32768 - 1) + 1;

Please just use `strlen()`.  `cmd` will be NULL or a nil terminated string.  If the linter complains, then we will ignore it.

mdavidsaver:
As a general comment, static code analysis tools are meant to help us.  But they are not perfect, and imo. developers need to understand what the tool is triggering on before acting, then exercise judgement about whether this is a valid concern, and decide on an appropriate response.  Which may well be ignoring.

Looking more closely at this case:

```c
    const size_t cmd_len = strnlen(cmd, 32768 - 1) + 1;
    struct cmditem *const item = mallocMustSucceed(sizeof(struct cmditem) + cmd_len, "atInit");
    item->cmd = (char *)(item + 1);
    memcpy(item->cmd, cmd, cmd_len);
```

What would happen if a very long command string were somehow introduced?

The `strnlen()` call would truncate, and `memcpy()` would then not copy in a trailing nil.  So an attempt to address the Codacy warnings about handling of a possible unterminated C string in fact introduces just such a bug.  oops

Reading `32768` triggers my mental "magic number" alarm.  Usage of `strnlen()` requires the developer to have the additional information about the maximum input buffer length.  This is information is not encoded in the source, which is why the Codacy linter does not know.  We as developers need to recognize this lack of information in code, and seek it out in other contexts.

@jerzyjamroz I hope you do not see the above unduly harsh on you specifically.  I have been annoyed by Codacy a few times recently, and you have the misfortune of providing the proverbial [straw](https://en.wikipedia.org/wiki/Straw_that_broke_the_camel%27s_back).

mdavidsaver:
This may be a reasonable occasion to promote [a great podcast](https://www.pushkin.fm/podcasts/cautionary-tales/the-thrill-and-the-drop-a-first-date-rollercoaster).  One message of that episode is "don't second guess the machine unless you know more that it does".  Here I would turn that around.  As programmers, we need to know more so that we can second guess the machine.

Ok.  End for rant...

jerzyjamroz:
@mdavidsaver , Thanks for the detailed explanation!

mdavidsaver (approved):
Ok.  I think this PR is ready to go!

anjohnson:
Core Group: This functionality looks good now, could we add a test for this functionality, and a Release Note entry to document it please?

For the Release Note, see the #628 PR on how to add it in a separate file.

jerzyjamroz:
The file ```atInitTest.cpp``` has been added to the tests.

jerzyjamroz:
The file ```documentation/new-notes/PR-558.md``` has been added as requested.

documentation/new-notes/PR-558.md
```diff
@@ -0,0 +1,24 @@
+### New ```atInit``` IOC Shell Command Added
```
anjohnson: Inline `code` formatting should use single back-quotes, using 3 of them is intended for multi-line code blocks so some Markdown formatters might not output what you're expecting.

documentation/new-notes/PR-558.md
```diff
@@ -0,0 +1,24 @@
...
+
+The pull request ```PR #558``` introduces a new IOC shell command `atInit` to EPICS Base, which allows users to schedule arbitrary commands to be executed automatically after the IOC initialization phase (`iocInit`).
```
anjohnson: Please word this assuming the PR has already been merged and it's being read by someone looking at the 7.0.10 release. You can mention the PR, in which case I would link to it as well, something like:

```markdown
This release incorporates [PR #558](https://github.com/epics-base/epics-base/pull/558)
which added a new IOC shell command `atInit`. This command allows startup scripts to
schedule arbitrary commands to be executed automatically after the IOC initialization
phase (`iocInit`).
```

documentation/new-notes/PR-558.md
```diff
@@ -0,0 +1,24 @@
...
+
+The module allows you to write better-structured ioc shell files to include in your startup scripts without tracking where iocInit is located (and how IOC is deployed) e.g.:
```
anjohnson: "The module"?

Maybe you could mention the other implementations of this functionality that this command can replace; look back in the PR comments for them.

documentation/new-notes/PR-558.md
```diff
@@ -0,0 +1,24 @@
...
+- to achieve the best maintainability,
+- to improve writing boot sequences,
+- to improve IOC startup flexibility and scripting capabilities.
+
+#### Features
+
+- Define commands that run after IOC initialization completes.
+- Overwrites ```autosave``` to ensure proper configuration of the required PV.
```
anjohnson: I don't see anything in the code related to autosave, this line doesn't make sense.

documentation/new-notes/PR-558.md
```diff
@@ -0,0 +1,24 @@
...
+- Supports any valid IOC shell command as an argument.
+- Example usages:
+  - `atInit "dbpf <PV> <VAL>"`
+  - `atInit "date"`
```
anjohnson: You might want to give an example where the command to be run takes a string argument and the string contains one or more spaces, showing how to escape double-quotes or use single quotes for the inner argument instead. Please test that, because I am concerned that it might come up against #362 which we haven't fixed properly IIRC, so escaping the inner double-quotes like `\"` might not work properly. The commands `dbl` or `dbgrep` would be suitable as they can both take a space-separated list of field names.

documentation/new-notes/PR-558.md
```diff
@@ -0,0 +1,24 @@
...
+
+#### Testing
+
+- Unit tests have been added to verify:
+  - Registration and execution of commands scheduled with `atInit`.
+  - Proper handling of empty or invalid commands without crashing the IOC.
+  - Execution order and environment variable effects.
```
anjohnson: We don't normally document what the unit tests actually check, at least not in the Release Notes. I won't insist that you delete this part if you want to include it, but it could easily get out of date in a year or two's time.

anjohnson (changes requested):
Sorry that my comments below are longer than the notes that I'm commenting on. These are generally minor, but more people will read the notes than will look at the code so I think they're important.

Thanks!

documentation/new-notes/PR-558.md
jerzyjamroz: The behavior related to `autosave` is inherited because the `atInit` commands run after the `initHookAfterIocRunning` hook, which is triggered after both iocInit and autosave complete.
I updated the description to:
`- Executes following `iocInit` and `autosave` initialization (important for proper PV configuration).`

documentation/new-notes/PR-558.md
jerzyjamroz: Done, I added: `atInit "dbpf $(P)EvtClkSource-Sel 'Upstream (fanout)'"`

documentation/new-notes/PR-558.md
jerzyjamroz: Removed.

jerzyjamroz:
@anjohnson, thank you for your comments - I’m learning from your feedback as well. I have addressed all your points and updated the PR accordingly.

documentation/new-notes/PR-558.md
jerzyjamroz: And, I added the [relevant test](https://github.com/epics-base/epics-base/blob/5d7330476046259c4c29b9b13ab30ccf01c7895d/modules/libcom/test/atInitTest.cpp#L36) as well.

anjohnson:
Core Group: Final command rename to "afterIocRunning" please.

jerzyjamroz:
> Core Group: Final command rename to "afterIocRunning" please.

Done.

anjohnson (approved):
Thanks!

jerzyjamroz:
The next step would be to evaluate whether implementing an `initHookAtShutdown` with `SIGTERM` support (which is not currently supported) is a good idea ([more info](https://github.com/epics-base/epics-base/pull/558#issuecomment-2547950355)), or not? Let me know your opinion.

dirk-zimoch:
`initHookAtShutdown` should probably run its items in reverse order.
I think the implementation of a `SIGTERM` handler calling `epicsExit()` is a different topic. It may change behavior a lot. Stuff may hang inside `epicsExit()`. Thus, the first thing `epicExit()` should do is to uninstall the `SIGTERM` handler.

dirk-zimoch:
Also handle the case that modules may install their own `SIGTERM` handler. Maybe make sure EPICS installs its `SIGTERM` handler relatively late and checks if other handlers are already installed and then makes sure those are called too?

mdavidsaver:
> `initHookAtShutdown` with `SIGTERM` support (which is not currently supported) is a good idea ([more info](https://github.com/epics-base/epics-base/pull/558#issuecomment-2547950355)), or not? Let me know your opinion.

I would like to see some concrete use cases.  What I have heard so far, disabling IRQs, seem to me like cases where a driver `epicsExit()` hook should be preferred.

Are there cases where an some a driver lacks sufficient information for a complete shutdown?

jeonghanlee:
> Thanks!

When do we have the 7.0.10 base release date? I love to have this functionality!

jerzyjamroz:
@mdavidsaver 

> a driver `epicsExit()` hook should be preferred.

At the moment, `epicsExit()` is never executed if one `kill`s the IOC process.
I agree that if one designs the driver, one can handle the `SIGTERM` in it. But then, every implementation would differ. The idea behind this `"shutdown hook"` is to have a unified solution.
One solution to elaborate is to register another hook linked to `SIGTERM`, e.g. `initHookAtSigterm`, but then it does not handle the scenario if one exits IOC.

`epicsExit() hook` - I am not sure how to understand it explicitly.

> I would like to see some concrete use cases.

Every time one wants to ensure that hardware is properly disabled while IOC is not running, so the safe maintenance work of a certain system can start, e.g. 
```
initHookAtShutdown "dbpf $(P)SoftSeq0Disable-Cmd 1”
initHookAtShutdown "dbpf $(P)SoftSeq1Disable-Cmd 1"

initHookAtShutdown "dbpf $(P)Ena-Sel Disabled"
```

anjohnson:
The `epicsAtExit()` API was provided for EPICS code to use for shutdown events, and most existing drivers already register with it; adding a second API for what looks like almost identical functionality doesn't seem like a good idea.

Has anyone tried catching `SIGTERM` or friends and having that call `epicsExitLater()`, assuming that's legal from a `signal()` handler? Please discuss the result in a separate issue or PR.

mdavidsaver:
> call `epicsExitLater()`, assuming that's legal from a `signal()` handler?

It is not.  (at least so far)

Neither `epicsExit()` nor `epicsExitLater()` can safely be called from a signal handler.

dirk-zimoch:
I have done that. Totally illegal. Totally works. I have blocked the signal for all threads except the main thread (I think, need to look up the implementation) because calling `epicsExit()` from a random thread looked too daring. But it is so „unsafe“ code that I never considered to make it a PR.

dirk-zimoch:
See #671.

Issue #557, opened 2024-10-20
Inconsistency with CP link and asynchronous record


mdavidsaver:
The following contrived DB example ends with `a:done.VAL == 0`.  I expect that the final result should be `3` (`1|2`), however the initial update of the `a:calc.INPB` CA link is dropped since that record is still processing as a result of `a:one.PINI`.

To my mind this is a bug since it violates the guarantee of eventual consistency made for EPICS subscriptions.

```
record(longout, "a:one") {
    field(PINI, "YES")
    field(VAL, "1")
    field(FLNK, "a:calc")
    field(TPRO, "1")
}
record(longout, "a:two") {
    field(VAL, "2")
    field(TPRO, "1")
}
record(calcout, "a:calc") {
    field(INPA, "a:one MS")
    field(INPB, "a:two CP MS")
    field(CALC, "A|B")
    field(ODLY, "1")
    field(OUT , "a:done PP MS")
    field(TPRO, "1")
}
record(longin, "a:done") {
    field(TPRO, "1")
}
```

```
$ softIoc -d async-calink.db 
Starting iocInit
############################################################################
## EPICS R7.0.8.2-DEV
## Rev. R7.0.8.1-146-g5fa58673ab437f74fd36-dirty
## Rev. Date Git: 2024-10-20 09:34:51 -0700
############################################################################
_main_: dbProcess of 'a:one'
_main_: dbProcess of 'a:calc'
iocRun: All initialization complete
scanOnce: dbProcess of Active 'a:calc' with RPRO=0
epics> cbLow: dbProcess of 'a:done'
epics> dbgf a:done
DBF_LONG:           0 = 0x0   
epics>
```

mdavidsaver:
I see this behavior as related to what I addressed with `dbPutField()` and `PUTF / RPRO` (https://code.launchpad.net/~epics-core/epics-base/+git/prop-putf/+merge/336468) to allow chaining of async. records following a remote PUT.

In short, I think that the `scanOnceCallback()` calls made from `dbCa.c` should somehow set `PUTF` (or `RPRO`) to trigger the re-processing behavior if the target record is busy.

This would in effect generalize `PUTF` to mean that processing was initiated by an event external to the local database.

anjohnson:
Why are you setting `a:calc.ODLY`?

[...]

PR #559, opened 2024-10-23
CP link RPRO

mdavidsaver:
Attempt to address #557 by setting PUTF/RPRO for CP link processing resulting from a libca event.

~~Add `scanOnceCallback4()` with flag field and flag `SCAN_ONCE_PUTF` to accomplish this.~~

Move `dbProcess()` to dbCa worker.

### DIFF
--- a/modules/database/src/ioc/db/dbCa.c
+++ b/modules/database/src/ioc/db/dbCa.c
@@ -115,10 +115,9 @@ static int dbca_chan_count;
  * During link modification or IOC shutdown the pca->plink pointer (guarded by caLink.lock)
  * is used as a flag to indicate that a link is no longer active.
  *
- * References to the struct caLink are owned by the dbCaTask, and any scanOnceCallback()
- * which is in progress.
+ * References to the struct caLink are owned by the dbCaTask.
  *
- * The libca and scanOnceCallback callbacks take no action if pca->plink==NULL.
+ * The libca callbacks take no action if pca->plink==NULL.
  *
  *   dbCaPutLinkCallback causes an additional complication because
  *   when dbCaRemoveLink is called the callback may not have occured.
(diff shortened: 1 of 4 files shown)


mdavidsaver:
As a note, I considered (and might prefer) an alternate approach of moving the `dbProcess()` from the Once worker to the dbCa worker.  I didn't try this initially due to `dbAccess.h` vs. `db_access.h` complications...

mdavidsaver:
> As a note, I considered (and might prefer) an alternate approach of moving the `dbProcess()` from the Once worker to the dbCa worker. I didn't try this initially due to `dbAccess.h` vs. `db_access.h` complications...

I re-worked this PR as described, and like the result better.

The first iteration still exists as https://github.com/mdavidsaver/epics-base/tree/cp-link-rpro-take1

anjohnson:
Agree this looks a lot cleaner, does it completely resolve the problems you’ve been seeing?

mdavidsaver:
Rebased.

> ... does it completely resolve the problems you’ve been seeing?

Yes.

modules/database/src/ioc/db/dbCa.c
```diff
@@ -1273,6 +1246,13 @@ static void dbCaTask(void *arg)
                     printLinks(pca);
                 }
             }
+            if (link_action & CA_DBPROCESS) {
+                dbCommon *prec;
+                epicsMutexMustLock(pca->lock);
+                prec = pca->plink->precord;
+                epicsMutexUnlock(pca->lock);
+                db_process(prec);
```
mdavidsaver, line 1254: I missed calls to `dbScanLock()/dbScanUnlock()` here.

mdavidsaver: Symptom is occasional `Logic Error: processTarget 1` from `dbDbLink.c` when several `dbPutLink()` calls race.

mdavidsaver: Corrected with 8735a7b17cce8b2d39c6a9bcefa5ece005e64623

mdavidsaver:
As a note.  I have encountered a large IOC with lots of (unintentional) CA links which overflows the Once queue at the default size with Base 7.0.9, and not with this PR.

PR #560, opened 2024-10-24
replaced the html version of msi document with markdown.

ttkorhonen:
Replaced the html version with Markdown. No changes in content (at least intentional), some formatting and link targets for future use (easier referencing from other documents on readthedocs.) 
As submitted, this PR will remove the html version. The PR can be modified if the html should still be kept.


mdavidsaver (approved):
I emphatically approve!

modules/database/src/ioc/dbtemplate/msi.md
```diff
@@ -0,0 +1,384 @@
+# msi: Macro Substitution and Include Tool
+
+(msitool)=
+## Introduction
+
+msi is a general purpose macro substitution/include tool.
+It accepts as input an ascii template file. It looks for lines containing two reserved
+command names: `include` and `substitute`. It also looks for and performs
+substitutions on macros of the form `\$(var)` and `\${var}`. It uses the
```
anjohnson: Putting back-slashes `\` inside back-tick quotes like this leaves the back-slashes in the output, which the [HTML version](https://epics.anl.gov/base/R7-0/8-docs/msi.html) doesn't have and would be confusing.

modules/database/src/ioc/dbtemplate/msi.md
```diff
@@ -0,0 +1,384 @@
+# msi: Macro Substitution and Include Tool
+
+(msitool)=
```
anjohnson, line 3: This `(msitool)=` string doesn't look like normal Markdown syntax to me, does the Sphinx Markdown processor understand it?

modules/database/src/ioc/dbtemplate/msi.md
```diff
@@ -0,0 +1,384 @@
...
+implements.
+
+msi also allows substitutions to be specified via a separate
+substitution file. This substitution file allows the same format as the
+substitution files accepted by the EPICS IOC's dbLoadTemplate command.
+
+
+## Command Syntax
+
+`msi -V -g -o _outfile_ -I _dir_ -M _subs_ -S _subfile_ _template_`
```
anjohnson: Markdown doesn't recognize underscore characters `_` inside back-ticks to generate italics in code-quotes. I would just leave off the underscores around these parameter names.

modules/database/src/ioc/dbtemplate/msi.md
```diff
@@ -0,0 +1,384 @@
...
+    Multiple macro values can be specified in one substitution
+    parameter, or in multiple -M parameters. For example:
+
+        msi -M "a=aval,b=bval" -Mc=cval template
+
+    specifies that in the template file each occurrence of:
+
+        $(a) or ${a} is replaced by aval
+        $(b) or ${b} is replaced by bval
+        $(c) or ${c} is replaced by cval
```
anjohnson: These 3 lines are being formatted as a code-block literal because of the indent. Making them an indented bulleted list like this should prevent that:

        + `$(a)` or `${a}` is replaced by _aval_
        + `$(b)` or `${b}` is replaced by _bval_
        + `$(c)` or `${c}` is replaced by _cval_

resulting in:

+ `$(a)` or `${a}` is replaced by _aval_
+ `$(b)` or `${b}` is replaced by _bval_
+ `$(c)` or `${c}` is replaced by _cval_

modules/database/src/ioc/dbtemplate/msi.md
```diff
@@ -0,0 +1,384 @@
...
+    {
+        c="\"cc\""
+        b=bb
+        a="aa"
+    }
+
+Within a substitutions file, the file name may appear inside double
+quotation marks; these are required if the name contains certain
+characters or environment variable macros of the form \${ENV_VAR} or
+\$(ENV_VAR), which will be expanded before the file is opened.
```
anjohnson: For display by GitHub either the underscore in `ENV_VAR` needs a back-slash to escape it or those two macros should be inside back-ticks (without escaping the dollar sign).

anjohnson (changes requested):
Some small formatting fixes to the Markdown. If you ask GitHub to show this file as a Rich Diff you'll see what it thinks the formatted document should look like. To do that, in the right of the file's header in the "Files changed" tab click the Page icon just left of the "Viewed" check-box.

With those applied this should be good, thanks!

modules/database/src/ioc/dbtemplate/msi.md
ttkorhonen: Readthedocs (sphinx) handles these without escaping the underscore. But I can insert back-ticks.

modules/database/src/ioc/dbtemplate/msi.md
ttkorhonen: OK.

modules/database/src/ioc/dbtemplate/msi.md
ttkorhonen: Same as above, back-ticks inserted later.

modules/database/src/ioc/dbtemplate/msi.md
ttkorhonen: This is a MyST directive. See https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#cross-references
This makes cross-referencing easier and more robust.

modules/database/src/ioc/dbtemplate/msi.md
ttkorhonen: OK, I put the back-ticks in later and missed that effect.

modules/database/src/ioc/dbtemplate/msi.md
ttkorhonen: Another option would be to leave out the back-ticks and keep the italics. But let us go with the back-ticks.

anjohnson (approved):

PR #565, opened 2024-11-13
Add ABORT_ON_ASSERT flag to CONFIG_SITE

JJL772:
This flag causes EPICS to call abort() on assertion failure rather than suspend the executing thread. With the epicsThreadSuspendSelf() behavior, an IOC can end up in a difficult to detect error state where one or more threads has essentially crashed due to an assertion failure. In many cases, it may be desirable to kill the entire IOC instead of a single thread.


mdavidsaver:
Rather than a one size fits all epics-base build option, maybe this would be better as an `ENV_PARAM` like `EPICS_CA_ADDR_LIST` and friends?

JJL772:
AFAIK `getenv` doesn't provide any guarantees when called from C++ static init. So if an assert fails during static initialization, it may not respect the setting in the environment.

This may not really matter for ENV_PARAMs since they provide a default though.

mdavidsaver:
> ... So if an assert fails during static initialization, it may not respect the setting in the environment.

Can you say more about where this concern comes from?  Any links to citations?

JJL772:
> Can you say more about where this concern comes from? Any links to citations?

You can disregard my previous comment, I misread part of the POSIX standard :facepalm: 

> The value of an environment variable is a string of characters. For a C-language program, an array of strings called the environment shall be made available **when a process begins.**

I missed the bolded part... `getenv` should be safe to call during static initialization.

I'll switch this over to use CONFIG_SITE_ENV.

anjohnson:
EPICS sites like APS which build for multiple targets in the same source tree will want to be able to configure different defaults for different targets. Embedded IOCs (VxWorks, RTEMS) may want to try to limp along, whereas Workstation IOCs might be fine with aborting. IMO we should set those as target-specific defaults for the common targets, and the Release Notes entry for this change should show how to change them and/or add more.

The `bldEnvData.pl` script which turns the `CONFIG_SITE_ENV` values into a file `envData.c` for compilation currently reads up to 3 files to get the default parameter values: `configure/CONFIG_ENV`, `configure/CONFIG_SITE_ENV` and `configure/os/CONFIG_SITE_ENV.$(T_A)` (the last one is specific to each target architecture). We don't ship any instances of that last file, but it will be used if present.

I'm wondering if we should make it read yet another file for per OS-class defaults; the most complicated part of doing that might be deciding what the filename should be, since `bldEnvData.pl` already knows the OS class. This addition would implement reading files `configure/os/CONFIG_SITE_ENV.$(OS_CLASS)` before the per-target one, allowing overrides:

```patch
diff --git a/modules/libcom/src/env/bldEnvData.pl b/modules/libcom/src/env/bldEnvData.pl
index f7b4c7497..ceac02236 100644
--- a/modules/libcom/src/env/bldEnvData.pl
+++ b/modules/libcom/src/env/bldEnvData.pl
@@ -55,6 +55,11 @@ close SRC;
 #
 my @configs = ("$config/CONFIG_ENV", "$config/CONFIG_SITE_ENV");
 
+if ($opt_s) {
+    my $config_arch_env = "$config/os/CONFIG_SITE_ENV.$opt_s";
+    push @configs, $config_arch_env
+        if -f $config_arch_env;
+}
 if ($opt_t) {
     my $config_arch_env = "$config/os/CONFIG_SITE_ENV.$opt_t";
     push @configs, $config_arch_env
```

mdavidsaver:
> EPICS sites like APS which build for multiple targets in the same source tree will want to be able to configure different defaults for different targets.

Fair enough.  To my mind though, this added use case need not be a blocker on this PR.  I think it would be acceptable to add an `ENV_PARAM`, with a default to the present suspend behavior.  Then, a site like APS could add the OS_CLASS specific behavior when/if it became desirable.

anjohnson:
> APS could add the OS_CLASS specific behavior when/if it became desirable.

Fine by me, discussion about what the default behavior should be on specific OSs can happen without blocking this PR, and I'll be happy to apply my change to `bldEnvData.pl` and add the necessary `configure/os/CONFIG_SITE_ENV.*` files afterwards.

@JJL772 please add `EPICS_ABORT_ON_ASSERT` to `configure/CONFIG_SITE_ENV` with a value of `NO`, and also add it to `modules/libcom/src/env/envDefs.h` as another `ENV_PARAM`. The code in `osdAssert.c` should call `envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &doAbort)` to [fetch the value](https://docs.epics-controls.org/projects/base/en/latest/envDefs_h.html#_CPPv421envGetBoolConfigParamPK9ENV_PARAMPi) at runtime, so no compile-time macro will be needed.

JJL772:
Thanks for the suggestions @anjohnson, I've applied those changes.

configure/CONFIG_SITE
```diff
@@ -178,5 +178,9 @@ LINKER_USE_RPATH = YES
 # All root directories are considered to be the same.
 LINKER_ORIGIN_ROOT = $(INSTALL_LOCATION)
 
+# Call abort() on assertion failure instead of suspending the current thread
+#  Must be either YES or NO.
+#ABORT_ON_ASSERT = YES
+
```
anjohnson: These are no longer needed, please remove.

modules/libcom/src/osi/Makefile
```diff
@@ -102,6 +102,10 @@ Com_SRCS += osdSignal.cpp
 
 Com_SRCS += osdEnv.c
 
+ifeq ($(ABORT_ON_ASSERT),YES)
+osdAssert_CPPFLAGS += -DABORT_ON_ASSERT
+endif
+
```
anjohnson: Not needed.

modules/libcom/src/osi/os/default/osdAssert.c
```diff
@@ -50,6 +52,12 @@ void epicsAssert (const char *pFile, const unsigned line,
     errlogPrintf("Please E-mail this message to %s or to tech-talk@aps.anl.gov\n",
         pAuthorName);
 
-    errlogPrintf("Calling epicsThreadSuspendSelf()\n");
-    epicsThreadSuspendSelf ();
+    if (envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &shouldAbort) == 0 && shouldAbort) {
+        errlogPrintf("Calling abort()\n");
```
anjohnson, line 56: Please use `printf(stderr, ...)` here, the process is about to die and errlog might not wake up in time to do anything.

anjohnson (commented):
Core Group: This looks good, just some cleanup comments to finish it off.

Thanks!

modules/libcom/src/osi/os/default/osdAssert.c
JJL772: Do you want the rest of the `errlogPrintf`'s changed in this function? It might be better to use `errlogFlush()` before `abort()`.

mdavidsaver: > Do you want the rest of the `errlogPrintf`'s changed in this function?

This is a fair point.  The earlier usages of errlog are just fine in the suspend case, but will also likely be lost in the abort case.

> It might be better to use `errlogFlush()` before `abort()`.

Agree.

modules/libcom/src/osi/os/default/osdAssert.c
```diff
@@ -50,6 +52,13 @@ void epicsAssert (const char *pFile, const unsigned line,
     errlogPrintf("Please E-mail this message to %s or to tech-talk@aps.anl.gov\n",
         pAuthorName);
 
-    errlogPrintf("Calling epicsThreadSuspendSelf()\n");
-    epicsThreadSuspendSelf ();
+    if (envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &shouldAbort) == 0 && shouldAbort) {
+        errlogPrintf(stderr, "Calling abort()\n");
```
mdavidsaver:
```
  ../osi/os/default/osdAssert.c:56:22: error: passing argument 1 of ‘errlogPrintf’ from incompatible pointer type [-Werror=incompatible-pointer-types]
     56 |         errlogPrintf(stderr, "Calling abort()\n");
        |                      ^~~~~~
        |                      |
        |                      FILE *
  In file included from ../error/errlog.h:33,
                   from ../error/epicsPrint.h:16,
                   from ../osi/os/default/osdAssert.c:16:
  ../error/errlog.h:100:58: note: expected ‘const char *’ but argument is of type ‘FILE *’
    100 | LIBCOM_API int errlogPrintf(EPICS_PRINTF_FMT(const char *pformat), ...)
        |                                              ~~~~~~~~~~~~^~~~~~~

```

JJL772: I had changed that to an `fprintf` and then back to `errlogPrintf`...fixed now.

mdavidsaver:
> Core Group: This looks good, just some cleanup comments to finish it off.

I think these have now been done.

mdavidsaver:
Merged as 721e9cc3a7e388903567eb0490293857f53db159

PR #566, opened 2024-11-14
Clear NAMSG together with NSTAT and NSEV

exzombie:
Previously, there was no way to clear AMSG when setting a no-alarm status: it kept the last error message, and it is strange to have a record that has no alarm while showing an error. With this change, the NAMSG field is treated the same as NSTA and NSEV: they are cleared when their content is moved to STAT, SEVR, and AMSG.


anjohnson:
This change makes perfect sense to me, I would go so far as to say the current code is buggy.

mdavidsaver:
Look correct.

PR #568, opened 2024-11-19
Propagate AMSG through MSS links

### LINKED ISSUE #567: Propagating AMSG through links
I'd like to attach an error message to alarm status of records whose values come from an aSub. I treat aSub records as "private", in the sense that users have no business looking at or touching their fields. They only interact with records that are linked from an aSub and represent its inputs and outputs. When the subroutine encounters an error, I set its alarm status and use the MSS attribute on output links to make the problem visible to clients.

In cases where a situation that a message needs to be shown to the user arises often, we have always used a dedicated message record, with the aSub writing to it through one of its outputs. However, for "this should (almost) never happen" kind of situations, we just write to the IOC log and tell people to go look there. They shouldn't need to do it too often. However, it would be *really* convenient to be able to use AMSG for this purpose and let people look at that field instead of having to look at the IOC console.

But, as noted above, the person seeing the alarm will see it on the interface records, and is probably not even aware that there is a subroutine generating that alarm. For this reason, it would be good to be able to propagate AMSG through a link. I propose to extend the behavior of the MSS link attribute to do that. I don't think it's necessary to introduce another attribute: if you are overwriting STAT, you either also like overwriting AMSG, or you just don't care about AMSG. That's my view, at least.

And that's why I'm opening this issue instead of a pull request: looking at the code, extending `recGblInheritSevr()` doesn't look difficult, but I would first like to know whether others share my view of using the MSS attribute for this.

exzombie:
Fixes #567

MS and MSI links do not propagate STAT and therefore do not propagate AMSG, either. CA, CP and CPP links also do not propagate AMSG, but the reason is technical: the message is not available over Channel Access. Apparently, it is available over PVA (`pvmonitor` shows it without doing anything special) so it should be possible to propagate it over PVA links, but unless I'm mistaken that belongs into the pvxs module.

The new `recGblInheritSevrMsg()` function is not a formatting function (unlike `recGblSetSevrMsg()`) because it is supposed to be used only for inheriting the already rendered message and nothing else.

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -22,6 +22,16 @@ should also be read to understand what has changed since earlier releases:
 
 ## Changes made on the 7.0 branch since 7.0.8.1
 
+### The AMSG error message propagates through MSS links
+
+A database link with the MSS attribute will now propagate not only SEVR and
+STAT, but also AMSG. This field contains additional information that complements
+STAT. Links with MS or MSI attributes do not propagate STAT, and therefore do
+not propagate AMSG, either.
+
+Channel Access links do not propagate AMSG, regardless of the MSS attribute,
+because the message is not available over Channel Access.
+
 ### DBE_PROPERTY event rate changed
 
 Updating property fields now only post DBE_PROPERTY events if the
--- a/modules/database/src/ioc/db/dbDbLink.c
+++ b/modules/database/src/ioc/db/dbDbLink.c
@@ -226,9 +226,10 @@ static long dbDbGetValue(struct link *plink, short dbrType, void *pbuffer,
     }
 
     if (!status && precord != dbChannelRecord(chan))
-        recGblInheritSevr(plink->value.pv_link.pvlMask & pvlOptMsMode,
+        recGblInheritSevrMsg(plink->value.pv_link.pvlMask & pvlOptMsMode,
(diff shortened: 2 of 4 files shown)


anjohnson:
No idea why `epicsThreadPriorityTest` failed on MacOS, I will retry that one.

```
  epicsThreadPriorityTest.tap .. 
  not ok  6 - epicsEventWaitWithTimeout returned 1
  Failed 1/7 subtests 
```

anjohnson (approved):
Looks good to me, thanks!

mdavidsaver:
The usage looks fine, but the implementation isn't what I expected.  I would think that `recGblInheritSevr()` would become a wrapper around `recGblInheritSevrMsg()` with a `NULL` message.

exzombie:
> The usage looks fine, but the implementation isn't what I expected. I would think that `recGblInheritSevr()` would become a wrapper around `recGblInheritSevrMsg()` with a `NULL` message.

Yeah, I can see why you would expect that, it's a common pattern. It's not obvious for this case, though, because MSS is the exception and the only case that actually propagates the message. It seemed appropriate to emphasize that to the reader of the implementation (the caller has no idea either way). But I guess that can be done with a comment just as well :)

anjohnson:
Core Group: Approved, thanks!

PR #569, opened 2024-11-25
Add debug build configuration for os x mX arch

simon-ess:
Parallel to the darwin-x86 configuration, here is some darwin-aarch64-debug configuration.

### DIFF
--- a/configure/os/CONFIG.darwin-aarch64-debug.Common
+++ b/configure/os/CONFIG.darwin-aarch64-debug.Common
@@ -1,0 +1,11 @@
+# CONFIG.darwin-aarch64-debug.Common
+#
+# Definitions for darwin-aarch64-debug host builds - darwin-aarch64 target build with debug compiler flags
+# Sites may override these definitions in CONFIG_SITE.darwin-aarch64-debug.Common
+#-------------------------------------------------------
+
+include $(CONFIG)/os/CONFIG.darwin-aarch64.Common
+
+# Removes -O optimization and adds -g compile option
+HOST_OPT=NO
+
--- a/configure/os/CONFIG.darwin-x86-debug.Common
+++ b/configure/os/CONFIG.darwin-x86-debug.Common
@@ -1,0 +1,11 @@
+# CONFIG.darwin-x86-debug.Common
+#
+# Definitions for darwin-x86-debug host builds - darwin-x86 target build with debug compiler flags
+# Sites may override these definitions in CONFIG_SITE.darwin-x86-debug.Common
+#-------------------------------------------------------
+
+include $(CONFIG)/os/CONFIG.darwin-x86.Common
+
+# Removes -O optimization and adds -g compile option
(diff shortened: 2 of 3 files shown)


anjohnson:
I suspect we might want to provide a file `CONFIG.darwin-aarch64-debug.darwin-aarch64-debug` too, for building with the right flags when `EPICS_HOST_ARCH` is set to `d-a64-debug` (look at `CONFIG.linux-arm-debug.linux-arm-debug`). That `darwin-x86-debug` doesn't have one probably shows that the `d-x86-debug` target isn't used much (if at all), or that setting `HOST_OPT=NO` doesn't have much effect nowadays.

mdavidsaver:
Core Group: Looks fine.

configure/os/CONFIG.darwin-x86-debug.Common
```diff
@@ -0,0 +1,11 @@
+# CONFIG.darwin-x86-debug.darwin-x86-debug
```
anjohnson: Please fix the filename:

`s/darwin-x86-debug$/Common/`

configure/os/CONFIG.darwin-x86-debug.Common
```diff
@@ -0,0 +1,11 @@
...
+#
+# Definitions for darwin-x86 host - darwin-x86 target build with debug compiler flags
```
anjohnson: The settings in this file (`CONFIG.darwin-x86-debug.Common`) apply to all builds when the host-arch is `darwin-x86-debug`. The wording of the above comment could be confusing because most cross-targets would *not* be built with debug flags. Please use:

`# Definitions for darwin-x86-debug host builds`

configure/os/CONFIG.darwin-aarch64-debug.Common
```diff
@@ -0,0 +1,11 @@
+# CONFIG.darwin-aarch64-debug.darwin-aarch64-debug
```
anjohnson: Please fix the filename:

`s/darwin-aarch64-debug$/Common/`

configure/os/CONFIG.darwin-aarch64-debug.Common
```diff
@@ -0,0 +1,11 @@
...
+#
+# Definitions for darwin-aarch64 host - darwin-aarch64 target build with debug compiler flags
```
anjohnson: The settings in this file (`CONFIG.darwin-aarch64-debug.Common`) apply to all builds when the host-arch is `darwin-aarch64-debug`. The wording of the above comment could be confusing because most cross-targets would _not_ be built with debug flags. Please use:

`# Definitions for darwin-aarch64-debug host builds`

anjohnson (changes requested):
Nit-picking but I think important...

anjohnson (approved):

anjohnson:
Core Group: Approved.

PR #570, opened 2024-11-25
fix recursion bug

dirk-zimoch:
Previously, linking a field of e.g. a calc record to itself (calc.INPA = calc.A) directly or through a chain of other records that support to follow the link for attributes like units, precision caused an infinite recursion and crashed the ioc.

Affected records: calc, calcout, seq, sub and aSub

Also added support for seq link 0 fields which had been added meanwhile.

### DIFF
--- a/modules/database/src/std/rec/aSubRecord.c
+++ b/modules/database/src/std/rec/aSubRecord.c
@@ -302,6 +302,16 @@ static long get_outlinkNumber(int fieldIndex) {
     return -1;
 }
 
+#define do_recursion_safe(func, plink, ...) \
+do { \
+    dbScanLock((dbCommon*)prec); \
+    DBLINK link = *(plink); \
+    (plink)->lset = NULL; \
+    func(&link, __VA_ARGS__); \
+    (plink)->lset = link.lset; \
+    dbScanUnlock((dbCommon*)prec); \
+} while(0)
+
 static long get_units(DBADDR *paddr, char *units)
 {
     aSubRecord *prec = (aSubRecord *)paddr->precord;
@@ -309,12 +319,12 @@ static long get_units(DBADDR *paddr, char *units)
 
     linkNumber = get_inlinkNumber(dbGetFieldIndex(paddr));
     if (linkNumber >= 0) {
-        dbGetUnits(&prec->inpa + linkNumber, units, DB_UNITS_SIZE);
+        do_recursion_safe(dbGetUnits, &prec->inpa + linkNumber, units, DB_UNITS_SIZE);
         return 0;
     }
     linkNumber = get_outlinkNumber(dbGetFieldIndex(paddr));
     if (linkNumber >= 0) {
-        dbGetUnits(&prec->outa + linkNumber, units, DB_UNITS_SIZE);
+        do_recursion_safe(dbGetUnits, &prec->outa + linkNumber, units, DB_UNITS_SIZE);
     }
     return 0;
 }
@@ -322,27 +332,25 @@ static long get_units(DBADDR *paddr, char *units)
 static long get_precision(const DBADDR *paddr, long *pprecision)
 {
     aSubRecord *prec = (aSubRecord *)paddr->precord;
+    long status;
     int fieldIndex = dbGetFieldIndex(paddr);
     int linkNumber;
+    short precision;
 
     *pprecision = prec->prec;
     linkNumber = get_inlinkNumber(fieldIndex);
     if (linkNumber >= 0) {
-        short precision;
-
-        if (dbGetPrecision(&prec->inpa + linkNumber, &precision) == 0)
-            *pprecision = precision;
+        do_recursion_safe(status = dbGetPrecision, &prec->inpa + linkNumber, &precision);
+        if (status == 0) *pprecision = precision;
         return 0;
     }
-
     linkNumber = get_outlinkNumber(fieldIndex);
     if (linkNumber >= 0) {
-        short precision;
-
-        if (dbGetPrecision(&prec->outa + linkNumber, &precision) == 0)
-            *pprecision = precision;
-    } else
-        recGblGetPrec(paddr, pprecision);
+        do_recursion_safe(status = dbGetPrecision, &prec->outa + linkNumber, &precision);
+        if (status == 0) *pprecision = precision;
+        return 0;
+    }
+    recGblGetPrec(paddr, pprecision);
     return 0;
 }
 
@@ -354,16 +362,16 @@ static long get_graphic_double(DBADDR *paddr, struct dbr_grDouble *pgd)
 
     linkNumber = get_inlinkNumber(fieldIndex);
     if (linkNumber >= 0) {
-        dbGetGraphicLimits(&prec->inpa + linkNumber,
-            &pgd->lower_disp_limit,
-            &pgd->upper_disp_limit);
+        do_recursion_safe(dbGetGraphicLimits, &prec->inpa + linkNumber,
+                                &pgd->lower_disp_limit,
+                                &pgd->upper_disp_limit);
         return 0;
     }
     linkNumber = get_outlinkNumber(fieldIndex);
     if (linkNumber >= 0) {
-        dbGetGraphicLimits(&prec->outa + linkNumber,
-            &pgd->lower_disp_limit,
-            &pgd->upper_disp_limit);
+        do_recursion_safe(dbGetGraphicLimits, &prec->outa + linkNumber,
+                                &pgd->lower_disp_limit,
+                                &pgd->upper_disp_limit);
     }
     return 0;
 }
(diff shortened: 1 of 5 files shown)


anjohnson:
There's lot of repetition here, would it be possible to put the recursion checks into dbLink.c or even into dbDbLink.c? That way it would work for out-of-tree record types as well.

mdavidsaver:
> ... linking a field of e.g. a calc record to itself ...

What isn't the existing PACT logic breaking this loop?

dirk-zimoch:
> What isn't the existing PACT logic breaking this loop?

It doesn’t. It fills the stack and then crashes.
This loop does not happen when processing the record but when doing a ca get using a DBR type with meta-data like prec, egu, etc.

dirk-zimoch:
> would it be possible to put the recursion checks into dbLink.c or even into dbDbLink.c

I have to temporarily manipulate the link but dbGetUnits() etc have const link arguments.

To be exact, I have to store the fact somewhere that this link has already been visited. I (ab)use lset for this. I found no other place. I could create a thread-local list or something of already visited links and search that one, but that seemed too complicated.

anjohnson:
I might be Okay with removing `const` from the metadata `dbGetBlah()` routines in dbLink.c, as long as the `struct lset` function pointers aren't touched. Would that make the change smaller & cleaner?

mdavidsaver:
> This loop does not happen when processing the record but when doing a ca get using a DBR type with meta-data like prec, egu, etc.

So this issue can be triggered remotely?

What can I do to replicate this behavior?

mdavidsaver:
> What can I do to replicate this behavior?

@dirk-zimoch Can you attach a .db file which I can use with `softIoc`?

anjohnson:
```
record(calc, "calc") {
  field(CALC, "VAL+1")
  field(INPA, "calc.A")
}
```

Then
```
Cox$ caget -d DBR_GR_DOUBLE calc.A
Read operation timed out: some PV data was not read.
calc.A
    Native data type: DBF_DOUBLE
    Request type:     DBR_GR_DOUBLE
    Element count:    1
    Value:            0
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Units:            
    Precision:        0
    Lo disp limit:    0
    Hi disp limit:    0
    Lo alarm limit:   0
    Lo warn limit:    0
    Hi warn limit:    0
    Hi alarm limit:   0
    Warning: "Virtual circuit disconnect"
    Context: "op=0, channel=calc.A, type=DBR_GR_DOUBLE, count=1, ctx="anlvpn173.aps.anl.gov:5064""
    Source File: ../getCopy.cpp line 91
    Current Time: Mon Nov 25 2024 13:23:05.643728000
Unexpected problem with CA circuit to server "anlvpn173.aps.anl.gov:5064" was "Connection reset by peer" - disconnecting
```
The IOC (darwin-aarch64 in my case) shows:
```
epics> Bus error: 10
Cox$
```
I'm sure pvget would trigger the same bug.

anjohnson:
The link doesn't have to loop directly to itself, any loop of DB links would do this. A CA link does prevent the problem though, and this is a rather unusual configuration.

mdavidsaver:
So the recursion is `dbGet() -> getOptions() -> get_units() -> dbGetUnits() -> dbDbGetUnits() -> dbGet()`.

```
... repeated ...
#25121 dbGet (paddr=(dbAddr*)(0x560f756d62a8, "calc.A"), dbrType=dbrType@entry=10, 
    pbuffer=pbuffer@entry=0x7f238ebff530, options=options@entry=0x7f238ebff520, nRequest=nRequest@entry=0x7f238ebff528, 
    pflin=pflin@entry=0x0) at ../db/dbAccess.c:925
#25122 0x00007f239da19e1c in dbDbGetUnits (plink=<optimized out>, units=0x7f238ebff758 "", unitsSize=16)
    at ../db/dbDbLink.c:335
#25123 0x00007f239da1d4ca in dbGetUnits (plink=<optimized out>, units=units@entry=0x7f238ebff758 "", 
    unitsSize=unitsSize@entry=16) at ../db/dbLink.c:507
#25124 0x00007f239da9f239 in get_units (paddr=<optimized out>, units=0x7f238ebff758 "") at ../rec/calcRecord.c:177
#25125 0x00007f239da0dd51 in getOptions (pflin=0x0, options=0x7f238ebff700, poriginal=<synthetic pointer>, 
    paddr=(dbAddr*)(0x560f756d6200, "calc.A")) at ../db/dbAccess.c:381
#25126 dbGet (paddr=paddr@entry=(dbAddr*)(0x560f756d6200, "calc.A"), dbrType=dbrType@entry=10, 
    pbuffer=pbuffer@entry=0x7f238ebff750, options=options@entry=0x7f238ebff700, nRequest=nRequest@entry=0x7f238ebff708, 
    pflin=pflin@entry=0x0) at ../db/dbAccess.c:925
#25127 0x00007f239da11065 in dbChannelGet (chan=chan@entry=(dbChannel*)(0x560f756d61f8, "calc.A"), type=type@entry=10, 
    pbuffer=pbuffer@entry=0x7f238ebff750, options=options@entry=0x7f238ebff700, nRequest=nRequest@entry=0x7f238ebff708, 
    pfl=pfl@entry=0x0) at ../db/dbChannel.c:630
#25128 0x00007f239da25ce9 in dbChannel_get_count (chan=chan@entry=(dbChannel*)(0x560f756d61f8, "calc.A"), 
    buffer_type=<optimized out>, pbuffer=0x7f239c4790a0, nRequest=nRequest@entry=0x7f238ebffc78, pfl=pfl@entry=0x0)
    at ../db/db_access.c:565
#25129 0x00007f239da51219 in read_reply (pArg=pArg@entry=0x7f238ebffcd0, dbch=(dbChannel*)(0x560f756d61f8, "calc.A"), 
    eventsRemaining=eventsRemaining@entry=1, pfl=pfl@entry=0x0) at ../rsrv/camessage.c:537
#25130 0x00007f239da51b8c in read_notify_action (mp=0x7f238ebffd50, pPayload=0x7f239c475098, client=0x7f2388001508)
    at ../rsrv/camessage.c:719
#25131 0x00007f239da52f6b in camessage (client=client@entry=0x7f2388001508) at ../rsrv/camessage.c:2520
#25132 0x00007f239da50424 in camsgtask (pParm=0x7f2388001508) at ../rsrv/camsgtask.c:116
#25133 0x00007f239d9aea5a in start_routine (arg=0x7f2384006130) at ../osi/os/posix/osdThread.c:442
#25134 0x00007f239d4a81c4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#25135 0x00007f239d527ac0 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:100
```

mdavidsaver:
My first thought is to try to break this loop in `dbDbLink.c` instead of in every record support.  At the point where `dbGet()` would be called.

dirk-zimoch:
> and this is a rather unusual configuration

Very unusual. So much that it went unnoticed for 10 years. Not illegal though. Only pretty useless.

mdavidsaver:
Also, I guess this issue goes all the way back to 1e028d68d885cea30978b0069d6cd6442651bda3 (3.16.0.1)?

dirk-zimoch:
> At the point where dbGet() would be called.

The loop does not necessarily start at the root, where dbGet is called. Imagine a figure 6 instead of 0.

mdavidsaver:
> ... and this is a rather unusual configuration

mmmmm.  I know I have seen, and presented, this pattern in training classes for many years.

```
record(calc, "tick") {
    field(SCAN, "1 second")
    field(CALC, "A+1")
    field(INPA, "tick.A NPP")
}
```

dirk-zimoch:
> So the recursion is dbGet() -> getOptions() -> get_units() -> dbGetUnits() -> dbDbGetUnits() -> dbGet().

Ah, yes. You are right. There is a dbGet call in each loop.

mdavidsaver:
> mmmmm. I know I have seen, and presented, this pattern in training classes for many years.

oh, I see.  Not quite.  The issue is `A` linking to itself.  `A` linking to `VAL` is fine.

dirk-zimoch:
I would use:

record(calc, "tick") {
    field(SCAN, "1 second")
    field(CALC, "A+1")
    field(INPA, "tick NPP")
}

Link to VAL, not to A.

dirk-zimoch:
> as long as the struct lset function pointers aren't touched

Touching lset (and restoring it later) is the whole point — and the trick to break the loop. I protect the record from being processed while its link is crippled with dbScanLock. Once unlocked the link is usable again.

Or do you mean as long as I don‘t change the signature of the functions in lset? That I don‘t need.

anjohnson:
@mdavidsaver The original issue started around 27e80be2e54765000471144b7ff7067f565dd16e (3.15-ish)

@dirk-zimoch Abusing the `plink->lset` pointer like you do seems nasty but the function pointers inside the `struct lset` could still be `const`, which is what I want to avoid. An alternative might be to set a bit in the `plink->flags` for recursion testing, but that still requires the `link *` arguments to `dbGetUnits()` and friends to have the `const` removed from them.

anjohnson:
Ooh, maybe I'm wrong and they can't be const any more. Yuck.

dirk-zimoch:
> would it be possible to put the recursion checks into dbLink.c or even into dbDbLink.c? That way it would work for out-of-tree record types as well.

I can make an alternative PR changing the dbGetBlah() functions to non const plink arguments and manipulate lset there. However, out-of-tree record types would not benefit because they usually don‘t call dbGetBlah() in their get_blah() rset function (and thus neither have this bug nor this feature). But dbGetBlah() would become slower for everyone, not only for records with this feature.
Also, this feature cannot be generalized because we cannot know which field is associated with which input link. Only the record type knows. That would only be possible if we change the (input) link structure (and dbd file syntax) to contain the field the value goes to. With that change, not each record type would have to implement reading links into fields for itself. But that is probably far future (EPICS R8 or higher).

mdavidsaver:
> I can make an alternative PR changing the dbGetBlah() functions to non const plink arguments ...

imo. This would be an unacceptable breaking change to a public API.

To my mind, "cheating" by casting away `const` could be acceptable as an internal detail to Base (where all records and links are allocated).  Instead of messing with `lset`, How about adding another `struct link::flags` bit?  eg. `DBLINK_FLAG_LSACT` as in Link Support ACTive.

The logic in `dbDbLink.c` where `dbGet()` is called would be to instead error if this new flag is already set.  Then set it before calling `dbGet()`, and then clear afterwards.

anjohnson:
Would adding routines that provide recursion protection be possible (without ever having to cast away constness)?

BTW we can't assume that every link field has an associated field that the value gets read into. The printf record has no such fields, and there may be some other more obscure ones that don't immediately spring to mind.

mdavidsaver:
> Also, this feature cannot be generalized because we cannot know which field is associated with which input link.  ...

True, although I think per-link granularity could be sufficient in practice.  This would also prevent a situation where eg. `dbDbGetControlLimits()` somehow calls `dbDbGetGraphicLimits()` on the same link, but I don't think this would be a practical limitation.  (needs more thought through)

mdavidsaver:
> Would adding routines that provide recursion protection be possible (without ever having to cast away constness)?

I see this situation as another special case for DB links, like lock sets, which it isn't practically possible for an external link support to replicate.  If this holds, then I don't see justification for adding to the public `lset` API.


> BTW we can't assume that every link field has an associated field that the value gets read into.

Right, which is why I am suggesting a flag in `struct link`.

dirk-zimoch:
The recommended alternative solution is in #571.

mdavidsaver:
Closed in favor of #571

PR #571, opened 2024-12-02
Recursion bug v2

dirk-zimoch:
Previously, linking a field of e.g. a calc record to itself (calc.INPA = calc.A) directly or through a chain of other records that support to follow the link for attributes like units, precision caused an infinite recursion and crashed the ioc.

Affected records: calc, calcout, seq, sub and aSub

Also added support for seq link 0 fields which had been added meanwhile.

This is an alternative implementation to #570.

### DIFF
--- a/modules/database/src/ioc/db/dbDbLink.c
+++ b/modules/database/src/ioc/db/dbDbLink.c
@@ -232,19 +232,41 @@ static long dbDbGetValue(struct link *plink, short dbrType, void *pbuffer,
     return status;
 }
 
+/* Some records get options (precsision, units, ...) for some fields
+ * from an input link. We need to catch the case that this link
+ * points back to the same field or we will end in an infinite recursion.
+*/
+static long dbDbGetOptionLoopSafe(const struct link *plink, short dbrType,
+    void *pbuffer, long option)
+{
+    /* We need to cast away const to set the flags.
+       That's ok because we know that plink is never actually readonly.
+       And we reset everything to its original state.
+    */
+    struct link *mutable_plink = (struct link *)plink;
+    long status = S_dbLib_badLink;
+    dbChannel *chan = linkChannel(plink);
+    DBADDR *paddr = &chan->addr;
+    long number_elements = 0;
+
+    dbScanLock(paddr->precord);
+    if (!(mutable_plink->flags & DBLINK_FLAG_VISITED)) {
+        mutable_plink->flags |= DBLINK_FLAG_VISITED;
+        status = dbGet(paddr, dbrType, pbuffer, &option, &number_elements, NULL);
+        mutable_plink->flags &= ~DBLINK_FLAG_VISITED;
+    }
+    dbScanUnlock(paddr->precord);
+    return status;
+}
+
 static long dbDbGetControlLimits(const struct link *plink, double *low,
         double *high)
 {
-    dbChannel *chan = linkChannel(plink);
-    DBADDR *paddr = &chan->addr;
     struct buffer {
         DBRctrlDouble
         double value;
     } buffer;
-    long options = DBR_CTRL_DOUBLE;
-    long number_elements = 0;
-    long status = dbGet(paddr, DBR_DOUBLE, &buffer, &options, &number_elements,
-            NULL);
+    long status = dbDbGetOptionLoopSafe(plink, DBR_DOUBLE, &buffer, DBR_CTRL_DOUBLE);
 
     if (status)
         return status;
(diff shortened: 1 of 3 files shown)


modules/database/src/ioc/db/dbDbLink.c
```diff
@@ -232,19 +232,41 @@ static long dbDbGetValue(struct link *plink, short dbrType, void *pbuffer,
...
     dbChannel *chan = linkChannel(plink);
     DBADDR *paddr = &chan->addr;
+    long number_elements = 0;
+
+    dbScanLock(paddr->precord);
+    if (!(mutable_plink->flags & DBLINK_FLAG_VISITED)) {
+        mutable_plink->flags |= DBLINK_FLAG_VISITED;
+        status = dbGet(paddr, dbrType, pbuffer, &option, &number_elements, NULL);
+        mutable_plink->flags &= ~DBLINK_FLAG_VISITED;
+    }
```
mdavidsaver, line 257: This looks good to me.

modules/database/src/std/rec/seqRecord.c
```diff
@@ -282,10 +282,10 @@ static void processCallback(epicsCallback *arg)
 static long get_units(DBADDR *paddr, char *units)
 {
     seqRecord *prec = (seqRecord *) paddr->precord;
-    int fieldOffset = dbGetFieldIndex(paddr) - indexof(DLY1);
+    int fieldOffset = dbGetFieldIndex(paddr) - indexof(DLY0);
 
     if (fieldOffset >= 0)
-        switch (fieldOffset & 2) {
+        switch (fieldOffset & 3) {
```
mdavidsaver, line 288: I am having trouble understanding what the intent of this seqRecord meta-data mapping was and is.

- DLYx <- static `"s"`
- DOLx <- empty string?
- DOx <- dbGetUnits(DOLx)
- LNKx <- dbGetUnits(DOLx)

dirk-zimoch: The function is never called for the links LNKx or DOx. (I think)

dirk-zimoch: The original code ( & 2) depended on the assumption that the function is only called for fields DLYx (fieldOffset & 3 == 0) or DOx (fieldOffset & 3 == 2) but never for the links DOLx or LNKx. Thus, I had tested only & 2. But I found it "safer" to actually test both bits to ensure that the links DOLx and LNKx never return a constant value or call dbGet*() but instead don't return any units, precision, etc. The change is probably not necessary, but after the introduction of DLY0/DO0/DOL0/LNK0 broke something, I wanted to be more robust against future changes.

dirk-zimoch: So the meta-data mapping is:
 * DLYx  <- units=`"s"`; prec=seqDLYprecision
 * DOLx <- default: don't set units (i.e. whatever the caller prepared, usually `""`), prec=PREC (as before)
 * DOx  <- dbGetUnits(DOLx), dbGetPrecision(DOLx)
 * LNKx  <- default: don't set units, prec = PREC

anjohnson:
Nice code simplification, LGTM!

Do we have any unit tests for this feature?

dirk-zimoch:
> Do we have any unit tests for this feature?

Of course not.

modules/database/src/std/rec/seqRecord.c
mdavidsaver: > The function is never called for the links LNKx or DOx. (I think)

I add `printf("%s.%s calling %s\n", paddr->precord->name, paddr->pfldDes->name, __func__);` to each of the `get_*` metadata functions in `seqRecord.c`, load a single seqRecord named `seq`, and then run:

```
caget -d GR_DOUBLE seq.{DLY0,DOL0,DO0,LNK0}
```

which causes `softIoc` to print:

```
seq.DLY0 calling get_units
seq.DLY0 calling get_precision
seq.DLY0 calling get_graphic_double
seq.DLY0 calling get_alarm_double
seq.DOL0 calling get_units
seq.DOL0 calling get_graphic_double
seq.DOL0 calling get_alarm_double
seq.DO0 calling get_units
seq.DO0 calling get_precision
seq.DO0 calling get_graphic_double
seq.DO0 calling get_alarm_double
seq.LNK0 calling get_units
seq.LNK0 calling get_graphic_double
seq.LNK0 calling get_alarm_double
```

mdavidsaver: > So the meta-data mapping is: ...

This seems reasonable for `get_units()`, and indeed the others as well.  So I think your change is fixing a bug introduced by [27e80be2e54765000471144b7ff7067f565dd16e](https://github.com/epics-base/epics-base/commit/27e80be2e54765000471144b7ff7067f565dd16e#diff-1fb0f72ef42fe5df5cf554d9651244eeed42c1ea892245053a965cad9302e0c4R432).

anjohnson:
Core group: Approved.

PR #575, opened 2024-12-15
Allow RTEMS non epics task to set task priority

hjunkes:
If RTEMS is used with the POSIX API, the init task 'POSIX_Init()' is unfortunately given the priority '2'. This corresponds to the second lowest POSIX prio (RTEMS pthread prio 256). In order to give the 'rtems-init-task' the appropriate IOCsh prio a call of 'epicsThreadSetPriority()' is necessary: 'epicsThreadSetPriority(epicsThreadGetIdSelf(), epicsThreadPriorityIocsh);' However, 'POSIX_Init()' itself is not an epics task.

### DIFF
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -949,7 +949,7 @@ POSIX_Init ( void *argument __attribute__((unused)))
      */
 
     /* PosixMaxPrio == 100 */
-    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY * epicsThreadPriorityIocsh / 100;
+    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - ( RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY) * epicsThreadPriorityIocsh / 100;
     rtems_task_priority old;
     sc = rtems_task_set_priority (RTEMS_SELF, mainPrio, &old);
     assert(sc == RTEMS_SUCCESSFUL);
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -957,7 +957,7 @@ POSIX_Init ( void *argument __attribute__((unused)))
 
     param.sched_priority = (sched_get_priority_max(policy)
                             - sched_get_priority_min(policy))
-                         * epicsThreadPriorityIocsh / 100
+                         * epicsThreadPriorityIocsh / (epicsThreadPriorityMax - epicsThreadPriorityMin) 
                          + sched_get_priority_min(policy);
 
     sc = pthread_setschedparam(pthread_self(), policy, &param);
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -947,15 +947,20 @@ POSIX_Init ( void *argument __attribute__((unused)))
      * corresponds to the second lowest POSIX prio (RTEMS pthread
      * prio 254). This task shoud have IOCsh prio.
      */
-
-    /* PosixMaxPrio == 100 */
-    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY
-                                 - (RTEMS_MAXIMUM_PRIORITY - RTEMS_MINIMUM_PRIORITY)
-                                 * epicsThreadPriorityIocsh / 100;
-    rtems_task_priority old;
-    sc = rtems_task_set_priority (RTEMS_SELF, mainPrio, &old);
+    pthread_attr_t attr;
+    struct sched_param  param;
+    int policy;
+    sc  = pthread_attr_init(&attr);
     assert(sc == RTEMS_SUCCESSFUL);
-    printf("Priority changed from %d -> %d\n", old, mainPrio);
+    sc = pthread_attr_getschedpolicy(&attr, &policy);
+    assert(sc == RTEMS_SUCCESSFUL);
+
+    param.sched_priority = (sched_get_priority_max(policy)
+                            - sched_get_priority_min(policy))
+                         * epicsThreadPriorityIocsh / 100;
+
+    sc = pthread_setschedparam(pthread_self(), policy, &param);
+    assert(sc == RTEMS_SUCCESSFUL);
 
     initConsole ();
 
@@ -1037,11 +1042,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
     rtems_bsd_setlogpriority("debug");
     on_exit(default_network_on_exit, NULL);
 
-    /* Let other tasks run to complete background work */
-    sc = rtems_task_set_priority (RTEMS_SELF, RTEMS_MAXIMUM_PRIORITY/2, &old);
-    assert(sc == RTEMS_SUCCESSFUL);
-    printf("Priority changed from %d -> %d\n", old, RTEMS_MAXIMUM_PRIORITY/2);
-
     sc = rtems_bsd_initialize();
     assert(sc == RTEMS_SUCCESSFUL);
 
@@ -1103,10 +1103,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
         printf("time from ntp : %s.%09ld UTC\n", timeBuff, now.tv_nsec);
       }
     }
-
-    sc = rtems_task_set_priority (RTEMS_SELF, mainPrio, &old);
-    assert(sc == RTEMS_SUCCESSFUL);
-    printf("Priority changed at end of network init from %d -> %d\n", old, mainPrio);
 
 #else // Legacy stack, old network initialization
     char *cp;
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -827,15 +827,6 @@ default_network_on_exit(int exit_code, void *arg)
 }
 
 static void
-default_network_set_self_prio(rtems_task_priority prio)
-{
-        rtems_status_code sc;
-
-        sc = rtems_task_set_priority(RTEMS_SELF, prio, &prio);
-        assert(sc == RTEMS_SUCCESSFUL);
-}
-
-static void
 default_network_dhcpcd(void)
 {
     static const char default_cfg[] = "clientid test client\n";
@@ -950,6 +941,20 @@ POSIX_Init ( void *argument __attribute__((unused)))
     struct timespec  	now;
     char timeBuff[100];
 
+    /*
+     * If RTEMS is used with the POSIX API, the init task
+     * 'POSIX_Init()' is unfortunately given the priority '2'. This
+     * corresponds to the second lowest POSIX prio (RTEMS pthread
+     * prio 256). This task shoud have IOCsh prio.
+     */
+
+    /* PosixMaxPrio == 100 */
+    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY * epicsThreadPriorityIocsh / 100;
+    rtems_task_priority old;
+    sc = rtems_task_set_priority (RTEMS_SELF, mainPrio, &old);
+    assert(sc == RTEMS_SUCCESSFUL);
+    printf("Priority changed from %d -> %d\n", old, mainPrio);
+
     initConsole ();
 
     /*
@@ -996,10 +1001,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
     if (epicsRtemsInitPostSetBootConfigFromNVRAM(&rtems_bsdnet_config) != 0)
         delayedPanic("epicsRtemsInitPostSetBootConfigFromNVRAM");
 #endif
-    /*
-     * Override RTEMS Posix configuration, it gets started with posix prio 2
-     */
-    epicsThreadSetPriority(epicsThreadGetIdSelf(), epicsThreadPriorityIocsh);
 
     /*
      * Create a reasonable environment
@@ -1035,7 +1036,9 @@ POSIX_Init ( void *argument __attribute__((unused)))
     on_exit(default_network_on_exit, NULL);
 
     /* Let other tasks run to complete background work */
-    default_network_set_self_prio(RTEMS_MAXIMUM_PRIORITY - 1U);
+    sc = rtems_task_set_priority (RTEMS_SELF, RTEMS_MAXIMUM_PRIORITY/2, &old);
+    assert(sc == RTEMS_SUCCESSFUL);
+    printf("Priority changed from %d -> %d\n", old, RTEMS_MAXIMUM_PRIORITY/2);
 
     sc = rtems_bsd_initialize();
     assert(sc == RTEMS_SUCCESSFUL);
@@ -1098,6 +1101,11 @@ POSIX_Init ( void *argument __attribute__((unused)))
         printf("time from ntp : %s.%09ld UTC\n", timeBuff, now.tv_nsec);
       }
     }
+
+    sc = rtems_task_set_priority (RTEMS_SELF, mainPrio, &old);
+    assert(sc == RTEMS_SUCCESSFUL);
+    printf("Priority changed at end of network init from %d -> %d\n", old, mainPrio);
+
 #else // Legacy stack, old network initialization
     char *cp;
     if ((cp = getenv("EPICS_TS_NTP_INET")) != NULL)
@@ -1111,12 +1119,14 @@ POSIX_Init ( void *argument __attribute__((unused)))
     {
         unsigned int p;
         if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
-                                            == epicsThreadBooleanStatusSuccess)
+                                            != epicsThreadBooleanStatusSuccess)
         {
-            rtems_bsdnet_config.network_task_priority = p;
+            p = RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY * epicsThreadPriorityScanLow / 100;
         }
+        printf(" This is network task prio (RTEMS) : %d , OSI Prio %d\n", p, epicsThreadPriorityScanLow);
+        rtems_bsdnet_config.network_task_priority = p;
     }
-    printf("\n***** Initializing network (Legacy Stack)  *****\n");
+    printf("\n***** Initializing network (Legacy Stack) with prio %d  *****\n", rtems_bsdnet_config.network_task_priority);
     rtems_bsdnet_initialize_network();
     printf("\n***** Network Status  *****\n");
     rtems_netstat(3);
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -945,11 +945,13 @@ POSIX_Init ( void *argument __attribute__((unused)))
      * If RTEMS is used with the POSIX API, the init task
      * 'POSIX_Init()' is unfortunately given the priority '2'. This
      * corresponds to the second lowest POSIX prio (RTEMS pthread
-     * prio 256). This task shoud have IOCsh prio.
+     * prio 254). This task shoud have IOCsh prio.
      */
 
     /* PosixMaxPrio == 100 */
-    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - ( RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY) * epicsThreadPriorityIocsh / 100;
+    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY
+                                 - (RTEMS_MAXIMUM_PRIORITY - RTEMS_MINIMUM_PRIORITY)
+                                 * epicsThreadPriorityIocsh / 100;
     rtems_task_priority old;
     sc = rtems_task_set_priority (RTEMS_SELF, mainPrio, &old);
     assert(sc == RTEMS_SUCCESSFUL);
@@ -1121,7 +1123,8 @@ POSIX_Init ( void *argument __attribute__((unused)))
         if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
                                             != epicsThreadBooleanStatusSuccess)
         {
-            p = RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY * epicsThreadPriorityScanLow / 100;
+            p = RTEMS_MAXIMUM_PRIORITY
+              - (RTEMS_MAXIMUM_PRIORITY - RTEMS_MINIMUM_PRIORITY) * epicsThreadPriorityScanLow / 100;
         }
         printf(" This is network task prio (RTEMS) : %d , OSI Prio %d\n", p, epicsThreadPriorityScanLow);
         rtems_bsdnet_config.network_task_priority = p;
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -1116,14 +1116,9 @@ POSIX_Init ( void *argument __attribute__((unused)))
 
     if (rtems_bsdnet_config.network_task_priority == 0)
     {
-        unsigned int p;
-        if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
-                                            != epicsThreadBooleanStatusSuccess)
-        {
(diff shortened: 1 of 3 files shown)


anjohnson:
Is that warning message going to be printed every time an IOC boots on RTEMS? I don't think I'd want that extra noise in the output. However can't the rtems-init-task call the appropriate Posix thread API itself to set its own priority?

hjunkes:
Personally, I like it when a few status messages appear when booting. Then you can see that something is happening :-)
(If no console is connected, it doesn't bother)

I can also set the priority with the RTEMS POSIX API.

Since the EPICS API is also used very early on in the existing code rtems_init, e.g.

```
    /*
     * More environment: iocsh prompt and hostname
     */
    {
        char hostname[1024];
        gethostname(hostname, 1023);
        char *cp = mustMalloc(strlen(hostname)+3, "iocsh prompt");
        sprintf(cp, "%s> ", hostname);
        epicsEnvSet ("IOCSH_PS1", cp);
        epicsEnvSet("IOC_NAME", hostname);
    }
    ```
    
    This can also be found during network initialization:
    
    ```
     if (rtems_bsdnet_config.network_task_priority == 0)
    {
        unsigned int p;
        if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
                                            == epicsThreadBooleanStatusSuccess)
        {
            rtems_bsdnet_config.network_task_priority = p;
        }
    }
    ```
    
    I found it consistent to also use the EPICS API for prio set.

**Comment by mdavidsaver** [comment]
This change seems like it would render the `isRealTimeScheduled` unused on RTEMS.  If we go that route, I think it would be better to just get rid of it entirely.  eg. 

```c
#ifndef __rtems__
   if(!pthreadInfo->isRealTimeScheduled) return;
#endif
```

That said, remember there is scheduler policy also to consider.  aka. `SCHED_FIFO` vs. `SCHED_RR` vs. `SCHED_OTHER`. 

When `epicsThreadSetPriority()` encounters a non-epics thread, should it set `SCHED_FIFO` as well as priority?

mdavidsaver:
> Personally, I like it when a few status messages appear when booting.

But this message may well be printed after boot, when it would become a nuisance.

Until or unless Base gets [controllable debug logging](https://epics-base.github.io/pvxs/util.html#logging) I do not like adding unnecessary noise.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -938,6 +929,20 @@ POSIX_Init ( void *argument __attribute__((unused)))
...
 
+    /*
+     * If RTEMS is used with the POSIX API, the init task
+     * 'POSIX_Init()' is unfortunately given the priority '2'. This
+     * corresponds to the second lowest POSIX prio (RTEMS pthread
+     * prio 256). This task shoud have IOCsh prio.
+     */
+
+    /* PosixMaxPrio == 100 */
+    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY * epicsThreadPriorityIocsh / 100;
```
mdavidsaver: I'm confused by this expression.  As written it looks like increasing `epicsThreadPriorityIocsh` would decrease `mainPrio`, which seems backwards.

Also, I find references to a [`RTEMS_MINIMUM_PRIORITY`](https://github.com/search?q=RTEMS_MINIMUM_PRIORITY&type=code) which appears to have a non-zero value.

hjunkes: ok, RTEMS_MINIMUM_PRIORITY is defined as 1.
The calculation of the main priority should therefore be as follows:
```
rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - ( RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY) * epicsThreadPriorityIocsh / 100;
```
Please note that the POSIX API also 'uses' the RTEMS-classic priorities for RTEMS.
This is what a running system looks like:
```
ID       NAME                 SHED PRI STATE  MODES    EVENTS WAITINFO
------------------------------------------------------------------------------
0b010001  _main_              UPD   23 READY  P:T:nA 00002000
0b010002  PVAS timers         UPD  191 SEM    P:T:nA   NONE
0b010003  osdNTP_Runner       UPD   27 READY  P:T:nA   NONE
0b010004  taskwd              UPD  229 SEM    P:T:nA   NONE
0b010005  osdNTP_Monitor      UPD  128 SEM    P:T:nA   NONE
0b010006  NTPTimeSync         UPD   27 SEM    P:T:nA   NONE
0b010007  errlog              UPD  229 SEM    P:T:nA   NONE
0b010008                      UPD   27 SEM    P:T:nA   NONE
0b010009  timerQueue          UPD   77 SEM    P:T:nA   NONE
0b01000a  cbLow               UPD  105 SEM    P:T:nA   NONE
0b01000b  cbMedium            UPD   93 SEM    P:T:nA   NONE
0b01000c  cbHigh              UPD   75 SEM    P:T:nA   NONE
0b01000d  dbCaLink            UPD  128 SEM    P:T:nA   NONE
0b01000e  PVAL                UPD  128 SEM    P:T:nA   NONE
0b01000f  PDB-event           UPD  206 SEM    P:T:nA   NONE
0b010010  pvAccess-client     UPD  166 READY  P:T:nA   NONE
0b010011  UDP-rx 0.0.0.0:     UPD  128 WK     P:T:nA   NONE   sbwait
0b010012  UDP-rx 141.14.1     UPD  128 WK     P:T:nA   NONE   sbwait
0b010013  UDP-rx 141.14.1     UPD  128 WK     P:T:nA   NONE   sbwait
0b010014  UDP-rx 224.0.0.     UPD  128 WK     P:T:nA   NONE   sbwait
0b010015  scanOnce            UPD   85 SEM    P:T:nA   NONE
0b010016  scan-10             UPD  103 SEM    P:T:nA   NONE
0b010017  scan-5              UPD  100 SEM    P:T:nA   NONE
0b010018  scan-2              UPD   98 SEM    P:T:nA   NONE
0b010019  scan-1              UPD   95 SEM    P:T:nA   NONE
0b01001a  scan-0.5            UPD   93 SEM    P:T:nA   NONE
0b01001b  scan-0.2            UPD   90 SEM    P:T:nA   NONE
0b01001c  scan-0.1            UPD   88 READY  P:T:nA   NONE
0b01001d  CAS-TCP             UPD  209 WK     P:T:nA   NONE   accept
0b01001e  CAS-UDP             UPD  214 WK     P:T:nA   NONE   sbwait
0b01001f  CAS-beacon          UPD  211 TIME:IS P:T:nA   NONE   Nanosleep
0b010020  CAC-event           UPD  125 SEM    P:T:nA   NONE
0b010021  TCP-acceptor        UPD  128 WK     P:T:nA   NONE   accept
0b010022  UDP-rx 0.0.0.0:     UPD  128 WK     P:T:nA   NONE   sbwait
0b010023  UDP-rx 141.14.1     UPD  128 WK     P:T:nA   NONE   sbwait
0b010024  UDP-rx 141.14.1     UPD  128 WK     P:T:nA   NONE   sbwait
0b010025  UDP-rx 224.0.0.     UPD  128 WK     P:T:nA   NONE   sbwait
```

mdavidsaver: I am still confused about the subtraction in (abbreviated): `MAX - MAX*prioIocsh/100`

```c
#define epicsThreadPriorityIocsh        91
```

If I were to change `91` to `90`, do you intend that `mainPrio` should increase or decrease?

Since the basic relationship is $\frac{rtemsPrio-MIN}{MAX-MIN} == \frac{epicsPrio}{epicsMax}$

I expect to see something like:

```c
rtems_task_priority mainPrio 
 = RTEMS_MINIMUM_PRIORITY
 + (RTEMS_MAXIMUM_PRIORITY-RTEMS_MINIMUM_PRIORITY)
 * epicsThreadPriorityIocsh/epicsThreadPriorityMax;
```

(since this would be integer math, it would round down)

hjunkes: If we change epicsThreadPriorityIocsh from 91 to 90, I expect an increase of mainPrio (RTEMS higher Number -> lower prio)
You can see that here:
```
0b010016  scan-10             UPD  103 SEM    P:T:nA   NONE
0b010017  scan-5              UPD  100 SEM    P:T:nA   NONE
0b010018  scan-2              UPD   98 SEM    P:T:nA   NONE
0b010019  scan-1              UPD   95 SEM    P:T:nA   NONE
0b01001a  scan-0.5            UPD   93 SEM    P:T:nA   NONE
0b01001b  scan-0.2            UPD   90 SEM    P:T:nA   NONE
0b01001c  scan-0.1            UPD   88 READY  P:T:nA   NONE
```

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -938,6 +929,20 @@ POSIX_Init ( void *argument __attribute__((unused)))
...
 
+    /*
+     * If RTEMS is used with the POSIX API, the init task
+     * 'POSIX_Init()' is unfortunately given the priority '2'. This
+     * corresponds to the second lowest POSIX prio (RTEMS pthread
+     * prio 256). This task shoud have IOCsh prio.
+     */
+
+    /* PosixMaxPrio == 100 */
+    rtems_task_priority mainPrio = RTEMS_MAXIMUM_PRIORITY - ( RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY) * epicsThreadPriorityIocsh / 100;
```
anjohnson: Core Group:

`RTEMS_MAXIMUM_PRIORITY - RTEMS_MAXIMUM_PRIORITY` = 0.

This needs a little more testing.

kiwichris:
I think fiddling a POSIX resource via the classic RTEMS API is peeking under the hood of RTEMS. If you use the POSIX API it will always have to work as you expect across all versions of RTEMS.

kiwichris:
> Is that warning message going to be printed every time an IOC boots on RTEMS? I don't think I'd want that extra noise in the output. However can't the rtems-init-task call the appropriate Posix thread API itself to set its own priority?

I find the warnings on a console a pain because you need to go and check each one when something is wrong.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -1099,12 +1109,15 @@ POSIX_Init ( void *argument __attribute__((unused)))
...
         unsigned int p;
         if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
-                                            == epicsThreadBooleanStatusSuccess)
+                                            != epicsThreadBooleanStatusSuccess)
         {
-            rtems_bsdnet_config.network_task_priority = p;
+            p = RTEMS_MAXIMUM_PRIORITY
+              - (RTEMS_MAXIMUM_PRIORITY - RTEMS_MINIMUM_PRIORITY) * epicsThreadPriorityScanLow / 100;
         }
+        printf(" This is network task prio (RTEMS) : %d , OSI Prio %d\n", p, epicsThreadPriorityScanLow);
```
kiwichris: There are shell commands in RTEMS that can report this type of value for those times when you need to look.

kiwichris:
I have reviewed the change however it is not clear what it is attempting to do?

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -1023,7 +1026,9 @@ POSIX_Init ( void *argument __attribute__((unused)))
     on_exit(default_network_on_exit, NULL);
 
     /* Let other tasks run to complete background work */
-    default_network_set_self_prio(RTEMS_MAXIMUM_PRIORITY - 1U);
+    sc = rtems_task_set_priority (RTEMS_SELF, RTEMS_MAXIMUM_PRIORITY/2, &old);
+    assert(sc == RTEMS_SUCCESSFUL);
+    printf("Priority changed from %d -> %d\n", old, RTEMS_MAXIMUM_PRIORITY/2);
```
kiwichris: The comment says:
```
/* Let other tasks run to complete background work */
```
On SMP anything able to run would be running. It is not clear to me what this code is doing?

hjunkes:
@Chris Thanks for the review

I now only use the posix API to modify the priority of the main task.

I threw out the thing with ‘/* Let other tasks run to complete background work */’. I had taken this from the libbsd-examples.

No more output on the console about priority customisation.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -938,6 +929,27 @@ POSIX_Init ( void *argument __attribute__((unused)))
...
+    struct sched_param  param;
+    int policy;
+    sc  = pthread_attr_init(&attr);
+    assert(sc == RTEMS_SUCCESSFUL);
+    sc = pthread_attr_getschedpolicy(&attr, &policy);
+    assert(sc == RTEMS_SUCCESSFUL);
+
+    param.sched_priority = (sched_get_priority_max(policy)
+                            - sched_get_priority_min(policy))
+                         * epicsThreadPriorityIocsh / 100;
```
mdavidsaver: @hjunkes I'm sorry to keep poking at this, but it seems like the basics of the linear equation.  I see the slope.  Where is the offset?

Consider the edge cases.

if `epicsThreadPriorityIocsh=0`, then the result needs to be `sched_get_priority_min(policy)`.

If `epicsThreadPriorityIocsh=100`, then the result needs to be `sched_get_priority_max(policy)`.

I think you need to change this expression to:

```c
    param.sched_priority = (sched_get_priority_max(policy) - sched_get_priority_min(policy))
                         * epicsThreadPriorityIocsh / 100
                         + sched_get_priority_min(policy);
```

modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c
```diff
@@ -33,7 +34,7 @@ void epicsThreadShowInfo(epicsThreadOSD *pthreadInfo, unsigned int level)
         if(pthreadInfo->tid) {
             int status;
             status = pthread_getschedparam(pthreadInfo->tid,&policy,&param);
-            if(!status) priority = param.sched_priority;
+            if(!status) priority = RTEMS_MAXIMUM_PRIORITY - param.sched_priority;
```
mdavidsaver, line 37: Is this expression mixing units?  `param.sched_priority` is on the posix scale (cf. `sched_get_priority_min() ... _max()`).  Will this always be the same as the RTEMS priority scale of `RTEMS_MAXIMUM_PRIORITY`?

modules/libcom/RTEMS/posix/rtems_init.c
hjunkes: Yes, you are absolutely right, I have been sloppy here.

modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c
hjunkes: The call 
```
   sc = pthread_setschedparam(pthread_self(), policy, &param);
    assert(sc == RTEMS_SUCCESSFUL);
 ```
makes it right. 

I followed Chris's recommendation and the result looks ok.

rt pthread
```
ID       NAME                 SHED PRI STATE  MODES    EVENTS WAITINFO
------------------------------------------------------------------------------
0b010001  _main_              UPD   25 READY  P:T:nA 00002000
...
```

**Comment by mdavidsaver** [comment]
Taking a step back.  Do I have the following priority ranges and usages in the PR complete and correct?

- EPICS
  - `epicsThreadPriorityMin -> epicsThreadPriorityMax` (inclusive)
  - `epicsThreadPriorityIocsh` the target priority for `main()`
- RTEMS
  - `RTEMS_MINIMUM_PRIORITY` -> `RTEMS_MAXIMUM_PRIORITY` (inclusive)
  - Value of `rtems_bsdnet_config.network_task_priority`
- POSIX priorities
  - `sched_get_priority_min(policy) -> sched_get_priority_max(policy)` (inclusive)
  - Values of `sched_param::sched_priority`

Now I should note that the EPICS thread priority scale is _supposed_ to be `0 -> 99` inclusive, although there places where this is incorrectly taken as `0 -> 100` inclusive.  So historical usage is not consistent.

Also, all of these are priorities are integers, so there is potentially a question of rounding when converting between scales.  (`epicsThreadHighestPriorityLevelBelow()` and `epicsThreadLowestPriorityLevelAbove()` are meant to handle this)

As I understand it, the current goals of this PR are:

1. Adjust the priority of the initial task (aka. `main()`) to be `epicsThreadPriorityIocsh` on the EPICS scale.
2. Correct `rtems_bsdnet_config.network_task_priority` to a value on the RTEMS scale
  - Currently being incorrectly assigned with an EPICS priority.

@hjunkes Have I missed anything?

File: modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c (hunk 6 above)
Comment (mdavidsaver) [reply modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c:37]: > ... the result looks ok.

This is not really the criteria for success here :)

I think about the three different priority scales involved by analogy with different distance units.  So the expression `RTEMS_MAXIMUM_PRIORITY - param.sched_priority` to me reads like subtracting feet from [furlongs](https://en.wikipedia.org/wiki/Furlong).  A nonsensical operation.  The fact that the three scales/units are not so different as feet and furlongs hinders us by making it harder to judge correctness from the numeric results.

Comment (kiwichris) [reply modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c:37]: I agree with this analogy and the comment about the ability to judge correctness. My long term concern is changes inside RTEMS could subtly effect things and no one notices.

The EPICS priority to platform priority calculation is an EPICS one and I do not know if an interface is exported or widely needed. All I am suggesting with this change is being consistent by using the platform priority, ie POSIX in this case rather than introducing a third priority mapping.

**Comment by kiwichris** [comment]
>     2. Correct `rtems_bsdnet_config.network_task_priority` to a value on the RTEMS scale

This is only valid for the legacy stack. The libbsd stack currently does not provide a means to adjust network priorities. I do not know the history of `rtems_bsdnet.h` in LibBSD and why it is there. From a brief check it does not seem to be used.

**Comment by kiwichris** [comment]
> @chris Thanks for the review
> 
> I now only use the posix API to modify the priority of the main task.

Great.
 
> I threw out the thing with ‘/* Let other tasks run to complete background work */’. I had taken this from the libbsd-examples.

Does EPICS have any specific needs about network initialization, for example interfaces needing valid IP addresses? I have no idea what preconditions it needs before starting up.

> No more output on the console about priority customisation.

Thanks.

**Comment by kiwichris** [comment]
I have raised https://gitlab.rtems.org/rtems/rtos/rtems-release/-/issues/15 in RTEMS to look at providing a mapping API.

File: modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c (hunk 6 above)
Comment (kiwichris) [reply modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c:37]: I have reviewed the EPICS priority calculation and the way RTEMS 6 is handling priorities. The old assumption of the scheduler's priority range being limited by an `unsigned char` has gone and the priority range now depends on the scheduler being used.

The EPICS POSIX thread support correctly handles the scheduler's min and max priority values scaling the EPICS priority ranges over the scheduler's range. The problem is fitting this scaled range into any class API threads that exist. An example is the legacy and libbsd threads.

BSPs like the beatnik or MVME2700 are non-SMP builds of RTEMS so the default scheduler is a simple priority based one. SMP builds default to the EDF scheduler as libbsd requires it and the priority range is `INT_MAX`. That means a libbsd internal thread has hard coded the classic API priority of `100` and that maps to a POSIX priority of `2147483547`.

It is not yet clear to me how to handle this.

File: modules/libcom/RTEMS/posix/rtems_init.c [hunk 7]
```diff
@@ -1111,12 +1118,15 @@ POSIX_Init ( void *argument __attribute__((unused)))
     {
         unsigned int p;
         if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
-                                            == epicsThreadBooleanStatusSuccess)
+                                            != epicsThreadBooleanStatusSuccess)
         {
-            rtems_bsdnet_config.network_task_priority = p;
+            p = RTEMS_MAXIMUM_PRIORITY
+              - (RTEMS_MAXIMUM_PRIORITY - RTEMS_MINIMUM_PRIORITY) * epicsThreadPriorityScanLow / 100;
```
anjohnson: Core Group: @mdavidsaver suggests using positive scale, not negative.

```C
p = RTEMS_MINIMUM_PRIORITY +
    (RTEMS_MAXIMUM_PRIORITY - RTEMS_MINIMUM_PRIORITY) * epicsThreadPriorityScanLow / 100;
```

May need to cast priorities to `int` if they are `unsigned`.
Use the `epicsThreadPriorityMax` (which is 99) instead of `100`.

mdavidsaver: I was mistaken this morning.  Heinz has this mapping correct once `100` is replaced with `(epicsThreadPriorityMax - epicsThreadPriorityMin)` (aka. 99).

Looking in [`include/rtems/rtems/tasks.h`](https://github.com/RTEMS/rtems/blob/2ec78303ebed718ef4f696a6698c8f003ef12c93/cpukit/include/rtems/rtems/tasks.h#L297-L313), I see that MAXIMUM and MINIMUM refer to numeric value, not increasing priority.  (RTEMS classic priority ~= *NIX niceness?)

```
 * @brief This runtime constant represents the lowest (least important) task
 *   priority of the scheduler with index zero.
 */
#define RTEMS_MAXIMUM_PRIORITY _RTEMS_Maximum_priority()
...
 * @brief This compile time constant provides the highest (most important) task
 *   priority settable by the API.
 */
#define RTEMS_MINIMUM_PRIORITY 1
```

note: unless I hear otherwise, I will assume that the value of `_RTEMS_Maximum_priority()` will not change at runtime.

So the limits of the epics <-> rtems classic priority mapping are:

* epicsThreadPriorityMin <-> RTEMS_MAXIMUM_PRIORITY
* epicsThreadPriorityMax <-> RTEMS_MINIMUM_PRIORITY

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -1111,12 +1118,15 @@ POSIX_Init ( void *argument __attribute__((unused)))
     {
         unsigned int p;
         if (epicsThreadHighestPriorityLevelBelow(epicsThreadPriorityScanLow, &p)
-                                            == epicsThreadBooleanStatusSuccess)
+                                            != epicsThreadBooleanStatusSuccess)
```
mdavidsaver: Why call epicsThreadHighestPriorityLevelBelow at all?  The result `p` will always be an EPICS priority.  I think the mapping below should always be used starting with epicsThreadPriorityScanLow.

modules/libcom/src/osi/os/RTEMS-posix/osdThreadExtra.c
mdavidsaver: I think that usage of `RTEMS_MAXIMUM_PRIORITY` here is incorrect.  This code is mapping a POSIX to an EPICS priority for display purposes.  RTEMS classic priorities are not involved.

```c
priority = (( param.sched_priority - sched_get_priority_min() )
   * (epicsThreadPriorityMax - epicsThreadPriorityMin))
   / (sched_get_priority_max() - sched_get_priority_min())
  + epicsThreadPriorityMin;
```

(I find myself thinking it would be simpler all around to calculate these scalings as `float`.  It might be some years before the accumulated extra CPU time needed exceeds what we have spent on this PR)

ralphlange:
@hjunkes: Please try to rebase instead of back-merging the target branch.
Even though squashing will eventually hide the additional repetitive commits, they make things harder to read in the intermediate stages.

hjunkes:
@Ralph:I have made a 'sync fork' ... is that rebase?

ralphlange:
> @ralph:I have made a 'sync fork' ... is that rebase?

No.

Rebase is an action on your PR branch, where - instead of merging new commits from the target branch - you reapply the commits of your PR branch to the current state of the target branch.
(Like ripping off a branch and then glueing it to another place of the tree. Further up, in most cases.)

Your commits will get new hashes, but comments, dates and author will be persisted.

anjohnson:
> rebase instead of back-merging

@hjunkes You do that by running `git pull --rebase` instead of `git pull`. If you have some local edits in your workspace that you haven't or don't want to commit yet the rebase option won't like them, but you can run `git pull --rebase --autostash` instead to have it save the changes and re-apply them after doing the rebase. You might have to resolve conflicts from the rebase operation more than once if you're unlucky, but the output from `git status` always tells you where you are and what to do next.

You can configure git to apply those flags automatically (search for `pull.rebase` in the output from `git help config`), but there are many different options which I often get confused by so I just add those flags manually.

After you have committed your local changes and want to publish the new version you will have to use `git push --force` to push them back up to your branch on GitHub.

ralphlange:
Or you can use a good GUI and not care about the too-many command line options.
(I like SmartGit, where you get a free license when you work regularly on open source projects. https://www.syntevo.com/smartgit/)

PR #577, opened 2024-12-22
Add EPICS_DIAGNOSE_AS

JJL772:
* Added `EPICS_DIAGNOSE_AS` for clang's `diagnose_as_builtin` attribute. https://clang.llvm.org/docs/AttributeReference.html#diagnose-as-builtin. This will allow the compiler to produce fortify diagnostics for the marked declarations. 
* Applied `EPICS_DIAGNOSE_AS` to various `epicsStdio.h` and `epicsString.h` functions.

### DIFF
--- a/modules/libcom/src/misc/epicsString.h
+++ b/modules/libcom/src/misc/epicsString.h
@@ -91,24 +91,28 @@ LIBCOM_API size_t epicsStrnEscapedFromRawSize(const char *buf, size_t len);
  *
  * Implements strcmp from the C standard library, except is case insensitive
  */
+EPICS_DIAGNOSE_AS_STRCASECMP(1, 2)
 LIBCOM_API int epicsStrCaseCmp(const char *s1, const char *s2);
 
 /** \brief Does case-insensitive comparision of two strings
  *
  * Implements strncmp from the C standard library, except is case insensitive
  */
+EPICS_DIAGNOSE_AS_STRNCASECMP(1, 2, 3)
 LIBCOM_API int epicsStrnCaseCmp(const char *s1, const char *s2, size_t len);
 
 /** \brief Duplicates a string
  *
  * Implements strdup from the C standard library.  Calls mallocMustSucceed() to allocate memory
  */
+EPICS_DIAGNOSE_AS_STRDUP(1)
 LIBCOM_API char * epicsStrDup(const char *s);
 
 /** \brief Duplicates a string
  *
  * implements strndup from the C standard library. Calls mallocMustSucceed() to allocate memory
  */
+EPICS_DIAGNOSE_AS_STRNDUP(1, 2)
 LIBCOM_API char * epicsStrnDup(const char *s, size_t len);
 
 /** \brief Prints escaped version of string
--- a/modules/libcom/src/osi/compiler/clang/compilerSpecific.h
+++ b/modules/libcom/src/osi/compiler/clang/compilerSpecific.h
@@ -30,6 +30,48 @@
 /* Expands to a 'const char*' which describes the name of the current function scope */
 #define EPICS_FUNCTION __PRETTY_FUNCTION__
 
+/*
+ * Allows the compiler to apply fortify diagnostics to marked functions.
+ * Introduced in clang 14.
+ * Ref: https://clang.llvm.org/docs/AttributeReference.html#diagnose-as-builtin
+ */
+#if __has_attribute(diagnose_as_builtin)
+#define EPICS_DIAGNOSE_AS(...) __attribute__((diagnose_as_builtin(__VA_ARGS__)))
+
+#if __has_builtin(__builtin_strncasecmp)
+#define EPICS_DIAGNOSE_AS_STRNCASECMP(_1, _2, _3) EPICS_DIAGNOSE_AS(__builtin_strncasecmp, _1, _2, _3)
+#endif
+
+#if __has_builtin(__builtin_strdup)
+#define EPICS_DIAGNOSE_AS_STRDUP(_1) EPICS_DIAGNOSE_AS(__builtin_strdup, _1)
+#endif
+
+#if __has_builtin(__builtin_strndup)
+#define EPICS_DIAGNOSE_AS_STRNDUP(_1, _2) EPICS_DIAGNOSE_AS(__builtin_strndup, _1, _2)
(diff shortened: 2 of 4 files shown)


mdavidsaver:
From a quick look, it doesn't seem that GCC has or is adding support for `__attribute__((diagnose_as_builtin` as of 2024.  So for the present, this is a clang-only feature.

Could these annotations be added without introducing such a generic, yet specific, formula?

So far we have preferred fairly specific macros like `EPICS_PRINTF_STYLE` instead of something more generic like `EPICS_ATTRIBUTE(...)`.

eg. Could `__builtin_printf` be folded into the existing `EPICS_PRINTF_STYLE`?  Would this change anything?

modules/libcom/src/cxxTemplates/tsSLList.h
```diff
@@ -311,13 +311,13 @@ inline bool tsSLIterConst<T>::valid () const
...
 {
-    return this->pEntry == rhs.pConstEntry;
+    return this->pEntry == rhs.pEntry;
 }
 
 template < class T >
 inline bool tsSLIterConst<T>::operator != (const tsSLIterConst<T> &rhs) const
 {
-    return this->pEntry != rhs.pConstEntry;
+    return this->pEntry != rhs.pEntry;
```
mdavidsaver: Well, this fixes a latent compile error introduced by b67619c0fc2475ae6d9156771c77ae4c2b2559f7, almost 24 years ago.

Can you say how you noticed this?  Did you find a compilation error?

It should be reiterated that usage of `cxxTemplated/` is not recommended for new code.

JJL772: I just compiled with clang 19.1.5. Not sure why it decided to start evaluating this code though.

mdavidsaver: I won't insist on a separate PR for this straightforward change.  Still, could you quote the specific error, which I assume will some variation on "no such member pConstEntry", and attach the full error output?  This will help google to help others to find this issue.

JJL772:
```
clang++  -D_GNU_SOURCE -D_DEFAULT_SOURCE          -D_X86_64_ -DUNIX  -Dlinux   -DBUILDING_libCom_API   -O3 -g   -Wall      -mtune=generic     -m64  -fPIC -I. -I../O.Common -I. -I../osi/compiler/clang -I../osi/compiler/default -I. -I../osi/os/Linux -I../osi/os/posix -I../osi/os/default -I.. -I../as -I../bucketLib -I../calc -I../cvtFast -I../cppStd -I../cxxTemplates -I../dbmf -I../ellLib -I../env -I../error -I../fdmgr -I../flex -I../freeList -I../gpHash -I../iocsh -I../log -I../macLib -I../misc -I../osi -I../pool -I../ring -I../taskwd -I../timer -I../yacc -I../yacc -I../yajl -I../../../../include/compiler/clang -I../../../../include/os/Linux -I../../../../include         -c ../cxxTemplates/resourceLib.cpp
In file included from ../cxxTemplates/resourceLib.cpp:15:
In file included from ../cxxTemplates/resourceLib.h:43:
../cxxTemplates/tsSLList.h:314:32: error: no member named 'pConstEntry' in 'tsSLIterConst<T>'
  314 |     return this->pEntry == rhs.pConstEntry;
      |                            ~~~ ^
../cxxTemplates/tsSLList.h:320:32: error: no member named 'pConstEntry' in 'tsSLIterConst<T>'
  320 |     return this->pEntry != rhs.pConstEntry;
      |                            ~~~ ^
2 errors generated.
```

JJL772: Actually, I've pulled this commit into #579, just so there's less going on in this PR. Sorry for the spam.

anjohnson:
Do these new decorations do anything to the [Doxygen](https://epics.anl.gov/base/R7-0/8-docs/doxygen/epics_string_8h.html) and [read-the-docs](https://docs.epics-controls.org/projects/base/en/latest/epicsString_h.html) generated documentation pages? Please make sure they will not be affected.

JJL772:
> @anjohnson Do these new decorations do anything to the [Doxygen](https://epics.anl.gov/base/R7-0/8-docs/doxygen/epics_string_8h.html) and [read-the-docs](https://docs.epics-controls.org/projects/base/en/latest/epicsString_h.html) generated documentation pages? Please make sure they will not be affected.

Doxygen output seems to be unchanged.

> So far we have preferred fairly specific macros like EPICS_PRINTF_STYLE instead of something more generic like EPICS_ATTRIBUTE(...).
> eg. Could __builtin_printf be folded into the existing EPICS_PRINTF_STYLE? Would this change anything?

@mdavidsaver I'm not sure that this can be cleanly rolled into `EPICS_PRINTF_STYLE`. If `EPICS_PRINTF_STYLE` is applied to i.e. `epicsSnprintf`, the `diagnose_as_builtin(__builtin_printf,f)` would not match the function's behavior. Using `diagnose_as_builtin(__builtin_snprintf)` for `epicsSnprintf` is much preferred because the compiler will be aware of the output buffer and its bounds.

For now, I added individual macros for each of the `__builtin` types to resolve compile errors on older versions of clang. (`EPICS_DIAGNOSE_AS_PRINTF`, `EPCIS_DIAGNOSE_AS_STRDUP`, etc...). This isn't very clean though...

JJL772:
Clang will not generate bounds checks for the annotated functions with `-D_FORTIFY_SOURCE=2`: https://godbolt.org/z/h766da5bT

However it will generate pretty detailed warnings: https://godbolt.org/z/hjvb1EGh4

I'll have to play around with clang-tidy to see how this affects its analysis.

anjohnson:
It's interesting that Clang's warning message says "memcpy" even though the call was to "thing" Potentially a bit confusing, unfortunately:

```
<source>:9:5: warning: 'memcpy' will always overflow; destination buffer has size 2, but size argument is 3 [-Wfortify-source]
    9 |     thing(test, "abc", 3);
      |     ^
```

anjohnson:
Core Group: Too specific to Clang, we don't feel the need for help diagnosing problems provided by these flags by that compiler only, which are unlikely to be replicated in GCC.

PR #580, opened 2025-01-07
Add SHRLIB_INSTALLS mechanism

ralphlange:
@henriquesimoes found in ralphlange/opcua-binary-builder#1 that the existing mechanisms for `BIN_INSTALLS` and `LIB_INSTALLS` do not use the correct file permissions and/or install locations for shared libraries.

This PR adds a new setting `SHRLIB_INSTALLS` for shared libraries.

### DIFF
--- a/configure/CONFIG_ADDONS
+++ b/configure/CONFIG_ADDONS
@@ -68,8 +68,9 @@
 #    OBJS_HOST     host system object files to build and install
 #    OBJS_IOC      ioc system object files to build and install
 #    USR_INCLUDES  include directories
-#    BIN_INSTALLS  binaries to install
-#    LIB_INSTALLS  library binaries to install
+#    BIN_INSTALLS      binaries to install
+#    LIB_INSTALLS      library binaries to install
+#    SHRLIB_INSTALLS   shared library binaries to install
 #    RCS           win32 resource files for building libraries and prods
 #    PROD_RCS      win32 resource files for building prods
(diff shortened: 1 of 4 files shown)


ralphlange:
Needs verification/testing and a matching PR for the documentation.

anjohnson:
I don't believe this will support Windows DLLs as proposed, although you might not care about that for opcua. If you don't want to add that, it should be mentioned in the Release Notes.

A Windows DLL needs 2 files to be installed, a `.lib` (`.dll.a` on MinGW/Cygwin) which goes into `$(INSTALL_LIB)` with r-- permissions and provides the stub routines that callers get linked with, and a `.dll` for `$(INSTALL_BIN)` which contains the shared library code and gets r-x permissions. Use `git grep DLLSTUB` to find out more.

ralphlange:
Of course it will need more work to work on Windows...

...where the situation is even a bit more complicated, as the stub is only needed if you want to link against the DLL. If you provide additional second-level-dependency DLLs that you don't actively need to link against, these won't need/have the stub.
As you pointed out, stub names depend on the compiler chain to be used.

I will use wildcards to install stubs only if they are provided, using the names determined by `$(T_A)`.
For installing static and shared libraries in parallel (possible using MinGW), it's perfectly appropriate to handle it separately and ask for both `SHRLIB_INSTALLS` and `LIB_INSTALLS` to be set.

PR #581, opened 2025-01-07
Post monitors from compress record when resetting it

ericonr:
This way clients receive updates with an empty array after writing into the RES field.

### DIFF
--- a/documentation/RELEASE_NOTES.md
+++ b/documentation/RELEASE_NOTES.md
@@ -22,6 +22,11 @@ should also be read to understand what has changed since earlier releases:
 
 ## Changes made on the 7.0 branch since 7.0.8.1
 
+### Post monitors from compress record when it's reset
+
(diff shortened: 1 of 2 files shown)


ericonr:
@simon-ess , since you touched these recently

simon-ess:
This seems a simple enough change, is it possible to add a test of this behaviour?

mdavidsaver:
Original discussion https://epics.anl.gov/tech-talk/2024/msg01373.php

ericonr:
> This seems a simple enough change, is it possible to add a test of this behaviour?

@simon-ess that would require adding tests for the compress record which use a CA client, right? Or is there some other way I could observe the side effects of a `monitor(prec)` call?

I don't have any experience using the libCA C API, and it isn't used in the current `compress` tests.

anjohnson:
Core Group: Approved.

PR #582, opened 2025-01-07
RFC: API to renumber socket FDs for code that uses select() with libCa

anjohnson:
The fd assigned to a new socket is normally the lowest fd available. The select() system call can only manage sockets where fd < FD_SETSIZE. The EPICS PCAS uses select(), so can only serve clients that come in on low-numbered sockets. The CA Gateway calls libCa as well as PCAS, so to maximize the clients PCAS can serve, sockets created by libCa should be allocated an fd of FD_SETSIZE or above.

For systems that support the fcntl() F_DUPFD_CLOEXEC this commit adds a routine epicsSocketCreateSelect() to osiSock that creates a socket with fd < FD_SETSIZE, and modifies epicsSocketCreate() to create sockets for libraries like libCa that don't use select() and get fds >= FD_SETSIZE.

Most of the code of interest here is in osi/os/posix/osdSock.c. The new `attachWarning()` and `clearMulticastAll()` routines are just moving code that would otherwise have to be duplicated.
The new bit is the call `fcntl ( sock, F_DUPFD_CLOEXEC, FD_SETSIZE )` which duplicates the fd given in `sock` to a new fd greater than or equal to the 3rd argument, and also sets `FD_CLOEXEC` at the same time. By inserting this into `epicsSocketCreate()` I can force sockets created by libCa out of the fd range which the PCAS `select()` call can handle, so the PV Gateway can support nearly `FD_SETSIZE` clients while still connecting to all the IOCs it needs to.

I modify the PCAS `casIntfIO()` constructor to use `epicsSocketCreateSelect()` for the server socket so it will also work with `select()`. This PR would require a similar modification to any other EPICS software which uses both libCa and `select()`.

An alternative approach might be to change libCa to call a routine that explicitly returns a socket with a high fd.


mdavidsaver:
I am not enthusiastic about our adding new API, and runtime cost to `epicsSocketCreate()`, intended solely to support `select()`, which is the oldest and most deprecated of the socket multiplexer syscalls.

Before accepting this change I would want to see a serious attempt made to update `class fdManager` to use `poll()` or other newer mechanism which removes the `FD_SETSIZE` limitation.  (As I understand it, only vxWorks and RTEMS w/ legacy networking do not provide at least `poll()`)

anjohnson:
I replaced the previous version of this with a cleaner implementation. It still adds a public API to osiSock.h, but now the renumbering of FDs is opt-in, and could be used on sockets returned by `epicsSocketAccept()` too. The second commit changes the CA client calls from `epicsSocketCreate()` to a local static method that checks for the environment variable `EPICS_CAC_AVOID_LOW_FDS` at startup and only calls `epicsSocketRenumber()` when it exists.

Not sure if my class method `cac::CreateSocket()` is the best place for that code, but it works there. I didn't bother to change the repeater or casr which also create sockets, they don't need to cooperate with other application code.

mdavidsaver:
> ... serious attempt made to update `class fdManager` to use `poll()` ...

See #587

anjohnson:
Abandoning this now that #587 is available.

PR #584, opened 2025-01-13
Fix compile error in setBootConfigFromNVRAM.c with libbsd

JJL772:
Several declarations such as `rtems_bsdnet_bootp_boot_file_name` come from `rtems/rtems_bsdnet.h`, so it needs to be included regardless of networking stack.

### DIFF
--- a/modules/libcom/RTEMS/setBootConfigFromNVRAM.c
+++ b/modules/libcom/RTEMS/setBootConfigFromNVRAM.c
@@ -92,7 +92,7 @@ splitNfsMountPath(char *nfsString)
     }
 }
 
-#if defined(HAVE_MOTLOAD)
+#if defined(HAVE_MOTLOAD) && defined(RTEMS_LEGACY_STACK)
 
 /*
  * Motorola MOTLOAD NVRAM Access
@@ -220,7 +220,7 @@ setBootConfigFromNVRAM(void)
         epicsEnvSet("TZ", cp);
 }
 
-#elif defined(HAVE_PPCBUG)
+#elif defined(HAVE_PPCBUG) && defined(RTEMS_LEGACY_STACK)
 /*
  * Motorola PPCBUG NVRAM Access
  */
@@ -320,7 +320,7 @@ setBootConfigFromNVRAM(void)
     splitRtemsBsdnetBootpCmdline();
 }
 
-#elif defined(__mcf528x__)
+#elif defined(__mcf528x__) && defined(RTEMS_LEGACY_STACK)
 
 static char *
 env(const char *parm, const char *defaultValue)


mdavidsaver:
@JJL772 Which RTEMS version and BSP are you building with?

It looks like [`rtems_bsdnet.h`](https://github.com/RTEMS/rtems-libbsd/blob/main/rtemsbsd/include/rtems/rtems_bsdnet.h) is provided by libbsd.  However, I don't think any of our CI builds cover `setBootConfigFromNVRAM.c` with libbsd.

@hjunkes Thoughts?

JJL772:
> Which RTEMS version and BSP are you building with?

RTEMS 6 with uC5282, beatnik (mvme6100) and mvme3100.

hjunkes:
Yes, we can to adopt this. RTEMS6 now has the 'add-ons' **rtems-net-legacy** and **rtems-rtems-libbsd**. In both there is now a rtems/rtems_bsdnet.h and in both there ist the following definition:

```
extern char *rtems_bsdnet_bootp_boot_file_name;
```

But only rtems-legacy-stack has a definition for this:

```
./rtems-net-legacy/rtems/rtems_glue.c:char *rtems_bsdnet_bootp_boot_file_name = 0;
```

This section can be found in modules/libcom/RTEMS/posix in rtems_init.c:

```
...
#ifndef RTEMS_LEGACY_STACK
/* these settings are needed by the rtems startup
 * may provide by dhcp/bootp
 * or environments from the "BIOS" like u-boot, motboot etc.
 */
struct rtems_static_ifconfig {
   char *ip_address;
   char *ip_netmask;
};
/* friggle hpj 10.10.24 */
struct rtems_static_ifconfig rtems_static_ifconfig = {NULL,NULL};

char rtemsInit_NTP_server_ip[16] = "";
char bootp_server_name_init[128] = "1001.1001@10.0.5.1:/epics";
char bootp_boot_file_name_init[128] = "/epics/myExample/bin/RTEMS-beatnik/myExample.boot";
char bootp_cmdline_init[128] = "/epics/myExample/iocBoot/iocmyExample/st.cmd";

char *rtems_bsdnet_bootp_server_name = bootp_server_name_init;
char *rtems_bsdnet_bootp_boot_file_name = bootp_boot_file_name_init;
char *rtems_bsdnet_bootp_cmdline = bootp_cmdline_init;
struct in_addr rtems_bsdnet_bootp_server_address;
epicsEventId   dhcpDone;
#endif
...
```

You can see from the comment that this is still a 'cramp'. At the moment Ran Hong (APS) and I are still working on making the reading of NVRAM universal. 

```
Hi Heinz,

I just put my codes to the following repositories.
For the RTEMS6 with u-boot for qoriq bsp, use this
[Files · uboot · Ran Hong / RTEMS-RH · GitLab](https://gitlab.rtems.org/hongran86/rtems-rh/-/tree/uboot?ref_type=heads)
Use the uboot branch.
For the epics base for mvme2500, I modified on top of CJ's development and saved them to
[hongran/epics-base (github.com)](https://github.com/hongran/epics-base)

I need to make a few changes to the build configuration files to enable critical features. Today I just rebuilt my IOC and make sure it is working. In the next few days I will go through the RTEMS6 building procedure and then write a wiki page.

As for the EPICS code, currently it is still very messy. I think probably we need to discuss how should we handle the options: DHCP vs static IP, different methods of accessing boot parameters for different boards, libbsd vs legacy networking.

Cheers,

Ran Hong
```

mdavidsaver:
> This section can be found in modules/libcom/RTEMS/posix in rtems_init.c:

https://github.com/epics-base/epics-base/blob/144f9756eac4e800f18d3d9a35f0d42a712614fd/modules/libcom/RTEMS/posix/rtems_init.c#L106-L109

Right, so with the libbsd stack, `rtems_bsdnet_bootp_boot_file_name` and friends becomes internal definition in Base.

In this case, I don't think we should depend on prototypes from libbsd headers.

JJL772:
It seems like the code also relies on `rtems_bsdnet_config` from rtems_bsdnet.h, but it's not defined anywhere by EPICS base.

As a workaround for now, I've disabled NVRAM support when building with libbsd. Per @hjunkes's comment, it sounds like someone from APS is working on a proper implementation.

kiwichris:
This PR is needed to let `epics-base` build with LibBSD. It does not help with LibBSD support other than letting work happen around LibBSD  builds. I would like to test and fix the [`iocStats` PR 61](https://github.com/epics-modules/iocStats/pull/61) where @hjunkes posted a failure with LibBSD. I would prefer to work is a suitable base than apply possible fixes when fixing issues in other repos.

LibBSD support is being discussed in #595 so the effect of this PR would be resolved in that related work.

anjohnson:
@JJL772 Is SLAC wanting to run RTEMS 6 on uC5282 with the libbsd networking? Is there enough RAM on the board to actually do that?

kiwichris:
> @JJL772 Is SLAC wanting to run RTEMS 6 on uC5282 with the libbsd networking? Is there enough RAM on the board to actually do that?

Is this the Arcturus uC5282 with 16M of memory?

It will use more memory however It depend on the feature set enabled for LibBSD. I have not worked on hardware needing to have low memory but it has been done. Does it need NFSv4?

JJL772:
> @JJL772 Is SLAC wanting to run RTEMS 6 on uC5282 with the libbsd networking? Is there enough RAM on the board to actually do that?

We may run uC5282 on RTEMS 6, but not with libbsd.

kiwichris:
> > @JJL772 Is SLAC wanting to run RTEMS 6 on uC5282 with the libbsd networking? Is there enough RAM on the board to actually do that?
> 
> We may run uC5282 on RTEMS 6, but not with libbsd.

What stack are you planning on using?

anjohnson:
Is there any overlap between this PR and #853? @brendanchandler could you take a look please and comment here?

brendanchandler:
> Is there any overlap between this PR and #853? @brendanchandler could you take a look please and comment here?

Yes, #853 will fix the issue this pull request addresses.

anjohnson:
#853 has been merged, so I'm closing this as we think the issues have been addressed.
Please reopen if there's anything we missed.

PR #585, opened 2025-01-16
Test cross linux-arm with QEMU user mode

mdavidsaver:
It seems like as of QEMU 9.2, the Linux process emulation now covers the necessary `ioctl()` for Base unit tests, and even softIoc.  In part I think this is because the Linux version of `osiSockDiscoverBroadcastAddresses()` now uses `getifaddrs()` (which in turn uses netlink) instead of the `ioctl()` interface.

The `IP_MULTICAST_ALL` socket option isn't working, but that isn't essential.

```
$ qemu-arm-static -L /usr/arm-linux-gnueabihf  ./bin/linux-arm/softIoc -d tick.db 
Starting iocInit
## EPICS R7.0.8.2-DEV
## Rev. R7.0.8.1-179-gcdce2db211cbb8d47b03
## Rev. Date Git: 2025-01-10 18:52:54 -0800
Warning: Unable to clear IP_MULTICAST_ALL (err=92).  This may cause problems on multi-homed hosts.
CAS: CA beacon send to 192.168.71.255:5065 ERROR: Operation not permitted
CAS: CA beacon send to 10.234.10.255:5065 ERROR: Operation not permitted
epics> dbl
cnt
epics>
```

A draft because I'm not sure this is a useful addition, and because setting CROSS_COMPILER_RUNTEST_ARCHS isn't having the desired effect in the GHA builds.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -204,7 +204,7 @@ jobs:
     - name: "apt-get install"
       run: |
         sudo apt-get update
-        sudo apt-get -y install qemu-system-x86 g++-mingw-w64-x86-64 gdb
+        sudo apt-get -y install qemu-system-x86 qemu-user g++-mingw-w64-x86-64 gdb
       if: runner.os == 'Linux'
     - name: Prepare and compile dependencies
       run: python .ci/cue.py prepare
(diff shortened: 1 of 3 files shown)


mdavidsaver:
> A draft because I'm not sure this is a useful addition, and because setting CROSS_COMPILER_RUNTEST_ARCHS isn't having the desired effect in the GHA builds.

After correcting for https://github.com/epics-base/epics-base/pull/599#issuecomment-2683776023 I see tests run for cross linux-arm for a local build, but still not on GHA...

PR #586, opened 2025-01-16
Add fd manager test

mdavidsaver:
Replace fdmgrTest with an actual test of the fdManager code.

I don't claim this as exhaustive test coverage.  I think it does cover some of the quirks of fdManager, whether they are relevant or not.  eg. `fdManager::process(minTime)` will wait the full `minTime` if there are expired timers, but not I/O.

I hope this will be of some use when extending/rewriting fdManager to use `poll()` a la. https://github.com/epics-base/epics-base/pull/582#issuecomment-2578300774 .


mdavidsaver:
Included in, and not redundant to, #587

PR #587, opened 2025-01-20
fdManager changed to use poll()

dirk-zimoch:
The implementation using select() limits file descriptors to FD_SETSIZE, typically 1024. This number is too low for some applications, for example for the CA gateway.
Therefore, use poll() instead on all architectures that support it.


mdavidsaver:
Requested examples:

Detecting RTEMS "legacy" network stack, which only has `select()`

https://github.com/epics-base/epics-base/blob/e4ad4becde5ebe76b2251bef0fdfccd4fd1101ea/modules/libcom/RTEMS/posix/rtems_init.c#L90-L94

Windows has ~~`poll()`~~ [`WSAPoll()`](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsapoll).  The different names can be handled with some macro [rewriting](https://github.com/epics-base/pvxs/blob/ff1d6510cbf84229f7018eacf4008ecd85dce565/test/testsock.cpp#L30-L35).

mdavidsaver:
I would recommend switching to to implementation hiding (aka [PImpl](https://en.cppreference.com/w/cpp/language/pimpl)) to avoid introducing additional complexity to `fdManager.h`, and potentially enabling two different versions of `fdManager.cpp` if the two implementations end up being different enough.

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -268,6 +336,17 @@ void fdManager::installReg (fdReg &reg)
...
     }
+#ifdef FDMGR_USE_POLL
+    // keep enough event slots for all fds to become available at once
+    if (++nfds > pollfdsCap) {
+        if (pollfdsCap == 0)
+            pollfdsCap = 16;
+        else
+            pollfdsCap *= 2;
+        delete[] pollfds;
+        pollfds = new struct pollfd[pollfdsCap];
```
mdavidsaver: `std::vector<>` will do this for us, and defend against memory leaks.

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -31,27 +31,41 @@ using std :: max;
...
-const unsigned mSecPerSec = 1000u;
-const unsigned uSecPerSec = 1000u * mSecPerSec;
+static const unsigned mSecPerSec = 1000u;
+static const unsigned uSecPerSec = 1000u * mSecPerSec;
+
+#ifdef FDMGR_USE_POLL
+static const int PollEvents[] = { // must match fdRegType 
+    POLLRDNORM | POLLRDBAND | POLLIN | POLLHUP | POLLERR,
+    POLLWRBAND | POLLWRNORM | POLLOUT | POLLERR,
+    POLLPRI};
```
mdavidsaver: I do not think it is desirable to request events absent specific handling.  If one of these unhandled events does occur, the result would be spinning in a tight loop.  eg. `POLLRDBAND` and `POLLWRBAND` are not handled by a regular `recv()`.

mdavidsaver: fyi. how libevent maps `poll()` events https://github.com/libevent/libevent/blob/9a6a799e7002aae09d5e7cb244bdb913eed75a04/poll.c#L212-L220 .

dirk-zimoch: I took the mapping from select to poll from here:
https://www.man7.org/linux/man-pages/man2/select.2.html
See under „Notes“ the paragraph „ Correspondence between select() and poll() notifications“.

mdavidsaver: Fair.  I guess we are looking for bug-for-bug compatibility...

dirk-zimoch:
Modified to use std::vector and hide implementation details.

dirk-zimoch:
Tested with CA gateway on RHEL8 using 1815 file descriptors.

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -19,39 +19,108 @@
...
+// hopefully its a reasonable guess that poll()/select() and epicsThreadSleep()
 // will have the same sleep quantum
 //
-LIBCOM_API fdManager::fdManager () : 
-    sleepQuantum ( epicsThreadSleepQuantum () ),
-        fdSetsPtr ( new fd_set [fdrNEnums] ),
-        pTimerQueue ( 0 ), maxFD ( 0 ), processInProg ( false ),
-        pCBReg ( 0 )
+LIBCOM_API fdManager::fdManager () :
+    priv(new fdManagerPrivate(*this))
```
dirk-zimoch, line 122: Oops, memory leak.

dirk-zimoch: switching to unique_ptr ...

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -79,62 +147,115 @@ LIBCOM_API fdManager::~fdManager()
...
+#endif
         ++iter;
     }
 
     if ( ioPending ) {
+#ifdef FDMGR_USE_POLL
+        if (minDelay * mSecPerSec > INT_MAX)
+            minDelay = INT_MAX / mSecPerSec;
+
+        int status = poll(&priv->pollfds.front(), // ancient C++ has no vector.data()
```
mdavidsaver: wrt. "ancient C++".  The pattern I have seen used most is `&vec[0]`.

More significantly, be careful as either one is UB if `vec.empty()`.  The MSVC library has `assert()` that this is so.

Is it true that `ioPending!=0` implies `!priv->pollfds.empty()`?

dirk-zimoch: Here, ioPending == pollfds.size().

dirk-zimoch:
Rebased on top of #586 (Add fd manager test).

dirk-zimoch:
The Windows builds fail  the fdManagerTest with
`fdManager: poll() failed because "An invalid argument was supplied. "`
Any idea? If Windows poll does not work, I can switch back to select.

dirk-zimoch:
MS says `WSAINVAL` is returned when none of the fds are valid or some other condition that does not form a meaningful sentence: "if the WSAPOLLFD structures pointed to by the fdarray parameter when requesting socket status."

dirk-zimoch:
I think I have to filter out the bits that are meaningful only in `revents` but not in `events`. Linux ignores them but Windows is offended.

modules/libcom/src/fdmgr/fdManager.cpp
mdavidsaver: unique_ptr is introduced with c++11, when auto_ptr was removed.  fyi. for compatibility I have used the auto_ptr name with a conditional typedef from unique_ptr.  eg.

https://github.com/epics-base/epics-base/blob/065fe7cab62bc6b88e32e61e022e910554b05c7f/modules/ca/src/client/udpiiu.h#L44-L51

mdavidsaver:
> Any idea? If Windows poll does not work, I can switch back to select.

fyi. asyn uses WSAPoll on Windows.

https://github.com/epics-modules/asyn/blob/0ed37b298307f3c32718bd6b28fcdf4bd459bacf/asyn/drvAsynSerial/drvAsynIPPort.c#L77-L80

dirk-zimoch:
I wonder if I should filter the poll flags only on Windows (for efficiency)?
Linux, MacOS X and RTEMS explicitly state that the filtered flags are "only returned in `revents`; ignored in `events`", while Windows writes:

> The [WSAPOLLFD](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/ns-winsock2-wsapollfd) structure must only contain a combination of the above flags that are supported by the Winsock provider [POLLRDBAND, POLLRDNORM, POLLWRNORM]. Any other values are considered errors and WSAPoll will return SOCKET_ERROR. [...] If the POLLPRI flag is set on a socket for the Microsoft Winsock provider, the WSAPoll function will fail.

dirk-zimoch:
@mdavidsaver, your [fd manager tests](#586 ) introduce several Codacy warnings. Mind to clean up?

mdavidsaver:
> several Codacy warnings. Mind to clean up?

These all looks like nuisance warnings to me.  I'll see what I can do to quiet them.

mdavidsaver:
> I wonder if I should filter the poll flags only on Windows (for efficiency)?

I don't think this would make a practical difference, so maybe do what makes the source code shorter?

dirk-zimoch:
> These all looks like nuisance warnings to me

True. But then why do we have a test for it? Someone has set up Codacy and has decided that checks should fail if we use for example C-style typecasts in C++. If that is a nuisance than I suggest to remove those checks. We have enough failure-noise in our checks.

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -19,122 +19,256 @@
...
+#ifdef FDMGR_USE_SELECT
+    fd_set fdSets[fdrNEnums];
+    SOCKET maxFD;
+#endif
+
+    //
+    // Set to fdreg when in call back
+    // and nill otherwise
+    //
+    volatile fdReg* pCBReg;
```
mdavidsaver, line 82: What is the intent of adding `volatile` here?

I don't think it would have any practical effect.  fdManager has never been reentrant.

mdavidsaver:
As a note, this PR will conflict with part of #298 making the `fileDescriptorManager` singleton "lazy".  I think this PR should take priority.

mdavidsaver:
> Mind to clean up?

I have removed the dreaded C casts.  Hopefully this will please the Codacy.

ralphlange:
> > Mind to clean up?
> 
> I have removed the dreaded C casts. Hopefully this will please the Codacy.

Thank you! It definitely pleases me. IMHO, they should have dropped the C casts at some point.

modules/libcom/src/fdmgr/fdManager.cpp
anjohnson: Should this snippet become a new epicsAutoPtr.h header? We're already using it twice in modules/ca and once in modules/database (except they're defining it in `namespace ca`). Make it a separate PR though.

modules/libcom/src/fdmgr/fdManager.cpp
dirk-zimoch: To silence a Codacy warning. It did not believe that calling a callback function could change the value.

modules/libcom/src/fdmgr/fdManager.cpp
dirk-zimoch: Yes. But I found that out of scope for this PR. I can rebase this on top of such a PR.

modules/libcom/test/fdManagerTest.cpp
```diff
@@ -0,0 +1,317 @@
...
+        if(sd!=INVALID_SOCKET)
+            epicsSocketDestroy(sd);
+    }
+    void swap(Socket& o) {
+        std::swap(sd, o.sd);
+    }
+    osiSockAddr bind()
+    {
+        osiSockAddr addr;
+        memset(&addr, 0, sizeof(addr));
```
dirk-zimoch: Codacy complains about using memset to clear a buffer, because it is not a safe way to delete sensitive information as the compiler may optimize out any modification of a variable which goes out of scope without being read.
https://docs.codeant.ai/application_rules/c/lang/security#insecure-use-memset
However, this is not the case here. It is clearly initialization not scrubbing. I think this is a false positive by Codacy. It looks like Codacy lazily flags any use of memset(ptr,0,cnt) instead of checking if the variable is accessed later before going out of scope. Or maybe it (silently) argues that addr is not read, only addr.sa is, so clearing any other member of addr may be optimized away.

Anyway, I will try if initializing addr to {0} instead will make Codacy shut up.

dirk-zimoch: Works.

anjohnson:
Needs a rebase, probably due to a conflict in Release Notes.

dirk-zimoch:
Rebased.

modules/libcom/src/fdmgr/fdManager.cpp
mdavidsaver: imo. we should not introduce this unique_ptr -> auto_ptr alias as a Base API at this point.  I see 7.1 in the fairly near future, with a c+11 minimum, which will render this workaround obsolete.

dirk-zimoch:
CA gateways of SLS are now running for 3 weeks with this patch. No incidents so far.

anjohnson:
Core Group: ANJ to test this at APS if possible.

modules/libcom/src/fdmgr/fdManager.cpp
mdavidsaver: Well, codacy can not know that fdmanager is not designed to be reentrant, and having `volatile` here doesn't hurt anything.  So ok.

modules/libcom/src/fdmgr/fdManager.cpp
mdavidsaver: Since there is only one usage in a public header, this could just be changed to:

```c++
#if __cplusplus>=201103L
    std::unique_ptr <struct fdManagerPrivate> priv;
#else
    std::auto_ptr <struct fdManagerPrivate> priv;
#endif
```

Since I sense that Dirk may be loosing patience with me, I won't ask him to change this.  If I care enough, I will do this myself after this PR is merged.

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -222,82 +364,87 @@ void fdReg::destroy()
...
-// fdManager::installReg ()
+// fdManager::installReg()
 //
-void fdManager::installReg (fdReg &reg)
+void fdManager::installReg(fdReg &reg)
 {
-    this->maxFD = max ( this->maxFD, reg.getFD()+1 );
-    // Most applications will find that its important to push here to
+#ifdef FDMGR_USE_SELECT
+    priv->maxFD = std::max(priv->maxFD, reg.getFD()+1);
```
mdavidsaver, line 405: I notice that `maxFD` is not recomputed in `removeReg()`, so it will never decrease.  However, this was the case before.  So not a regression.

modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -203,9 +345,9 @@ LIBCOM_API void fdManager::process (double delay)
          * of select()
          */
         epicsThreadSleep(minDelay);
-        this->pTimerQueue->process(epicsTime::getCurrent());
+        priv->pTimerQueue->process(epicsTime::getCurrent());
     }
-    this->processInProg = false;
+    priv->processInProg = false;
```
mdavidsaver, line 350: As a note.  I looked for, but did not find any new `return` added above, so I think the `processInProg` would be cleared in all cases excepting an exception.  This was the case before, meaning that an exception from a user callback was/is not recoverable.   I don't think any new `throw` are implicitly added, so no objections.

mdavidsaver (approved):
I think this PR is ready, although I agree that more testing with cagateway is desirable since the new fdManagerTest is not comprehensive.

dirk-zimoch:
> Core Group: ANJ to test this at APS if possible.

I checked our (PSI) facility schedule. I would be able to upgrade the SwissFEL gateways in April after the collaboration meeting. Thus, I would feel confident to say this is good if no problem shows by end of May.

anjohnson:
DZ: Stable running at SwissFEL.

Core Group: approved, please rebase and read #628 for how to move the Release entry to a new file.

dirk-zimoch:
Rebased and Release info moved to new file.

anjohnson (approved):
Thanks, merging.

mdavidsaver:
I suspect this is a false positive [failure](https://ci.appveyor.com/project/mdavidsaver/epics-base/builds/52248022/job/o4ear5oo8hpwkdob).  Although I am not immediately sure how.

```
fdManagerTest.tap ............ 
not ok  2 - trig.expired
Failed 2/13 subtests 
```

anjohnson:
I agree that it's probably a false positive.

In the `testOnlyTimer()` code below, it looks like the `trig.expired` test is relying on the 0.1 second `trig_timer` delay having fired before the `mgr.process(0.2)` call returns.

```C++
    ScopedTimer trig_timer(mgr.createTimer()),
                never_timer(mgr.createTimer());
    epicsTime now(epicsTime::getCurrent());
    trig_timer.timer.start(trig, now+0.1);
    never_timer.timer.start(never, now+9999999.0);
    mgr.process(0.2);
    testOk1(trig.expired);
    testOk1(!never.expired);
```

That code could be a source of problems, since we know that on Appveyor background threads don't always get to run as quickly as they do on other systems. I would certainly increase the process time, maybe make it 2 or even 5 seconds?

mdavidsaver:
> Appveyor background threads don't always get to run as quickly

I don't think taking too long would cause this failure.  Rather, something is not waiting long enough.

The `testOnlyTimer()` case is meant to exercise the `epicsThreadSleep()` code path in `fdManager::process()` (as opposed to the `select()/poll()` path).  So I think the only ways this test fails is if `epicsThreadSleep(minDelay)` sleeps for __less__ than 0.1 seconds.  Or if the sleep is long enough, but `epicsTime::getCurrent()` somehow returns an earlier time.

mdavidsaver:
Thinking further...  Since epicsTimer is involved, then https://github.com/epics-base/epics-base/issues/106 probably comes into play.  Bottom line, this test failure almost certainly not a regression caused by this PR.

anjohnson:
The fdManager test has been failing frequently on Appveyor, the last 2 times this was the contents of the failed fdManager.tap file:
```tap
1..13
ok  1 - Did nothing
not ok  2 - trig.expired
ok  3 - !never.expired
ok  4 - connect() 0 0
ok  5 - readable
ok  6 - OneShot event mask 3
ok  7 - readable
ok  8 - testing == testing
ok  9 - expired
ok 10 - !readable
ok 11 - tmo.expired
ok 12 - readable
```

The code for that failing test is this:
```cpp
void testOnlyTimer()
{   
    fdManager mgr;
    Expire trig, never;
    ScopedTimer trig_timer(mgr.createTimer()),
                never_timer(mgr.createTimer());
    epicsTime now(epicsTime::getCurrent());
    trig_timer.timer.start(trig, now+0.1);
    never_timer.timer.start(never, now+9999999.0);
    mgr.process(0.2); 
    testOk1(trig.expired);
    testOk1(!never.expired);
}       
```

It doesn't particularly surprise me that Appveyor fails when only given 0.2 seconds, we've solved similar issues in the past by giving it longer delays, but never really worked out what's actually happening on their systems. Does anyone object to my changing that 0.2 to be 1 or 2 seconds, or maybe call `mgr.process(0.2)` more than once before failing the `trig.expired` check? There is a similar call to `empty.process(0.1)` in the `testEmpty()` routine which could also be increased to avoid any similar issues there; all of the other tests of fdManager wait for 1 or 5 seconds.

Given a choice between this test taking a few extra seconds everywhere and it failing on Appveyor, I'd take the slow-up every time, although we could also adjust the timeout values we use based on whether `EPICS_TEST_IMPRECISE_TIMING` is set.

dirk-zimoch:
I think that for tests performance is not an issue. Simply increase the timeout and hopefully be done with it. It takes no extra time where the trigger fires in time already now and the extra time on Appveyor is negligible compared to their job queuing time.

PR #588, opened 2025-01-22
import CAref function and return codes reference to doxygen comments

minijackson:
Import all function documentations, return code reference into the header files, as doc comments, in the hopes of moving the `CAref.html` manual to https://docs.epics-controls.org, which will be done in a later PR.

Also added doc comments to some other low-hanging fruits, for example when plain comments where near, or when it was obvious to do so.

I haven't removed the function reference from `CAref.html` since, if everyone agrees, I would like to migrate it in its entirety at a later point.

### DIFF
--- a/documentation/Makefile
+++ b/documentation/Makefile
@@ -65,6 +65,11 @@ libcom_HEADERS += yajl_common
 libcom_HEADERS += yajl_gen
 libcom_HEADERS += yajl_parse
 
+ca_HEADERS += cadef
+ca_HEADERS += caerr
+ca_HEADERS += caeventmask
+ca_HEADERS += db_access
+
 database_HEADERS += chfPlugin
 database_HEADERS += dbChannel
 database_HEADERS += dbCommon
@@ -128,7 +133,7 @@ menu_HEADERS += menuScan
 menu_HEADERS += menuSimm
 menu_HEADERS += menuYesNo
 
-HEADER_TYPES = libcom database record menu
+HEADER_TYPES = libcom ca database record menu
 
 HEADER_MD_FILES = $(foreach t, $(HEADER_TYPES), \
     $(addsuffix _h.md, $($t_HEADERS)))
--- a/documentation/ca-API.rst
+++ b/documentation/ca-API.rst
@@ -1,0 +1,7 @@
+Channel Access C/C++ APIs
+=========================
+
+.. toctree::
+   :maxdepth: 1
+   :caption: CA Header Files
+
--- a/documentation/index.rst
+++ b/documentation/index.rst
@@ -25,6 +25,7 @@ EPICS Base Documentation
    :caption: C/C++ Headers
 
    libcom-api
+   ca-api
    database-api
    record-api
    menu-api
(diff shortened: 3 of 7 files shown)


documentation/Makefile
```diff
@@ -16,9 +16,13 @@ libcom_HEADERS += alarm
 libcom_HEADERS += alarmString
 libcom_HEADERS += asTrapWrite
 libcom_HEADERS += bucketLib
+libcom_HEADERS += cadef
+libcom_HEADERS += caerr
+libcom_HEADERS += caeventmask
```
anjohnson: Channel Access is not part of libCom. All these CA header filenames should be added to a new `ca_HEADERS` variable, add `ca` to `HEADER_TYPES`, and create a new file ca-API.rst

anjohnson (changes requested):
Thanks, this looks like a great start. I won't comment on the text since I assume that came from the CA Reference manual, I just want to ensure that this goes into its own API section, not into the libcom section.

documentation/Makefile
minijackson: Good point, done.

minijackson:
I know there's a checkbox in the ReadTheDocs project settings for building pull requests, would it be possible to activate it? I don't have access to the epics-base project.

ralphlange:
> I know there's a checkbox in the ReadTheDocs project settings for building pull requests, would it be possible to activate it? I don't have access to the epics-base project.

Aye. Check.

mdavidsaver:
> Thanks, this looks like a great start ...

I very much agree.  Thanks for taking the time to do this.

minijackson:
Here's the link to the ReadTheDocs build of the PR: https://epics--588.org.readthedocs.build/projects/base/en/588/ca-api.html

anjohnson:
Formatting looks nice.

There are lots of macros (and legacy routines) that have no description in the output, but might have been grouped with related items inside the header file. Could they be grouped in the output by adding the appropriate Doxygen constructs for that to the headers?

There is lots of descriptive text in the CARef.html file which is essential documentation, where do you think that should be moved to? I'm not expecting it to be part of this PR, I just want to see your overall plan for that.

minijackson:
> Could they be grouped in the output by adding the appropriate Doxygen constructs for that to the headers?

I thought I wasn't able to make grouping work, but it turns out I was using the wrong king of group. Using `\defgroup` / `\ingroup` / `\addtogroup` does *not* work. I had to use what Doxygen calls "[Member Groups](https://www.doxygen.nl/manual/grouping.html#memgroup)", with `@{` / `@}` / `\name`

I've added an example in my latest commit, but I'd be more comfortable if someone more knowledgeable about this file chooses how to group things, and how to name the groups.

> There is lots of descriptive text in the CARef.html file which is essential documentation, where do you think that should be moved to? I'm not expecting it to be part of this PR, I just want to see your overall plan for that.

My plan, if you're all okay with it, is to put everything prosaic in epics-docs, and everything that looks like generated reference documentation in epics-base.

So for epics-base:

- C/C++ documentation (this PR)
- CLI reference (#590)

And for epics-docs:

- Channel Access configuration (with the list of environment variables)
- Building an application
- Troubleshooting
- Function Call Interface General Guidelines

I'm have the PR for epics-docs pretty much done, but I prefer to do it after things are merged in epics-base, so that I can use and test Intersphinx links.

PR #590, opened 2025-01-24
docs: Channel Access CLI reference

minijackson:
Follow-up on #588, imported the CLI reference from `CAref.html` into their own Markdown files, next to the source code, so that they have a higher chance of being updated if the source code is changed.

Also added `msi.md` to the index (with some light changes), since it was already here.

Link to the PR built by ReadTheDocs: https://epics--590.org.readthedocs.build/projects/base/en/590/ca-cli.html

### DIFF
--- a/documentation/Makefile
+++ b/documentation/Makefile
@@ -138,6 +138,7 @@ RTD_SRC = $(COMMON_DIR)/rtd-src
 
 DOCS += README.md
 DOCS += RELEASE_NOTES.md
+DOCS += ca-cli.md
 
 include $(TOP)/configure/RULES
 
--- a/documentation/ca-cli.md
+++ b/documentation/ca-cli.md
@@ -5,4 +5,5 @@
 :caption: Tools
 
 caget
+cainfo
 ```
--- a/documentation/ca-cli.md
+++ b/documentation/ca-cli.md
@@ -18,4 +18,5 @@ acctst
 catime
 casw
 caEventRate
+ca_test
 ```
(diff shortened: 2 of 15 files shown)


minijackson:
Oh before I forgot, in the CAref.html manual, there's documentation about an `excas` command:

> This is an example CA server that is sometimes used for testing purposes. An example server can be created with the makeBaseApp Perl script, as described in the Application Developer’s Guide.

But I couldn't find it in epics-base. Does somebody knows anything about it?

anjohnson:
`excas` was part of the Portable Channel Access Server code, which was unbundled into a separate support module in the [epics-modules](https://github.com/epics-modules/pcas) organization.

If the excas documentation can easily be extracted maybe you could create a PR to add it there? Nothing fancy though, it probably just needs a .html file adding to the `src/templates/top/caServerApp` directory with a Makefile `HTMLS += excas.html` entry to install it.

There are many other command-line programs (I wouldn't call the CA ones an "interface" though) that come with Base, and it would be nice to collect all of these together. Some of them were described in the AppDevGuide which Timo has converted, while some of the Perl scripts in the `bin/<arch>` have POD documentation built-in. That can be viewed by running for example `perldoc bin/<arch>/podToMD.pl`. Could these be added to the Command Line Reference section?

Thanks!

minijackson:
> If the excas documentation can easily be extracted maybe you could create a PR to add it there? Nothing fancy though, it probably just needs a .html file adding to the src/templates/top/caServerApp directory with a Makefile HTMLS += excas.html entry to install it.

Just checked, it seems the pcas already has an [`excas.html`](https://github.com/epics-modules/pcas/blob/master/src/template/top/caServerApp/excas.html) file.

ralphlange:
For eight years, yes.

PR #591, opened 2025-01-28
Fix an internal link

ttkorhonen:
Fixed the link that points to the Device Support Interface section, internal to the document. I assume that is the intention here; in any case, the original definition in the pod file was not working.


anjohnson (approved):
Agreed, thanks!

PR #592, opened 2025-02-03
Add dbServerStats() API

anjohnson:
This adds a routine `dbServerStats()` for getting channel and client counts from a named server layer that registered with the IOC's dbServer API and provided a `stats()` method.

The IOC's Channel Access server registers using the name `rsrv`, the pvxs server as `qsrv2` (the registered name appears in the output from the iocsh `dbsr` command).

Code in modules such as iocStats can use `#ifdef HAS_DBSERVER_STATS` to detect Base versions which provide this routine.

### DIFF
--- a/modules/database/test/ioc/db/dbServerTest.c
+++ b/modules/database/test/ioc/db/dbServerTest.c
@@ -40,6 +40,8 @@ void oneReport(unsigned level)
 void oneStats(unsigned *channels, unsigned *clients)
 {
     oneState = STATS_CALLED;
+    *channels = 2;
+    *clients = 1;
 }
 
 int oneClient(char *pbuf, size_t len)
@@ -128,8 +130,9 @@ MAIN(dbServerTest)
     char name[16];
     char *theName = "The One";
     int status;
+    unsigned ch=0, cl=0;
 
-    testPlan(25);
+    testPlan(29);
 
     /* Prove that we handle substring names properly */
     epicsEnvSet("EPICS_IOC_IGNORE_SERVERS", "none ones");
@@ -163,7 +166,12 @@ MAIN(dbServerTest)
 
     testDiag("Checking server methods called");
     dbsr(0);
-    testOk(oneState == REPORT_CALLED, "dbsr called report()");
+    testOk(oneState == REPORT_CALLED, "dbsr called one::report()");
+    testOk(dbServerStats("none", &ch, &cl) != 0, "Stats: unknown name rejected");
+    testOk(dbServerStats("no-routines", &ch, &cl) != 0, "Stats: no-routine rejected");
+    testOk(dbServerStats("one", &ch, &cl) == 0 && oneState == STATS_CALLED,
+        "dbServerStats('one') called one::stats()");
+    testOk(ch == 2 && cl == 1, "Stats: counts returned as expected");
 
     oneSim = NULL;
     name[0] = 0;
--- a/modules/database/test/ioc/db/dbServerTest.c
+++ b/modules/database/test/ioc/db/dbServerTest.c
@@ -132,7 +132,7 @@ MAIN(dbServerTest)
     int status;
     unsigned ch=0, cl=0;
 
-    testPlan(32);
+    testPlan(35);
 
     /* Prove that we handle substring names properly */
     epicsEnvSet("EPICS_IOC_IGNORE_SERVERS", "none ones");
@@ -154,6 +154,9 @@ MAIN(dbServerTest)
     testDiag("Registering dbServer 'disabled'");
     testOk(dbRegisterServer(&disabled) == 0, "Registration accepted");
 
+    testOk(dbServerStats("one", &ch, &cl) == -1 && oneState == NOTHING_CALLED,
+            "dbServerStats returns error before IOC running");
+
     testDiag("Changing server state");
     dbInitServers();
     testOk(oneState == INIT_CALLED, "dbInitServers");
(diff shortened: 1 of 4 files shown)


simon-ess:
LGTM. Is it worth adding the dbServer api calls to QSRV(1)?

ralphlange:
> LGTM. Is it worth adding the dbServer api calls to QSRV(1)?

IMHO: No. I'd see it as incentive to move to the newer stack.

simon-ess:
I thought that might be the case, but I wasn't sure how long this would be held back in that case, particularly given the request from @hanlet for iocStats.

mdavidsaver (changes requested):
How is a caller supposed to know what name to pass to `dbServerStats()`?  Hard coded name strings?

For that matter, is this per server level of granularity useful?

@simon-ess Do you have time to try to use this API in iocStats?

ralphlange:
> How is a caller supposed to know what name to pass to `dbServerStats()`? Hard coded name strings?

Obviously, the caller needs a function to get the list of registered servers.

> For that matter, is this per server level of granularity useful?

Hard to tell beforehand.
The caller can always add up numbers to get summary information, but will never know by-server stats if they are not available in the API.
Also, server management would need to loop over every server in the list of registered servers anyway, so why would it hide by-server stats from the caller?

anjohnson:
> How is a caller supposed to know what name to pass to dbServerStats()? Hard coded name strings?

Yes. With some effort they could be provided in the `st.cmd` script, but for iocStats they will probably be hard-coded into the `ioc.template` file as well as in the display screens. As my second paragraph in the description above says,

> the registered name appears in the output from the iocsh `dbsr` command

so they don't require spelunking through source code to find.

If the caller is similar to iocStats it has to have already loaded record instances for each server's count PV that it publishes. It can't load new record instances at iocInit (after all the servers have registered themselves) so getting a list of the server names at run-time isn't useful, it needs to know them earlier than that.

We can add another API to get all the registered server names when someone wants to create software which publishes the statistics as an NTtable say. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

> is this per server level of granularity useful?

Yes, iocStats already publishes the PVs `$(IOC):CA_CONN_CNT` and `$(IOC):CA_CLNT_CNT`. I want similar PVs `$(IOC):PVA_CONN_CNT` and `$(IOC):PVA_CLNT_CNT` so I can see how many clients are using each protocol.

Is @hanlet currently using QSRV or QSRV2?

mdavidsaver:
> ... I can see how many clients are using each protocol.

Right, which isn't quite what a caller would get.  I remain skeptical that per-server client counts will be the common use case, and I certainly don't think this is the only use case.

Can you add a way to sum the counts for all servers, both expected and unexpected?  eg. `name=NULL` or `name="*"`

ralphlange:
> Can you add a way to sum the counts for all servers, both expected and unexpected? eg. `name=NULL` or `name="*"`

+1
That would be a reasonable alternative to a call returning the registered servers.

ralphlange:
> Right, which isn't quite what a caller would get. I remain skeptical that per-server client counts will be the common use case, and I certainly don't think this is the only use case.

True.
Still: if the iocStats data ever gets a PVA representation - you don't really need the database here, do you? -  wouldn't that be an array of name-plus-numbers structures for the 'servers' part?

modules/database/test/ioc/db/dbServerTest.c
```diff
@@ -163,8 +166,23 @@ MAIN(dbServerTest)
...
+    testOk(dbServerStats("one", &ch, &cl) == 0 && oneState == STATS_CALLED,
+        "dbServerStats('one') called one::stats()");
+    testOk(ch == 2 && cl == 1, "Stats: counts returned as expected");
+
+    ch = 0; cl = 0; oneState = NOTHING_CALLED;
+    testOk(dbServerStats(NULL, NULL, &cl) == 0 && oneState == STATS_CALLED,
+        "dbServerStats(NULL, &cl) called one::stats()");
+    testOk(dbServerStats(NULL, &ch, NULL) == 0 && oneState == STATS_CALLED,
+        "dbServerStats(NULL, &ch) called one::stats()");
+    testOk(ch == 2 && cl == 1, "Stats: counts returned as expected");
```
mdavidsaver: I can imagine seeing:

```
not ok 42 Stats: counts returned as expected
```

and thinking... thank you, now what were they?

anjohnson: Ack, I'm fixing the Codacy issue too.

hanlet:
Good morning Mark and Andrew,
I'm not familiar with the CI/CD procedure, tests, or this documentation, but I had been having a seg fault
failure with a non-updated version of this module.  At the core meeting this week, Michael mentioned
that there was a scenario when rsrv was turned off that would cause seg faults; when I set the
SERVER_IGNORE="", so as to not turn off rsrv, that my seg fault in this module went away.

Perhaps this is helpful,
Pierrick

md: edited to remove email debris...

mdavidsaver:
Gosh, that urldefense snooper makes a real mess...  @hanlet are those URLs relevant to your message, or incidental email reply debris?

mdavidsaver:
A shorter implementation.

I find myself wondering if it would be useful to instead return the number of dbServer instances which contributed to the count?  (so -1 if not running, 0 for no match/no servers, positive num. of servers)

```c
int dbServerStats(const char *name, unsigned *channels, unsigned *clients)
{
    dbServer *psrv = (dbServer *)ellFirst(&serverList);
    unsigned tch = 0, tcl = 0, nmatch = 0;

    if (state != running || !psrv)
        return -1;

    for (; psrv; psrv = (dbServer *)ellNext(&psrv->node))
    {
        if(psrv->stats && (!name || strcmp(name, psrv->name) == 0)) {
            unsigned lch = 0, lcl = 0;
            psrv->stats(&lch, &lcl);
            tch += lch;
            tcl += lcl;
            nmatch++;
            if(name)
                break; /* no duplicates allowed in */
        }
    }
    if (nmatch) {
        if (channels) *channels = tch;
        if (clients) *clients = tcl;
        return 0;
    }
    return -1;
}
```

anjohnson:
So `return nmatch` instead, sure. That's also another statistic which could be reported by iocStats. Conditionally setting the stats returned makes less sense with the count, so I've removed the `if (nmatch)` too.

mdavidsaver (approved):
Now looks ok to me.

Issue #596, opened 2025-02-06
casStatsFetch() crashes if RSRV not initialized


mdavidsaver:
I find that `casStatsFetch()` unconditionally tries to lock a mutex `clientQlock`, which is allocated from `rsrv_init()`.  This presents a problem when RSRV is ignored per dbServer and `rsrv_init()` does not run.

The same applies to other functions in `rsrv.h`.

@simon-ess @hanlet fyi. I expect trouble with iocStats

mdavidsaver:
> The same applies to other functions in `rsrv.h`.

Only `casClientInitiatingCurrentThread()`.  `casr()` was fixed way back ad792ae19a29773c9d91bc269a4d83d2d2e972db.

> ... I expect trouble with iocStats

https://github.com/epics-modules/iocStats/pull/21 accounts for calling `casStatsFetch()` too early.  However, with RSRV disabled, too early never ends.  So the advice to wait for `initHookAfterCaServerInit` is no longer sufficient.

anjohnson:
Fixed by #597.

PR #597, opened 2025-02-06
make casStatsFetch() safe-ish when RSRV not initialized

mdavidsaver:
See #596.  Testing these global statics is not reentrant, however prior to `rsrv_init()`, or if it will never be called, then there is no racer.

### DIFF
--- a/modules/database/src/ioc/rsrv/camsgtask.c
+++ b/modules/database/src/ioc/rsrv/camsgtask.c
@@ -163,6 +163,8 @@ void camsgtask ( void *pParm )
 
 int casClientInitiatingCurrentThread ( char * pBuf, size_t bufSize )
 {
+    if ( ! rsrvCurrentClient )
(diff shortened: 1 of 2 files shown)


mdavidsaver:
Applied as 7a6e11cae0f46cdb4da148c278aed3a6daeadb08

modules/database/src/ioc/rsrv/camsgtask.c
```diff
@@ -163,6 +163,8 @@ void camsgtask ( void *pParm )
 
 int casClientInitiatingCurrentThread ( char * pBuf, size_t bufSize )
 {
+    if ( ! rsrvCurrentClient )
+        return RSRV_ERROR; /* not yet initialized, or disabled via dbServer */
```
anjohnson, line 167: Breaks the build on VS-2012 and older, this adds a statement before declarations inside a block.

PR #599, opened 2025-02-11
no longer need extern "C" { } around epicsExport macros

dirk-zimoch:
It is a common error to forget `extern "C" { }` around  `epicsExportAddress`, `epicsExportRegistrar` and `epicsRegisterFunction` in 3rd party C++ code (e.g. device drivers), because it works on most architectures except on Windows. (But it causes warnings when using -std=c++23 because of the deprecated C-style type casts.)

This change provides C++ compatible versions of those macros which no longer require (but still allow) to be wrapped in `extern "C" { }`.


mdavidsaver:
> ... but still allow ...

It seems that `extern "C" {extern "C" {...}}` is understood by the oldest version of all of the compilers I tried on https://godbolt.org/ .

mdavidsaver:
I can't find any usage of either of these macros in epics-base from c++ code.  I think some test coverage is necessary.

Would `dbHeaderTest.c` aka. `dbHeaderTestxx.cpp` be sufficient?  This would only be a syntax test.

dirk-zimoch:
I noticed the missing coverage because base uses those macros only in C. I had renamed some of those files from *.c to *.cpp to test but of course not pushed those modifications.
I was not sure in which module folder to put those tests. I tried libCom but for a realistic use case I wanted to export dset. But that is not defined in libCom.
Also, how to properly do a syntax-only test?

dirk-zimoch:
I see a warning in PVAServerRegister.cpp caused by this. I will have a look.

dirk-zimoch:
> extern "C" {extern "C" {...}}

Totally valid C++. This allows to wrap #include in extern “C” and still have extern “C” inside the header file.

mdavidsaver:
> Also, how to properly do a syntax-only test?

https://github.com/epics-base/epics-base/blob/1d19ba4cc2e8c9ab17288b684022a0682e398bde/modules/database/test/std/rec/dbHeaderTest.c#L8-L10

I think it would be enough to add some variable definitions and exports after the `#include` s.

dirk-zimoch:
In order not to pollute dbHeaderTest.c too much, I  have created a new epicsExportTest.c and epicsExportTestxx.cpp using the same approach.

dirk-zimoch:
Extended the test to check for proper functionality of the macros in both, C and C++.

modules/libcom/src/misc/epicsExport.h
```diff
@@ -75,12 +75,16 @@ typedef void (*REGISTRAR)(void);
...
  */
+#ifdef __cplusplus
+#define epicsExportAddress(typ, obj) \
+    extern "C" { epicsShareExtern typ *EPICS_EXPORT_POBJ(typ,obj); } \
+    epicsShareDef typ *EPICS_EXPORT_POBJ(typ, obj) = reinterpret_cast<typ *>(&obj)
+#else
 #define epicsExportAddress(typ, obj) \
     epicsShareExtern typ *EPICS_EXPORT_POBJ(typ,obj); \
     epicsShareDef typ *EPICS_EXPORT_POBJ(typ, obj) = (typ *) (char *) &obj
+#endif
```
anjohnson: Could something like this be made to work, to avoid a lot of the duplication here and below?

```C++
#ifdef __cplusplus
#  define EXTERN_C(STMT) extern "C" { STMT }
#  define EXPORT_CAST(TYPE, OBJ) reinterpret_cast<TYPE>(OBJ)
#else
#  define EXTERN_C(STMT) STMT
#  define EXPORT_CAST(TYPE, OBJ) (TYPE) (OBJ)
#endif

#define epicsExportAddress(typ, obj) \
    EXTERN_C(epicsShareExtern typ *EPICS_EXPORT_POBJ(typ,obj);) \
    epicsShareDef typ *EPICS_EXPORT_POBJ(typ, obj) = EXPORT_CAST(typ *, &obj)

// ...

#undef EXPORT_CAST
#undef EXTERN_C
```

mdavidsaver: I think I prefer Dirk's current version as it is shorter, and doesn't involve temporary macro definitions in header files.

dirk-zimoch: > doesn't involve temporary macro definitions

That was my concern. You never know which macros have already been defined when this header is included.

dirk-zimoch:
There ist still something wrong with the Windows linking :-(
Don‘t merge yet!

mdavidsaver:
> There ist still something wrong with the Windows linking :-(

Are you running into a situation of trying to dllimport symbols in the executable?  (I think not possible)

mdavidsaver:
> Are you running into a situation of trying to dllimport symbols in the executable? (I think not possible)

I can successfully build with cross mingw by spliting out the exports from `epicsExportTest.c` and linking them into the `dbRecStdTest` library which is already built in the same directory.  However, only by linking in one copy, which defeats the combined C and C++ part of this test.  Perhaps there is some further trick to avoid exporting duplicate symbols?

dirk-zimoch:
I had successfully built for Windows as a cross architecture, running MSVC 2019 `cl` in `wine` on my RHEL host. Thus I am surprised that it does not build in the CI environment.
Somehow, I cannot run the test, because I do not get any *Test.exe, even though I have windows-x64 in my `CROSS_COMPILER_RUNTEST_ARCHS`. I get epicsExportTest.obj and epicsExportTestxx.obj though.

dirk-zimoch:
Next try.
Concerning windows in `CROSS_COMPILER_RUNTEST_ARCHS`: I get .exe files in modules/libcom/test/O.windows-x64/ but not in any other modules/*. That makes it a bit difficult to test locally.

dirk-zimoch:
I got the modules/database tests cross compiled for Windows (had to fix modules/database/test/std/rec/Makefile). Then both, epicsExportTest.exe and epicsExportTestxx.exe build and run without error!
No idea why it fails in the CI environment.

```
$ wine epicsExportTestxx.exe
1..26
ok  1 - registrar1 executed
ok  2 - registrar2 executed
ok  3 - registrar3 executed
ok  4 - registrar4 executed
# Testing if dsets and functions are found
Starting iocInit
## EPICS R7.0.8.2-DEV
## Rev. Breaking_change-901-g1ba9f6211ee35e83028e-dirty
## Rev. Date Git: 2025-02-19 11:32:13 +0100
# Testing if registrars have run
ok  5 - Variable i1 counted registrars: 2 == 2
ok  6 - Variable i2 counted registrars: 2 == 2
# Testing if variables are accessible
ok  7 - Setting i1 = 4 in iocsh
ok  8 - Setting i2 = 5 in iocsh
ok  9 - Variable i1: 4 == 4
ok 10 - Variable i2: 5 == 5
# Testing if dsets work correctly
ok 11 - dbGetField("li1", 5) -> -1 == -1
ok 12 - dbGetField("li2", 5) -> -2 == -2
ok 13 - dbPutField("li1.PROC", 5, ...) -> 0 (Ok)
ok 14 - dbPutField("li2.PROC", 5, ...) -> 0 (Ok)
ok 15 - dbGetField("li1", 5) -> 5 == 5
ok 16 - dbGetField("li2", 5) -> 5 == 5
# Testing if functions work correctly
ok 17 - dbGetField("asub1.A", 5) -> 1 == 1
ok 18 - dbGetField("asub1.B", 5) -> -2 == -2
ok 19 - dbGetField("asub2.A", 5) -> 3 == 3
ok 20 - dbGetField("asub2.B", 5) -> -4 == -4
ok 21 - dbPutField("asub1.PROC", 5, ...) -> 0 (Ok)
ok 22 - dbPutField("asub2.PROC", 5, ...) -> 0 (Ok)
ok 23 - dbGetField("asub1.A", 5) -> 1 == 1
ok 24 - dbGetField("asub1.B", 5) -> 2 == 2
ok 25 - dbGetField("asub2.A", 5) -> 3 == 3
ok 26 - dbGetField("asub2.B", 5) -> 4 == 4

    Results
    =======
       Tests: 26
      Passed:  26 = 100.00%
```

You can see, it is the same commit which fails in CI.

dirk-zimoch:
Looks like mingw is different from native Windows build. I need to set up a local mingw environment to test that...

mdavidsaver:
> Concerning windows in `CROSS_COMPILER_RUNTEST_ARCHS`: I get .exe files in modules/libcom/test/O.windows-x64/ but not in any other modules/*. That makes it a bit difficult to test locally.

I have noticed the same before (https://github.com/epics-base/epics-base/pull/585), and investigated without success.  This time I see a difference.

https://github.com/epics-base/epics-base/blob/7bd3e7aa2e2483bc4a2878742504fac2651487b0/modules/libcom/test/Makefile#L297-L300

https://github.com/epics-base/epics-base/blob/7bd3e7aa2e2483bc4a2878742504fac2651487b0/modules/database/test/ioc/db/Makefile#L210-L213

Note the `_RTEMS` in the second case...

dirk-zimoch:
I fail to  understand why the mingw builds fail (windows builds succeed). It has something to do with this windows dllimport/dllexport bullshit but what? Even with the original epicsExport.h file and only building epicsExportTest.c, not epicsExportTestxx.cpp, mingw fails wit `undefined reference to __imp_pvar_...` linker errors for all the exported stuff.

Is something wrong with the way I use `#include <epicsExport.h>` in epicsExportTest.c? I include it last. so what goes wrong here?

dirk-zimoch:
> Note the `_RTEMS` in the second case..

Exactly. I removed the `_RTEMS` and then the cross Windows tests run.

mdavidsaver:
> It has something to do with this windows dllimport/dllexport bullshit but what?

Let me rephrase https://github.com/epics-base/epics-base/pull/599#issuecomment-2667525872.  You are trying to dllimport/export symbols which are not in a DLL.  There is no import library for the executable, which is where the `__imp_` symbols would be found (for mingw builds at least...).

You could make use of `TESTLIBRARY = dbRecStdTest`.  Since there are already many usages of epicsExport from C code, maybe just add a .cpp file with some exports to that library?

Alternately, maybe just revert to the idea of a syntax only check of exporting?

dirk-zimoch:
rebased onto latest 7.0

dirk-zimoch:
Something is wrong with the RTEMS tests. @mdavidsaver please have a look.

` /bin/bash: line 0: test: 4.9: integer expression expected`

I suspect the following has to do with my test, but what goes wrong for RTEMS only?
```
/home/runner/work/epics-base/epics-base/modules/database/test/std/rec/O.RTEMS-pc686-qemu
[...]
Bailout called.  Further testing stopped:  Failed to load test database
FAILED--Further testing stopped: Failed to load test database
```

dirk-zimoch:
> Something is wrong with the RTEMS tests.

`TESTFILES` was missing. Fixed.

dirk-zimoch:
Problem with motor driver:
```
In file included from ../O.7.0.9_Common/motorRecord.h:515,
                 from ../motorApp/MotorSrc/motorRecord.cc:207:
/usr/local/epics/base-7.0.9/include/epicsExport.h:82:54: error: 'reinterpret_cast' from type 'volatile int*' to type 'int*' casts away qualifiers
   82 |     epicsShareDef typ *EPICS_EXPORT_POBJ(typ, obj) = reinterpret_cast<typ *>(&obj)
      |                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../motorApp/MotorSrc/motorRecord.cc:215:13: note: in expansion of macro 'epicsExportAddress'
  215 | extern "C" {epicsExportAddress(int, motorRecordDebug);}
      |             ^~~~~~~~~~~~~~~~~~
```
I.e. using C++ style cast with `volatile` variables does not work. Would need `const_cast<>` first to cast away `volatile`.

dirk-zimoch:
Fixed. And tests added for `volatile` and `const` variables.
While exporting `const int` or `const double` variables to the ioc shell is probably dangerous, because it does not create a read-only shell variable as it better should, I see that some people want to declare their exported variables `volatile`. And declaring `rset`, `dset` and `drvet` tables `const` would make a lot of sense.

dirk-zimoch:
Hm... Locally, all tests succeed. Strange that they fail now in ci.

documentation/RELEASE_NOTES.md
```diff
@@ -18,6 +18,11 @@ __This version of EPICS has not been released yet.__
 
 __Add new items below here__
 
+### epicsExport simplifications
+
+`epicsExportAddress()`, `epicsExportRegistrar()` and `epicsRegisterFunction()`
+no loger require to be wrapped in `extern "C" { }` in C++ code.
```
mdavidsaver: `loger` -> `longer`

modules/database/test/std/rec/Makefile
```diff
@@ -227,6 +227,30 @@ TARGET_SRCS += dbHeaderTest.cpp
...
+epicsExportTest_LIBS = epicsExportTestLib
+TESTLIBRARY += epicsExportTestxxLib
+epicsExportTestxxLib_SRCS += epicsExportTestxx.cpp
+epicsExportTestxxLib_LIBS += dbCore Com
+TESTPROD_HOST += epicsExportTestxx
+TESTS += epicsExportTestxx
+TARGETS += epicsExportTestxx$(OBJ)
+epicsExportTestxx_SRCS += epicsExportTestMain.c
+epicsExportTestxx_SRCS += epicsExportTestIoc_registerRecordDeviceDriver.cpp
+epicsExportTestxx_LIBS = epicsExportTestxxLib
```
mdavidsaver, line 250: Please add a blank line before each `TESTLIBRARY` and `TESTPROD`.

modules/database/test/std/rec/epicsExportTestMain.c
```diff
@@ -0,0 +1,92 @@
...
+    testdbReadDatabase("epicsExportTest.db", 0, 0);
+    testIocInitOk();
+
+    testDiag("Testing if dsets work correctly");
+    testdbGetFieldEqual("li1", DBF_LONG, -1);
+    testdbGetFieldEqual("li2", DBF_LONG, -2);
+    testdbPutFieldOk("li1.PROC", DBF_LONG, 1);
+    testdbPutFieldOk("li2.PROC", DBF_LONG, 1);
+    testdbGetFieldEqual("li1", DBF_LONG, 5);
+    testdbGetFieldEqual("li2", DBF_LONG, 5);
```
mdavidsaver, line 62:
```
   epicsExportTestxx.tap ..... 
  not ok  7 - dbPutField("li1.PROC", 5, ...) -> 0x2020005 (<Unknown code>)
  not ok  8 - dbPutField("li2.PROC", 5, ...) -> 0x2020005 (<Unknown code>)
  not ok  9 - dbGetField("li1", 5) -> 5 == -1
  not ok 10 - dbGetField("li2", 5) -> 5 == -2
```

Some details about this test failure may be found in the earlier stderr.

```
perl -CSD epicsExportTestxx.t -tap > epicsExportTestxx.tap
  make[5]: Leaving directory '/__w/epics-base/epics-base/modules/database/test/std/rec/O.linux-x86_64'
  device support dset1 not found
  device support dset2 not found
  recGblRecordError: longin: init_record Error (514,3) PV: li1
  recGblRecordError: longin: init_record Error (514,3) PV: li2
  recGblRecordError: read_longin Error (514,5) PV: li1
  recGblRecordError: read_longin Error (514,5) PV: li2
```

So this failure is due to an earlier error during `epicsExportTestIoc_registerRecordDeviceDriver()`.  Which could be detected since this function actually returns an `int` (see registerRecordDeviceDriver.pl`).

I suspect the ultimate cause is some c++ name mangling.

modules/libcom/src/misc/epicsExport.h
```diff
@@ -75,12 +71,24 @@ typedef void (*REGISTRAR)(void);
  *
  * \param typ Object's data type.
  * \param obj Object's name.
- *
- * \note C++ code needs to wrap with @code extern "C" { } @endcode
  */
+#ifdef __cplusplus
+namespace epics {
```
mdavidsaver: A common c++ convention would place definitions like the following in a namespace like `epics::detail` or `epics::impl` to at least hint these they should not be used explicitly to API callers.

modules/libcom/src/misc/epicsExport.h
```diff
@@ -75,12 +71,24 @@ typedef void (*REGISTRAR)(void);
...
+    template<typename T>
+    static inline T* ignore_cv_qualifier(volatile T* x) { return const_cast<T*>(x); }
+    template<typename T>
+    static inline T* ignore_cv_qualifier(const T* x) { return const_cast<T*>(x); }
+    template<typename T>
+    static inline T* ignore_cv_qualifier(T* x) { return x; }
```
mdavidsaver: A note for the future.  `ignore_cv_qualifier()` has the same function as the c++11 [`std::remove_cv`](https://en.cppreference.com/w/cpp/types/remove_cv), although a somewhat different invocation.  Since this would (I think) depend on `decltype()`, it can't yet be used directly.

For the present, I think `ignore_cv_qualifier()` may actually not be inline'd in all cases (compiler versions) since `inline` is only a hint.  The c++11 `constexpr` modifier would ensure this, but we're not c++11 yet...

Maybe just keeping the heavy handed `(typ *) (char *)` casting would be a more straightforward solution/workaround until then?

modules/libcom/src/misc/epicsExport.h
```diff
@@ -75,12 +71,24 @@ typedef void (*REGISTRAR)(void);
...
+}
+#define epicsExportAddress(typ, obj) \
+    extern "C" { epicsShareExtern typ *EPICS_EXPORT_POBJ(typ,obj); } \
+    epicsShareDef typ *EPICS_EXPORT_POBJ(typ, obj) = reinterpret_cast<typ *>(epics::ignore_cv_qualifier(&obj))
```
mdavidsaver: fyi. when referencing into C++ namespaces from a C macro, best practice is to start from the global namespace be prefixing `::`.  So eg. `epics::ignore_cv_qualifier` would become `::epics::ignore_cv_qualifier``.

mdavidsaver (changes requested):

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: We cannot assume C++11 yet. I could branch depending on C++ version.  And maybe rename `ignore_cv_qualifier` to `remove_cv`.

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: OK.

modules/database/test/std/rec/epicsExportTestMain.c
dirk-zimoch: None of my architectures, including RHEL8, RHEL9, Windows(MSVC 2019) and mingw failed. I will push some variants to see what causes it....

documentation/RELEASE_NOTES.md
dirk-zimoch: Oops. Fixed.

modules/database/test/std/rec/Makefile
dirk-zimoch: OK. I also moved it before the dbHeaderTest, so that is is not under the comment
`# These are compile-time tests, no need to link or run`
as we actually run it now.

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: Ok.

modules/database/test/std/rec/epicsExportTestMain.c
dirk-zimoch: Shouldn't mangling problems cause compiler errors?

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: > `ignore_cv_qualifier()` has the same function as the c++11 `std::remove_cv`, although a somewhat different invocation. 

No.  `std::remove_cv` can only be uses to compare types qualifier-less. But not to cast pointers.

dirk-zimoch: > Maybe just keeping the heavy handed (typ *) (char *) casting would be a more straightforward solution/workaround until then?

According to my tests, that seems to be the only way to make the CI tests succeed. I still do not understand, why only half of the test architectures in don't work and why every test always succeeds on my local system.
I still don't like this solution.

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: I gave up on this, going back to C-style cast.

modules/database/test/std/rec/epicsExportTestMain.c
dirk-zimoch: Using C-style casts solved the problem.

dirk-zimoch: Testing the result of `epicsExportTestIoc_registerRecordDeviceDriver()` now. It has always returned 0, even when the test fails.

dirk-zimoch:
Finished, I hope.

modules/libcom/src/misc/epicsExport.h
```diff
@@ -94,11 +94,15 @@ typedef void (*REGISTRAR)(void);
  \endcode
  *
  * \param fun Registrar function's name.
- *
- * \note C++ code needs to wrap with @code extern "C" { } @endcode
  */
+#ifdef __cplusplus
+#define epicsExportRegistrar(fun) \
+    extern "C" { extern epicsShareFunc REGISTRAR EPICS_EXPORT_PFUNC(fun); } \
+    REGISTRAR EPICS_EXPORT_PFUNC(fun) = reinterpret_cast<REGISTRAR>(&fun)
```
mdavidsaver, line 109: Thinking about #211, I ask myself: will this initializer become a global constructor?

I think my answer is "no", at least not on any currently supported architecture.

Quoting [cppreference.com](https://en.cppreference.com/w/cpp/language/reinterpret_cast)

> Unlike static_cast, but like const_cast, the reinterpret_cast expression does not compile to any CPU instructions (except when converting between integers and pointers, or between pointers on obscure architectures where pointer representation depends on its type). It is primarily a compile-time directive which instructs the compiler to treat expression as if it had the type target-type. 

Which is more ambiguous than I expected.  Since this is a function pointer cast, I think that "obscure architectures" would in practice only be an issue for someone attempting eg. a WebAssembly port.  There the C cast would have the same issue.

modules/libcom/src/misc/epicsExport.h
```diff
@@ -35,10 +35,6 @@
 #define epicsExportSharedSymbols
 #include <shareLib.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
```
mdavidsaver: I also found myself wondering if the enclosing `extern "C"` could effect the `REGISTRAR` typedef.  Surprisingly, the answer is "yes".  However, only for [SunCC](https://docs.oracle.com/cd/E19059-01/wrkshp50/805-4956/bajdcjch/index.html).

While Base no longer supports that compiler, I think it would be better to avoid an unnecessary change.

("unnecessary" because it does not effect the macro definitions)

mdavidsaver (changes requested):
I think this PR is almost ready.

modules/libcom/src/misc/epicsExport.h
mdavidsaver: ... so please re-add the `extern "C"` block.

dirk-zimoch: It went in the way of my template functions. But they are gone now anyway.
I can re-add it. With a smaller scope.

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: Maybe that was the problem with my "cast away const magic" template. Even though it was a static inline function containing only cost_cast, it probably was treated as an initializer function by some compilers. Maybe constexpr will help. I will try that later....

dirk-zimoch:
Interesting. Better but not good yet.

dirk-zimoch:
It seems there is no solution "between" C-style cast and a C++11 constexpr casing function. I am going back to C-style cast then. We can change this in 7.1 with C++11.

modules/libcom/src/misc/epicsExport.h
dirk-zimoch: Done.

dirk-zimoch:
Can someone with access to appveyor please restart the timed out build?

mdavidsaver:
> Can someone with access to appveyor please restart the timed out build?

Now restarted.

anjohnson:
@dirk-zimoch to rebase without the merge commit and we'll merge properly then, don't wait for Appveyor.

Issue #538, opened 2024-08-23
Missing function attribute EPICS_NORETURN


dirk-zimoch:
The function `testAbort()` does not return. It should be decorated accordingly. Therefore, a macro `EPICS_NORETURN` or similar needs to be defined in compilerSpecific.h for each compiler type. The gcc version would be `__attribute__((noreturn))` but I don't know what other compilers use.

dirk-zimoch:
The MSVC version seems to be `__declspec(noreturn)`.

mdavidsaver:
`EPICS_NORETURN` seems a reasonable a addition to `compilerDependencies.h` / `compilerSpecific.h`.

Another candidate application would be to `cantProceed()`.

#155

dirk-zimoch:
I have let gcc find out which functions don't return using `-Wsuggest-attribute=noreturn`. It found:
- `acctst()`
- `flexfatal()`
- `lerrif()`
- `lerrsf()`
- `cantProceed()`
- `epicsExit()`
- `testAbort()`
- `terminal_start()`
- `cast_server()`
- `rsrv_online_notify_task()`

The yacc and flex files already use a macro `NORETURN` defined for gcc as `__attribute__((noreturn))` for some functions that do not return but not for all (ending by calling other `NORETURN` functions).

To cover Windows complicates the task slightly because `__declspec(noreturn)` must be specified *before* the function and `__attribute__((noreturn))` *after* the function. Thus I ended up for gcc and clang with
```
#define EPICS_NORETURN(f) f __attribute__((noreturn))
```
and for MSVC with
```
#define EPICS_NORETURN(f) __declspec(noreturn) f
```
decorating the functions like this:
```
LIBCOM_API void EPICS_NORETURN(epicsExit(int status));
```
or this:
```
LIBCOM_API EPICS_NORETURN(void epicsExit(int status));
```
Both works.

dirk-zimoch:
See PR #540.

ericonr:
I believe this issue can be closed, right?

PR #603, opened 2025-02-19
detect and handle backslash as part of multi-line string

### LINKED ISSUE #602: multiline strings in iocsh files
It would be a good feature to have support for multiline strings in st.cmd / iocsh scripts.

Currently I noticed that when iocsh reads a script it does not like the line to be terminated with `\` (backslash). If it were able to concatenate the lines that end with backslash it would help the readability of very long lines in startup files.
Then this would be possible:

```
dbLoadRecords "db/dbSubExample.db",\
      "user=ahoj,  \
foo=bar ,\
  unused=hahhaha,    \
more=macros,\
  andmore=....,\
var=$(VAR1)\
"
```

Would something like this be acceptable to introduce or are there known issues and reasons why not have it?

hinxx:
Initial attempt to allow handling of the multi-line strings in the iocsh files. See also #602 .


mdavidsaver:
How long of a (continued) line have you tested with?  Longer than `MAC_SIZE` (256 bytes)?

hinxx:
Yes, this is what I get.

file:
```
epicsEnvSet "LONGERTHAN256" "XX \
asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf \
1234123123412341234123412341234123412341234123412341234123412341234\
987987987987987987987987987987987987987987987987987987987987987987\
mbvmbnvmbnvmnbvmnbvmnbvmbnvmnbvbnmvmbnvnbmvmnbvmbnvmnbvmnbvmnbvmnbv\
poiupoiupoiupoiupoiupioupoiupoiupoiupoiupoiupoiupoiupoiupoiupoiupoiu\
765765765765765765765765765765765765765765765765765765765765765765765\
ghfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgf XX"

epicsEnvSet FOO "$(LONGERTHAN256)"

```

iocsh:
```
epicsEnvSet "LONGERTHAN256" "XX asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf 1234123123412341234123412341234123412341234123412341234123412341234987987987987987987987987987987987987987987987987987987987987987987mbvmbnvmbnvmnbvmnbvmnbvmbnvmnbvbnmvmbnvnbmvmnbvmbnvmnbvmnbvmnbvmnbvpoiupoiupoiupoiupoiupioupoiupoiupoiupoiupoiupoiupoiupoiupoiupoiupoiu765765765765765765765765765765765765765765765765765765765765765765765ghfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgf XX"
epicsEnvSet FOO "XX asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf 1234123123412341234123412341234123412341234123412341234123412341234987987987987987987987987987987987987987987987987987987987987987987mbvmbnvmbnvmnbvmnbvmnbvmbnvmnbvbnmvmbnvnbmvmnbvmbnvmnbv"

```

There is no errors in the iocsh, the `FOO` value is just clipped at 256 chars.


On a related note, I should probably add new tests for this feature. Would `modules/libcom/test/epicsStringTest.c` be a good place?

anjohnson:
`epicsStringTest.c` contains tests for `modules/libcom/src/misc/epicsString.c`, so isn't the right place to test your new functionality. We don't currently have tests for the gnuReadline.c code, so a new test program is probably required for them.

We'll need to be careful to not try to build & run this test program when the EPICS build is configured without readline or on targets which don't have it.

hinxx:
> We don't currently have tests for the gnuReadline.c code, so a new test program is probably required for them.

OK.
I'm thinking that the test should mimic `iocshBody()` such that creates the `ReadlineContext`. Then it repetitively calls `epicsReadline` on read lines; but nothing more than that (not expanding macros, identifying and executing commands).

> We'll need to be careful to not try to build & run this test program when the EPICS build is configured without readline or on targets which don't have it.

Would that implies that only builds with readline would be able to handle multiline strings? I was hoping that it would not be limited to readline.
For example `epicsReadline`, in case of missing `osd` does similar loop in fetching chars from input as seen in the `gnuReadline.cpp` `osdReadline` ; I guess that is in case of `EPICS_COMMANDLINE_LIBRARY_EPICS`. I'm not familiar with any other `osdReadline.c`, but I can take a closer look.

ralphlange:
Don't worry about fixing VS2010/2012. We're dropping those two. (see #606)

hinxx:
I'll be ignoring failed vs builds.

I don't know why the two Ub-20 gcc + RT fail. The error says it can not open `../multiline-input.txt`. Why does it work for all other builds? I also notice the the rest of Ub-20 gcc builds (successful) say `Action 'test-results' skipped as per configuration`.

Not sure what CodeQL means with `This 'call to strcpy' with input from  may overflow the destination.` I fail to see that would be happening due to changes I made.

The Check EditorConfig says `33:49: the final newline is missing` in `multiline-input.txt` which is intentional to show that a line a file not ending with an empty line is not parsed. Chatbot says yaml config could include an exception for this file not check for trailing new line.

The Codacy 'issues' I'm also not sure what to do about.

mdavidsaver:
> ... The error says it can not open `../multiline-input.txt`. Why does it work for all other builds?  ...

Add to `TESTFILES`.

https://github.com/epics-base/epics-base/blob/d1a94bb715e62f760681593d99431ec6e79d32b0/modules/libcom/test/Makefile#L268-L272

anjohnson:
I suspect Codacy doesn't realize that `testAbort()` never returns. If you bear that in mind, the errors seem to make more sense.

We should annotate that in the epicsUnitTest.h file, but we don't appear to have an `EPICS_NORETURN` macro but could probably use one.

mdavidsaver:
> There is no errors in the iocsh, the `FOO` value is just clipped at 256 chars.

This is what I expected to see, although I think I was wrong in my thinking about precisely why/where this happens.

If I instead do:

```
echo "epicsEnvSet LONGERTHAN256 XX \
asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf \
1234123123412341234123412341234123412341234123412341234123412341234\
987987987987987987987987987987987987987987987987987987987987987987\
mbvmbnvmbnvmnbvmnbvmnbvmbnvmnbvbnmvmbnvnbmvmnbvmbnvmnbvmnbvmnbvmnbv\
poiupoiupoiupoiupoiupioupoiupoiupoiupoiupoiupoiupoiupoiupoiupoiupoiu\
765765765765765765765765765765765765765765765765765765765765765765765\
ghfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgfhgf XX"
```

Then echo prints the whole string, including the trailing `XX`.

So the 256 character truncation happens when expanding `$(LONGERTHAN256)`.  This can also be shown by:

```
system "env | grep LONGERTHAN256"
```

or

```
echo "$(LONGERTHAN256) $(LONGERTHAN256) !!"
```

... which does print the trailing `!!`.

So the whole long line passes through `macDefExpand()` just fine.  The 256 character limit applies to individual macro expansions, not the whole input string.

Where am I going with this?  I think I have convinced myself that these multi-line strings would not impose any new limitations.

hinxx:
> We should annotate that in the epicsUnitTest.h file, but we don't appear to have an EPICS_NORETURN macro but could probably use one.

Dirk has been looking into this in #538 and #540. I'll just ignore these messages..

anjohnson:
I was about to ask you to fix the [EditorConfig check failure](https://github.com/epics-base/epics-base/actions/runs/13564804997/job/37915413368?pr=603), but on looking at it ...
```
modules/libcom/test/multiline-input.txt:
33:49: the final newline is missing
99 this one will not appear 'cause of no newlin
```
... I realized you're deliberately using a file with no newline at the end as part of a test. That could be quite brittle (i.e. easily broken by someone who doesn't understand that's by design here who adds a newline), but I guess the CI would then fail that test. However I'm concerned that this file will subsequently fail that check on every future build. Can you add an exception to the `.editorconfig` for that file, I'm hoping that would silence the CI check.

The [CodeQL failure report](https://github.com/epics-base/epics-base/pull/603/checks?check_run_id=37915772660) is actually very detailed. If you click the "Show Paths" button in the Annotations box for that result it demonstrates (in 49 steps!) how it thinks this could overflow the buffer. I haven't looked through it in detail and I doubt this is a big issue since we're dealing with user input here which shouldn't be potentially malicious. I'd be happier if someone could do that and report back here before we dismiss the alert.

Thanks for the pointer to #540, that needs a review by the Core Group soon, which might clean up some of the Codacy warnings.

hinxx:
> Can you add an exception to the .editorconfig for that file, I'm hoping that would silence the CI check.

Done; test looks good now.

> If you click the "Show Paths" button in the Annotations box for that result it demonstrates (in 49 steps!) how it thinks this could overflow the buffer.

I fail to see the button. This is how my screen looks like:
Screenshot_2025-02-28_09-49-35

mdavidsaver:
> I fail to see the button. This is how my screen looks like:

Well, that is interesting.  I see:

image

I don't recall seeing "Show paths" previously.  (could be inattention on my part...)

anjohnson:
That's probably only visible to maintainers then, for security reasons. The "Show more details" link adds an explanation from the rule that failed.

@dirk-zimoch has updated #540 and I hope to merge that soon. I'd like to see what effect that has on this PR before we decide what to do about these failures.

mdavidsaver:
I consider the specific linter complaint about `strcpy()` to be a false positive.  A Real Human Person can see that the destination buffer is sized by `strlen()` only a few lines above.

https://github.com/epics-base/epics-base/blob/13d6ca598cca495b2e559e808392ed22265c57e3/modules/libcom/src/iocsh/iocsh.cpp#L853-L858

It isn't clear to me if there is an actual issue hiding in the reported 49 step "Path".  My best guess is this line in gnuReadline.c ?

```c
            line[linelen++] = c;
```

Perhaps the linter has not picked up on the `realloc()` just above?  Or because adding `50` does not explicitly capture a relationship between `linesize` and `linelen`?

```c
            if ((linelen + 1) >= linesize) {
                char *cp;
                linesize += 50;
```

It would be asking a lot of this linter to infer the relationship implied by `line[linelen++] = c`.

Maybe changing to `linesize = linelen+50` would quiet the warning?

hinxx:
> Maybe changing to linesize = linelen+50 would quiet the warning?

Changed.. lets see what happens.

> Perhaps the linter has not picked up on the realloc() just above? Or because adding 50 does not explicitly capture a relationship between linesize and linelen?

Funny how this line of code was not touched by me. Why did linter not fail earlier?

BTW, could one run these tests locally without needing to make a PR change & push & wait for CI to fail?

mdavidsaver (approved):
> Funny how this line of code was not touched by me. Why did linter not fail earlier?

As far as I can tell, this PR is flagged because the detected code path ending in this `strcpy()` in iocsh.cpp has intermediate steps which are changed by this PR.  Looking through that code path again, it might be that the "root cause" is the linter being unable to conclude that `macExpandString()` always adds a nil terminator.  I find this understandable as I can't prove this to myself by reading through macCore.c.  Which is I think anyway irrelevant to merging this PR.

I think this PR is good enough to merge.

hinxx:
Thanks for looking into it @mdavidsaver ! I've added some more test cases in the last commit.

hinxx:
> That's probably only visible to maintainers then, for security reasons. The "Show more details" link adds an explanation from the rule that failed.


FWIW, I managed to get a hold of the these details by looking that the Security tab on my fork of epics-base. Not sure if needed to enable anything on the Github Actions  / Settings to get it.

Screenshot_2025-04-06_13-43-56

anjohnson:
Core Group: We think this is ready to merge, would you write a Release Note entry (see #628 to do that as a separate file) for it. Rebase is optional if you like.

modules/libcom/test/Makefile
```diff
@@ -339,6 +339,20 @@ TESTS += nonEpicsThreadPriorityTest
 endif
 endif
 
+# HK: improve detection of valid command line libraries
+ifeq ($(COMMANDLINE_LIBRARY),READLINE)
+TESTPROD_HOST += readlineTest
+readlineTest_SRCS += readlineTest.c
+TESTS += readlineTest
+TESTFILES += ../multiline-input.txt ../multiline-expect.txt
+endif
```
anjohnson: Core Group: Instead of duplicating these `ifeq()` blocks, please use this Makefile conditional which uses a list of acceptable names:

```Makefile
ifneq ($(filter READLINE EPICS,$(COMMANDLINE_LIBRARY)),)
TESTPROD_HOST += readlineTest
...
endif
```

Some older linux OSs needed additional choices for `COMMANDLINE_LIBRARY` to adjust the OS libraries and compiler flags, using the names `READLINE_CURSES` and `READLINE_NCURSES`. Our Cygwin support also used yet another name, but we don't support Cygwin any more. However our more recent automatic readline discovery process doesn't support these so it's likely they aren't required any more, and I wouldn't bother adding them here (any user who manually configures those settings just doesn't run this self-test).

hinxx: Done.

anjohnson (approved):
Thanks, this looks good to me now.

Issue #428, opened 2023-10-03
Undocumented OUTLINK behavior change in IOCs


daykin:
The Application Developers' guide has always stated:
>Only CA is appropriate for an output link. The write to a field over channel access causes processing as specified in Channel Access Puts to Passive Scanned Records.

In older versions of base (e.g. 3.15.9), a developer who would like their output link to process the destination record, but inelegantly creates an output link with CPP instead of CA, is still somewhat bailed out, as "PP" matches. 
https://github.com/epics-base/epics-base/blob/R3.15.9/src/ioc/dbStatic/dbStaticLib.c#L2177

But, in versions with the newer dbLinkInfo struct (since commit https://github.com/epics-base/epics-base/commit/200355bc0030267431161018547180c292378fff on 7.0 branch), links are filtered according to the logic in dbParseLink, which silently negates CPP to NPP. 

I am wondering if it might be a good idea to add some verbiage in the documentation and/or a runtime warning.

mdavidsaver:
> I am wondering if it might be a good idea to add some verbiage in the documentation and/or a runtime warning.

This sounds like a good project.  Will you take it on?

ralphlange:
Well spotted! Make sure to mention that this applies only to local target links,

If the target of the OUT link is remote, it will always use CA, the `PP` property of the target field decides the processing policy and the PP/NPP link property of the OUT link is not relevant.

daykin:
I am happy to put something together. Currently on the Phoebus track at codeathon, so maybe not this week.

PR #608, opened 2025-02-21
Warn to stderr when discarding CPP modifier for outlink

daykin:


daykin:
Addresses #428. Is this excessively wordy?

ralphlange:
Should also apply to CP flagged output links, shouldn't it?
https://docs.epics-controls.org/en/latest/process-database/EPICS_Process_Database_Concepts.html#channel-access-links

daykin:
> Should also apply to CP flagged output links, shouldn't it? https://docs.epics-controls.org/en/latest/process-database/EPICS_Process_Database_Concepts.html#channel-access-links

Oh, duh, yes. I should add that...

daykin:
> > Should also apply to CP flagged output links, shouldn't it? https://docs.epics-controls.org/en/latest/process-database/EPICS_Process_Database_Concepts.html#channel-access-links
> 
> Oh, duh, yes. I should add that...

Although, I think in the case of CP the behavior doesn't change between versions. It would still fall through in 3.15.9, since `strstr('CPP','PP')` hits but `strstr('CP','PP')` does not. In any case, it might nonetheless be a good idea to warn for both, as it's not immediately obvious what happened to the flags.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -2380,7 +2380,17 @@ long dbParseLink(const char *str, short ftype, dbLinkInfo *pinfo)
         /* filter modifiers based on link type */
         switch(ftype) {
         case DBF_INLINK: /* accept all */ break;
-        case DBF_OUTLINK: pinfo->modifiers &= ~pvlOptCPP; break;
+        case DBF_OUTLINK:
+            if(pinfo->modifiers & (pvlOptCPP|pvlOptCP)){
+                fprintf(stderr, "Warning: dbParseLink: Output link to %s was specified with \n\
```
mdavidsaver: I think this should be switched to `errlogPrintf()`.  Generally we avoid synchronous logging when it can happen after `iocInit()`.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -2380,7 +2380,17 @@ long dbParseLink(const char *str, short ftype, dbLinkInfo *pinfo)
...
+                CP/CPP, which has no effect. \n\
+                Only CA affects local output links. \n\
+                Remote output link behavior is governed by the target record.\n",
+                pinfo->target);
+            }
+            pinfo->modifiers &= ~pvlOptCPP;
+            pinfo->modifiers &= ~pvlOptCP;
```
mdavidsaver:
```c
pinfo->modifiers &= ~(pvlOptCP|pvlOptCPP);
```

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -2380,7 +2380,16 @@ long dbParseLink(const char *str, short ftype, dbLinkInfo *pinfo)
         /* filter modifiers based on link type */
         switch(ftype) {
         case DBF_INLINK: /* accept all */ break;
-        case DBF_OUTLINK: pinfo->modifiers &= ~pvlOptCPP; break;
+        case DBF_OUTLINK:
+            if(pinfo->modifiers & (pvlOptCPP|pvlOptCP)){
+                errlogPrintf("Warning: dbParseLink: Output link to %s was specified with \n\
+                CP/CPP, which has no effect. \n\
+                Only CA affects local output links. \n\
+                Remote output link behavior is governed by the target record.\n",
```
mdavidsaver: @daykin Have you checked to see what this message looks like when printed?

fyi. indentation can be more clearly expressed with a multiline string literal using the implied [concatenation](https://en.cppreference.com/w/cpp/language/string_literal#Concatenation) behavior.

```c
errlogPrintf("Warning: dbParseLink: Output link to %s was specified with \n"
             "         CP/CPP, which has no effect.\n"
...
);
```

The message itself seems to me a bit verbose when spread over 4 lines.

Also, I think this message is missing the most important piece of information, the source link record and field.  This is known at each of the non-test call sites.  (annoyingly in different forms: `dbAddr*`, `DBENTRY*`, `dbCommon*`/`dbFldDes*`)

`dbParseLink()` is not public API, so its signature can be changed.

daykin: > dbParseLink() is not public API, so its signature can be changed.

That is a good thought, it is probably best to add that info to the signature, as far as this issue is concerned.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -2380,7 +2380,13 @@ long dbParseLink(const char *str, short ftype, dbLinkInfo *pinfo)
         /* filter modifiers based on link type */
         switch(ftype) {
         case DBF_INLINK: /* accept all */ break;
-        case DBF_OUTLINK: pinfo->modifiers &= ~pvlOptCPP; break;
+        case DBF_OUTLINK:
+            if(pinfo->modifiers & (pvlOptCPP|pvlOptCP)){
+                errlogPrintf("Warning: Discarding CP/CPP modifier in CA output link from %s to %s.\n",
+                recname, pinfo->target);
```
mdavidsaver: Please also include the source link field name.  Also, in my example I forgot `ERL_WARNING`.   eg.

```c
errlogPrintf(ERL_WARNING ": Discarding CP/CPP modifier in CA output link from %s.%s to %s.\n",
```

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -1095,7 +1095,7 @@ static long dbPutFieldLink(DBADDR *paddr,
         return S_db_badDbrtype;
     }
 
-    status = dbParseLink(pstring, pfldDes->field_type, &link_info);
+    status = dbParseLink(pstring, pfldDes->field_type, &link_info, precord->name);
```
mdavidsaver: In this context `pfldDes->name` has the field name.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -2618,7 +2624,7 @@ long dbPutString(DBENTRY *pdbentry,const char *pstring)
             dbLinkInfo link_info;
             DBLINK *plink = (DBLINK *)pfield;
 
-            status = dbParseLink(pstring, pflddes->field_type, &link_info);
+            status = dbParseLink(pstring, pflddes->field_type, &link_info, dbGetRecordName(pdbentry));
```
mdavidsaver: In this context, call `dbGetFieldName(pdbentry)`.

mdavidsaver (changes requested):
One last minor change.  Please add the source link field name to the message.

daykin:
@mdavidsaver I saw another user on tech-talk ran into this (https://epics.anl.gov/tech-talk/2025/msg00895.php). Any remaining tasks you'd like to see before merging?

modules/database/test/ioc/db/dbPutLinkTest.c
```diff
@@ -121,6 +121,16 @@ static void testLinkParse(void)
...
 
+    info.modifiers |= pvlOptCPP;
+    dbParseLink("something CPP", DBF_OUTLINK, &info, "dummy","OUT");
+    testOk(info.modifiers == 0, "CPP modifier was discarded");
+    dbFreeLinkInfo(&info);
+
+    info.modifiers |= pvlOptCP;
+    dbParseLink("something CP", DBF_OUTLINK, &info, "dummy","OUT");
+    testOk(info.modifiers == 0, "CP modifier was discarded");
+    dbFreeLinkInfo(&info);
```
mdavidsaver, line 132: While not critical, wrapping with `eltc()` will avoid printing the warning.  eg.

```c
eltc(0);
dbParseLink(...); // would make noise
eltc(1);
```

mdavidsaver (approved):
Looks ok to me.

PR #610, opened 2025-02-25
fix for archs where sizeof(void*)>sizeof(long), i.e. windows-x64

dirk-zimoch:


mdavidsaver:
Why not stored the result of `strtoull()` in an `unsigned long long`, as is indicated by the function name?

dirk-zimoch:
Because that triggers warnings when casting the `unsigned long long` to a shorter `epicsThreadId` (pointer) on 32 bit architectures. Casting the `unsigned long long` (64 bit integer) result of `stroull()` to the shorter `uintptr_t` (32 bit integer) does not trigger such warnings and casting `uintptr_t` to a pointer like `epicsThreadId` should never trigger a warning. I have not tried on 32 bit Windows, though. Maybe the 64 to 32 bit integer cast triggers warnings there. Should we select `stroull()` or `stroul()` depending on pointer size?

mdavidsaver:
> we select `stroull()` or `stroul()` depending on pointer size?

Maybe prefer the `epicsParseUInt32/64()` functions from `epicsStdlib.h`?

dirk-zimoch:
> Maybe prefer the `epicsParseUInt32/64()` functions

Rather not. These do not return their result but expect a pointer to store the result, which is of different type for the two functions (of course), none of them `uintptr_t`, which then requires even more `#ifdef` and  casting craziness.

dirk-zimoch:
I am now avoiding the temporary integer variable altogether.

mdavidsaver:
> Rather not. These do not return their result but expect a pointer to store the result ...

Right, because they properly detect, and clearly communicate, a parse failure.

I accept that this PR would not make this situation any worse.  Free form entry of pointer addresses is already playing with fire.

anjohnson:
Suggestion: We could change the epicsThread iocsh commands to not accept thread IDs any more, making this issue moot. IIRC these are really only needed on VxWorks, at APS we normally use the target shell anyway so we wouldn't care. This knocks out 2 problems at once.

dirk-zimoch:
Sometimes different threads have the same name. How to handle those without using the numeric tid? But I agree that the code needs to become more robust. Entering an invalid tid must never crash but only print an error, like entering an invalid name. That means the called functions must never simply dereference the epicsThreadId (pointer) but compare it against the list of valid ids. Maybe a function epicsThreadIdValid() could help to check in the but that still would be a race condition. Better do it in the called functions, even if that decreases performance. A „safe“ wrapper to be used by the iocsh functions may help. Anyway, that is beyond the scope of this PR.

dirk-zimoch:
The scope of this PR was to allow to enter 64 bit tids when pointers have  64 bits but long ints have only 32. That‘s all.

dirk-zimoch:
Now always using `strtoull()`, even if pointers and `long` have the same number of bit and/or `long long` has more bits than a pointer.

mdavidsaver (approved):
Looks fine to me.  And Appveyor is progressing.

anjohnson:
Core Group: Approved.

PR #611, opened 2025-02-26
support CROSS_COMPILER_RUNTESTS_ARCHS other than RTEMS

dirk-zimoch:
Several Makefiles (not all) have added tests only for RTEMS if `CROSS_COMPILER_RUNTEST_ARCHS` matched the current `T_A`. The tests must be added independent of the OS for other than RTEMS RUNTEST_ARCHS to work.

### DIFF
--- a/modules/database/test/ioc/db/Makefile
+++ b/modules/database/test/ioc/db/Makefile
@@ -208,8 +208,8 @@ TESTSPEC_RTEMS = dbTestHarness.boot; epicsRunDbTests
 
 TESTSCRIPTS_HOST += $(TESTS:%=%.t)
 ifneq ($(filter $(T_A),$(CROSS_COMPILER_RUNTEST_ARCHS)),)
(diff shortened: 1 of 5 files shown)


mdavidsaver (approved):
Based on testing of #585 I think this change is necessary, although apparently not sufficient to run cross linux-arm tests with GHA/ci-scripts.

anjohnson (approved):
Can you change the assignments `TESTPROD =` into appending `TESTPROD +=` — that won't change these particular cases, but if tests ever gets added that _are_ specific to RTEMS, assigning to `TESTPROD` will drop them (see configure/CONFIG_ADDONS).

Other than that this looks right.

dirk-zimoch:
> Can you change the assignments `TESTPROD =` into appending `TESTPROD +=`

Done, as well in modules/libcom/test/Makefile.

anjohnson:
As @mdavidsaver said, these changes are necessary but not sufficient, we'll need another PR for it to work on GHA.

PR #613, opened 2025-03-04
Add `fields` argument to dbgrep()

FreddieAkeroyd:
Allows optionally printing out the value of a field for each record found e.g.  `dbgrep("*PRESSURE*","VAL")`

### DIFF
--- a/modules/database/src/ioc/db/dbIocRegister.c
+++ b/modules/database/src/ioc/db/dbIocRegister.c
@@ -235,14 +235,16 @@ static void dblaCallFunc(const iocshArgBuf *args) { iocshSetError(dbla(args[0].s
 
 /* dbgrep */
 static const iocshArg dbgrepArg0 = { "pattern",iocshArgStringRecord};
-static const iocshArg * const dbgrepArgs[1] = {&dbgrepArg0};
-static const iocshFuncDef dbgrepFuncDef = {"dbgrep",1,dbgrepArgs,
-                                           "List record names matching pattern.\n"
+static const iocshArg dbgrepArg1 = { "field",iocshArgStringRecord};
+static const iocshArg * const dbgrepArgs[2] = {&dbgrepArg0,&dbgrepArg1};
+static const iocshFuncDef dbgrepFuncDef = {"dbgrep",2,dbgrepArgs,
+                                           "List record names matching pattern and optionally print a field value. \n"
                                            "The pattern can contain any characters that are legal in record names as well as:\n"
                                            " - \"?\", which matches 0 or one characters.\n"
                                            " - \"*\", which matches 0 or more characters.\n\n"
-                                           "Example: dbgrep(\"*gpibAi*\")\n"};
-static void dbgrepCallFunc(const iocshArgBuf *args) { iocshSetError(dbgrep(args[0].sval));}
+                                           "Example: dbgrep(\"*gpibAi*\")\n"
+                                           "         dbgrep(\"*gpibAi*\",\"VAL\")\n"};
+static void dbgrepCallFunc(const iocshArgBuf *args) { iocshSetError(dbgrep(args[0].sval,args[1].sval));}
 
 /* dbgf */
 static const iocshArg dbgfArg0 = { "record name",iocshArgStringRecord};
--- a/modules/database/src/ioc/db/dbIocRegister.c
+++ b/modules/database/src/ioc/db/dbIocRegister.c
@@ -235,15 +235,15 @@ static void dblaCallFunc(const iocshArgBuf *args) { iocshSetError(dbla(args[0].s
 
 /* dbgrep */
 static const iocshArg dbgrepArg0 = { "pattern",iocshArgStringRecord};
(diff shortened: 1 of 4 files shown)


modules/database/src/ioc/db/dbIocRegister.c
```diff
@@ -235,14 +235,16 @@ static void dblaCallFunc(const iocshArgBuf *args) { iocshSetError(dbla(args[0].s
 
 /* dbgrep */
 static const iocshArg dbgrepArg0 = { "pattern",iocshArgStringRecord};
-static const iocshArg * const dbgrepArgs[1] = {&dbgrepArg0};
-static const iocshFuncDef dbgrepFuncDef = {"dbgrep",1,dbgrepArgs,
-                                           "List record names matching pattern.\n"
+static const iocshArg dbgrepArg1 = { "field",iocshArgStringRecord};
```
anjohnson: I don't think this argument should be `iocshArgStringRecord` since that tells the iocsh completion code to try to complete it as a record name. It should just be `iocshArgString`.

modules/database/src/ioc/db/dbTest.c
```diff
@@ -305,8 +306,25 @@ long dbgrep(const char *pmask)
...
+                if (field == NULL) {
+                    puts(pname);
+                }
+                else {
+                    const char *pvalue = "<field not present>";
+                    status = dbFindField(pdbentry, field);
+                    if (status) {
+                        if (!strcmp(field, "recordType")) {
+                            pvalue = dbGetRecordTypeName(pdbentry);
+                        }
```
anjohnson: I don't believe this is necessary, try requesting a field named `RTYP`.

modules/database/src/ioc/db/dbTest.c
```diff
@@ -305,8 +306,25 @@ long dbgrep(const char *pmask)
...
-            if (epicsStrGlobMatch(pname, pmask))
-                puts(pname);
+            if (epicsStrGlobMatch(pname, pmask)) {
+                if (field == NULL) {
+                    puts(pname);
+                }
+                else {
+                    const char *pvalue = "<field not present>";
+                    status = dbFindField(pdbentry, field);
+                    if (status) {
```
anjohnson: Could you compare this with the code for `dbl()` which accepts a space-separated list of field names and prints them all — it would be nice if this accepted the same kind of list. The main difference is that `dbl` takes `*` or a record type name in the first argument, instead of the record name pattern.

anjohnson (changes requested):
I like this idea; some comments inline below on changes to keep uniformity with the `dbl` command.

Thanks!

modules/database/src/ioc/db/dbIocRegister.c
FreddieAkeroyd: Ah yes, careless cut and paste on my part - thanks

modules/database/src/ioc/db/dbTest.c
FreddieAkeroyd: I took this from the `dbl` code that seems to allow both `RTYP` and `recordType`, is the `recordType`  just historical and pre-dates `RTYP` and so for any new code like this there is no need/expectation of `recordType` working?

modules/database/src/ioc/db/dbTest.c
FreddieAkeroyd: Sure - will add that feature

modules/database/src/ioc/db/dbTest.c
anjohnson: Oops, well I have no memory of that code snippet in `dbl`, despite working on it about 17 years ago. It almost certainly predates the addition of the `RTYP` record attribute, which was a more general solution to record-type identification that works remotely. The CA library also had a special request type for the record type-name of a channel, which nobody uses now. I guess this does no harm, but unless it's documented for `dbl` I'd probably consider removing it from there sometime.

FreddieAkeroyd:
I've noticed one slight oddity in behaviour that was present too in the original `dbl` - if you have leading and trailing space characters e.g. `" DESC "` you get extra `VAL` items printed. Is this OK or did you want me to refactor to avoid this, maybe changing to use `epicsStrtok_r`

anjohnson:
Hmm, what fun:
```
epics> dbl * ""
anj:BaseVersion
anj:exit
epics> dbl * " "
anj:BaseVersion, "7.0.9.1-DEV", "7.0.9.1-DEV"
anj:exit, "0", "0"
```
That would be a (minor) bug in the command's argument parser. I could see printing VAL once in that case as understandable and not problematic, but twice? I feel an Oscar Wilde quote coming on:

> To lose one parent may be regarded as a misfortune; to lose both looks like carelessness.

If you're willing to fix that as you suggest I'd be happy, thanks!
In which case I won't re-run the failed Appveyor build right now.

FreddieAkeroyd:
yes i thought two was one too many :-) Hopefully now all fixed using `epicsStrtok_r`

PR #606, opened 2025-02-21
ci: drop AppVeyor builds on VS2010/VS2012

ralphlange:
SSIA

PR #615, opened 2025-03-12
Don't use __attribute__((noreturn)) on VxWorks

dirk-zimoch:
VxWorks does not mark `abort()` or `exit()` as noreturn. Thus, functions declared noreturn which end in a call to those functions cause a compiler warning on vxWorks.

### DIFF
--- a/modules/libcom/src/osi/compiler/gcc/compilerSpecific.h
+++ b/modules/libcom/src/osi/compiler/gcc/compilerSpecific.h
@@ -60,6 +60,9 @@
 /*
  * No return marker
  */
+#ifndef vxWorks
+// VxWorks does not mark abort() or exit() noreturn!
 #define EPICS_NORETURN __attribute__((noreturn))
+#endif
 
 #endif  /* ifndef compilerSpecific_h */


modules/libcom/src/osi/compiler/gcc/compilerSpecific.h
```diff
@@ -60,6 +60,9 @@
 /*
  * No return marker
  */
+#ifndef vxWorks
+// VxWorks does not mark abort() or exit() noreturn!
```
mdavidsaver, line 64: With #606 merged, we can allow single line comments in C code now?

ralphlange: You mean C++ comments?

anjohnson: The MSVC-2010 compiler was the last one complaining about those. I just double-checked the VxWorks 6.9 compiler, it has no problem with "C++-style" or "single-line" `//` comments in a .c file.

dirk-zimoch: I will change it to a C style comment.

anjohnson: Why? That's no longer necessary, we have dropped support for those old compilers now.

dirk-zimoch: I don't actually like to find C++ style comments in C code. But I did not notice immediately because no compiler complains any more. Even Codacy is fine with that. Alas, where have the good old times gone?

mdavidsaver: > You mean C++ comments?

Technically these have been standard since C99.


> ... Alas, where have the good old times gone?

`gcc -Werror=kids-these-days ...` ;)

PR #563, opened 2024-11-11
Enhance Access Security with METHOD and AUTHORITY support, and add YAML parsing

george-mcintyre:
note: Part of EPICS Secure PVAccess work

Primary changes:
1. **Support for METHOD and AUTHORITY:**
   - Added new fields `METHOD("type"[,...])` and `AUTHORITY("common_name"[,...])` to access security rules for PV authorization based on connection method and certificate authority.
   - Introduced new API functions `asAddClientX`, `asChangeClientX`, and `asTrapWriteBeforeWithDataX` in the `libcom` library.
   - Added a new access type `RPC` to control PV access via RPC (currently operates as `WRITE`).
   - Updated ASG (Access Security Group) processing to incorporate these new fields.

Additional update:
2. **YAML Support:**
   - Added lexer and parser rules for handling YAML-formatted access security configurations.
   - Introduced new functions to manage YAML parsing in ACF files.
   - Introduce new YAML schema file for validation of new YAML format in supported editors

These changes provide fine-grained control over access security based on connection methods and certificate authorities, while also introducing the flexibility of YAML for configuring access security settings.

PR #618, opened 2025-03-17
Add TLS and certificate-based access security features

george-mcintyre:
### Add TLS and certificate-based access security features

Extended access security to support new fields (`METHOD`, `AUTHORITY`) and options (`ISTLS`) for certificate-based connection authorization. Introduced a new `RPC` access type and updated related APIs (`asAddClientX`, `asTrapWriteWithDataX`) for enhanced control. Adjusted tests and documentation to accommodate new functionalities and ensure backward compatibility.


george-mcintyre:
Replaces #653

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -791,9 +791,11 @@ static int write_action ( caHdrLargeArray *mp,
         return RSRV_ERROR;
     }
 
-    asWritePvt = asTrapWriteWithData ( pciu->asClientPVT,
+    asWritePvt = asTrapWriteWithDataX ( pciu->asClientPVT,
```
mdavidsaver: CA does not, and never will, have a concept of RPC call.  and imo. never will support TLS transport.  So I think the changes under rsrv/ should be removed as proof of API backwards compatibility.

modules/libcom/src/as/asLibRoutines.c
```diff
@@ -63,22 +63,27 @@ static ASG *asAsgAdd(const char *asgName);
...
+ *  - The version pointed to by pasbase is kept as is but locked against changes
+ *  - A new version is created and pointed to by pasbasenew
+ *  - If anything goes wrong. The original version is kept. This results is some
+ *    wasted space but at least things still work.
+ *  - If the new access security configuration is successfully read then:
+ *    * the old memberList is moved from old to new.
+ *    * the old structures are freed.
+ *
+ * @param arg Argument (not used)
+ */
```
mdavidsaver, line 86: Public API documentation comments should appear in the header, otherwise they will be ignored by doxygen.

modules/libcom/src/as/asLibRoutines.c
```diff
@@ -361,8 +366,52 @@ void epicsStdCall asPutMemberPvt(ASMEMBERPVT asMemberPvt,void *userPvt)
...
+ * @param asMemberPvt Pointer to the member structure
+ * @param asl Access level
+ * @param user User name
+ * @param host Host name
+ * @return Status code
+ *          `S_asLib_asNotActive` if access security is not active,
+ *          `S_asLib_badMember` if the member is not provided,
+ *          `S_asLib_noMemory` if there is not enough memory to allocate the client structure,
+ *          or the status from `asComputePvt` which will be 0 if the client is successfully added
+ */
```
mdavidsaver: Rather than an almost copy paste for `asAddClient()` and `asAddClientX()`, I would prefer to document the former as being a short-hand for `asAddClientX(pasClientPvt, asMemberPvt, asl, user, "ca", NULL, host, 0)`.

modules/libcom/src/as/asLib.h
```diff
@@ -92,6 +99,13 @@ LIBCOM_API long epicsStdCall asAddClient(
 /*client must provide permanent storage for user and host*/
 LIBCOM_API long epicsStdCall asChangeClient(
     ASCLIENTPVT asClientPvt,int asl,const char *user,char *host);
+/*client must provide permanent storage for user, host, method, and authority */
+LIBCOM_API long epicsStdCall asAddClientX(
+    ASCLIENTPVT *asClientPvt,ASMEMBERPVT asMemberPvt,
+    int asl,const char *user,char *method,char *authority,char *host, int isTLS);
+/*client must provide permanent storage for user, host, method, and authority */
+LIBCOM_API long epicsStdCall asChangeClientX(
+    ASCLIENTPVT asClientPvt,int asl,const char *user,char *method,char *authority, char *host, int isTLS);
```
mdavidsaver, line 169: I'm concerned about adding additional arguments to already long argument lists.  Especially when the new arguments are not simply appended.  Is it time to switch to passing in the arguments as a structure?

modules/libcom/test/aslibtest.c
```diff
@@ -111,6 +166,220 @@ static void testHostNames(void)
...
+static void testRulesDumpOutput(void)
+{
+    testDiag("testRulesDumpOutput()");
+    testOk1(asInitMem(method_auth_config, NULL)==0);
+
+    char *buf = NULL;
+    size_t size = 0;
+
+    // DEFAULT
+    FILE *fp = open_memstream(&buf, &size);
```
mdavidsaver: As @anjohnson [mentioned previously](https://github.com/epics-base/epics-base/pull/563#discussion_r1846982443), `open_memstream` is not portable to all supported targets.

It is also unnecessary.  This test is allowed to write files to `$PWD`.

mdavidsaver (changes requested):
Some specific comments, including some from #563.  I also intend to write up a more conceptual discussion of what an extended credential model can/should be.

modules/database/src/ioc/rsrv/camessage.c
george-mcintyre: Agreed: Addressed in commit 56fdf388

modules/libcom/src/as/asLibRoutines.c
george-mcintyre: Resolved by 24c78119

modules/libcom/test/aslibtest.c
george-mcintyre: Resolved with #ce377d73

modules/libcom/src/as/asLibRoutines.c
george-mcintyre: Addressed in f6d4c08e

modules/libcom/src/as/asLib.h
george-mcintyre: I agree that we should change the signature but I hope we can address the format of ACF files to make them forwards compatible (future proof/extensible).  e.g. allowing them to syntactically accept arbitrary entries for `RULES` section but having the parser discard entries that it does not implement.  This would allow us to introduce changes like this one without breaking legacy clients.    At that time I think it would be good to change the signature because in any case it would be a breaking change for legacy clients (so why not do it at the same time)

george-mcintyre:
Replaced with #641

Issue #426, opened 2023-09-23
Problem with CA nameserver and CA_V413 protocol


mdavidsaver:
https://epics.anl.gov/tech-talk/2023/msg01394.php

> I recently had to diagnose a problem with our local CA nameserver. We use a fork of the EPICS CA nameserver, which exhibits the same problem.
> 
> The symptom was that certain clients would receive monitor updates with element counts of zero, even for scalar PVs, when the PVs were resolved through a new build of the CA nameserver. This only occurred with clients built against R3.14.12 or newer, a nameserver built against R3.14.12 or newer, and an ioc built against R3.14.11 or older. (Note that I use the term "ioc" throughout to refer to any CA server in order to avoid confusion over which server is being described.) We identified camonitor and PyEpics as two clients that were affected. Notably caget was not.
> 
> I traced the problem to a change made in CA_V413, that allows  clients to use a zero element count in a CA_PROTO_EVENT_ADD or CA_PROTO_READ_NOTIFY messages. Prior to 413 an element count of zero in a request is supposed to always trigger an error. 413 allows a request for zero elements to which the server responds with the actual number of elements available. Using the element count of zero seems to have become the default when a client is communicating with an ioc that supports V413. Everything works fine when the client and ioc handle name resolution directly. The channel falls back to the lowest common denominator of CA minor protocol version.
> 
> The problem occurs when a CA nameserver built against a newer EPICS redirects newer clients to older iocs. The nameserver always answers the client with a message that correctly gives the ioc's address, but includes the nameserver's own protocol version. When the client then opens a channel with the ioc, the client has the false belief that the ioc has a later protocol version that it really does. This never seemed to create a problem for us until CA_V413, but when a 413 client connects to a <= 412 ioc it uses the zero element count which the ioc handles by returning updates with zero elements. (This itself seems like another bug in that for V412 and earlier an element count of zero is supposed to generate an error.)
> 
> My fix was to create patches to libca and libcas and modify the nameserver to use them. The libca patch adds a ca_host_minor_protocol(chid) function to the CA client API that allows a client to request the minor protocol version of a connected server. The client side of the CA nameserver uses it to fetch and store the minor protocol version from each ioc. The libcas patch adds an overload of pvExistReturn() that allows the server to respond to a client with both a specified network address and a specified minor protocol version number. The server side of the CA nameserver uses this to respond to clients with both the stored address and the stored protocol version. With these changes in place the problem disappears.
> 
[...]

PR #621, opened 2025-03-25
Forcing CA protocol version for clients connecting through the nameserver

klemenv:
This is a proposed patch to address issue #426 solely on the CA client side, no other changes to the name server are required. It works by letting user specify the oldest IOC/CA version used on the network. When not enabled (default), it has no impact on the existing code. When enabled at compile time, it tries to detect if UDP response is coming from the nameserver, in which case it ignores the CA version from the UDP response message and instead uses the compile-time value. When connecting to CA server 13+, the server properly advertises its version when TCP channel is established and overrides previous version from UDP. In all other cases it uses the compile-time value - frankly, this may introduce other version-compatibility problems if IOC is of different version. Luckily there has been few CA protocol changes, ie the oldest IOC at SNS is 3.14.8 which is CA version 11.

### DIFF
--- a/modules/ca/src/client/udpiiu.cpp
+++ b/modules/ca/src/client/udpiiu.cpp
@@ -725,6 +725,33 @@ bool udpiiu :: searchRespAction (
         serverAddr.ia.sin_addr = addr.ia.sin_addr;
     }
 
+#ifdef MIN_IOC_CA_VER_WHEN_REDIRECT
+    // Try to detect the nameserver redirecting us to a different IP
+    if (memcmp(&addr.ia.sin_addr, &serverAddr.ia.sin_addr, sizeof(struct in_addr)) != 0) {
+
+        // Additional check: nameserver should be explicitly defined in the search list
+        tsDLIter < SearchDest > iter ( _searchDestList.firstIter () );
+        while ( iter.valid () )
+        {
+            osiSockAddr tmpAddr;
+            iter->getAddr(tmpAddr);
+            if (memcmp(&addr.ia.sin_addr, &tmpAddr.ia.sin_addr, sizeof(struct in_addr)) != 0 &&
+                addr.ia.sin_port == tmpAddr.ia.sin_port) {
+
+                char from[64];
+                char to[64];
+                ipAddrToDottedIP(&addr.ia, from, sizeof(from));
+                ipAddrToDottedIP(&serverAddr.ia, to, sizeof(to));
+                printf("Redirected to %s, nameserver %s version %u, min IOC version %u\n", to, from, minorVersion, MIN_IOC_CA_VER_WHEN_REDIRECT);
+
+                minorVersion = MIN_IOC_CA_VER_WHEN_REDIRECT;
+                break;
+            }
+            iter++;
+        }
+    }
+#endif
+
     if ( CA_V42 ( minorVersion ) ) {
        cacRef.transferChanToVirtCircuit
             ( msg.m_available, msg.m_cid, 0xffff,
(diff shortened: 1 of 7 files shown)


modules/ca/src/client/udpiiu.cpp
```diff
@@ -725,6 +725,33 @@ bool udpiiu :: searchRespAction (
         serverAddr.ia.sin_addr = addr.ia.sin_addr;
     }
 
+#ifdef MIN_IOC_CA_VER_WHEN_REDIRECT
+    // Try to detect the nameserver redirecting us to a different IP
+    if (memcmp(&addr.ia.sin_addr, &serverAddr.ia.sin_addr, sizeof(struct in_addr)) != 0) {
```
tboegi: Does it make sense to avoid memcmp() here and use something similar
to the code in repeater.cpp ?

`inline bool repeaterClient::identicalAddress ( const osiSockAddr &fromIn )
{
    if ( fromIn.sa.sa_family == this->from.sa.sa_family ) {
        if ( fromIn.ia.sin_port == this->from.ia.sin_port) {
            if ( fromIn.ia.sin_addr.s_addr == this->from.ia.sin_addr.s_addr ) {
                return true;
            }
        }
    }
    return false;
}
`

modules/ca/src/client/udpiiu.cpp
```diff
@@ -725,6 +725,33 @@ bool udpiiu :: searchRespAction (
...
+
```
tboegi, line 733: Does it make sense to avoid memcmp() here and use something similar
as coded in repeater.cpp ?

`inline bool repeaterClient::identicalAddress ( const osiSockAddr &fromIn )
{
    if ( fromIn.sa.sa_family == this->from.sa.sa_family ) {
        if ( fromIn.ia.sin_port == this->from.ia.sin_port) {
            if ( fromIn.ia.sin_addr.s_addr == this->from.ia.sin_addr.s_addr ) {
                return true;
            }
        }
    }
    return false;
}

`

klemenv: I was not aware that repeater.cpp already does this. I used memcmp because of concern for sockaaddr possibly having platform dependent implementation (POSIX, BSD, WIN32). If that has been addressed (and used in repeater), I'm happy to change it to the expanded version as it's more intuitive albeit longer.

mdavidsaver (changes requested):
This change feels very hacky to me.  I don't like the idea of supporting it in Base going forward.

Still, I see value to keeping it available for sites stuck with a name server and older CA client versions.  So maybe switch it to Draft?

modules/ca/src/client/udpiiu.cpp
mdavidsaver: A couple of points.

IPv4 addresses `sockaddr_in::in_addr::s_addr`, aka `in_addr_t` can be compared as an integer.  No need for `memcmp()`.

Also, while not immediately relevant.  Be careful about comparing the whole `sockaddr_in` with `memcmp()` as the `sockaddr_in::sin_zero` member should not be compared (it may not always be zero...).  I found this out the hard way with the RTEMS legacy network stack...

klemenv:
I agree, this is a hack. But there's limited ways how to deal with old clients and servers that break its own protocol, and EPICS base is full of code supporting very early CA versions. There seem to be a lot of effort put in over the years to ensure EPICS IOCs with various CA protocols can talk to each other, and here we're just trying to continue with that tradition. Unfortunately upgrading IOCs to newer versions of EPICS it's not always easy or even feasible, and without this or any other patch we're forced to use ancient EPICS for new installations of nameservers.

Making this compile time option ensures that sites that don't need this are not impacted.

anjohnson:
Core Group: We don't like this approach, we prefer the modified 425 approach as described [here](https://github.com/epics-base/epics-base/pull/425#issuecomment-2775095927).

PR #626, opened 2025-04-04
Replace dbgrep by dbglob, leaving dbgrep as an alias

simon-ess:
Possible future plan to make dbgrep actually use regular expressions.

### DIFF
--- a/modules/database/src/ioc/db/dbIocRegister.c
+++ b/modules/database/src/ioc/db/dbIocRegister.c
@@ -233,18 +233,26 @@ static const iocshFuncDef dblaFuncDef = {"dbla",1,dblaArgs,
                                          "Example: dbla(\"alia*\")\n"};
 static void dblaCallFunc(const iocshArgBuf *args) { iocshSetError(dbla(args[0].sval));}
 
+/* dbglob */
+static const iocshArg dbglobArg0 = { "pattern",iocshArgStringRecord};
+static const iocshArg dbglobArg1 = { "fields",iocshArgString};
+static const iocshArg * const dbglobArgs[2] = {&dbglobArg0,&dbglobArg1};
+static const iocshFuncDef dbglobFuncDef = {"dbglob",2,dbglobArgs,
+                                           "List record names matching pattern and optionally print field values. \n"
+                                           "The pattern can contain any characters that are legal in record names as well as:\n"
+                                           " - \"?\", which matches 0 or one characters.\n"
+                                           " - \"*\", which matches 0 or more characters.\n\n"
+                                           "Example: dbglob(\"*gpibAi*\")\n"
+                                           "         dbglob(\"*gpibAi*\",\"VAL DESC\")\n"};
+static void dbglobCallFunc(const iocshArgBuf *args) { iocshSetError(dbglob(args[0].sval,args[1].sval));}
+
 /* dbgrep */
 static const iocshArg dbgrepArg0 = { "pattern",iocshArgStringRecord};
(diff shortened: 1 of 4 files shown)


modules/database/src/ioc/db/dbTest.c
```diff
@@ -283,7 +283,7 @@ long dbli(const char *pattern)
     return 0;
 }
 
-long dbgrep(const char *pmask)
+long dbglob(const char *pmask)
```
anjohnson: This removes `dbgrep()` as a VxWorks function/command.

simon-ess: D'oh

mdavidsaver (commented):
I have no objection to this new name.

modules/database/src/ioc/db/dbTest.h
```diff
@@ -29,6 +29,8 @@ DBCORE_API long dbla(const char *pmask);
 /* list infos */
 DBCORE_API long dbli(const char *patern);
 /*list records with mask*/
+DBCORE_API long dbglob(const char *pmask,const char *fields);
+/*list records with mask; alias for dbglob*/
```
anjohnson: Minor: The "mask" in both these comments isn't really the right word, could you change them to "pattern" or "glob pattern".

modules/database/src/ioc/db/dbIocRegister.c
```diff
@@ -233,18 +233,26 @@ static const iocshFuncDef dblaFuncDef = {"dbla",1,dblaArgs,
...
+                                           " - \"?\", which matches 0 or one characters.\n"
+                                           " - \"*\", which matches 0 or more characters.\n\n"
+                                           "Example: dbglob(\"*gpibAi*\")\n"
+                                           "         dbglob(\"*gpibAi*\",\"VAL DESC\")\n"};
+static void dbglobCallFunc(const iocshArgBuf *args) { iocshSetError(dbglob(args[0].sval,args[1].sval));}
+
 /* dbgrep */
 static const iocshArg dbgrepArg0 = { "pattern",iocshArgStringRecord};
 static const iocshArg dbgrepArg1 = { "fields",iocshArgString};
 static const iocshArg * const dbgrepArgs[2] = {&dbgrepArg0,&dbgrepArg1};
```
anjohnson: Since `dbgrep` is now an alias for `dbglob`, you could re-use `dbglobArgs` in the `dbgrepFuncDef` structure, since the arg's for both must be identical. The help text is different though, so you can't reuse `dbglobFuncDef` in the
```C
iocshRegister(&dbgrepFuncDef,dbgrepCallFunc);
```
line below though.

anjohnson (changes requested):
This will of course need a Release Note, and (since we aren't yet using Doxygen annotations from the dbTest.h file for documenting the iocsh commands) there should also be changes to the [IOC Test Facilities](https://docs.epics-controls.org/en/latest/appdevguide/IOCTestFacilities.html) document for the next release.

I'm working on a release-notes branch for accepting individual release notes as separate files, you're welcome to wait for that to be merged before finishing this PR if you like.

modules/database/src/ioc/db/dbTest.h
anjohnson: … and I've just noticed that the prototype argument names used in the header are also using "mask", maybe modify them too?

simon-ess:
I have added a release snippet. As I understand how git manages directories, this should be mergeable both before #628 or after it.

PR #627, opened 2025-04-04
fix epicsStrtod for 32 bit architectures

dirk-zimoch:
On 64 bit architectures, double values could be parsed as  64 bit hex integers (with possible precision loss) but on 32 bit architectures only 32 bit hex values worked. Allow 64 bit hex values on 32 bit architectures too.


anjohnson:
Core group: approved.

PR #628, opened 2025-04-11
Conflict-free release note entries for GitHub pull requests

### LINKED ISSUE #598: PR Conflicts in Release Notes
There are getting annoying, I will take on the issue of creating a subdirectory for adding individual markdown files for each entry, which at release time a script will concatenate into a file for that release (and the individual files deleted).

anjohnson:
This PR replaces the single `documentation/RELEASE_NOTES.md` file in the
EPICS source tree with a mechanism that should prevent merge conflicts when
new entries are added from multiple pull requests.

In this approach each pull request adds a separate Markdown file to the
`documentation/new-notes` directory with a unique name. All these files get
combined into a single `RELEASE-<version>.md` file when a release is made and
the `new-notes` directory is emptied ready for further development.

Instructions on creating new entries are provided in a `README.txt` file in
the `documentation/new-notes` directory. Running `make` in the `documentation`
directory will combine all the new entries into a `RELEASE_NOTES.md` file that
then gets installed into the `doc` top-level directory. The `make` command
will also install the older `RELEASE-<version>.md` files into that `doc`
directory.


anjohnson:
To-Do:
+ The generated RELEASE_NOTES.md file is [published by RTD here](https://epics--628.org.readthedocs.build/projects/base/en/628/RELEASE_NOTES.html) but we should also add links to the older Release-<version>.md files in the release, either as links in the output file or in a separate doc - @minijackson?
+ Codacy is overly opinionated about the Markdown, I disagree with many of its suggestions but it might have some suggestions that could be used.
+ @simon-ess asked whether the file generation could be done without using a Perl script. It was easier to start in Perl but looks possible, although it would make the Makefile quite ugly.

simon-ess:
Part of my question is whether or not this needs a script at all, yes, and secondarily whether or not it needs to be perl (although since you have written it...). As we discussed, I don't think that we need a script to make new entries, people can really just do that by hand.

Anyhow, given that this is meant to be a tool used by the people performing a release, I think we could easily write this in any language we want as it is does not need to be supported on every system. The same is true for any of our documentation generation, to be honest: this doesn't have to be something that VxWorks, or an old version of RTEMS or whatnot can actually do. We could use python 3.13 for any of these if we want to, since this will typically done by a much more limited subset of people, making it quite different than other parts of the build system.

minijackson:
For the release notes in RTD, I think it would be better to put release notes into their own directory, and we could link them like so:

````markdown
```{toctree}
:caption: Release notes
:glob:
:hidden:
:reversed:
:titlesonly:

release-notes/*
```
````

I'm unsure of the usefulness of the `RELEASE_NOTES.md` file after this PR. Why not make the script take a `RELEASE-<version>.md` command-line argument and remove `RELEASE_NOTES.md` entirely? The process would then be:

- People add release notes into `documentation/release-notes/new-notes/PR-<xxx>.md`
- When a new release happens:
  - The release manager person runs `./documentation/release-notes/make-notes -o ./documentation/release-notes/RELEASE-<version>.md`
  - The release manager tracks this new file using git
  - The new release notes would be automatically picked up by RTD

**Comment by anjohnson** [comment]
@minijackson Thanks for your comments.

The EPICS build philosophy is that all files which are intended for use downstream be installed into or under a specific `$(INSTALL_LOCATION)/something` directory. I will strongly resist attempts to require users or modules to delve inside the Base source tree since that limits any structural changes to the sources that we might want to do in the future.

My new rules in this PR install the generated `RELEASE_NOTES.md` file into `doc`. Actually the only thing that running plain `make` in the `documentation` directory does is to create that file and install it and the other `.md` files. It's possible to make the `documentation` directory a regular directory build target in the top-level `Makefile` so those files will always be installed by the build, and I like the idea of doing that.

When building sources from an official release, the `RELEASE_NOTES.md` file isn't needed because the `RELEASE-<version>.md` file will have been included and the `documentation/new-notes` directory shouldn't have any `.md` files left. I haven't implemented the code to finish the release part yet, that will be coming.

Between releases though I don't want to publish each of the `new-notes/*.md` files separately, and combining them was the main purpose of this PR. I needed a name for the generated file, and I didn't want to use `RELEASE-$(EPICS_SHORT_VERSION).md` because that version won't have been released yet — the next Base release might have a slightly different version number anyway. RTD was already publishing and pointing to `RELEASE_NOTES.html` so it was easiest to keep that name.

Adding links to the other versioned files to the generated file should be easy, I'll give that a try next.

**Comment by anjohnson** [comment]
The `RELEASE_NOTES.md` file is now a generated index with links to all the included `RELEASE-<version>.md` files found in this build, as well as containing the new notes entries between tagged releases.

The Perl script inserts the older releases into the index in the correct order (descending).
While they do appear in the side-bar on RTD, those are links to the line in the `RELEASE_NOTES.md` file, if someone can advise how to change the generated Markdown to make them link directly to the other file I would appreciate help with that. I suspect that using `toctree` for those files as suggested by @minijackson would present them in the wrong order.

**Comment by anjohnson** [comment]
Core group: Undo changes to top Makefile.

**Comment by anjohnson** [comment]
Fixed the issue with the sidebar, the generated `RELEASE_NOTES.md` file now contains directives for the Sphinx myst processor to include all of the `RELEASE-<version>.md` files with appropriate adjustments to the start line and header levels. This results in the published notes all appearing on the same webpage, and the sidebar TOC shows the right structure.

I also added some more fixes to the dbd.pod files, and updated the ReleaseChecklist instructions. I think this PR is ready to merge now.

-----

Recent Sphinx & plugin versions still aren't able to build this without crashing, although I have cleaned up some of the warnings I was getting from them.

One issue that they don't like is there being an instance of [an enum](https://epics--628.org.readthedocs.build/projects/base/en/628/db_access_h.html#_CPPv415dbr_value_class) with the same name as the enum's tag-name. My workaround is to turn the tagged enum into a typedef, but that change isn't included in this PR:

```patch
diff --git a/modules/ca/src/client/access.cpp b/modules/ca/src/client/access.cpp
index 58cc4bca5..e53d0f078 100644
--- a/modules/ca/src/client/access.cpp
+++ b/modules/ca/src/client/access.cpp
@@ -944,7 +944,7 @@ const unsigned short dbr_value_size[LAST_BUFFER_TYPE+1] = {
 };
 
 //extern "C"
-const enum dbr_value_class dbr_value_class[LAST_BUFFER_TYPE+1] = {
+const dbr_value_class_e dbr_value_class[LAST_BUFFER_TYPE+1] = {
     dbr_class_string,       /* string max size              */
     dbr_class_int,          /* short                        */
     dbr_class_float,        /* IEEE Float                   */
diff --git a/modules/ca/src/client/db_access.h b/modules/ca/src/client/db_access.h
index 0fd9c5691..5d48ae113 100644
--- a/modules/ca/src/client/db_access.h
+++ b/modules/ca/src/client/db_access.h
@@ -554,13 +554,14 @@ LIBCA_API extern const unsigned short dbr_value_size[];
 
 #ifndef db_accessHFORdb_accessC
 /* class for each type's value */
-enum dbr_value_class {
+typedef enum {
                 dbr_class_int,
                 dbr_class_float,
                 dbr_class_string,
-                dbr_class_max};
+                dbr_class_max
+} dbr_value_class_e;
 
-LIBCA_API extern const enum dbr_value_class dbr_value_class[LAST_BUFFER_TYPE+1];
+LIBCA_API extern const dbr_value_class_e dbr_value_class[LAST_BUFFER_TYPE+1];
 
 /*
  * ptr to value given a pointer to the structure and the DBR type
```

anjohnson:
Core Group: Approved.

PR #629, opened 2025-04-11
caget: convert INT to SHORT in dbr type.

ericonr:
This makes the interface consistent with the caget help message, which documents DBR*_INT types as separate items, though they are equivalent to SHORT types.

The INT to SHORT replacement happens before any calls to dbr_text_to_type, because it's impossible for a type ending in "INT" to be parsed successfully by dbr_text_to_type. And the longest DBR type name is shorter than 20 chars, so 30 chars is definitely a safe bet for temporary buffers (and matches the size used elsewhere in the function).

We rename the 'str' buffer to 'dbrStr' to help with disambiguation.

---

I considered updating the help message, but that was much more manual work to align properly >:)

### DIFF
--- a/modules/ca/src/tools/caget.c
+++ b/modules/ca/src/tools/caget.c
@@ -417,12 +417,24 @@ int main (int argc, char *argv[])
                                 /* Argument (type) may be text or number */
             if (sscanf(optarg, "%d", &type) != 1)
             {
-                dbr_text_to_type(optarg, type);
+                const char *typeStr = optarg;
+                size_t len = strlen(typeStr);
+                char intToShortStr[30];
+                if (len >= 3 && len < (sizeof intToShortStr) - 2)                /* There's room to replace INT with SHORT (2 extra chars) */
+                {
+                    if (strcmp(typeStr + len - 3, "INT") == 0)                /* Replace INT with SHORT */
+                    {
+                        strcpy(intToShortStr, typeStr);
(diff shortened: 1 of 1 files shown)


anjohnson:
Core Group: Can you explain what you think the problem is here? A `DBR_INT` value is defined to be the same size as a `DBR_SHORT`, in CA both are symbolic names for the same data type; this API was designed when many CPUs and C compilers defined an int as 16 bits.

ericonr:
@anjohnson it's related to this tech-talk thread: https://epics.anl.gov/tech-talk/2023/msg01713.php

`caget` currently does not accept `DBR_INT` as a type identifier, even though it documents it in the help output. This PR converts `caget -d X_INT` into `caget -d `X_SHORT`, so the type parsing functions can deal with it properly. I realize now that this motivation wasn't documented in the commit message.

mdavidsaver:
> `caget` currently does not accept `DBR_INT` as a type identifier

Ah.  So you want to handle the following?

```sh
$ ./bin/linux-x86_64/caget -d STS_INT foo
Requested dbr type out of range or invalid - ignored. ('caget -h' for help.)
...
```

ericonr:
Indeed :)

ericonr:
I included

> Despite being documented in the caget help message as possible arguments
for the '-d' option, passing DBR*_INT types to caget using their names
(instead of numerical values) does not work.

In the commit to hopefully make my intention clearer.

PR #631, opened 2025-04-14
define more editorconfig properties

dirk-zimoch:
Ensure intent style space for source code.
Fixed those files which had tab indents.

Ensure lines have no trailing whitespace.
Trimmed trailing whitespace from all file which had them.


.editorconfig
```diff
@@ -6,3 +6,19 @@ root = true
 # as some compilers complain about files not ending in newline
 [*]
 insert_final_newline = true
+end_of_line = lf
```
anjohnson: I don't think this particular setting would work well for users developing on Windows, did you add it based on personal experience?

Our `$(TOP)/.gitattributes` file currently tells Git to check out most text files using the native EOL for the host OS, and convert them back to line-feeds when committing changes. Since Git has been getting this right for a long time I'm not in favor of changing it. I suspect there are editors on Windows that could get confused, and I'd rather leave this to Git.

dirk-zimoch: It is to defend against users developing on Windows mixing in their '\r\n' line terminators in case they mis-configure git. I am currently checking where it breaks and what I have to change to make the current code match. Maybe I switch this PR to a draft for the moment...

dirk-zimoch: I have cloned this branch in windows and see what you meant.
I have removed end_of_line from .editorconfig.

.editorconfig
```diff
@@ -6,3 +6,31 @@ root = true
...
+[*.cc]
+indent_style = space
+
+[*.cpp]
+indent_style = space
+
+[*.h]
+indent_style = space
+
+[*.hpp]
```
JJL772: These can be combined into one section with
```editorconfig
[*.{c,cc,cpp,h,hpp}]
indent_style = space
```

dirk-zimoch: „Available since EditorConfig Core 0.11.0“
I wanted to stay backward compatible.
On the other hand, 0.11.0 is really old now.

mdavidsaver:
> Fixed those files which had tab indents.

How was this done?  Some tool?  `sed`?

dirk-zimoch:
Text editor :-)
I manually edited all files that eclint complained about.

dirk-zimoch:
PS: If you compare with whitespace changes hidden, you will see that the changes in the affected files disappear.

.editorconfig
anjohnson: Core: Please combine the patterns, rebase. 0.11.0 is 12+years old now.

dirk-zimoch: Done.

anjohnson:
Core Group: Trim trailing spaces on C/... sources.

dirk-zimoch:
I had to fix some of the 3rd party files (boost) to remove trailing spaces.

dirk-zimoch:
Rebased onto current 7.0 HEAD.

dirk-zimoch:
Rebased. (and new whitespaces fixed).

dirk-zimoch:
I noticed that the editorconfig job uses a quite old version of eclint: 0.5.0
I have 2.8.1 on my RHEL9 system!

Maybe [greut/eclint-action](https://github.com/greut/eclint-action]) is a bit outdated or uses an outdated container?
Should we rather implement the action ourselves by installing the latest eclint and running that?

mdavidsaver:
> Maybe [greut/eclint-action](https://github.com/greut/eclint-action) is a bit outdated or uses an outdated container?

That "action" git repository has not been updated in 2 years.  Same for what I think is the repository behind the docker image used https://github.com/greut/eclint .

> I have 2.8.1 on my RHEL9 system!

Any idea where RH got this source from?  Presumably this is from a fork or alternative project?

dirk-zimoch:
I installed eclint with `sudo npm install -g eclint`, not as an rpm.
`nmp search eclint` says it got it from https://npm.im/eclint.
That page claims that eclint version 2.8.1 is 7 years old.
Also, on its "Versions" page, there is no version 0.5.0. I don't know where greut/eclint-action got its version from.

ralphlange:
Yep. Different implementations.
The 2.8.1 (2018) is JavaScript/TypeScript, while the 0.5.0 (2023) is written in Go.

simon-ess:
Well that's not confusing _at all_

dirk-zimoch:
If it is this one: https://gitlab.com/greut/eclint -- I only get "500 Internal Server Error" (with greetings from cloudflare)

ralphlange:
https://github.com/greut/eclint

dirk-zimoch:
Hm... even the syntax is different. The JavaScript version requires a command `check` and the go version must be used without. Nice "replacement".

But at least it no longer complains about empty files having no final newline.

anjohnson:
The idea of us installing anything directly from npm makes me nervous given their security history, but I've never actually done that myself.

Here's a potentially silly idea (ignore me if I've misunderstood the conversation above), could we change this job to be run in say a Rocky-9 docker container, assuming that would have the newer version of eclint available as an RPM so we wouldn't have to pull it from npm ourselves?

ralphlange:
The "newer" version (with the higher number) is an older implementation.

Using the 0.5.0 (or 0.5.1) version's container image sounds like the most appropriate thing to do.

dirk-zimoch:
Rebased and new whitespaces in iocsh.h fixed.

dirk-zimoch:
I noticed another difference between the two eclint variants: the go implementation ignores `indent_style = space` if `indent_size` is not specified as well. The JavaScript variant can handle this. Thus, our test may not trigger if people add tabs again. Need to test this.
The other difference was that the go variant does not complain about empty files having no final newline. That was an improvement.

PR #634, opened 2025-04-16
Always search Homebrew and MacPorts directories

dirk-zimoch:
This allows other modules like pvxs to find files there, too. Still, readline is recogized correctly.

### DIFF
--- a/configure/os/CONFIG_SITE.darwinCommon.darwinCommon
+++ b/configure/os/CONFIG_SITE.darwinCommon.darwinCommon
@@ -6,24 +6,17 @@
 # These settings are designed for users of Homebrew.
 # Users of other third-party package managers are welcome to
 # provide patches appropriate for their manager.
-ifneq (,$(wildcard /opt/homebrew))
-  # Default location on aarch64
-  HOMEBREW_DIR = /opt/homebrew
-else ifneq (,$(wildcard /usr/local/Homebrew))
-  # Default location on x86_64
-  HOMEBREW_DIR = /usr/local
-else ifneq (,$(wildcard /opt/local/include/readline))
-  # MacPorts
-  READLINE_DIR = /opt/local
-endif
+HOMEBREW_DIR := $(shell brew --prefix 2>/dev/null)
+INCLUDES += $(HOMEBREW_DIR:%=-I%/include)
+LDFLAGS += $(HOMEBREW_DIR:%=-L%/lib)
 
 # Look for Homebrew's readline
 ifneq (,$(wildcard $(HOMEBREW_DIR)/opt/readline))
-  READLINE_DIR = $(HOMEBREW_DIR)/opt/readline
+  INCLUDES_READLINE += -I$(HOMEBREW_DIR)/opt/readline/include
+  LDFLAGS_READLINE += -L$(HOMEBREW_DIR)/opt/readline/lib
 endif
 
-# Use GNU readline if it's avaiilable
-ifneq (,$(wildcard $(READLINE_DIR)/include/readline/readline.h))
-  INCLUDES_READLINE = -I$(READLINE_DIR)/include
-  LDFLAGS_READLINE = -L$(READLINE_DIR)/lib
-endif
+# Look for MacPorts
+MACPORT_DIR := $(patsubst %/bin/port,%,$(shell which port))
+INCLUDES += $(MACPORT_DIR:%=-I%/include)
+LDFLAGS += $(MACPORT_DIR:%=-L%/lib)


configure/os/CONFIG_SITE.darwinCommon.darwinCommon
```diff
@@ -6,24 +6,13 @@
...
-else ifneq (,$(wildcard /usr/local/Homebrew))
-  # Default location on x86_64
-  HOMEBREW_DIR = /usr/local
-else ifneq (,$(wildcard /opt/local/include/readline))
-  # MacPorts
-  READLINE_DIR = /opt/local
-endif
+#HOMEBREW_DIR := $(shell brew --prefix 2>/dev/null)
+INCLUDES += $(HOMEBREW_DIR:%=-I%/include)
+LDFLAGS += $(HOMEBREW_DIR:%=-L%/lib)
```
anjohnson: Does leaving the `HOMEBREW_DIR` line commented out put strange values into the `INCLUDES` and `LDFLAGS` variables here? Did you comment that line out because it dies unless `brew` is installed?

The point of the code you're replacing was that the user shouldn't have to edit this `CONFIG_SITE` file at all if they have Homebrew installed in either of the two standard locations as described in the comments. I don't want to undo that convenience.

anjohnson (commented):
Please confirm that removing the lines which set `READLINE_DIR` still works in all cases. I don't see a specific reason why we should need it given the `toolchain.c` check, but I'd like to know your reasoning for the comment above "Still, readline is recogized correctly."

configure/os/CONFIG_SITE.darwinCommon.darwinCommon
dirk-zimoch: Oops, that should not have gone in. It was my test that everything works even if homebrew is not installed. I need to remove the comment.

dirk-zimoch:
My reasoning for "Still, readline is recogized correctly" is that I have built it and could confirm that I still get readline support without explicitly setting READLINE_DIR.

dirk-zimoch:
Hmm... I have completely removed homebrew and still can build EPICS base with no problems and still seem to have readline.
So homebrew is not really needed?

anjohnson:
Apple provide a library that pretends to be libreadline and supports line-editing and recall, but AFAIK it doesn't handle the command completion that has been added to iocsh recently. Please run `otool -L` on your IOC binary to see which readline library it's being linked to, and check whether it supports completion.

anjohnson:
Actually you may need to run `otool -L` against the libCom.*.dylib to see which readline it links to:
```
Kato$ otool -L /Users/anj/Software/epics/base-7.0/bin/darwin-x86/softIoc
/Users/anj/Software/epics/base-7.0/bin/darwin-x86/softIoc:
	/Users/anj/Software/epics/base-7.0/lib/darwin-x86/libdbRecStd.3.24.1.dylib (compatibility version 3.24.1, current version 3.24.1)
	/Users/anj/Software/epics/base-7.0/lib/darwin-x86/libdbCore.3.24.1.dylib (compatibility version 3.24.1, current version 3.24.1)
	/Users/anj/Software/epics/base-7.0/lib/darwin-x86/libca.4.14.6.dylib (compatibility version 4.14.6, current version 4.14.6)
	/Users/anj/Software/epics/base-7.0/lib/darwin-x86/libCom.3.24.1.dylib (compatibility version 3.24.1, current version 3.24.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1800.101.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
Kato$ ldd /Users/anj/Software/epics/base-7.0/lib/darwin-x86/libCom.3.24.1.dylib
/Users/anj/Software/epics/base-7.0/lib/darwin-x86/libCom.3.24.1.dylib:
	/Users/anj/Software/epics/base-7.0/lib/darwin-x86/libCom.3.24.1.dylib (compatibility version 3.24.1, current version 3.24.1)
	/usr/local/opt/readline/lib/libreadline.8.dylib (compatibility version 8.2.0, current version 8.2.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1800.101.0)
```

dirk-zimoch:
You're right, without homebrew I have line editing and history but no tab completion.
`otool -L libCom.dylib` shows: 
`/usr/lib/libedit.3.dylib (compatibility version 2.0.0, current version 3.0.0)`

If homebrew is available but does not have the readline package, the situation is the same, of course. But base still builds without errors. Should we print a message in that case?

dirk-zimoch:
Now tested with both, Homebrew and MacPorts (and without). Linking confirmed with `otool -L libCom.dylib`.

I have built pvxs with libevent provided by either, Homebrew or MacPort. Both work without the need to configure pvxs (see also https://github.com/epics-base/pvxs/pull/112)

In case a library is provided by both, Homebrew and MacPorts, the Homebrew installation takes precedence.

It is also possible to mix for example Homebrew's readline with MacPort's libevent. Just in case someone wants to walk that road.

The install locations of both, Homebrew and MacPorts, are now detected, rather than assumed which is more important, and easier, for Homebrew. Installing MacPorts in a different location is discouraged and only possible with a manual installation, but we never know...

dirk-zimoch:
rebased onto the fixes for "Fedora Latest"

PR #636, opened 2025-04-17
Add support for EPICS_DB_INCLUDE_PATH to dbLoadTemplate

daneos:
This allows `dbLoadTemplate` to search through either a custom-provided list of paths, or through paths provided by the environment variable `EPICS_DB_INCLUDE_PATH`.


ralphlange:
I'm not sure I fully understand the scope of this PR.

I understand the additional argument to the `dbLoadTemplate` command. (That would need changes in the documentation, right?)

But EPICS_DB_INCLUDE_PATH already exists and is supported - what exactly is the added functionality?

daneos:
My understanding is that `EPICS_DB_INCLUDE_PATH` is supported for `dbLoadRecords` (or more generally everything that uses `dbReadCOM` under the hood) but not for `dbLoadTemplate`.

I am not the author of the patch though, so input from @simon-ess would be greatly appreciated.

mdavidsaver:
From a quick test, it looks like `dbLoadTemplate` already respects `EPICS_DB_INCLUDE_PATH` for `file`.

```
mkdir xyz
touch xyz/foo.template
cat <<EOF > xyz/test.substitutions
file "foo.template" {
{X="Y"}
}
```

```
$ ./bin/linux-x86_64/softIoc 
epics> dbLoadTemplate xyz/test.substitutions
filename="../dbStatic/dbLexRoutines.c" line number=287 dbRead opening file foo.template
ERROR failed to load 'foo.template'
dbLoadRecords("foo.template", X="Y")
Substitution file error: Error while reading included file
line 3: '}'
epics>
```

```
$ EPICS_DB_INCLUDE_PATH=$PWD/xyz ./bin/linux-x86_64/softIoc 
epics> dbLoadTemplate xyz/test.substitutions
epics> 
```

anjohnson:
As @mdavidsaver said `dbLoadTemplate` does already allow `file` entries to be found using `EPICS_DB_INCLUDE_PATH` since that search is built into the routines it calls for each file. This PR allows the `.substitutions` file itself to be found using `EPICS_DB_INCLUDE_PATH`, which isn't supported at the moment. I would have expected your st.cmd files to know exactly where their `.substitutions` files are and so be able to prepend their path to the filename, but I don't use your build system.

Does ESS actually need to be able to provide a path on the `dbLoadTemplate` command-line as well as using the environment variable for the search path? I'm a bit more reluctant to add arguments like this unless you have a specific need for both.

mdavidsaver:
> This PR allows the `.substitutions` file itself to be found ...

Ah.  I think this has value.

> I would have expected your st.cmd files to know exactly where their `.substitutions` files are ...

I have encountered support modules installing `.substitutions` files in `<TOP>/db/`.  For applications referencing these support modules, a search path makes sense to me.

anjohnson:
> I have encountered support modules installing `.substitutions` files in `<TOP>/db/`.

Indeed, the epics-modules LabJack and measComp modules for example. The sample st.cmd file for the LabJack T7 IOC [uses this line](https://github.com/epics-modules/LabJack/blob/ecddc41285f13ac401725cbd4c5c2dce11d6099d/iocBoot/iocLabJack_T7/st.cmd#L26) to load the appropriate databases:
```
dbLoadTemplate("$(LABJACK)/db/LabJack_T7.substitutions", "P=$(PREFIX), PORT=$(PORT), WDIG_POINTS=$(WDIG_POINTS), WGEN_POINTS=$(WGEN_POINTS)")
```
Inside it lists a series of files [like this](https://github.com/epics-modules/LabJack/blob/ecddc41285f13ac401725cbd4c5c2dce11d6099d/LabJackApp/Db/LabJack_T7.substitutions#L2):
```
file "$(LABJACK)/db/LabJack_device.template" { pattern
```
The `LABJACK` environment variable is set by the `envPaths` file generated from the IOC's `configure/RELEASE` file. I assume the E3 build system provides something similar for pointing to the IOC's support modules.

If a `.substitutions` file only uses filenames without paths in its `file` statements and they all come from the module's `<TOP>/db/`, this approach works instead:
```
cd "$(LABJACK)/db"
dbLoadTemplate("LabJack_T7.substitutions", "P=$(PREFIX), PORT=$(PORT), ...)
```

I'd be interested in hearing about cases where neither of these techniques work.

daneos:
> This PR allows the .substitutions file itself to be found using EPICS_DB_INCLUDE_PATH

Yes, this is correct.

> Does ESS actually need to be able to provide a path on the dbLoadTemplate command-line

I'm not sure, I will check our IOCs for that.

> The LABJACK environment variable is set by the envPaths file generated from the IOC's configure/RELEASE file. I assume the E3 build system provides something similar for pointing to the IOC's support modules.

For e3, `require` does most of the heavy lifting setting up paths and environment variables, and this includes `EPICS_DB_INCLUDE_PATH`.

ralphlange:
FWIW:
I found having paths inside .substitutions "file" directives was clashing with the EPICS_DB_INCLUDE_PATH too often and regularly broke setups relying on the order of directories in the include path env variable.
So for CODAC@ITER: no paths in the "file" command; everything is found via EPICS_DB_INCLUDE_PATH; substitutions files are always below $TOP and called from st.cmd using their full path.

mdavidsaver:
> I'd be interested in hearing about cases where neither of these techniques work.

I'm sure it won't surprise you to learn that I have a different preference for structuring `st.cmd`.  To match your example, this would look like:

```
epicsEnvSet("EPICS_DB_INCLUDE_PATH", "$(EPICS_DB_INCLUDE_PATH):$(LABJACK)/db")
dbLoadTemplate("LabJack_T7.substitutions", "P=$(PREFIX), PORT=$(PORT), ...)
```

I have given occasional thought to teaching `convertRelease.pl` to set `EPICS_DB_INCLUDE_PATH` automatically in `envPaths`.  Or similarly adding `epicsEnvAppend()` to mitigate the length limitation imposed by macLib.

The root of this preference goes back to the epics-deb packaging, where I could prefix my scripts with:

```
epicsEnvSet("EPICS_DB_INCLUDE_PATH", "$(PWD)/../../db:/usr/lib/epics/db")
```

ralphlange:
Ah, that's the common ancestor.

anjohnson:
I'm never surprised by [TMTOWTDI](https://en.wiktionary.org/wiki/TMTOWTDI), "a motto associated with the [Perl](https://en.wiktionary.org/wiki/Perl#English) programming language", as against Python's "there should be one — and preferably only one — obvious way to do it" guideline in [PEP-20](https://www.python.org/dev/peps/pep-0020/).

I have no objections to `dbLoadTemplate` searching `EPICS_DB_INCLUDE_PATH` for `.substitutions` files.

I do still want to hear whether the added command-line argument to provide a local override to that path is actually needed though. It looks like this PR sets the search path to that (if given) when opening the `.substitutions` file, but it might _not_ be used when opening the individual template files, and that could be very confusing.

simon-ess:
Just to step into this after vacation:

The origin of this patch is that the PSI require that we inherited and use _overrides_ `dbLoadTemplate` with a completely new function including a re-implementation of the lex/yacc code, specifically to allow `dbLoadTemplate` to search for .substitutions files using a search path that is configured (in our/PSI's case, by require, but plausibly by convertRelease or some such tool).

We found this to be un-economical and though it made more sense to patch dbLoadTemplate instead of having our own custom implementation of it, and thought it also might be a useful feature in EPICS base; as @daneos said, almost all of the functions that try to read db-ish files from disc use this path, so it seems sensible to add it here as well.

Regarding the `path` variable, this comes back to the PSI implementation that we inherited. As near as we have been able to assess, we do not use it at all. However, it was present in the PSI version, and so @dirk-zimoch might have more input as to why this is there.

modules/database/src/ioc/dbtemplate/dbtoolsIocRegister.c
```diff
@@ -15,23 +15,24 @@
...
     "dbLoadTemplate",
-    2,
+    3,
     dbLoadTemplateArgs,
     "Load the substitution file given as first argument, apply the substitutions\n"
     "for each template in the substitution file, and load them using 'dbLoadRecords'.\n\n"
     "The second argument provides extra variables to substitute in the\n"
-    "template files (not the substitution file).\n\n"
+    "template files (not the substitution file). The third argument provides\n"
+    "a list of paths to search through for the subsitutions file.\n\n"
```
anjohnson: Also used for the .db file searches.

modules/database/src/ioc/dbtemplate/dbLoadTemplate.y
```diff
@@ -335,7 +339,7 @@ static int yyerror(char* str)
 
 static int is_not_inited = 1;
 
-int dbLoadTemplate(const char *sub_file, const char *cmd_collect)
+int dbLoadTemplate(const char *sub_file, const char *cmd_collect, const char *path)
```
anjohnson, line 342: This new argument is useful.

anjohnson (changes requested):
Core Group: Please adjust the help message, then we'll be happy.

modules/database/src/ioc/dbtemplate/dbtoolsIocRegister.c
```diff
@@ -15,23 +15,24 @@
...
     dbLoadTemplateArgs,
     "Load the substitution file given as first argument, apply the substitutions\n"
     "for each template in the substitution file, and load them using 'dbLoadRecords'.\n\n"
     "The second argument provides extra variables to substitute in the\n"
-    "template files (not the substitution file).\n\n"
+    "template files (not the substitution file). The third argument provides\n"
+    "a list of paths to search through for the subsitution and template files.\n\n"
     "See 'help dbLoadRecords' for more information.\n\n"
-    "Example: dbLoadTemplate db/my.substitutions 'user=myself,host=myhost'\n",
+    "Example: dbLoadTemplate db/my.substitutions 'user=myself,host=myhost' 'path/to/subst:path2/to2/subst2'\n",
```
mdavidsaver, line 31: How to document the addition of this argument?

If this was a C prototype, this would be a doxygen tag `\since UNRELEASED`.  I am unsure what to do about an IOC shell function through.  Maybe just add a line `Third argument added in UNRELEASED.`?

ralphlange: In the documentation - yes. (+1)

But here, in the online help?
If you're on an older system, you don't get the hint about the new functionality.
If you're on a new system, the introduction date is not that valuable when working with the system.

daneos: Should I adjust the help message then or leave it as is?

anjohnson: I agree with @ralphlange there doesn't need to be anything like `\since` in the help message, which should only document the running IOC.

The addition of this argument should be mentioned in a Release Note entry though, would you please add one. We are currently changing how we accept new entries; my PR #628 which implements the update is still being finalized but [the instructions here](https://github.com/epics-base/epics-base/pull/628/files#diff-65030e178047020d82f32b28d171cc4fdf0eb5ecbe10c9193f8edf9f7f8699bf) won't change and can be followed now. You will have to create a directory `documentation/new-notes` for your Markdown file called `PR-636.md`, but once both this and my PR get merged your note will be pulled into the published Release Notes file automatically.

simon-ess: See https://github.com/epics-base/epics-base/pull/626/files for an example of the release notes.

daneos: Added the release notes entry

anjohnson:
Core Group: Remove the dbltExpand program completely. Then approved.

simon-ess:
@anjohnson  Done, but appveyor run failed.

PR #637, opened 2025-04-17
Fix use-after-free issue on array filters

lucmaga:
This is a long issue we have are having with some complex IOCs at ESS. We have a code to reproduce it, from @simon-ess, https://gitlab.esss.lu.se/icshwi/pva-datarace-test. This reproduce with QSRV as PVA backend but we could not reproduce wtih QSRV2. Still IMO the bogus code is on the array filter as freePvt can be called even if there are still array references on the event queue.

---
This patch fix a use-after-free bug in the array filter that happens when the filter's private struct (`pvt`) is freed while there are still pending events referencing it. Specifically, this leads to the destruction of the `arrayFreeList` and its mutex, which is later accessed during field log cleanup.

When the queue thread eventually processes one of these `db_field_log` entries and attempts to free the array via `freeListFree()`, it tries to lock a now-destroyed mutex. This results in `epicsMutexMustLock`
failing with `EINVAL`, triggering an assertion and suspending the queue thread. The error looks like the following:

A call to 'assert(status == epicsMutexLockOK)'
    by thread 'PDB-event' failed in ../freeList/freeListLib.c line 169.
errlog: lost 1 messages
[    0x7f82695e7fb3]errlog: lost 3 messages
[    0x7f82695e10ce]errlog: lost 3 messages
[    0x7f82695cfb9a]errlog: lost 3 messages
[    0x7f826965c6e9]errlog: lost 3 messages
[    0x7f826965cce3]errlog: lost 3 messages
[    0x7f82695e2315]errlog: lost 3 messages
[    0x7f82691f7a92]errlog: lost 3 messages
[    0x7f82692797d0]errlog: lost 3 messages
EPICS Release EPICS R7.0.8.1-E3-7.0.8.1-patch.
Local time is 2025-03-20 08:53:04.260698231 CET
errlog: lost 2 messages
Thread PDB-event (0x558fad15ca90) suspended
callbackRequest: ERROR cbLow ring buffer full

To reproduce this data race for PVA was used a simple python script to rapidly open and close monitor on array PVs with a random filter. The same script cannot reproduce this on PVXS.

The fix is using a simple reference counter to ensure the pvt struct is never destroyed if there are still references for arrays on the arrayFreeList.

### DIFF
--- a/modules/database/src/std/filters/arr.c
+++ b/modules/database/src/std/filters/arr.c
@@ -28,6 +28,8 @@ typedef struct myStruct {
     epicsInt32 end;
     void *arrayFreeList;
     long no_elements;
+    long refCount;
+    unsigned int markToDestroy;
 } myStruct;
 
 static void *myStructFreeList;
@@ -48,6 +50,8 @@ static void * allocPvt(void)
     my->start = 0;
     my->incr = 1;
     my->end = -1;
+    my->refCount = 1;
+    my->markToDestroy = 0;
     return (void *) my;
 }
 
@@ -55,8 +59,15 @@ static void freePvt(void *pvt)
 {
     myStruct *my = (myStruct*) pvt;
 
-    if (my->arrayFreeList) freeListCleanup(my->arrayFreeList);
-    freeListFree(myStructFreeList, pvt);
+    my->refCount--;
+    if (my->refCount > 0) {
+        my->markToDestroy++;
+        return;
+    }
+    if (my->refCount <= 0 ) {
+        if (my->arrayFreeList) freeListCleanup(my->arrayFreeList);
+        freeListFree(myStructFreeList, pvt);
+    }
 }
 
 static int parse_ok(void *pvt)
@@ -69,8 +80,14 @@ static int parse_ok(void *pvt)
 
 static void freeArray(db_field_log *pfl)
 {
+    myStruct *my = (myStruct *)pfl->u.r.pvt;
+
     if (pfl->type == dbfl_type_ref) {
-        freeListFree(pfl->u.r.pvt, pfl->u.r.field);
+        freeListFree(my->arrayFreeList, pfl->u.r.field);
+        my->refCount--;
+    }
+    if (my->refCount <= 0 && my->markToDestroy){
+        freePvt(my);
     }
 }
 
@@ -126,7 +143,8 @@ static db_field_log* filter(void* pvt, dbChannel *chan, db_field_log *pfl)
             if (pfl->dtor) pfl->dtor(pfl);
             pfl->u.r.field = pTarget;
             pfl->dtor = freeArray;
-            pfl->u.r.pvt = my->arrayFreeList;
+            pfl->u.r.pvt = my;
+            my->refCount++;
         }
         /* adjust no_elements (even if zero elements remain) */
         pfl->no_elements = nTarget;


ralphlange:
Do I understand correctly that the `arrayFreeList` is destructed when the subscription goes away - while there are still updates in the outgoing queue? Meh.

And PVXS assumedly keeps the subscription and its filters around until all its data has been sent?
@mdavidsaver - does this make sense to you?

mdavidsaver:
> This reproduce with QSRV as PVA backend but we could not reproduce wtih QSRV2.

I suspect this is a result of a long standing, and hard to squash, bug in pvAccessCPP.  Due to the way in which reference counting (with `std::shared_ptr`) is done, it is possible for a pair of TX and RX threads to be "orphaned" on disconnect, leaving them to actually stop (or deadlock) at some later time.  So attempts at sequenced shutdown, as required by the filter code, do not always work.

I specifically designed PVXS to avoid this kind of shared_ptr ownership issues, so I am happy to hear that you have not been able to reproduce with QSRV2.

> Still IMO the bogus code is on the array filter as freePvt can be called

imo. "bug" is with pvAccessCPP, which does not (always) correctly sequence shutdown.

I do not think that reference counting `myStructFreeList` in `acc.c` would be sufficient.  There are other free lists involved.  See `dbChannelExit()`.

mdavidsaver:
> And PVXS assumedly keeps the subscription and its filters around until all its data has been sent?

```
$ ./bin/linux-x86_64/softIocPVX -d tick.db
...
epics> var atExitDebug 1
epics> exit
atExit rlExit((nil))
atExit exitDatabase((nil))
atExit twdShutdown((nil))
atExit errlogExitHandler(0x562a8a22a680)
atExit syncShutdown((nil))
atExit decShutdown((nil))
atExit dbndShutdown((nil))
atExit arrShutdown((nil))
```

With Base >= 7.0.4, PVXS/QSRV2 shuts down with the "de-init" hooks as part of `exitDatabase()`.  With older Base look, for two at exit handlers `pvxsExitBeforeIocShutdown()` and `pvxsExitAfterIocShutdown()`.  Both should happen before any of the filter Shutdown handlers.

The `pvxsExitBeforeIocShutdown()` handler fully stops the PVXS server, which should stop all worker threads (and join() them properly).

In short, the ordering required by `dbChannel` should be respected.

lucmaga:
Hi, thanks for reviewing.

> I suspect this is a result of a long standing, and hard to squash, bug in pvAccessCPP. Due to the way in which reference counting (with std::shared_ptr) is done, it is possible for a pair of TX and RX threads to be "orphaned" on disconnect, leaving them to actually stop (or deadlock) at some later time. So attempts at sequenced shutdown, as required by the filter code, do not always work.

This descriptions sounds like a possible explanation. In another IOC we tested, using the same script, it was way harder to trigger this behavior. Instead we saw lots of TX and RX threads in not OK mode. Interestingly this is not what I see on the IOC that can reproduce this.

I'm relatively new to this code so I would appreciate help to narrow this down. Is there a way to pin point that a subscription is connected with a specific TX/RX thread? I could try to understand what's the state of the thread when this issue occurs

> I do not think that reference counting myStructFreeList in acc.c would be sufficient. There are other free lists involved. See dbChannelExit().

`dbChannelDelete` will free the channel reference, `pevent->chan` in `dbEvent.h`. But that is not a problem here because `db_cancel_event` is called before it so field logs are just cleaned by the task thread, `modules/pva2pva/pdbApp/pdb.cpp:613`. I agree this does not solve the wrong workflow on PVA, but this fix the use after free issue on array filters if the event queue gets to this state. I though this was a valid state because `db_cancel_event` describes it as a possibility.

> imo. "bug" is with pvAccessCPP, which does not (always) correctly sequence shutdown.

I agree `pvAccessCPP` does not correctly shutdown. But as we have multiple consumers of db API, so it should enforce the correct behavior instead of expecting users too use it correctly. The problem is finding how to isolate the workflow, as `dbChannel` and `dbEvent` functions are called separately.

lucmaga:
> > And PVXS assumedly keeps the subscription and its filters around until all its data has been sent?
> 
> ```
> $ ./bin/linux-x86_64/softIocPVX -d tick.db
> ...
> epics> var atExitDebug 1
> epics> exit
> atExit rlExit((nil))
> atExit exitDatabase((nil))
> atExit twdShutdown((nil))
> atExit errlogExitHandler(0x562a8a22a680)
> atExit syncShutdown((nil))
> atExit decShutdown((nil))
> atExit dbndShutdown((nil))
> atExit arrShutdown((nil))
> ```
> 
> With Base >= 7.0.4, PVXS/QSRV2 shuts down with the "de-init" hooks as part of `exitDatabase()`. With older Base look, for two at exit handlers `pvxsExitBeforeIocShutdown()` and `pvxsExitAfterIocShutdown()`. Both should happen before any of the filter Shutdown handlers.
> 
> The `pvxsExitBeforeIocShutdown()` handler fully stops the PVXS server, which should stop all worker threads (and join() them properly).
> 
> In short, the ordering required by `dbChannel` should be respected.

What you mean is that all channels stays on until end of IOC? I'm failing to understand the relationship of this and the issue in question. I maybe be missing some context.

mdavidsaver:
> Is there a way to pin point that a subscription is connected with a specific TX/RX thread?

With pvAccessCPP, the TX/RX thread names include server IP address.


> ... this fix the use after free issue on array filters if the event queue gets to this state.

This fixes one specific case, there will be others.  The other filters follow the same pattern.  see `dbnd.c`.

I understand that the semantics of the dbChannel/dbEvent API are poorly documented.  And pvAccessCPP/pva2pva can violate these semantics.  However, retroactively trying to redefine these semantics does not seem to me like the answer.  (and would anyway be a large project)



> What you mean is that all channels stays on until end of IOC? I'm failing to understand the relationship of this and the issue in question. I maybe be missing some context.

In brief, PVXS/QSRV2 should always join worker threads and free resources in the correct order.

lucmaga:
> I understand that the semantics of the dbChannel/dbEvent API are poorly documented. And pvAccessCPP/pva2pva can violate these semantics. However, retroactively trying to redefine these semantics does not seem to me like the answer. (and would anyway be a large project)

Yes, I can see that. Maybe a good start would be to document this semantics. I would be willing to help with that.

> This fixes one specific case, there will be others. The other filters follow the same pattern. see dbnd.c.

Sure, this fixes only for array filters, but other filters do not have the same issue. The issue only occurs there because the internal `myStruct` has a free List inside that is used by `dtor`, in this case `freeArray`.  To summarize, `dtor` access `pvt` after it been freed.

The filters `dbnd.c`, `decimate.c`, `sync.c` and `utag.c` do not use a `dtor`. Only `ts.c` sets a `dtor` in some cases, besides `arr.c`, but those functions never access `pvt` struct. In the case of `ts.c` all filters are global variables, they are only destroyed on shutdown.

simon-ess:
Some comments after a discussion with @lucmaga :
* I do agree with @mdavidsaver that this does not seem like the best place to fix this issue; it is a whack-a-mole game to fix this one particular filter.
* A better idea we had was that it might be possible to move this "up one level" so that the filter code more generically can handle some reference counting and cleanup
* An additional concern with adding reference counting here: since this is trying to deal with the issue of data races and concurrency, we likely need to have some guard mechanism on the reference counts to make sure that we don't end up with a completely new set of problems here.

In general, it does seem like switching to PVXS would fix this specific issue. But @lucmaga 's point stands that there is ultimately some problem in the database API that allows for this behaviour. As such, until such time as one can provide a more general fix, it seems worth trying to future-proof and fix the (generic) filter code as well to avoid this case as well as any other future filters from running into similar issues.

lucmaga:
Just to close up the conversation @simon-ess, @mdavidsaver and @ralphlange. This idea, or anything alike will not be acceptable right? 
From the conversation, the only acceptable solution would be a rework of the dbChannel/dbEvent API.

anjohnson:
The preference concluded here was to fix the problem for all the filters. We already have an issue (#643) referring back to this PR so it shouldn't be completely forgotten if it's closed, although (assuming the changes here are sufficient and effective) I could see an argument for temporarily applying this and similar fixes to all the included filters until such time that we fix the higher-level problem.

mdavidsaver:
> The preference concluded here was to fix the problem for all the filters.

Not quite.  I do not see an issue with the filter code.  Rather, the APIs for making use of them are poorly documented, with known or suspected ordering violations by pvAccessCPP/pva2pva.  To my mind, these violations are the core problem to be fixed.  Or (my preference) we deprecate them in favor of PVXS.

mdavidsaver:
> ... I could see an argument for temporarily applying this and similar fixes to all the included filters until such time that we fix the higher-level problem.

I disagree.  My opinion might be changed if this sort of bug could be demonstrated with QSRV2.  Absent this, what I see is at attempt to patch Base to workaround a bug specific to QSRV1.

lucmaga:
An ordering issue should not lead to a use-after-free IMHO. It should at least be reliable that a malicious client could not exploit this.

lucmaga:
I can't see this going any further. I'm closing it for now.

PR #638, opened 2025-04-17
fix function pointer type according to new typed_dset

dirk-zimoch:
Broke in Fedora Latest with gcc 15.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -3456,7 +3456,7 @@ void  dbDumpDevice(DBBASE *pdbbase,const char *recordTypeName)
                     " - get_ioint_info()"
                 };
                 int i, n = pdevSup->pdset->number;
-                DEVSUPFUN *pfunc = (DEVSUPFUN*) &pdevSup->pdset->report;
+                long (**pfunc)(int) = &pdevSup->pdset->report;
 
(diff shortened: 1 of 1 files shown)


anjohnson:
Your version fails on many more targets than [mine did](https://github.com/anjohnson/epics-base/actions/runs/14518774891/job/40734270393). I just committed a fix for the dbLinkdset.c build, now waiting for that to build. If that succeeds I will probably push it directly without making a PR.

dirk-zimoch:
> Your version fails on many more targets than [mine did](https://github.com/anjohnson/epics-base/actions/runs/14518774891/job/40734270393).

Because mine was buggy ;-)

I still think that we should change the `pfunc` pointer type according to the new `typed_dset` instead of simply casting to the old `(DEVSUPFUN*)`. There was a reason to update to `typed_dset`, wasn't it?

mdavidsaver:
Builds with GCC 15 are succeeding as of a85ef1ad1c6ef2f7ad0b137f4580d117f36c2a26.  Is this PR still relevant?

anjohnson:
The last change left here is arguably a point of finesse. Given that the function pointer is only cast here so it can be printed its exact data type has no effect on the result. Thus the question to answer here should be:
- Is the use of and cast to a `DEVSUPFUN *` type easier to read than remembering that a `dset::report` is a "pointer to function taking int and returning long" which doesn't need a cast?

anjohnson:
Core Group: Not bothered.

PR #644, opened 2025-04-28
More colour, better error/warning messages

anjohnson:
Various changes to iocsh.cpp and others to add and improve the use of color.

- The IOC Shell output when running a startup script now displays comment lines in Cyan,
and Bold for command lines.
- The `softIoc -v` output also uses the same color scheme for the commands it shows.
- The default IOC Shell prompt is now green, but the choice of color can be edited in the
`configure/CONFIG_SITE_ENV` file – the default `ENV_PARAM` values can use the `ANSI_*()`
macros to select the text color.
- More error messages printed by the shell appear in Red, or use the red `ERROR` prefix.
- The word "Illegal" has been replaced with "Invalid" in several iocsh error messages.
- The iocsh `var` command now shows the data type of the registered variable.

image

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -353,17 +353,17 @@ struct Tokenize {
 
         if (redirect != NULL) {
             if(noise)
-                showError(filename, lineno, "Illegal redirection.");
+                showError(filename, lineno, ANSI_RED("Invalid redirection."));
             return true;
         }
         if (quote) {
             if(noise)
-                showError(filename, lineno, "Unbalanced quote.");
+                showError(filename, lineno, ANSI_RED("Unbalanced quote."));
             return true;
         }
         if (backslash) {
             if(noise)
-                showError(filename, lineno, "Trailing backslash.");
+                showError(filename, lineno, ANSI_RED("Trailing backslash."));
             return true;
         }
 
@@ -384,8 +384,8 @@ struct Tokenize {
             redirect->fp = fopen(redirect->name, redirect->mode);
             if (redirect->fp == NULL) {
                 int err = errno;
-                showError(filename, lineno, "Can't open \"%s\": %s.",
-                                            redirect->name, strerror(err));
+                showError(filename, lineno, ANSI_RED("Can't open '%s': %s"),
+                    redirect->name, strerror(err));
                 redirect->name = NULL;
                 // caller will clear tok.redirects
                 return -1;
@@ -814,13 +814,14 @@ cvtArg (const char *filename, int lineno, char *arg, iocshArgBuf *argBuf,
                 errno = 0;
                 argBuf->ival = strtoul (arg, &endp, 0);
                 if (errno == ERANGE) {
-                    showError(filename, lineno, "Integer '%s' out of range",
-                        arg);
+                    showError(filename, lineno,
+                        ANSI_RED("Integer '%s' out of range."), arg);
                     return 0;
                 }
             }
             if (*endp) {
-                showError(filename, lineno, "Illegal integer '%s'", arg);
+                showError(filename, lineno,
+                    ANSI_RED("Invalid integer '%s'."), arg);
                 return 0;
             }
         }
@@ -833,7 +834,8 @@ cvtArg (const char *filename, int lineno, char *arg, iocshArgBuf *argBuf,
         if (arg && *arg) {
             argBuf->dval = epicsStrtod (arg, &endp);
             if (*endp) {
-                showError(filename, lineno, "Illegal double '%s'", arg);
+                showError(filename, lineno,
+                    ANSI_RED("Invalid double '%s'."), arg);
                 return 0;
             }
         }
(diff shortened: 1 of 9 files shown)


modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -927,9 +930,9 @@ static void helpCallFunc(const iocshArgBuf *args)
             fputc('\n', epicsGetStdout());
         iocshTableUnlock ();
 
-        fprintf(epicsGetStdout(),
-                "\n"
-                "Type 'help <command>' to see the arguments of <command>.  eg. 'help db*'\n");
+        fprintf(epicsGetStdout(), "\n"
+            "Type 'help <command>' to see the arguments of <command>, "
+            "eg. 'help db*'\n");
```
mdavidsaver: A minor point, but isn't it preferable for readability for a string constant with two newlines to split across 3 source lines?

mdavidsaver (approved):
I like it.

simon-ess:
Is it possible to configure this at all? It seems that at the moment the only configurable portion is the `PS1`; however we are stuck with the default coloration/highlighting for all other output.

(But I agree with @mdavidsaver  - I think that this is a nice addition)

mdavidsaver:
Codacy flags a "Variable 'line' was passed to a function after being freed.", which imo. is an obvious false positive.

modules/libcom/src/iocsh/iocsh.cpp
anjohnson: I was trying to use an 80 column text-wrap width, but given the `ANSI_UNDERLINE()` string below it that's probably too pedantic here. I will fix.

anjohnson:
@mdavidsaver Thanks for confirming my thoughts about Codacy's report.

anjohnson:
I wondered if configurability would come up. This may show one reason it's desirable, although it doesn't seem quite as bad as I thought it might be:

image

Generalization of the `verbose_out()` function I added to `softIoc/softMain.cpp` might be a way to do that, it takes an enum value named after the type of text to be printed and looks up the ANSI color code for that in a table. Doing that across the code-base might be nice, but there are rather a lot of different kinds of ways that we provide and print messages, and I could see it being complicated to handle them all.

The `ANSI_ESC_` strings are mostly expanded out at compile-time, so they only add a few characters to the output and no additional levels of indirection/function calls.

mdavidsaver:
> I wondered if configurability would come up.

imo. schema control is a solid nice to have going forward, provided it does not add too much complexity.  I certainly do not see it as a blocker.

I see a reasonable contrast with both light on dark as well as dark on light color schemes as sufficient for this PR.  (which I do in fact see)

For #113 I did spend some time on contrast when selecting the colors.  Although I must admit that `ANSI_CYAN` is the hardest to read on a light background.  This is why I have PVXS use the darker `ANSI_BLUE` for info level log messages.

modules/libcom/src/iocsh/iocsh.cpp
anjohnson: The original wording doesn't mention glob patterns at all, and my fix for that grew to a second line anyway:
```patch
         fprintf(epicsGetStdout(), "\n"
-            "Type 'help <command>' to see the arguments of <command>, "
-            "eg. 'help db*'\n");
+            "Type 'help <glob>' for information about commands matching\n"
+            "the name or pattern <glob>, e.g. 'help db*'\n");
```

New commits coming soon.

anjohnson:
This image shows the use of `ANSI_BLUE` instead of `ANSI_CYAN` for comments:

image

I changed the code to use that instead, and added a release note. I think this PR is now ready (although Codacy still disagrees).

anjohnson:
Core Group: Configurability is desirable, but not essential to begin with (later PR). Use logical names for the things we configure, maybe re-use GNU's `LS_COLORS` and `GREP_COLORS` scheme. Consider removing the change to allow `ANSI_*` in `CONFIG_SITE_ENV`.

mdavidsaver:
A couple of thoughts...

wrt. scripts trying to parse stdout/stderr.  I think mainly tester scripts looking for the prompt string.  Here anything we do to change the default prompt will break a tester which assumes the current default.  I guess in this case our statement is that tester scripts should set `$IOCSH_PS1`?

I also find myself wondering if the `funopen()` / `fopencookie()` calls (cf. https://github.com/epics-base/epics-base/pull/248) might be useful as vectors for conditionally striping out escape sequences from stdout/err ?

anjohnson:
Rebased and made some final code adjustments.

In light of @mdavidsaver's comment above about test scripts setting `$(IOCSH_PS1)` (yes, they should always do that if they match the IOCSH prompt string to find the end of a command) I realize that trying to remove the color macros such as `ANSI_GREEN(…)` from the `configure/CONFIG_SITE_ENV` default value could break those scripts.

I have tightened up the bldEnvData.pl script to syntax-check the default values while still allowing `ANSI_ENV_<color>` and `ANSI_<color>("<string>")` macros in those default values; syntax errors in the parameter values should now halt the build at that script, not at the subsequent compilation of the envData.c file.

mdavidsaver:
With this PR, there are a lot more ANSI escapes going to stdout/err.  So much so that a capture via `tee` is hard to parse by eye.  However, I find that `less -R` will decode the color escapes nicely.

```sh
$ softIoc some.cmd 2>&1 | tee log
...

$ less -R log
```

anjohnson:
@mdavidsaver Typo, `less -R log`?

I have `$LESS` set to `-R -F -X`, although I rarely send IOC output to a log file. Is something like this worth adding to the Release Notes?

mdavidsaver:
> Typo, `less -R log`?

corrected.

PR #645, opened 2025-04-29
Fix failing windows conda builds

simon-ess:
Certain windows build environments (see
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1234387&view=results for an example) fail unless the echo command is properly quoted. This seems to be due to how Make handles various commands on non-posix systems; without the quotes it directly tries to run
```
echo foo bar
```
inside cmd.exe, which fails as echo is not a valid command in that context.

### DIFF
--- a/configure/RULES_MODULES
+++ b/configure/RULES_MODULES
@@ -41,7 +41,7 @@ RELEASE.host: $(RELEASE_LOCAL)
 
 $(RELEASE_LOCAL): Makefile $(CONFIG)/CONFIG_SITE \
     $(wildcard $(CONFIG)/CONFIG_SITE.local)
-	$(ECHO) Creating $@ with
+	$(ECHO) "Creating $@ with"
 	$(ECHO) "    $(PARENT_MODULE) = $(INSTALL_ABSOLUTE)"
 	@echo $(PARENT_MODULE) = $(INSTALL_ABSOLUTE)> $@
 


ralphlange:
Have we actually seen such failures in our builds?

ralphlange:
Also...
image

simon-ess:
I'm unaware of it in _our_ builds, but it definitely is failing in whatever environment is used to generate the conda builds.

Note that this is the only line that does an "echo" without quotes, so this change at least makes things more consistent as well.

ralphlange:
Looking at the error log you pointed to, it seems like make is trying to run
`echo Creating RELEASE.windows-x64.local with`
as an executable, which does not exist.

I.e., *not* calling `cmd.exe` is the problem. If your fix works, it looks like that make is scanning the command for quotes and switches to calling `cmd.exe` if there are.

That's a behaviour that I have not seen on Windows, neither under cmd nor Powershell nor MSYS2 nor Cygwin nor my QTCreator IDE. Seems to be really specific to that build.

"Consistency" sells.

ralphlange (approved):
Yeah, whatever.

simon-ess:
I might be slightly mis-diagnosing the exact issue, but we got a bit more information from cranking up the debug information and noting that the line without the quotes is being treated quite differently by make than those lines that do have quotes. Specifically, for lines that did have quotes we saw
```
2025-04-28T06:54:19.7120213Z echo "Installing ../../../../../_h_env/epics/templates/makeBaseExt/top/simpleExt/Makefile"
2025-04-28T06:54:19.7121815Z CreateProcess(NULL,C:/Program Files/Git/bin/sh.exe -c "echo \"Installing ../../../../../_h_env/epics/templates/makeBaseExt/top/simpleExt/Makefile\"",...)
```
whereas this one line without yielded
```
2025-04-28T06:54:25.2723084Z echo Creating RELEASE.windows-x64.local with
2025-04-28T06:54:25.2723697Z CreateProcess(NULL,echo Creating RELEASE.windows-x64.local with,...)
```
To be honest we don't fully know why this is behaving the way it is; from what I have seen looking through make's WIN32 code it _seems_ like the quotes should not yield different behaviour. But clearly in _some_ setups it does.

ralphlange:
More or less consistent with what I assumed.
It even calls `bash` from the Git installation instead of `cmd.exe` - just to make things less predictable.

FreddieAkeroyd:
Hmm .. this has reminded me of a local change I made to our system

https://github.com/ISISComputingGroup/EPICS-base/blob/69ef4134446ec2da9b46fc2e956c2b426d87a575/configure/os/CONFIG.win32-x86.Common#L30

PR #647, opened 2025-04-29
setBootConfigFromNVRAM.c: Fix warning due to discarded qualifiers & mismatched type

JJL772:


mdavidsaver (commented):
This looks like a straightforward change to pacify a compiler warning/error.

It does make me think that this code should be converted to use `epicsMMIO.h` accessors, or some RTEMS equivalents.

PR #648, opened 2025-05-05
RTEMS: Exclude BSD and legacy networking components from GeSys objects

JJL772:
Exclude some additional RTEMS components from GeSys objects.

While GeSys and Cexpsh do not (currently) support RTEMS 5+, GeSys-style objects are still useful with the RTEMS runtime linker.


mdavidsaver:
This seems straightforward enough, and I think unlikely to have side effects.

anjohnson (approved):
Are you guys are still using GeSys and CexpSh?

I have no objection to this change, nothing else should be using these Makefile variables.

JJL772:
For RTEMS 4.X we are still using it, and we intend to keep using it for RT6+. I'm planning to replace the Cexpsh loader with the RTEMS RTL, which can load these GeSys objects too.

anjohnson:
Will you be using the IOCSH to load st.cmd files, or will that still use the Cexp shell and require parentheses `()` around command arguments? I was hoping to drop them from our documentation since only Cexp users ever actually required them.

JJL772:
afaik all of our IOCSH scripts for RTEMS are executed by Cexp. It's probably OK to drop the parentheses from the docs though.

PR #649, opened 2025-05-05
Fix compile error on RTEMS-uC5282 w/RTEMS 6

JJL772:
This is a hacky workaround for an issue in newlib. _LDBL_EQ_DBL is normally defined by newlib.h, but its value is determined at configure time. The default -mcpu setting for m68k-rtems6-gcc supports long double (where sizeof long dobule == 12), while the Coldfire ISAs do not (sizeof long double == 8). Since newlib.h is shared for all m68k targets, this breaks some targets w/o long double support.

So, when building with -mcpu=5282, for example, _LDBL_EQ_DBL is not defined when it should be. ieefp.h has a sanity check that fails in this case, meanwhile libstdc++ completely explodes.

I'm kind of surprised this hasn't been an issue with other targets supported by GCC...

### DIFF
--- a/configure/os/CONFIG.Common.RTEMS-uC5282
+++ b/configure/os/CONFIG.Common.RTEMS-uC5282
@@ -13,6 +13,12 @@ ARCH_DEP_CFLAGS += -DMY_DO_BOOTP=NULL
 ARCH_DEP_CXXFLAGS_5 = -std=c++98
 ARCH_DEP_CXXFLAGS += $(ARCH_DEP_CXXFLAGS_$(RTEMS_VERSION))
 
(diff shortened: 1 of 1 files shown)


mdavidsaver:
@JJL772 Can you add a reference to an RTEMS or newlib bug report?  This would provide traceability on the issue, so that we could know when this workaround could be removed.

JJL772:
I (again) forgot to open the bug report until now... https://sourceware.org/bugzilla/show_bug.cgi?id=33750

Also added a reference in the comment.

anjohnson:
Core Group: There is an RTEMS MR that has been [merged for 6.3](https://gitlab.rtems.org/rtems/rtos/rtems/-/merge_requests/951) and to the [7 branch](https://gitlab.rtems.org/rtems/rtos/rtems/-/merge_requests/937) which resolves this. Close this after the 6.3 release.

PR #650, opened 2025-05-16
Silence some error log messages during test runs

simon-ess:
Several of the record tests do not silence output correctly, yielding e.g.
```
epicsExportTest.t ....... 1/31 Starting iocInit
```
instead of (e.g.)
```
simmTest.t .............. ok         
mbbioDirectTest.t ....... ok       
asyncproctest.t ......... ok     
subproctest.t ........... ok   
linkFilterTest.t ........ ok
```


anjohnson (approved):
Yay, thanks! This approach doesn't always work if the messages aren't printed using errlog, but we should use it it does.

PR #651, opened 2025-05-17
Record reference updates

anjohnson:
Both calc and calout record type docs were missing operators, and the event record has never
been modified to cover named events.

Also improved the formatting in postfix.h.

### DIFF
--- a/documentation/new-notes/PR-651.md
+++ b/documentation/new-notes/PR-651.md
@@ -1,0 +1,12 @@
+### Documentation Updates
+
+The reference documentation for the [event](eventRecord.html) record type
+has been updated to cover the use of named events which were added in Base
+3.14.12.3 and 3.15.1.
+
+Documentation for CALC expression evaluation has been updated for format
+enhancements and to add some missing operators.
+The best documentation for these expressions can be found in the
+[postfix.h](postfix_h.html) header in libCom, but both the
+[calc](calcRecord.html) and [calcout](calcoutRecord.html) record reference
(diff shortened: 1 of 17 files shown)


anjohnson:
Fixed some record-ref links to the database-concepts document which moved at some point. The server-side redirect which was set up resolved the file move, but can't handle references to specific document sections.

This is probably all I'm going to do in this PR, so marking it ready to review.

ttkorhonen:
This looks good as far as I can tell; a lot of important updates. In a further PR, I would change the record-ref link to use Sphinx interlinks, those are more robust against changes. That is better done separately, no need to change in this PR, as we once agreed to move the record-ref to the Base repo where the other parts already are.

anjohnson:
@ttkorhonen thanks.

I don't know if we can use Sphinks interlinks inside a `.dbd.pod` file yet, but we should have the ability to define new POD link types in our EPICS::PodMD Perl modules. I think that would let us translate something like `L<description|sphinx:getting-started/EPICS_Intro>` in a `.dbd.pod` file into whatever syntax Sphinx needs to see in the generated Markdown file.

I agree that's a separate project and PR though.

ttkorhonen:
@anjohnson , my comment above was wrong and in fact not what I actually was planning. The plan was rather to move the database-concept page to EPICS Base repo, and then use interlinks to refer to it from outside (i.e., docs repo). This way the link in pod files can stay local.
I just have not had time yet to make the corresponding PRs.

PR #653, opened 2025-05-19
rtems: Add interrupt lock support for RTEMS SMP builds

kiwichris:
This change lets EPICS build for RTEMS SMP BSPs. I have tested it on a Microzed with SMP and the `libComTestHarness` passed.

### DIFF
--- a/modules/libcom/src/osi/os/RTEMS/osdInterrupt.c
+++ b/modules/libcom/src/osi/os/RTEMS/osdInterrupt.c
@@ -18,25 +18,56 @@
 #include "epicsInterrupt.h"
 #include "epicsThread.h"
 
+#if defined(RTEMS_SMP)
+#include <rtems/sysinit.h>
+#include <epicsSpin.h>
+#endif /* RTEMS_SMP */
+
 #define INTERRUPT_CONTEXT_MESSAGE_QUEUE_COUNT    100
 
 static rtems_id interruptContextMessageQueue;
 
+/*
+ * SMP does not support interrupt disable and enable because
+ * interrupts are local to a processor. RTEMS provides locks to use
+ * with SMP however they do not fit the existing EPICS API as there is
+ * no way to hold and return a lock context.
+ *
+ * Use an EPICS spinlock and create it before the POSIX init thread
+ * runs.
+ */
+#if defined(RTEMS_SMP)
+static epicsSpinId isr_lock;
+static void create_isr_lock(void) {
+    isr_lock = epicsSpinMustCreate();
+}
+RTEMS_SYSINIT_ITEM(
+    create_isr_lock, RTEMS_SYSINIT_POSIX_USER_THREADS, RTEMS_SYSINIT_ORDER_FIRST);
+#endif /* RTEMS_SMP */
+
 int
 epicsInterruptLock (void)
 {
+#if defined(RTEMS_SMP)
+    epicsSpinLock (isr_lock);
+    return 0;
+#else  /* RTEMS_SMP */
     rtems_interrupt_level level;
-
     rtems_interrupt_disable (level);
     return level;
+#endif /* RTEMS_SMP */
 }
 
 void
 epicsInterruptUnlock (int key)
 {
+#if defined(RTEMS_SMP)
+    (void) key;
+    epicsSpinUnlock (isr_lock);
+#else /* RTEMS_SMP */
     rtems_interrupt_level level = key;
-
     rtems_interrupt_enable (level);
+#endif /* RTEMS_SMP */
 }
 
 int


modules/libcom/src/osi/os/RTEMS/osdInterrupt.c
```diff
@@ -18,25 +18,49 @@
...
+/*
+ * SMP does not support interrupt disable and enable because
+ * interrupts are local to a processor. RTEMS provides locks to use
+ * with SMP however they do not fit the existing EPICS API as there is
+ * no way to hold and return a lock context. The default EPCIS
+ * implementaton uses a mutex so lets use a Self-Contained mutex
+ * (faster) for SMP.
+ */
+#if defined(RTEMS_SMP)
+static rtems_mutex isr_lock = RTEMS_MUTEX_INITIALIZER("epics.isr");
```
mdavidsaver: @kiwichris Can you confirm that `rtems_mutex` is safe in interrupt context?

> ... RTEMS provides locks to use with SMP ...

For future reference, what are these alternative APIs?

kiwichris: You cannot lock a mutex within an interrupt handler context. I borrowed the idea from the EPICS Unix ports so I gather that would break existing drivers on EPICS that are playing with the interrupt mask in interrupts? Maybe [POSIX spinlocks](https://docs.rtems.org/docs/main/c-user/symmetric_multiprocessing_services.html#disabling-of-interrupts) would be a better choice? The EPICS API only allows saving the interrupt context as an `int` so the `rtems_interrupt_lock_context` cannot be used.

For the record calls such as [rtems-interrupt-disable()](https://docs.rtems.org/docs/main/c-user/interrupt/directives.html#rtems-interrupt-disable) are [not available in SMP](https://docs.rtems.org/docs/main/c-user/symmetric_multiprocessing_services.html#interrupt-service-routines-execute-in-parallel-with-threads). A disable would only be local to a core so the calls were removed from the API.

kiwichris: For the record this is the [Self-Contained API](https://github.com/epics-base/epics-base/pull/653#discussion_r2095927906) in RTEMS.

mdavidsaver: > You cannot lock a mutex within an interrupt handler context.

Thus my question since `epicsInterruptLock()` must be callable from an ISR.

> Maybe [POSIX spinlocks](https://docs.rtems.org/docs/main/c-user/symmetric_multiprocessing_services.html#disabling-of-interrupts) would be a better choice?

Have a look at `epicsSpin.h` and `posix/osdSpin.c`.

kiwichris: > > You cannot lock a mutex within an interrupt handler context.
> 
> Thus my question since `epicsInterruptLock()` must be callable from an ISR.
> 

Yes and it is valid. I reviewed the [default](https://github.com/epics-base/epics-base/blob/7.0/modules/libcom/src/osi/os/default/osdInterrupt.c) implementation and choose to follow that path to at least make the API consistent with other hosts that support SMP. I assume the default implementation' disable should not be called from an ISR?

RTEMS has a lock that can be used as a replacement on SMP but it has a context that is not compatible with the `int` returned from the disable call.

> > Maybe [POSIX spinlocks](https://docs.rtems.org/docs/main/c-user/symmetric_multiprocessing_services.html#disabling-of-interrupts) would be a better choice?
> 
> Have a look at `epicsSpin.h` and `posix/osdSpin.c`.

Thanks and nice.

Is it valid to say this API is only valid on an RTOS like RTEMS with a single core? Should the RTEMS SMP build generate a fatal error?

Any driver using interrupts ported from a single core board to an SMP one should be reviewed and updated to make sure it is OK on an SMP system.

mdavidsaver: > I assume the default implementation' disable should not be called from an ISR?

On the targets where this default would be used (freebsd, linux, osx, and windows) running from an ISR is not possible.

> Is it valid to say this API is only valid on an RTOS like RTEMS with a single core?

IMO, no.  I know of at least [one](https://github.com/search?q=repo%3Aepics-modules%2Fmrfioc2+epicsInterruptLock&type=code) EPICS driver (which I wrote) which uses the epicsInterrupt abstraction to support both RTEMS and Linux+UIO targets for a family of PCI devices.  In this case replacement on RTEMS with a global spinlock would work (since this is effectively what happens on Linux SMP).

kiwichris: > > I assume the default implementation' disable should not be called from an ISR?
> 
> On the targets where this default would be used (freebsd, linux, osx, and windows) running from an ISR is not possible.

On RTEMS with SMP we recommend your ISRs dispatch a thread as soon as possible and do its work there (if possible). It simplifies the interrupt handling and management and lets ISRs (running as threads) use available cores.

> > Is it valid to say this API is only valid on an RTOS like RTEMS with a single core?
> 
> IMO, no. I know of at least [one](https://github.com/search?q=repo%3Aepics-modules%2Fmrfioc2+epicsInterruptLock&type=code) EPICS driver (which I wrote) which uses the epicsInterrupt abstraction to support both RTEMS and Linux+UIO targets for a family of PCI devices. In this case replacement on RTEMS with a global spinlock would work (since this is effectively what happens on Linux SMP).

Nice. I will implement it with a spinlock. It is a great idea, so thanks.

kiwichris:
I ran the tests with the latest change on an SMP Zynq Microzed. The results are:
```
    EPICS Test Harness Results
    ==========================

All tests successful.
Programs=38, Tests=3161, 251 wallclock secs
```
I also updated my NTP tester with the following code:
```
    int k = epicsInterruptLock();
    epicsInterruptUnlock(k);
```

modules/libcom/src/osi/os/RTEMS/osdInterrupt.c
```diff
@@ -18,25 +18,56 @@
...
+ * interrupts are local to a processor. RTEMS provides locks to use
+ * with SMP however they do not fit the existing EPICS API as there is
+ * no way to hold and return a lock context.
+ *
+ * Use an EPICS spinlock and create it before the POSIX init thread
+ * runs.
+ */
+#if defined(RTEMS_SMP)
+static epicsSpinId isr_lock;
+void create_isr_lock(void) {
```
mdavidsaver: Can `create_isr_lock()` be `static`?

mdavidsaver (commented):
> I also updated my NTP tester with the following code:

Can you test a call from interrupt context?  (maybe a timer ISR?)

modules/libcom/src/osi/os/RTEMS/osdInterrupt.c
kiwichris: Yes. Updated. Thanks.

kiwichris:
> > I also updated my NTP tester with the following code:
> 
> Can you test a call from interrupt context? (maybe a timer ISR?)

The test I did was to make sure the spinlock was created. It was. RTEMS has tests for `SYSINIT`.

The implementation is based on the EPICS spinlock and I have assumed that and the POSIX spinlock tests should cover that for the use case here. The RTEMS test is here:

https://gitlab.rtems.org/rtems/rtos/rtems/-/blob/main/testsuites/psxtests/psxspin01/test.c?ref_type=heads

anjohnson:
Core Group: Implementation looks good, can we confirm (or enforce) that on RTEMS SMP osdSpin.c will always use the Posix spin-lock and not the mutex implementation?

mdavidsaver:
https://github.com/epics-base/epics-base/blob/e4fbd83c93a68a1d9cb863ff341735388d246347/modules/libcom/src/osi/os/posix/osdSpin.c#L24-L33

Would now need to ensure that `USE_SPIN` is defined in osdSpin.c so that the actual spin lock implementation is used with RTEMS 5+.

kiwichris:
> Would now need to ensure that `USE_SPIN` is defined in osdSpin.c so that the actual spin lock implementation is used with RTEMS 5+.

@mdavidsaver what do you mean by "actual implementation"?

mdavidsaver:
`posix/osdSpin.c` contains two implementations of the `epicsSpin*()` calls.  One using `pthread_spinlock_t`, and one using `pthread_mutex_t`.  As described in the quoted comment, the current logic will select the mutex implementation when RT scheduling is enabled.  While this is a reasonable choice for RT Linux to avoid PI, it is not for RTEMS where spin locks must be callable from an ISR.  So the quoted macro logic will need to be inverted (and should have been before now).  Assuming RTEMS defines the `_POSIX_SPIN_LOCKS` feature test macro, then the condition might be changed to:

```c
#if defined(DONT_USE_POSIX_THREAD_PRIORITY_SCHEDULING)  || defined(__rtems__)
```

anjohnson:
Core Group: Clarifying, we believe (as described in @mdavidsaver's last comment) this would do the wrong thing on RTEMS SMP targets.

kiwichris:
> Core Group: Clarifying, we believe (as described in @mdavidsaver's last comment) this would do the wrong thing on RTEMS SMP targets.

Noted and thanks. I will take a closer look at this.

anjohnson:
Core Group: Ping @kiwichris

kiwichris:
> Core Group: Ping @kiwichris

Work on RTEMS SMP on EPICS is currently unfunded. I did not this when I needed `libdebugger` running. It is in the queue of things to look into but it is a long list

anjohnson:
Okay, I'm going to mark this PR as Draft for now. Thanks!

PR #654, opened 2025-05-19
Record documentation syntax improvement

dirk-zimoch:
1. Make menu and field tables in generated .md files more readable by padding.
2. Improve readability of fields tables with repetitive fields in MD and HTML.

Some records have many fields of the same general pattern like `INPA` - `INPL`. Listing them all generates a repetitive table which takes a lot of space without adding much information.

Instead allow a more terse representation using `=fields INPA - INPL` which will render the rows for `INPA` and `INPL` normally and in between have a row with only a vertical ellipsis like this:
```
| ⋮                                                                 |
```


anjohnson:
While the RTD build succeeded, the output indicates there are POD errors shown in the generated [ComponentReference.html](https://epics--654.org.readthedocs.build/projects/base/en/654/ComponentReference.html) file:
image
This is from the end of the aSub reference doc:
image
I think you will need to use the POD escape mechanism to insert that character, try using `\E<VERTICAL ELLIPSIS>` to generate that.

dirk-zimoch:
The pages render correctly. (Almost. Except for columnspan.) However RTD complains that the encoding is not set (to something like utf-8, I guess) and thus assumes good old 7 bit ASCII. I am not sure if using `\E<VERTICAL ELLIPSIS>` will help. (I used `\x{22EE}`.)
I think I should rather try to add `=encoding` tags.

dirk-zimoch:
`=encoding utf8`
According to https://perldoc.perl.org/perlpod#Command-Paragraph

dirk-zimoch:
Hmm... The last commit was not really an improvement. I think I need help here.

dirk-zimoch:
Looks better now.

simon-ess:
Better, but see https://epics--654.org.readthedocs.build/projects/base/en/654/calcoutRecord.html - the `=encoding` has been passed on to the page.

modules/database/src/std/rec/calcRecord.dbd.pod
```diff
@@ -45,7 +45,7 @@ See L<Address
 Specification|https://docs.epics-controls.org/en/latest/guides/EPICS_Process_Database_Concepts.html#address-specification>
 for information on how to specify database links.
 
-=fields INPA, INPB, INPC, INPD, INPE, INPF, INPG, INPH, INPI, INPJ, INPK, INPL
+=fields INPA -INPL
```
simon-ess: A bit inconsistent spacing: `INPA -INPL` vs `ONVA - ONVU`

dirk-zimoch: And both work 😀

dirk-zimoch:
> the `=encoding` has been passed on to the page.

$#!+

No idea how to solve it then.

dirk-zimoch:
Additional blank lines around `=encoding` may help. Will try tomorrow.

anjohnson:
Did you try using `\E<VERTICAL ELLIPSIS>` inside the generated POD or `\N{VERTICAL ELLIPSIS}` in an HTML string yet (nicer than hex codes)? Perl does know about unicode character names, although you do need to tell it the output is unicode to avoid a warning (that's what those `-CSD` flags are for in the EPICS build system's `$(PERL)` variable):
```
MacOS$ perl -CSD -e 'print "ve = \N{VERTICAL ELLIPSIS}\n"'
ve = ⋮
MacOS$ perl -e 'print "ve = \N{VERTICAL ELLIPSIS}\n"'
Wide character in print at -e line 1.
ve = ⋮
```

anjohnson:
My bad, the names inside the POD `E<...>` (and there's no leading back-slash either) are HTML Entity names, so `E<hellip>` works in POD for a horizontal ellipsis `…`, but there's no HTML `&vellip;` equivalent AFAICS. However since you're using quoted strings inside Perl code the `\N{VERTICAL ELLIPSIS}` should work there.

dirk-zimoch:
I tried `\E<VERTICAL ELLIPSIS>` and it copied that literally into the .md file. Ugly, so I changed back. I have not tried `\N{VERTICAL ELLIPSIS}`. Nor did I know it exits. I am not familiar with perl. Or POD. Or RTD. All my changes so far are pure guesswork and only as good as my google-fu trying to hack together some working code 😏

dirk-zimoch:
If I can‘t get it working tomorrow, I will simply go for `...`

dirk-zimoch:
Extra newlines around `=encoding utf8` help.
And `\N{VERTICAL ELLIPSIS}` works.

modules/database/src/std/rec/calcRecord.dbd.pod
dirk-zimoch: Fixed.

anjohnson (approved):
Looks good to me.

I'd need to sit down with the Perl doc's to completely understand those new `split` regexes though!

dirk-zimoch:
> completely understand those new `split` regexes

Split at comma surrounded by optional spaces, as before, or at any (including none) space before or after a `-`, leaving the `-` itself in place (unlike the comma) to become an element of `@names`. This splits `A - Z` into 3 names `A`, `-`, `Z` regardless of the amount of space around the `-`.

anjohnson:
Core Group: Approved.

PR #624, opened 2025-04-04
ACF syntax forward compatibility

george-mcintyre:
### Description

This pull request introduces future-proofing for ACF parsing by allowing files to include unsupported sections without raising errors.  As long as the file is formatted according to the rules it will parse correctly.  
- Allowed and ignored:
  - RULE containing unsupported blocks (RULE ignored)
  - RULE with unsupported permissions (RULE ignored)
  - Top level items named other than supported keywords (may include optional parameter lists, blocks))
  - Sub items in ASG groups named other than supported ASG keywords (may include optional parameter lists, blocks)
  - Sub items in RULE body (may include optional parameter lists, blocks)
- Disallowed with error
  - Sub items in HAG list
  - Sub items in UAG list
  - non standard CALC
  - RULE flags can only be `TRAPWRITE`, or `NOTRAPWRITE`
 
### Checklist

- [x] Ensure backward compatibility.
- [x] Verify functionality with different ACF file examples.

PR #655, opened 2025-05-23
Extend calc to inputs A-U

dirk-zimoch:
Extend calc, calcout, and sub records to accept inputs A-U (just like aSub).
This also allows other applications which are using the calc API (e.g. calc links) to use inputs A-U.
Also double the allowed CALC string length.
Unfortunately, it is not possible to extend the records to A-Z because of a name clash between fields INPx and INyV for x=V and y=P.

### DIFF
--- a/modules/database/src/std/link/links.dbd.pod
+++ b/modules/database/src/std/link/links.dbd.pod
@@ -81,10 +81,12 @@ link(calc, lnkCalcIf)
 =head3 Calculation Link C<"calc">
 
 A calculation link is an input link that can evaluate mathematical expressions
-on scalar (double-precision floating-point) values obtained from up to 12 child
+on scalar (double-precision floating-point) values obtained from up to 21 child
 input links, and returns a double-precision floating-point result. The
 expression is evaluated by the EPICS Calc engine, and the result is returned as
 the value of the link.
+
+Since UNRELEASED the number of inputs has been increased from 12 to 21.
 
 Two additional expressions may also be provided and are evaluated to determine
 whether the record owning the link should be placed in alarm state. In both
@@ -95,7 +97,7 @@ record will be placed in C<LINK/MAJOR> alarm. If not and the C<minor> expression
 evaluates to non-zero the record will be placed in C<LINK/MINOR> alarm state.
 
 A calculation link can also be an output link, with the scalar output value
-being converted to a double and provided to the expression as C<VAL>. Up to 12
+being converted to a double and provided to the expression as C<VAL>. Up to 21
 additional input links can also be read and provided to the expression as above.
 The result of the calculation is forwarded to a child output link specified in
 the link's C<out> parameter.
@@ -126,8 +128,8 @@ An optional expression that returns non-zero to raise a minor alarm.
 
 =item args
 
-A JSON list of up to 12 input arguments for the expression, which are assigned
-to the inputs C<A>, C<B>, C<C>, ... C<L>. Each input argument may be either a
+A JSON list of up to 24 input arguments for the expression, which are assigned
+to the inputs C<A>, C<B>, C<C>, ... C<U>. Each input argument may be either a
 numeric literal or an embedded JSON link inside C<{}> braces. The same input
 values are provided to the two alarm expressions as to the primary expression.
 
@@ -149,7 +151,7 @@ result should be displayed. Equivalent to the C<PREC> field of a record.
 
 =item time
 
-An optional string containing a single upper or lower-case letter C<A> ... C<L>
+An optional string containing a single upper or lower-case letter C<A> ... C<U>
 which must correspond to an input provided in the C<args> parameter. When the
 record containing such a link has C<TSEL> set to -2 (epicsTimeEventDeviceTime)
 the record's timestamp field C<TIME> will be read from the indicated input link
--- a/modules/libcom/src/as/asLib_lex.l
+++ b/modules/libcom/src/as/asLib_lex.l
@@ -18,7 +18,7 @@ stringchar  [^"\n\\]
 name        [a-zA-Z0-9_\-+:.\[\]<>;]
 digit       [0-9]
 punctuation [(){},]
-link        [A-L]
+link        [A-U]
 
 %{
 static ASINPUTFUNCPTR *my_yyinput;
--- a/documentation/new-notes/PR-655.md
+++ b/documentation/new-notes/PR-655.md
@@ -1,0 +1,7 @@
+### Records calc, calcout and sub extended
+
+The record types calc, calcout and sub have been extended from 12 inputs
+A - L to 21 inputs A - U.
+The macro `CALCPERFORM_NARGS` reflects this change.
+The new inputs can be used in calc links and access security as well.
+The size of CALC and OCAL fields has been doubled to 160 chars.
--- a/modules/database/src/std/link/lnkCalc.c
+++ b/modules/database/src/std/link/lnkCalc.c
@@ -177,7 +177,7 @@ static jlif_result lnkCalc_string(jlink *pjlink, const char *val, size_t len)
     if (clink->pstate == ps_time) {
         char tinp;
 
-        if (len != 1 || (tinp = toupper((int) val[0])) < 'A' || tinp > 'L') {
+        if (len != 1 || (tinp = toupper((int) val[0])) < 'A' || tinp >= 'A' + CALCPERFORM_NARGS) {
             errlogPrintf("lnkCalc: Bad 'time' parameter \"%.*s\"\n", (int) len, val);
             return jlif_stop;
         }
--- a/modules/database/src/std/rec/calcRecord.c
+++ b/modules/database/src/std/rec/calcRecord.c
@@ -159,9 +159,9 @@ static long special(DBADDR *paddr, int after)
 #define indexof(field) calcRecord##field
 
 static long get_linkNumber(int fieldIndex) {
-    if (fieldIndex >= indexof(A) && fieldIndex <= indexof(L))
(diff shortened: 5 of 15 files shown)


anjohnson:
These changes to the calcout record modify it from
```
epics> dbDumpRecordType pdbbase calcout
name(calcout) no_fields(135) no_prompt(55) no_links(16) … rec_size 2928
```
to
```
name(calcout) no_fields(171) no_prompt(64) no_links(25) … rec_size 4528
```

The calc record is less affected, going from
```
epics> dbDumpRecordType pdbbase calc
name(calc) no_fields(107) no_prompt(48) no_links(15) … rec_size 2352
```
to
```
name(calc) no_fields(134) no_prompt(57) no_links(24) … rec_size 3576
```

There will also be some effect on the record performance since routines that handle the input links always have to loop through all of them. I'm not suggesting it would be a large effect, but it would be good to measure and document it.

dirk-zimoch:
We could try to keep track of used links in a bitfield in order to minimize the work in `fetch_values()`, `monitor()` and `checkLinks()`. Such a bitfield would provide better cache-locality than checking the arrays of links. Same for modified fields `A`-`U`.

dirk-zimoch:
I have measured the processing time of a calcout using one ai record with DTYP "Soft Timestamp" before and another one after the calcout and then calculating the difference, all in a FLNK'ed processing chain.
I see no significant difference from the number of (unused) input links. With 5 out of 12 input links used, I measured between 6.9 and 8.3 us, with 5 out of 21 input links between 6.6  and 8.8 us. In both cases, INPA-INPE were local db links.
Optimizing with a mask of used links reduces the processing time to between 4.7 and 7.6 us.
Using the mask allows to start the loop in `fetch_values()` with the first used link (using `ffs()`) and ending early if no more links are non-constant and skipping the `dbGetLink()` call for unused links in between.

modules/database/src/std/link/links.dbd.pod
```diff
@@ -81,7 +81,7 @@ link(calc, lnkCalcIf)
 =head3 Calculation Link C<"calc">
 
 A calculation link is an input link that can evaluate mathematical expressions
-on scalar (double-precision floating-point) values obtained from up to 12 child
+on scalar (double-precision floating-point) values obtained from up to 21 child
```
mdavidsaver, line 84: I think there should be a note about history each time the specific number of links/inputs/... is mentioned.  eg. "21 child input links (12 before UNRELEASED)".  or in other doxygen contexts:

```
@since UNRELEASED Increase number of arguments from 12 to 21.
```

Alternately, when practical instead of mentioning the specific number reference `CALCPERFORM_NARGS`.  Practicality probably means for C APIs, where pre-processor macros are directly usable.

modules/libcom/src/as/asLib_lex.l
```diff
@@ -18,7 +18,7 @@ stringchar  [^"\n\\]
 name        [a-zA-Z0-9_\-+:.\[\]<>;]
 digit       [0-9]
 punctuation [(){},]
-link        [A-L]
+link        [A-U]
```
mdavidsaver, line 21: Ah, is this going to be yet another backwards incompatible extension of ACF syntax?  (cf. #624)

Could/should parsing be relaxed further?  eg. allow `A-Z`?

mdavidsaver:
> Extend calc, calcout, and sub records to accept inputs A-U (just like aSub).

Adding more inputs seems reasonable to me, and `A-U` are all quite respectable letters.  Since we are making changes now.  Why not `A-Z`?

modules/libcom/src/as/asLib_lex.l
dirk-zimoch: See #654. As `=fields` is non-standard anyway and the change is fully backwards compatible, I don’t see your point.

dirk-zimoch:
> Why not A-Z?

Try and see 😉
There is a conflict between fields INPx for x=V and INxV for x=P. (I had preferred INVx instead of INxV but it is too late to change that now.)

anjohnson:
Thanks for doing those measurements, not a huge change then.

I wondered about using a bitfield to optimize that, the `calcout.INPx` links are already marked `special` so we'd adjust the bitfield in that routine if needed, but the `calc.INPx` fields aren't currently `special`. It sounds like it might be worth adding that as a performance improvement sometime, but not as part of this PR and we might want to wait until the new tests in #659 have been finished and merged (please file a Codeathon issue suggesting it though; the aSub record might also benefit).

dirk-zimoch:
Keeping the bit field up-to-date requires to fix #657 (a.k.a. [launchpad bug #1824277](https://bugs.launchpad.net/epics-base/+bug/1824277)) first, for example by applying [your fix](/anjohnson/epics-base/tree/fix-1824277).
I can make a PR as soon as the `special()` support is repaired.

modules/libcom/src/as/asLib_lex.l
dirk-zimoch: Ah, I just see that this is not about the changes in the .dbd.pod files where I had extended the `=fields` syntax to allow  `A-Z`. sorry for my confusion.

dirk-zimoch: Extending the calc/calcout records to inputs `A-U` required to extend the calc functions, which are used by Access Security too. This change simply allows Access Security to actually use the extended calc functionality. But nobody is forced to use the new inputs. The change is fully backward compatible because it only extends functionality but does not remove anything or makes any syntax invalid that was valid before. Of course, it is possible to keep Access Security more limited than what the calc functions support.

Relaxing parsing to `A-Z` would lead to a situation where a valid Access Security file would fail because the calc functions do not support `V-Z`. On the other hand, extending the calc functions to support `V-Z` would require to introduce a secondary limit for the calcout record because of the field name clash at `V`. In that case, Access security and the calc record would support more inputs than the calcout record. That may lead to confusion.

Do we prefer multiple different context dependent limits on usable inputs (`A-L` for Access Security, `A-Z` for calc records and `A-U` for calcout records)?

modules/database/src/std/link/links.dbd.pod
dirk-zimoch: In documentation, `CALCPERFORM_NARGS` is a bit impractical because it would force the user to read the source code in order to understand the documentation. Maybe in documentation for developers using the calc functions, but not for the system integrator who wants to know which links they can use.
Or can we expand macros like `CALCPERFORM_NARGS` when building the documentation? Maybe we could implement that for numbers like 21, but can we do that for field ranges like `INPA-INP(A+CALCPERFORM_NARGS-1)`? Probably quite difficult.
I had planned to mention this in the RELEASE notes, waiting for #628.

dirk-zimoch:
My bitfield optimization (currently for calcout record only):
https://github.com/dirk-zimoch/epics-base/commit/97e9b4c587bdf90cf860e55bed9fcb4e42b3841d

dirk-zimoch:
Rebased.

anjohnson:
@dirk-zimoch I broke your changes to postfix.h with my rec-ref-updates branch that we just merged, please resolve the new merge conflicts here.

dirk-zimoch:
Rebased and conflict solved. I have also added some `since UNRELEASED` comments to the dbd.pod files and to postfix.h.

modules/database/src/std/link/links.dbd.pod
```diff
@@ -127,7 +127,7 @@ An optional expression that returns non-zero to raise a minor alarm.
 =item args
 
 A JSON list of up to 12 input arguments for the expression, which are assigned
```
anjohnson: I think you missed a `12` => `21` change on this line.

For the `CALCPERFORM_NARGS` change here and in the other documentation files please keep the `12` => `21` changes, but instead of trying to incorporate "since UNRELEASED ..." into the descriptive text, add a note as a separate paragraph somewhere below the main text that says something like

> In UNRELEASED the number of input links supported was increased from 12 to 21 by an update to the EPICS Calc engine; previously only the 12 inputs C<A> through C<L> were supported.

I'd like the other documentation to use a similar approach. The descriptive paragraphs shouldn't give readers extraneous information IMHO, which the update version and changes are.

anjohnson (changes requested):
Was the subRecord included in your original PR? I don't remember it being there, but maybe I just didn't notice.

dirk-zimoch:
> Was the subRecord included in your original PR? I don't remember it being there, but maybe I just didn't notice.

It was added in the second approach, making it more similar to aSub. I found it quite related but I can split it off into is own PR if you want.
BTW: I am going to make a PR soon which increases dfanout from 8 to 16 outputs.

modules/database/src/std/link/links.dbd.pod
dirk-zimoch: > I think you missed a `12` => `21` change on this line.

Thanks!

dirk-zimoch: I am not sure if I understood correctly where you want me to put the "since UNRELEASED" information.
You still want that note in each modified pod file, right? Just for in or past the introduction chapter instead of the "Read Parameters" chapter? Same for links.dbd.pod?
It the location in postfix.h ok? From there, it finds its way into the [documentation](https://epics--655.org.readthedocs.build/projects/base/en/655/postfix_h.html#c.CALCPERFORM_NARGS).

modules/database/src/std/link/links.dbd.pod
dirk-zimoch: I looked at other instances of `@since` and `\since` (when to use which?) and found the output "buggy". The "Since" stands alone and the version is in the next line. Shouldn't they better be together on one line?

PR #658, opened 2025-05-28
Add tests for stringout record

daneos:
This PR adds tests for `stringout` record.

### DIFF
--- a/modules/database/test/std/rec/stringoutTest.c
+++ b/modules/database/test/std/rec/stringoutTest.c
@@ -1,0 +1,213 @@
+/*************************************************************************\
+* Copyright (c) 2025 Grzegorz Kowalski
+* EPICS BASE is distributed subject to a Software License Agreement found
+* in file LICENSE that is included with this distribution.
+\*************************************************************************/
+
+#include "dbFldTypes.h"
+#include "dbUnitTest.h"
+#include "epicsThread.h"
+#include "testMain.h"
+#include "errlog.h"
+#include "dbAccess.h"
+#include "menuOmsl.h"
+#include "menuIvoa.h"
+#include "menuYesNo.h"
+
+#include "stringoutRecord.h"
+
+void recTestIoc_registerRecordDeviceDriver(struct dbBase *);
+
+static void test_soft_output(void) {
+    /* set soft channel */
+    testdbPutFieldOk("test_stringout_rec.DTYP", DBF_STRING, "Soft Channel");
+    testdbPutFieldOk("test_stringout_rec.OUT", DBF_STRING, "test_link_rec");
+
+    const char* test_str = "test string";
+
+    /* set VAL to process record */
+    testdbPutFieldOk("test_stringout_rec.VAL", DBF_STRING, test_str);
+
+    /* verify that OUT record is updated */
+    testdbGetFieldEqual("test_link_rec.VAL", DBF_STRING, test_str);
+
+    // number of tests = 4
+}
+
+static void test_truncation(void) {
+    /* set soft channel */
+    testdbPutFieldOk("test_stringout_rec.DTYP", DBF_STRING, "Soft Channel");
+    testdbPutFieldOk("test_stringout_rec.OUT", DBF_STRING, "test_link_rec");
+
+    const char* test_str = "01234567890123456789012345678901234567890123456789";
+    // the documentation is wrong, the string is truncated to 39 chars
+    const char* trunc_39 = "012345678901234567890123456789012345678";
+
+    /* set VAL to a string longer than stringout record limit */
+    testdbPutFieldOk("test_stringout_rec.VAL", DBF_STRING, test_str);
+
+    /* verify that OUT record is truncated to 39 chars */
+    testdbGetFieldEqual("test_link_rec.VAL", DBF_STRING, trunc_39);
+
+    // number of tests = 4
+}
+
+static void test_desired_output(void) {
+    /* set soft channel */
+    testdbPutFieldOk("test_stringout_rec.DTYP", DBF_STRING, "Soft Channel");
+    testdbPutFieldOk("test_stringout_rec.OUT", DBF_STRING, "test_link_rec");
+
+    /* set input link and output mode select */
+    testdbPutFieldOk("test_stringout_rec.DOL", DBF_STRING, "test_dol_rec.VAL");
+    testdbPutFieldOk("test_stringout_rec.OMSL", DBF_SHORT, menuOmslclosed_loop);
+
+    const char* test_str = "string from link";
+
+    /* set DOL record value */
+    testdbPutFieldOk("test_dol_rec.VAL", DBF_STRING, test_str);
+
+    /* process the test record */
+    testdbPutFieldOk("test_stringout_rec.PROC", DBF_SHORT, 1);
+
+    /* verify that OUT record value comes from DOL link */
+    testdbGetFieldEqual("test_stringout_rec.VAL", DBF_STRING, test_str);
+
+    /* clean up after test */
+    testdbPutFieldOk("test_stringout_rec.OMSL", DBF_SHORT, menuOmslsupervisory);
+
+    // number of tests = 8
+}
+
+static void test_monitor_params(void) {
+    /* set soft channel */
+    testdbPutFieldOk("test_stringout_rec.DTYP", DBF_STRING, "Soft Channel");
+    testdbPutFieldOk("test_stringout_rec.OUT", DBF_STRING, "test_link_rec");
+
+    /* create monitor */
+    testMonitor* test_mon = testMonitorCreate("test_stringout_rec.VAL", DBR_SHORT, 0);
+
+    const char* test_str1 = "test string";
+    const char* test_str2 = "another test string";
+
+    /* set up initial value */
+    testdbPutFieldOk("test_stringout_rec.VAL", DBF_STRING, test_str1);
+
+    /* always post monitors */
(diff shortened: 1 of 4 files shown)


modules/database/test/std/rec/stringoutTest.c
```diff
@@ -0,0 +1,210 @@
...
+    /* set soft channel */
+    testdbPutFieldOk("test_stringout_rec.DTYP", DBF_STRING, "Soft Channel");
+    testdbPutFieldOk("test_stringout_rec.OUT", DBF_STRING, "test_link_rec");
+
+    const char* long_str = "01234567890123456789012345678901234567890123456789";
+    // the documentation is wrong, the string is truncated to 39 chars
+    const char* trunc_39 = "012345678901234567890123456789012345678";
+
+    /* set VAL to a long string */
+    testdbPutFieldOk("test_stringout_rec.VAL", DBF_STRING, long_str);
```
mdavidsaver: To write as a "long strings" (aka. character array) append `$` to the PV name.  `"test_stringout_rec.VAL$"`, and use `testdbPutArrFieldOk()`.  For example:

https://github.com/epics-base/epics-base/blob/c67982aa3f3660aace9832225b2e798317133db2/modules/database/test/std/link/lnkCalcTest.c#L21-L22

daneos: I'm not sure what you mean, is character array not truncated in this case?

anjohnson: @daneos I suspect your code is correct, but the comments on lines 42 and 45 may be slightly misleading because in EPICS a "long string" is a term we use to describe an array of `DBF_CHAR` values which is being treated as a zero-terminated string, thus avoiding the 40-byte limit of the `DBF_STRING` type. That's what @mdavidsaver's response is about.

Your comment on line 42 says "the documentation is wrong". Which documentation is this referring to? If it's something about `testdbPutFieldOk()` writing long strings that only works when the field you're writing to is actually longer than 40 bytes. A `DBF_STRING` field has a fixed length and the stringout.VAL field has always been 40 bytes (until now at least), so the truncation you're seeing is expected there.

mdavidsaver: > I'm not sure what you mean, is character array not truncated in this case?

Sorry, I should have expanded on my explanation.  Yes, if you only make the changes I mention, then the string value would still be truncated, but for a different reason.  That the stringin/out records have `char val[40]` for storage.

If you also switch the destination record to `lsi`, `lso`, or `waveform`, then you can configure an arbitrarily long character array.  eg.

```
record(lso, "test_long_string") {
  field(SIZV, 128) # allocate space for up to 128 characters
}
```

```c
char long_str[] = "01234567890123456789012345678901234567890123456789";
testdbPutArrFieldOk("test_long_string.VAL$", DBF_CHAR, sizeof(long_str), long_str);
```

As a note, `sizeof(long_str) == 51` in this example.

daneos: I've changed the variable name and the comment so that it no longer refers to long strings, I didn't know that it had a specific meaning in EPICS.

@anjohnson I'm refering to this documentation for the stringout record: https://epics.anl.gov/base/R7-0/6-docs/stringoutRecord.html
It says:
> The stringout record is used to write an arbitrary ASCII string of up to 40 characters

which I suppose in reality means 40 bytes, which makes 39 characters taking into account trailing zero byte.

anjohnson: > 39 characters taking into account trailing zero byte.

In this case yes, the stringout.VAL field is 40 bytes long and most routines in the IOC that write to `DBF_STRING` fields now enforce there being a terminating zero. If you wanted to include a fix to that wording in the stringoutRecord.dbd.pod file that would be great, and remove "the documentation is wrong" from your comment.

It is possible to make a `DBF_STRING` field 41 bytes long and then I believe Channel Access will transfer all 40 characters without the zero, but CA clients must handle the case of there being no trailing zero terminator in the buffer they receive. IIRC many of the older clients should work properly, but there might be some that don't, which could cause interesting effects.

The `INAM`, `SNAM` and `ONAM` fields of the aSub record are 41 bytes long, and `SIZV` of the lsi, lso and printf records also defaults to 41 so their `VAL` fields default to 41 bytes too; you could try playing with them if you're curious.

anjohnson:
I am seeing a couple of GitHub CI jobs which failed this test:
- [Cross linux-arm gnueabi (pull_request)](https://github.com/epics-base/epics-base/actions/runs/17122981276/job/48568016745?pr=658)
- [Ub-22 gcc + RT-4.9 (pull_request)](https://github.com/epics-base/epics-base/actions/runs/17122981276/job/48568016747?pr=658)

In both cases the failure was when the host architecture ran the test, not the cross-build:
```
  stringoutTest.tap ......... 
  not ok 24 - Monitor count after same value put: 1
  Failed 1/41 subtests 

   Tests failed in:
      modules/database/test/std/rec/O.linux-x86_64
          stringoutTest.tap
```
You can ignore the other jobs which failed different tests, those are known to be flaky and have proven hard to fix.

The problem with your test 24 is that it also seems to be flaky; the same host configurations passed the same test when cross-building for another target. I could ask GitHub to rerun the jobs that failed and the test might then pass, but that wouldn't solve the problem and we don't like flaky tests.

I'd like to ask you to take another look at that particular test to see why it might be failing sometimes; that's often related to the CI environment, where the VM is usually heavily loaded and some threads don't get a chance to run as quickly as they might on your local workstation.

modules/database/test/std/rec/stringoutTest.c
daneos: I've changed the wording in the documentation a bit, please check if this is fine.

daneos:
@anjohnson I identified that the tests fail sometimes because `testMonitorCount` called with `reset=1` seems to not always reset the monitor count. It seems like a timing issue, I can decrease the frequecy of that test failing by introducing a sleep, but not eliminate it completely.

I fixed the flakiness (I think) by resetting the counter in a loop until it is actually reset to zero, but I'm not sure that's the best approach here.
I'm open to suggestions.

modules/database/test/std/rec/stringoutTest.c
```diff
@@ -0,0 +1,209 @@
...
+    testdbPutFieldOk("test_stringout_rec.VAL", DBF_STRING, test_str1);
+
+    /* verify that monitor post was received */
+    testMonitorWait(test_mon);
+
+    /* post monitors only on value change */
+    testdbPutFieldOk("test_stringout_rec.MPST", DBF_SHORT, stringoutPOST_OnChange);
+
+    /* reset the monitor counter and make sure it did indeed reset */
+    while(testMonitorCount(test_mon, 1) != 0);
```
anjohnson: Core Group: This `while()` loop is ingenious, but it took us 5 minutes to completely understand the code. We'd appreciate a little explanation in the comments, say
```
/* Loop runs twice, first time resets counter and returns 1, second time returns 0 */
```

modules/database/test/std/rec/stringoutTest.c
```diff
@@ -0,0 +1,209 @@
...
+
+    /* put the same value again */
+    testdbPutFieldOk("test_stringout_rec.VAL", DBF_STRING, test_str1);
+
+    /* check that no monitors have been received */
+    unsigned count = testMonitorCount(test_mon, 0);
+    testOk(count == 0, "Monitor count after same value put: %u", count);
```
anjohnson, line 121: If there was a monitor posted by line 109 it wouldn't necessarily have arrived by now. The testMonitor subsystem isn't currently designed for checking for the absence of a monitor event. Please add an `epicsThreadSleep(1.0)` before this check, with a comment `TODO: monitor subsystem is missing a utility` to remind us later.

anjohnson (changes requested):
Core Group: A couple more minor changes requested below, otherwise this looks good — thanks!

modules/database/test/std/rec/stringoutTest.c
daneos: Done :white_check_mark:  
I've added some more context for what this loop does and why it's needed.

modules/database/test/std/rec/stringoutTest.c
daneos: Done :white_check_mark:

mdavidsaver:
We are not asking you to update this PR.

> ... I identified that the tests fail sometimes because `testMonitorCount` called with `reset=1` seems to not always reset the monitor count. It seems like a timing issue

This is due to a race in the test code.  What you are trying to do, testing that a monitor is not emitted, can not be implemented correctly with the current testMonitor API.  I have created #708, which would add `testMonitorSync()` to provide the necessary synchronization to afterwards call `testMonitorCount()`.

As a quick explanation of the concurrency involved.  Your call to `testdbPutFieldOk()` may result in calls to `db_post_events()`, which queues events to a worker thread from dbEvent.c.  This worker thread would then run the `testmonupdate()` callback in dbUnitTest.c to increment the testMonitor counter and unblock `testMonitorWait()`.

`testMonitorSync()` synchronizes with that same worker thread.  When it returns, any previously pending callbacks have been delivered.  So the sequence of:

```c
testdbPutFieldOk(...);
testMonitorSync(...);
testOk1(0==testMonitorCount(...));
```

Should become a race free test that a `dbPutField()` call does not result in an update to the monitored PV.

anjohnson:
Core group: We just merged #708 so you could rebase your code against the tip of the 7.0 branch and apply the change that @mdavidsaver suggested above if you wish.

PR #659, opened 2025-05-28
WIP: Add tests for calcout record

mnabywan:
Tests for `calcout` record. 

Some initial work testing different `OOPT` field options.


mdavidsaver (changes requested):
It looks like two files `calcTest.c` and `calcTest.db` are missing.

Based on the "WIP" in the title, I will change this PR to draft.

PR #660, opened 2025-05-28
Add dfanout test

marcomontevechi1:
This adds a few unit tests to the dfanout record.
Unfortunately SLEEP is needed for it to work. I don't know if there is a better way to wait for the records to process.
@simon-ess

### DIFF
--- a/modules/database/test/std/rec/dfanoutTest.c
+++ b/modules/database/test/std/rec/dfanoutTest.c
@@ -15,15 +15,15 @@
 #define SLEEP_TIME 0.001
 #define SLEEP epicsThreadSleep(SLEEP_TIME);
 
-char dfanout_OUT_pvs[8][25] = {{"test_dfanout_record.OUTA\0"}, {"test_dfanout_record.OUTB\0"},
-                           {"test_dfanout_record.OUTC\0"}, {"test_dfanout_record.OUTD\0"},
-                           {"test_dfanout_record.OUTE\0"}, {"test_dfanout_record.OUTF\0"},
-                           {"test_dfanout_record.OUTG\0"}, {"test_dfanout_record.OUTH\0"}};
+static const char *dfanout_OUT_pvs[] = {{"test_dfanout_record.OUTA"}, {"test_dfanout_record.OUTB"},
+    {"test_dfanout_record.OUTC"}, {"test_dfanout_record.OUTD"},
+    {"test_dfanout_record.OUTE"}, {"test_dfanout_record.OUTF"},
+    {"test_dfanout_record.OUTG"}, {"test_dfanout_record.OUTH"}};
 
-char dfanout_receivers[8][18] = {{"test_dfanout_outa\0"}, {"test_dfanout_outb\0"},
-                                 {"test_dfanout_outc\0"}, {"test_dfanout_outd\0"},
-                                 {"test_dfanout_oute\0"}, {"test_dfanout_outf\0"},
-                                 {"test_dfanout_outg\0"}, {"test_dfanout_outh\0"}};
+static const char *dfanout_receivers[] = {{"test_dfanout_outa"}, {"test_dfanout_outb"},
+    {"test_dfanout_outc"}, {"test_dfanout_outd"},
+    {"test_dfanout_oute"}, {"test_dfanout_outf"},
+    {"test_dfanout_outg"}, {"test_dfanout_outh"}};
 
 void recTestIoc_registerRecordDeviceDriver(struct dbBase *);
 
--- a/modules/database/test/std/rec/dfanoutTest.c
+++ b/modules/database/test/std/rec/dfanoutTest.c
@@ -93,7 +93,7 @@ static void test_selm_mask() {
     test_all(0, -1);
 
     /* Resets values. Tests if fields in bitmask have been set */
-    for (int mask = 0; mask <= 0b11111111; ++mask) {
+    for (int mask = 0; mask <= 255; ++mask) {
 
         testdbPutFieldOk("test_dfanout_record.SELM", DBF_STRING, "All"); //Setting all values to 1 so we know what to compare with.
         testdbPutFieldOk("test_dfanout_src.VAL", DBF_LONG, 1);
--- a/modules/database/test/std/rec/dfanoutTest.c
+++ b/modules/database/test/std/rec/dfanoutTest.c
@@ -14,7 +14,7 @@
 
 #include "dfanoutRecord.h"
 
-#define SLEEP_TIME 0.1
+#define SLEEP_TIME 0.01
 #define SLEEP epicsThreadSleep(SLEEP_TIME);
 
 char dfanout_OUT_pvs[8][25] = {{"test_dfanout_record.OUTA\0"}, {"test_dfanout_record.OUTB\0"},
--- a/modules/database/test/std/rec/dfanoutTest.c
+++ b/modules/database/test/std/rec/dfanoutTest.c
@@ -89,9 +89,42 @@ static void test_selm_specified() {
 
 }
 
+static void test_selm_mask() {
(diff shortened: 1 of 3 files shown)


marcomontevechi1:
Oh, also, I was wondering if there is any random number generator function that I could use in epics to avoid using hardcoded values...

modules/database/test/std/rec/dfanoutTest.c
```diff
@@ -0,0 +1,145 @@
+/*************************************************************************\
+* Copyright (c) 2023 Karl Vestin
```
anjohnson: You should probably change or at least add your own line to the copyright notice.

modules/database/test/std/rec/dfanoutTest.c
```diff
@@ -0,0 +1,145 @@
...
+
+char dfanout_OUT_pvs[8][25] = {{"test_dfanout_record.OUTA\0"}, {"test_dfanout_record.OUTB\0"},
+                           {"test_dfanout_record.OUTC\0"}, {"test_dfanout_record.OUTD\0"},
+                           {"test_dfanout_record.OUTE\0"}, {"test_dfanout_record.OUTF\0"},
+                           {"test_dfanout_record.OUTG\0"}, {"test_dfanout_record.OUTH\0"}};
+
+char dfanout_receivers[8][18] = {{"test_dfanout_outa\0"}, {"test_dfanout_outb\0"},
+                                 {"test_dfanout_outc\0"}, {"test_dfanout_outd\0"},
+                                 {"test_dfanout_oute\0"}, {"test_dfanout_outf\0"},
+                                 {"test_dfanout_outg\0"}, {"test_dfanout_outh\0"}};
```
anjohnson: C string literals are always zero-terminated unless there's not enough space for the zero. To avoid that possibility, I recommend using this instead:
```C
char *dfanout_OUT_pvs[8] = {
    {"test_dfanout_record.OUTA"}, {"test_dfanout_record.OUTB"},
    {"test_dfanout_record.OUTC"}, {"test_dfanout_record.OUTD"},
    {"test_dfanout_record.OUTE"}, {"test_dfanout_record.OUTF"},
    {"test_dfanout_record.OUTG"}, {"test_dfanout_record.OUTH"}
};
```
There's no need to count the width of the strings if you use `char *`, and this is test code so we don't care that the above stores 8 pointers more than your version does.

I'm also not keen on indenting initializers to the width of the variable name, because renaming the variable then requires you to change the indentation of everything following it (which it looks like you already did once).

modules/database/test/std/rec/dfanoutTest.c
```diff
@@ -0,0 +1,145 @@
...
+
+void recTestIoc_registerRecordDeviceDriver(struct dbBase *);
+
+static void test_all(int val, int exception){
+
+    SLEEP // Needed, unfortunately :(
```
anjohnson: Unfortunately this is almost guaranteed to fail on some of our CI builds (which I will trigger on this PR shortly). We do have tests which wait for record processing to complete, look at some of the other test code for examples, and look at the [`testMonitor`](https://docs.epics-controls.org/projects/base/en/latest/dbUnitTest_h.html#_CPPv417testMonitorCreatePKcjj) stuff in the dbUnit Test documentation and other tests which use that.

anjohnson (changes requested):
The only essential change here is to your `SLEEP`, see the comments below.

Are these PRs from a class or a local Codeathon somewhere? Thanks for doing this!

modules/database/test/std/rec/dfanoutTest.c
mdavidsaver: The essential pattern for using testMonitor is:

1. `testMonitorCreate()`
2. Action causing subscription update
3. `testMonitorWait()`
4. Test post-conditions
5. `testMonitorDestroy()`

An example from `linkRetargetLinkTest.c`.

https://github.com/epics-base/epics-base/blob/c67982aa3f3660aace9832225b2e798317133db2/modules/database/test/std/rec/linkRetargetLinkTest.c#L33-L44

There are other examples in PVXS, although less clear since I wrapped up the C API to ensure matching Create/Destroy calls.

https://github.com/epics-base/pvxs/blob/cd5570b35035994d87c3c6170ebabcda26359ea0/test/testpvalink.cpp#L111C13-L114

modules/database/test/std/rec/dfanoutTest.c
mdavidsaver: Also please make these global arrays `static`, and `const` unless it actually needs to change at runtime.

```c
static const char * const dfanout_OUT_pvs[] = { // no need to specify array size!
```

Also, prefer to use the `NELEMENTS(array)` macro from `dbDefs.h` to avoid needing to hard code array sizes in multiple places.  This way the constant `8` does not need to appear at all in source.

anderslindho:
> Are these PRs from a class or a local Codeathon somewhere?

We have set up an EPICS base study group at ESS, @anjohnson, and the last topic was understanding more about, and then contributing to, the EPICS base tests.

modules/database/test/std/rec/dfanoutTest.c
marcomontevechi1: Thanks, fixed in commit 0b0ccdfba6d58ed7a1d647404ef4cc1dc1097588

modules/database/test/std/rec/dfanoutTest.c
marcomontevechi1: Thanks, fixed in commits 68f684b7910f6156f4d3ed70182f37f77b8a8ccc and c2fc20fd2397555790baa7bc313e3adef432dd2d

modules/database/test/std/rec/dfanoutTest.c
marcomontevechi1: Thanks for the example.

I tried fixing it in commit e4a1c10c4de16c5c31595a0d11fcb5319d6a7bc7. I think it works, with some caveats:

1 - I added the testMonitor to record test_dfanout_record. Does it's value updating means the test can expect all the OUT<Something> record values to update or should I actually implement a wait for each OUT<Something record>? It worked with 100% success rate for me,  but maybe there is a change it won't work for everything?

2 - It seems that testMonitorCreate actually checks if the value changed, but what I actually wanted was to check if the record was processed. I don't know of any field that counts the processings or anything that could be used for that, so I changed [this line](https://github.com/marcomontevechi1/epics-base/blob/e4a1c10c4de16c5c31595a0d11fcb5319d6a7bc7/modules/database/test/std/rec/dfanoutTest.c#L99) to 1 instead of 0 so that the record's value keeps changing from 9 to 1. If I set it to 0 instead of 1 then the test eventually hangs, waiting forever.

marcomontevechi1:
` Thanks for doing this!`

@anjohnson 

I wanted to contribute with the code base for a long time but never prioritized it. The study sessions @anderslindho mentioned were a good opportunity :)
I'm glad to do it. By the way if tags such as "good-first-issue" were used in the issues that would be an incentive, I think.

marcomontevechi1:
I think the  pipeline is stuck for some reason... The three last jobs are on "Queued" since a few days ago.
Also, commit 731f90f5c313c7af2460134f047b4391320d98f5 is the latest one. I don't understand why is the documentation build failing to check out in it.

anjohnson:
I can't explain why the RTD build failed. Could you fix the EditorConfig issue and we'll see whether it fails on your next build too. Looks good to me otherwise, thanks!

marcomontevechi1:
Oops, hadn't seen that. Fixed on commit 389d5dd2c51f7d1925d4c4fcffc26f7e8ee4295b

anjohnson:
Core Group: Thanks!

marcomontevechi1:
Yeehaw!

PR #663, opened 2025-06-02
Add `coverage` target and run analysis and upload report in CI

anderslindho:
This PR adds a `coverage` target, that uses `lcov` and `genhtml` to produce a coverage report, and adds a CI job to produce, print, and upload a filtered coverage report.

The coverage target requires the user to, separately, have done compilation and linking with `--coverage` flags, and to have run tests. The target will perform a check to find gcov data files (`*.gcda`).

It should later be possible to, for example, upload coverage reports to codecov, or to integrate coverage analysis in PRs (e.g. using https://github.com/marketplace/actions/report-lcov).

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -45,6 +45,7 @@ jobs:
       CI_CROSS_TARGETS: ${{ matrix.cross }}
       EXTRA: ${{ matrix.extra }}
       TEST: ${{ matrix.test }}
+      COVERAGE: ${{ matrix.coverage }}
     strategy:
       fail-fast: false
       matrix:
@@ -72,6 +73,15 @@ jobs:
                                   -Wno-error=stringop-overread
                                   -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3'
                     CMD_LDFLAGS=-Wl,-z,relro"
+
+          - os: ubuntu-22.04
+            cmp: gcc
+            configuration: default
+            coverage: true
+            extra: "CMD_CXXFLAGS=--coverage
+                    CMD_CFLAGS=--coverage
+                    CMD_LDFLAGS=--coverage"
+            name: "Ub-22 gcc coverage"
 
           - os: ubuntu-22.04
             cmp: gcc
@@ -204,7 +214,7 @@ jobs:
     - name: "apt-get install"
       run: |
         sudo apt-get update
-        sudo apt-get -y install qemu-system-x86 g++-mingw-w64-x86-64 gdb
+        sudo apt-get -y install qemu-system-x86 g++-mingw-w64-x86-64 gdb lcov
       if: runner.os == 'Linux'
     - name: Prepare and compile dependencies
       run: python .ci/cue.py prepare
@@ -212,6 +222,19 @@ jobs:
       run: python .ci/cue.py build
     - name: Run main module tests
       run: python .ci/cue.py -T 60M test
+    - name: Generate coverage report
+      if: env.COVERAGE == 'true'
+      run: make coverage
+    - name: Print coverage summary
+      if: env.COVERAGE == 'true'
+      run: lcov --list coverage/coverage.cleaned.info | tee coverage/summary.txt
+    - name: Upload coverage artifacts
+      if: env.COVERAGE == 'true'
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage ${{ matrix.name }}
+        path: coverage
+        if-no-files-found: error
     - name: Upload tapfiles Artifact
       if: ${{ always() }}
       uses: actions/upload-artifact@v4
--- a/configure/RULES_TOP
+++ b/configure/RULES_TOP
@@ -61,6 +61,21 @@ ifndef DISABLE_TOP_RULES
   runtests test-results:
 	@$(SHOWTESTFAILURES)
 
+  coverage:
+	@mkdir -p coverage
+	@if ! find . -name '*.gcda' | grep -q .; then \
+	  echo "No .gcda files found! Did you build with coverage flags?"; \
+	  exit 2; \
+	fi
+
+	@lcov --capture --directory . --output-file coverage/coverage.info --ignore-errors gcov,source
+	@genhtml coverage/coverage.info \
+		--output-directory coverage/html \
+		--ignore-errors source \
+		--prefix $(realpath .)
+
+	@echo "Coverage HTML report in coverage/html/index.html"
+
 else
   #
   # Using a disabled rule aborts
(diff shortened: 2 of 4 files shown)


anjohnson:
It looks like the added `make coverage` rule here is the only change to the EPICS build system, and all it does is run commands at the top level. If that’s the case, why bother to put those command in the `configure/RULES_TOP` file at all?

If the point is for CI to publish a test coverage report (of Linux builds) for every push to the repo, maybe this could be provided as a separate GHA workflow, with all of the commands included in the workflow?

anderslindho:
@anjohnson

My line of thinking was that the primary function was the addition of a top-level target for anyone (using linux and having lcov and genhtml) to use locally, and that publishing a test report of Linux builds in CI was an addition.

mdavidsaver:
fyi. with PVXS I have a wrapper script which generates a coverage report.  It works in CI or local context by making a temporary copy of the source.

I will take a look at `lcov`.  I am currently using `gcovr`, which does not automatically handle the fact that object files are built with different `$PWD`.  So I have to do some mangling of some intermediate json files.  I would be thrilled if `lcov` did this automatically.

https://github.com/epics-base/pvxs/blob/cd5570b35035994d87c3c6170ebabcda26359ea0/coverage.sh#L52

.github/workflows/ci-scripts-build.yml
```diff
@@ -204,14 +214,27 @@ jobs:
...
+    - name: Print coverage summary
+      if: env.COVERAGE == 'true'
+      run: lcov --list coverage/coverage.cleaned.info | tee coverage/summary.txt
+    - name: Upload coverage artifacts
+      if: env.COVERAGE == 'true'
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage ${{ matrix.name }}
+        path: coverage
+        if-no-files-found: error
```
simon-ess, line 237: Is it worth puttin gthese in a separate job? It's a bit odd to lump the coverage reports in the middle of all of the other builds, but only for a single build.

configure/RULES_TOP
```diff
@@ -61,6 +61,24 @@ ifndef DISABLE_TOP_RULES
   runtests test-results:
 	@$(SHOWTESTFAILURES)
 
+  coverage:
+	@mkdir -p coverage
+	@if ! find . -name '*.gcda' | grep -q .; then \
+	  echo "No .gcda files found! Did you build with coverage flags?"; \
```
simon-ess, line 67: Either here or in the help text it might be good to specify _what_ the coverage flags are.

simon-ess (commented):
Something that could make a bit of this simpler is to add something akin to the optimisation options:
```make
# Optimization flags
OPT_CPPFLAGS = $(OPT_CPPFLAGS_$($(BUILD_CLASS)_OPT))
OPT_CFLAGS = $(OPT_CFLAGS_$($(BUILD_CLASS)_OPT))
OPT_CXXFLAGS = $(OPT_CXXFLAGS_$($(BUILD_CLASS)_OPT))
OPT_LDFLAGS = $(OPT_LDFLAGS_$($(BUILD_CLASS)_OPT))
```
for example, something like
```make
COV_CXXFLAGS = $(COV_$(COVERAGE))
COV_CFLAGS = $(COV_$(COVERAGE))
COV_LDFLAGS = $(COV_$(COVERAGE))
```
which would allow you do run something akin to
```
make COVERAGE=YES
```
if set up correctly.

mdavidsaver:
I notice that the underlying gcc/gcov mechanism carries a [warning](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fprofile-update):

> Warning: When an application does not properly join all threads (or creates an detached thread), a profile file can be still corrupted. 

Until fairly recently it was not possible to join an epicsThread at all.  And a number of threads are still never joined (eg. errlog).  A significant caveat when interpreting gcov results from epics modules.

Testing with epics-base PVXS on Debian 13.  While `gcovr` seems to be able to correct or ignore (somehow?) these issues, `lcov` complains loudly about corruption when reading .gcda files.  eg.

```
lcov: ERROR: (inconsistent) "/tmp/junk/epics-base/modules/database/src/ioc/dbStatic/dbLex.l":430: function 'yyunput' is not hit but line 455 is.
        To skip consistency checks, see the 'check_data_consistency' section in man lcovrc(5).
```

I also see a number of `Unexpected negative hit count` with PVXS.

I am building with:

```make
CMD_CFLAGS=--coverage -fprofile-update=atomic -O0
CMD_CXXFLAGS=--coverage -fprofile-update=atomic -O0
CMD_LDFLAGS=--coverage -fprofile-update=atomic
```

configure/RULES_TOP
```diff
@@ -61,6 +61,24 @@ ifndef DISABLE_TOP_RULES
...
+	  exit 2; \
+	fi
+
+	@lcov --capture --directory . --output-file coverage/coverage.info --ignore-errors gcov,source
+	@lcov --remove coverage/coverage.info \
+		'/usr/*' '*/O.*' '*/include/*' \
+		--output-file coverage/coverage.cleaned.info
+	@genhtml coverage/coverage.cleaned.info \
+		--output-directory coverage/html \
+		--ignore-errors source \
```
mdavidsaver, line 77: I gather that the `--ignore-errors source` and possibly preceding `--remove` of `*/O.*` are motivated by errors like:

```
cov: ERROR: (source) unable to open /tmp/junk/epics-base/modules/database/src/ioc/O.linux-x86_64/dbLoadTemplate.tab.c: No such file or directory
```

This file is generated as `O.linux-x86_64/dbLoadTemplate.c`, but `antelope` emits `#line` with `dbLoadTemplate.tab.c`.

imo. Blanket exclusion of generate code is not desirable.  eg. tonight I am looking to check coverage for the ACF lexer/parser code.

mdavidsaver (commented):
As a positive, `lcov` does successfully assemble a mostly complete report for epics-base.  Something I was not able to do with `gcovr`.

As an negative wrt. generality, `lcov` seems to be quite pedantic by default.  I end up having to pass a number of ignores:  `--ignore-errors inconsistent --ignore-errors corrupt --ignore-errors source --filter missing --ignore-errors range`.

As for the need to ensure consistent `#line` entries in generated files.  sigh...

mdavidsaver:
See https://github.com/epics-base/epics-base/issues/749

anderslindho:
I would appreciate it if we could resolve this MR, either by merge, requested changes from me, or closing. I don't really want to have MRs in my "inbox" that are a year+ old - I would now need to invest effort to pick this back up because of the time that has passed since I initially worked on this.

I have slightly different directions requested or insinuated by different maintainers and no reply or comment on my earlier clarification, so it's not quite clear what is expected here.

mdavidsaver:
@anderslindho If you ask for my answer today, it is that `make coverage` is not the right interface.  The need to set additional `*FLAGS` variables is not clear, and the use of `lcov` is too specific.

A better fit might be in https://github.com/epics-base/ci-scripts as a special command.  There it is clear how the additional dependencies could be managed, the Makefile variables injected, and the results presented.  It would also not be tied to the epics-base release cycle.

PR #665, opened 2025-06-04
Minimize host compilation

anjohnson:
A new `HOST_ARCH_BUILDS` variable controls what program types (`Host`, `Ioc` and/or `Command`) should be built for the host architecture. The default set in `configure/CONFIG_SITE` builds everything as we have always done, but it can now be set to just `Host` in say `configure/os/CONFIG_SITE.linux-x86_64.Common` alongside `CROSS_COMPILER_TARGET_ARCHS` if that host won't be running any IOCs.

To build command-line tools like `caget` and `ascheck` for the host as well, set it to `Host Command`.

To prevent say the `linux-arm` architecture from being self-built (requiring cross-builds only), set `HOST_ARCH_BUILDS` to the empty string in the `configure/os/CONFIG_SITE.linux-arm.Common` and any attempt to run `make` on that CPU will error out.

If nothing else using this could speed up our CI build jobs; we have lots of duplicate compilation at the moment.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -82,7 +82,7 @@ matrix:
 #---------------------------------#
 
 install:
-  - cmd: git submodule update --init --recursive
+  - cmd: git submodule update --init --recursive .ci
   - cmd: pip install git+https://github.com/mdavidsaver/ci-core-dumper#egg=ci-core-dumper
   - cmd: python .ci/cue.py prepare
 
--- a/.appveyor/epics-base-7.yml
+++ b/.appveyor/epics-base-7.yml
@@ -89,7 +89,7 @@ matrix:
 #---------------------------------#
 
 install:
-  - cmd: git submodule update --init --recursive
+  - cmd: git submodule update --init --recursive .ci
   - cmd: pip install git+https://github.com/mdavidsaver/ci-core-dumper#egg=ci-core-dumper
   - cmd: python .ci/cue.py prepare
 
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -83,7 +83,8 @@ jobs:
             cmp: gcc
             configuration: static
             cross: "windows-x64-mingw"
-            name: "Ub-22 gcc + MinGW, static"
+            extra: "HOST_ARCH_BUILDS=Host"
+            name: "Ub-22 X- MinGW, static"
 
           - os: ubuntu-22.04
             cmp: gcc
@@ -95,7 +96,7 @@ jobs:
             cmp: gcc
             configuration: static
             extra: "CMD_CFLAGS=-funsigned-char CMD_CXXFLAGS=-funsigned-char"
-            name: "Ub-22 gcc unsigned char"
+            name: "Ub-22 gcc uchar"
 
           - os: ubuntu-22.04
             cmp: clang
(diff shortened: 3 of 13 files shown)


anjohnson:
Known issues:

Some RTEMS cross-build CI jobs are trying to create test scripts that didn't used to:
```
  makeTestfile.pl: I don't know how to create scripts for testing RTEMS-uC5282 on linux-x86_64`
  makeTestfile.pl: I don't know how to create scripts for testing RTEMS-beatnik on linux-x86_64
  makeTestfile.pl: I don't know how to create scripts for testing RTEMS-xilinx_zynq_a9_qemu on linux-x86_64
```
Those need fixing before implementing in `ci-scripts-build.yml` (should be able to override `HOST_ARCH_BUILDS` using `extra:` in the job matrix).

Documentation is missing; stress that using the right `PROD_HOST`, 'PROD_IOC' or 'PROD_CMD' and friends in Makefiles is now essential.

mdavidsaver:
> If nothing else using this could speed up our CI build jobs; we have lots of duplicate compilation at the moment.

How many hours would be saved from an appveyor run?

anjohnson:
> How many hours would be saved from an appveyor run?

None, we don't do cross-builds on Appveyor. We have been adding cross-build targets to GHA though, and those jobs seem to take between +50% and +280% longer than the native builds to complete.

Someone was asking for this not too long ago, and the changes to implement it were relatively minor.

mdavidsaver:
> cross-build targets to GHA though, and those jobs seem to take between +50% and +280% longer than the native builds to complete.

It has also been my impression that host compiles are faster.  eg. cross mingw seems to be especially slow.

I did [something](https://github.com/epicsdeb/epics-base/blob/3964bbfe28bef45167ef83d63c767b827c3a8a42/debian/patches/0008-SKIP_HOST-build-option.patch) comparable circa 2012 for epicsdeb, where I was also packaging cross-builds for some modules.  I abandoned the idea fairly quickly as the time savings did not make up for the increased complexity of the build process.  So I am skeptical that this PR would result in enough practical benefit to offset the added complexity and churn.

anjohnson:
The failures in the latest version of this PR occur when GHA is trying to compile the PVA submodules after only compiling the host-specific code for the host target. It should be possible to prepare the necessary changes to those modules without breaking backwards compatibility, but I haven't got to working on them yet.

I guess this is an indication that downstream modules are likely to need similar fixes too. The speed-up to our own CI builds may still makes these changes worth doing, we'll have to see.

gilesknap:
@anjohnson this will be really useful in container builds where we want to keep the image small and only compile for the target architecture. (we don't need the submodules as we have switched to pvxs). 

Am I correct that we could just build `Command` in the native architecture for epics-base? AFAIR I'm building all of epics base for target and host in my RTEMS builds at present.

mdavidsaver:
> useful in container builds

DLS does cross-compiling for RTEMS in a container?

> we want to keep the image small and only compile for the target architecture

The one does not imply the other.  eg. for epicsdeb packaging, I originally included RTEMS cross-builds.  There the process was to build everything, then slice down to what was desired.  Of course deb/rpm packaging tool evolved to make this easy, while docker/podman have not (yet?).  But then, the quest for small container images frequently leads to some contorted scripting.

gilesknap:
> > useful in container builds
> 
> DLS does cross-compiling for RTEMS in a container?
> 
> The one does not imply the other. eg. for epicsdeb packaging, I originally included RTEMS cross-builds. There the process was to build everything, then slice down to what was desired. Of course deb/rpm packaging tool evolved to make this easy, while docker/podman have not (yet?). But then, the quest for small container images frequently leads to some contorted scripting.

Yes we build RTEMS in a container build. Its part of a process of keeping the development and management of all IOCs as similar as possible between are various targets.

In truth the size is not so important because the build stage hands off the built RTEMS binary to the runtime stage. The runtime connects to the crate, configures it and gets it to boot off of the binary. 

Size is only an issue because we use the build stage as our development environment but also build times are increased by building for the host architecture when we don't need it (we have separate container builds for x86, other arch).

> The one does not imply the other.

I'm not sure I understand, if I never build for the host arch then I have less binaries right?

anjohnson:
I worked out the GHA configuration changes to be able to set host-builds for individual jobs. This required creating a `configure/CONFIG_SITE.local` file, since the ci-scripts `cue.py` only appends `EXTRAS` to the command-line for the `build` step, but the configuration needs to be the same for the `test` and `test-results` steps as well.

These build jobs are significantly faster than the normal Base jobs because these aren't compiling the external submodules at the moment. Those modules will need fixes to their Makefiles before I enable them again and this PR could be merged. I marked the job-names with "X-" to indicate the ones where the host is only building `Host` targets. The "Xc" marks a job building both `Host` and `Cmd` targets.

@gilesknap
> Am I correct that we could just build Command in the native architecture for epics-base?

You will always need `Host` which installs the build scripts and compiles libCom and the flex & antelope programs (lex and yacc) that are written in C; adding `Cmd` adds the CA client library and catools. Everything else comes from building `Ioc`.

JJL772:
this PR would really help with packaging EPICS for embedded Linux with Yocto. we're starting to hit issues with EPICS modules and IOCs that pull in a lot of dependencies, as those dependencies need to be available on the build host as well. This isn't a problem for EPICS base and the more simple modules (everyone has libreadline-dev installed anyway), but it starts to become a problem with the areaDetector packages and whatnot.

PR #671, opened 2025-06-26
call atExit functions on SIGTERM and SIGINT

dirk-zimoch:
For Unix-like hosts install signal handlers for `SIGTERM` (what parent processes like procServ usually send to kill the IOC) and `SIGINT` (aka CTRL-C) which `close(STDIN_FILENO)`. That causes the IOC shell to terminate (whether using `readline()` or not) after calling the installed exit handlers.


modules/libcom/src/misc/epicsExit.c
```diff
@@ -203,6 +211,40 @@ LIBCOM_API void epicsExitLater(int status)
...
+{
+    /* try to clean up before exit */
+    if (atExitDebug)
+        fprintf(stderr, "\nTerminating signal %s (%d) cought by %s\n", strsignal(sig), sig, epicsThreadGetNameSelf());
+
+    /* restore original handlers */
+    sigaction(SIGINT,  &old_handler_SIGINT,  NULL);
+    sigaction(SIGTERM, &old_handler_SIGTERM, NULL);
+
+    epicsExitCallAtExits(); /* will only start executing handlers once */
```
mdavidsaver: It is not safe to call `epicsExitCallAtExits()` from a signal handler.  eg. at exit handlers frequently lock a mutex.

mdavidsaver (changes requested):

mdavidsaver:
My suggestion for a starting point is for interactive `iocsh()` to install/remove a signal handler which enables SIGINT to cause `iocsh()` to return.

modules/libcom/src/misc/epicsExit.c
dirk-zimoch: I know. This is just to show you what I have used to make it work for me after the discussion in #558.

jerzyjamroz:
Practical question, why not to add it into [epics-base/modules/database/src/template/top/iocApp/src/_APPNAME_Main.cpp](https://github.com/epics-base/epics-base/blob/f911f8ca3eb94e0a7be3e3ed26166cc63ad8dac4/modules/database/src/template/top/iocApp/src/_APPNAME_Main.cpp)?:
```
int main(int argc,char *argv[])
{
    //it can be other function - it is just to show the idea.
    // [OPTION1]: std::signal(SIGTERM, handle_sigterm);  // <------- diff
    if(argc>=2) {
        iocsh(argv[1]);
        epicsThreadSleep(.2);
    }
    iocsh(NULL); //<-- [OPTION2] 'std::signal' inside init of iocsh(NULL) (OR iocsh(argv[1]))
    epicsExit(0);
    return(0);
}
```
I think option 2 might be cleaner.

dirk-zimoch:
Does not fix the general problem that the user supplied atExit functions may call any system functions, not only the "[sighandler approved](https://man7.org/linux/man-pages/man7/signal-safety.7.html)" ones.

jerzyjamroz:
@dirk-zimoch , thanks for the clarification. Would it be possible to add a new hook to `handle_sigterm` e.g. `initHookAtSigterm` ?

dirk-zimoch:
I tried to break the iocsh main loop, using a global `terminate` variable set by the signal handler, in order to bring the ioc to a "normal" exit. But readline interferes. Maybe next week....

jerzyjamroz:
Or to leave exactly empty `handle_sigterm`, so `SIGTERM` will be just catched then standard EPICS `exit` procedures will happen due to the lack of e.g. `exit(0)` in the `handle_sigterm`?

dirk-zimoch:
The standard reaction to `SIGTERM` is to terminate the program without running `exit()`. That is the root of the problem we want to solve here.

dirk-zimoch:
[`_exit()`](https://man7.org/linux/man-pages/man2/_exit.2.html) is a "signal safe" function and does *not* call any exit handlers. It is called by `SIGINT`, `SIGTERM`and some [other](https://man7.org/linux/man-pages/man7/signal.7.html) signals (some write a core dump first) by default.  [`exit()`](https://man7.org/linux/man-pages/man3/exit.3.html) is the function that calls installed exit handlers but it is illegal to call it from an asynchronous signal handler, that is one caused by an external source like CTRL-C rather than one caused by executing a program instruction like `SIGBUS`.

jerzyjamroz:
@dirk-zimoch , thank you for the details. I prepared the proposal modification in [iocshOnce](https://github.com/jerzyjamroz/epics-base/commit/07945150f7eb14c9c10a33d606aa5c5a4d323a3f)

In summary:

* `SIGTERM` for 7.0 epics-base branch
```
epics> var atExitDebug 1
Terminated: 15
```

* Standard exit
```
epics> var atExitDebug 1
epics> exit
atExit errlogExitHandler(0x7fcd73405e90)
atExit rlExit(0x0)
atExit syncShutdown(0x0)
atExit decShutdown(0x0)
atExit dbndShutdown(0x0)
atExit arrShutdown(0x0)
```

* `SIGTERM` exit after update with [iocshOnce](https://github.com/jerzyjamroz/epics-base/commit/07945150f7eb14c9c10a33d606aa5c5a4d323a3f)

```
epics> var atExitDebug 1
epics> 
atExit errlogExitHandler(0x7faecb305060)
atExit rlExit(0x0)
atExit syncShutdown(0x0)
atExit decShutdown(0x0)
atExit dbndShutdown(0x0)
atExit arrShutdown(0x0)
```
So it looks OK.

dirk-zimoch:
I have added `SIGINT`, because that is what CTRL-C sends.
Does not work:

```
epics> var atExitDebug 1
epics> ^C
epics> 
epics> exit
atExit exitDatabase((nil))
atExit twdShutdown((nil))
atExit errlogExitHandler(0x9a5600)
atExit rlExit((nil))
atExit syncShutdown((nil))
atExit decShutdown((nil))
atExit dbndShutdown((nil))
atExit arrShutdown((nil))
atExit ClockTime_Shutdown((nil))
```

Does not exit at all. Not that I had expected it to work.
Readline prints the `^C` and then calls our handler, which does nothing.
Still reacts on keyboard input normally afterwards.
Can then be terminated with `exit` regularly calling the exit handlers.

dirk-zimoch:
Sending `SIGTERM` does not work either. Do you build with GNU readline?

dirk-zimoch:
I tried without readline (`COMMANDLINE_LIBRARY = EPICS`)  but it still does not work. I send `SIGTERM` and nothing happens. I press CTRL-C and nothing happens. The ioc does not terminate.

anjohnson:
The direction this PR is taking doesn't seem right to me.

The [signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html) man-page is useful to know what _can_ be called from a signal handler. Remember that many of our Posix OSI implementations are also used by MacOS and recent RTEMS versions, and the behavior on linux-arm may be different than linux-x86_64, so we do need to follow what's legal.

The LibCom routine `epicsExitLater()` currently creates a small low priority thread "exitLater" that calls `exitNow()` → `epicsExit()` to shut down the IOC nicely. Maybe we could change how that works; create that thread during early startup and have it wait in `sigwait()` for some signal TBD (would `SIGABRT` work?) first. Then `epicsExitLater()` and any signal handlers can send that (or call `abort()`) to trigger a shutdown, followed by a call to `sleep()` to give that time to work but provide a fall-back by returning if it doesn't.

mdavidsaver:
> Sending `SIGTERM` does not work either. Do you build with GNU readline?

I see libreadline installing signal handlers for SIGINT and SIGTERM on each call (and restores afterwards).

```sh
strace -e %signal -f ./bin/linux-x86_64/softIoc -d tick.db
...
[pid 2127196] rt_sigaction(SIGINT, {sa_handler=0x7f8f32cca7f0, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, 8) = 0
[pid 2127196] rt_sigaction(SIGTERM, {sa_handler=0x7f8f32cca7f0, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, {sa_handler=SIG_DFL, sa_
...
epics>
[pid 2127196] rt_sigaction(SIGINT, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, {sa_handler=0x7f8f32cca7f0, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, 8) = 0
[pid 2127196] rt_sigaction(SIGTERM, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, {sa_handler=0x7f8f32cca7f0, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f8f32e5b050}, 8) = 0
...
```

mdavidsaver:
> I see libreadline installing signal handlers for SIGINT and SIGTERM

Ah, there is a global function pointer hook (`rl_signal_event_hook`) for application code.

mdavidsaver:
> Practical question, why not to add it into [epics-base/modules/database/src/template/top/iocApp/src/_APPNAME_Main.cpp](https://github.com/epics-base/epics-base/blob/f911f8ca3eb94e0a7be3e3ed26166cc63ad8dac4/modules/database/src/template/top/iocApp/src/_APPNAME_Main.cpp)?:

Changes to the template would not propagate to then many many instances already made.

mdavidsaver:
> The direction this PR is taking doesn't seem right to me.

I think there is understanding that only a limited set of functions are safe to call from a signal handler.

@jerzyjamroz @dirk-zimoch Would you both agree that we can not retroactively limit `epicsAtExit()` handlers to call only such signal-safe functions?

dirk-zimoch:
> Would you both agree that we can not retroactively limit `epicsAtExit()` handlers to call only such signal-safe functions?

Agreed. Even `printf()` is unsafe. That does not mean it will not work. Most of the time. But telling iocsh to terminate its main loop is probably the best approach.

mdavidsaver:
> ... telling iocsh to terminate its main loop is probably the best approach.

fyi.  I looked into what it takes to interrupt `readline()` or `getc()`.  On (at least) Linux, readline calls `select()` + `read()`.  I can not find any statement about how `getc()` should handle signal interruption (`EINTR`).

Neither of the following seem safe in a multi-threaded environment.  Especially an MT environment like at IOC, with plugins, where we can not constrain what all can happen concurrently.  An inopportune `fork()` could inherit some very strange state.

It looks like BASH [makes use of](https://github.com/bminor/bash/blob/6794b5478f660256a1023712b5fc169196ed0a22/bashline.c#L4818-L4823) the special `sigsetjmp()` / [`siglongjmp()`](https://www.man7.org/linux/man-pages/man3/siglongjmp.3p.html).  This seems to be [documented](https://tiswww.cwru.edu/php/chet/readline/readline.html#Readline-Signal-Handling) behavior for libreadline, which mentions calling `rl_cleanup_after_signal()` after jumping out of a signal handler.  Of course in a process with multiple threads, jumping out of a signal handler seems an invitation for UB insanity unless it could be guaranteed that the handler is running on the thread which called `sigsetjmp()`.

Another idea might be to `close(0)` from the signal handler.  Maybe using `dup()` and then `dup2()` to keep stdin usable afterwards.  Although, even if this works, there would be a brief time when other threads would find FD zero to be closed.

mdavidsaver:
> ... telling iocsh to terminate its main loop is probably the best approach.

I'm now questioning if this can be done safely in an IOC.  Maybe it would be better to externalize?  eg. teach procServ to handle SIGTERM by first closing the child's stdin.

ralphlange:
Even though I'm always reluctant to add more complexity to procServ - this sounds like a good idea.

jerzyjamroz:
> But telling iocsh to terminate its main loop is probably the best approach.

@dirk-zimoch , I think, after the `SIGTERM`, the iocsh main loop is unsafe (I would need to analize it more but it is a guess).
This technique that I presented to [catch the SIGTERM only](https://github.com/jerzyjamroz/epics-base/blob/07945150f7eb14c9c10a33d606aa5c5a4d323a3f/modules/libcom/src/iocsh/iocsh.cpp#L1595) in order to avoid the instant `exit` which allows IOC to execute the `exit procedures` does not violate anything (results [here](https://github.com/epics-base/epics-base/pull/671#issuecomment-3012821222)).

dirk-zimoch:
> teach procServ to handle SIGTERM

Does not solve CTRL-C or sending signals from other programs than procServ.

jerzyjamroz:
@dirk-zimoch @mdavidsaver , 
1) I wrote improved version in [here](https://github.com/jerzyjamroz/epics-base/commit/69bc9a11eda52b8cb4f4614630a73bad87759b8a) so explicit implementation which works for `SIGTERM` with `sa.sa_flags = 0; // clear SA_RESTART to close epicsReadline() on signal`

2) It does not work for:
```
    sigaction(SIGINT,  &sa, NULL);
```
which I don't understand at the moment.

dirk-zimoch:
I first tried to do something with `rl_signal_event_hook` (e.g. return a non-zero value) but nothing brings `readline()` out of its loop.
I have then modified Jerzy's  signal handler to actually do something: `close(STDIN_FILENO);` That brings `readline()` out of its loop and then terminates the ioc shell cleanly. `close()` is a signal-safe function.

Here I hit CTRL-C:
```
epics> var atExitDebug 1
epics> ^CatExit exitDatabase((nil))
atExit twdShutdown((nil))
atExit errlogExitHandler(0x18c9a90)
atExit rlExit((nil))
atExit syncShutdown((nil))
atExit decShutdown((nil))
atExit dbndShutdown((nil))
atExit arrShutdown((nil))
atExit ClockTime_Shutdown((nil))
```
The same works when `SIGTERM` is sent to the ioc.

Works with `COMMANDLINE_LIBRARY = READLINE` as well as with  `COMMANDLINE_LIBRARY = EPICS`

mdavidsaver:
> I first tried to do something with `rl_signal_event_hook`

Sorry for wasting your time.  I now conclude that this hook is useless for our present purpose.

From the libreadline [documentation](https://tiswww.cwru.edu/php/chet/readline/readline.html#Readline-Signal-Handling).

> When one of these signals is received, ..., reset the signal handling to what it was before readline() was called, and resend the signal to the calling application.

So readline tries to be transparent to application signal handling.

mdavidsaver:
> I have then modified Jerzy's signal handler to actually do something: `close(STDIN_FILENO);`
> ...
> Works with `COMMANDLINE_LIBRARY = READLINE` as well as with `COMMANDLINE_LIBRARY = EPICS`

If we can live with the irreversability of this behavior, then we have a solution.  I think I am ok with this as I would not want SIGTERM to do anything other than exit the process as soon as practical.  I think the same would apply to SIGINT.

dirk-zimoch:
Thanks to Jerzy and Michael for your input.

Some thoughts on non-Linux systems:
Windows seems not to need this (at least not in my wine setup, but maybe someone with a real Windows system can confirm that CTRL-C calls the exit handlers and then terminates -> `var atExitDebug 1`)
VxWorks does not need it, I believe, because nobody uses those signals there.
RTEMS maybe not either for the same reason.
I have not tested MacOS yet but am quite confident.
Any other relevant architectures?

I don‘t know how to write a test for this. One would need to sent a signal and test if the exit handlers are called and if the IOC terminates.

Release notes will follow…

anjohnson:
Doesn't this make it impossible to use Ctrl-C to kill the IOC when some iocsh command hangs? That doesn't seem acceptable to me.

mdavidsaver:
> Doesn't this make it impossible to use Ctrl-C to kill the IOC ...

Ok.  That is a point against handling SIGINT.  (ctrl+c is SIGINT, not SIGTERM)

Or if a repeat is acceptable, making sure the first action of that handler is to restore the default behavior for a second ctrl+c.  (or call `sigaction()` with `SA_RESETHAND`)

dirk-zimoch:
Would it be acceptable to reset the signal handlers to their previous values (or simpler: to the defaults), so that hitting CTRL-C a second time would kill the IOC (without running exit handlers).

dirk-zimoch:
Or do something like `alarm(5)` in the signal handler?

anjohnson:
We'd have to document (and sites would have to train users) any need to press Ctrl-C twice in some cases. That doesn't seem ideal to me though.

Wasn't the original point to try to get the IOC to shut down nicely? Adding `alarm(5)` to the signal handler doesn't achieve that, and potentially hijacks any IOC whose exit handlers takes more than 5 seconds to perform their closedown actions.

Has anyone tried my suggestion to have the `SIGINT`/`SIGTERM` handler send a signal to a thread that's waiting in `sigwait()`, which then calls `epicsExit()` to do the shutdown? It's more work to implement, but should allow cleanup routines to be run in more situations. I withdraw my earlier suggest to try using `SIGABRT` though, that particular choice won't work.

dirk-zimoch:
Implemented @anjohnson's suggestion. Looks good on Linux and MacOS with and without readline.
Test with blocking command (here on Mac without readline):
```
epics> var atExitDebug 1
epics> epicsThreadSleep 1000
^CSignal Interrupt: 2 received by exitOnSignal
atExit exitDatabase(0x0)
atExit twdShutdown(0x0)
atExit errlogExitHandler(0x600000605c80)
atExit rlExit(0x0)
atExit syncShutdown(0x0)
atExit decShutdown(0x0)
atExit dbndShutdown(0x0)
atExit arrShutdown(0x0)
atExit done.

$ echo $?
130
```
The return code is 130 = 128 + 2 (SIGINT).

dirk-zimoch:
Made all debug output dependent on `atExitDebug`. Tests seemed to be confused by unexpected output.

dirk-zimoch:
Maybe the priority of the exitOnSignal thread should be `epicsThreadPriorityMax`?

dirk-zimoch:
@anjohnson's idea to let `epicsExitLater` send a signal to the waiting thread requires the existence of the signal handler in all architectures, in particular VxWorks, RTEMS and Windows. Thus, in this implementation, I left the `epicsExitLater` implementation as it is.

dirk-zimoch:
... I have tested it anyway and it works on Linux. However I had to use `kill(getpid(), SIGTERM)` rather than `raise(SIGTERM)`. And now `epicsExitLater` has two different implementations, one for Linux/MacOS and one for others. They may behave differently in certain situations.

anjohnson:
Thanks, glad to see that actually works (my attempt was seg-faulting). Can you move the signal code that's specific to Linux & Darwin into a libcom/src/osi/os/posix file, maybe in osdThread.c, to avoid the conditionals? We should probably have a discussion whether we want all EPICS programs which link against libCom to have these signal handlers enabled or not, do other Core Dev's have an opinion on that? If so, the signals could possibly be set up in the `once()` routine.

I would also like to see the "exitLater" thread routine from `exitLaterOnceFunc()` and your new thread routine combined into one, making the `epicsExitLater()` routine signal-safe. That way the exit cleanup thread stays in epicsExit.c but can be called from the signal handlers that would live in osdThread.c. You might need to be careful about initialization though, I hope the `once()` routine can be made to work.

anjohnson:
@dirk-zimoch I missed answering your point earlier about implementing signal handlers on other OSs. I always meant to encapsulate the signal-specific stuff inside an libCom/src/osi/os/posix implementation once we proved it works, so the epicsExit.c code would call os-dependent routines as needed. Different implementations would be needed for the other OSs, on VxWorks the `sigwait()` could become a `semTake()` etc.

mdavidsaver:
> ... whether we want all EPICS programs which link against libCom to have these signal handlers enabled or not ...

imo. no.  This new handler should be limited to the duration of an interactive IOC shell.  Installed on entry, and removed on exit.

mdavidsaver:
> ... so the epicsExit.c code would call os-dependent ...

libreadline already installs/removes signal handlers.  So I think it makes sense for SIGINT handling should be encapsulated as part of epicsReadline.

dirk-zimoch:
I see a problem combining `exitOnSignal` thread and the `exitLater` thread. The signal handling thread should be able to interrupt high priority threads busy-blocking the CPU, even if we have more such threads than we have CPU cores. Otherwise we would not be able to end the IOC with Ctrl-C. Thus, this thread needs to run with the highest possible priority.

The stated purpose of `epicsExitLater` is to "trigger process exit without blocking". I am not sure about its requirements. Its only use case in base is as a sub record routine in softIocExit.db. To achieve this, the handling thread runs on low priority, so that the calling higher priority thread may complete its work. Some documentation (beyond a commit message) which describes not only what `epicsExitLater` does but also why that may be useful would be good, I think. That would help not to break stuff on later refurbishments.

As the priority demands of the two threads are incompatible, I think we should keep these two separate. That would allow to keep the `epicsExitLater` behavior consistent across implementations and add `exitOnSignal` for those architectures where signal handling makes sense (i.e. Posix systems).

ralphlange:
> > ... whether we want all EPICS programs which link against libCom to have these signal handlers enabled or not ...
> 
> imo. no. This new handler should be limited to the duration of an interactive IOC shell. Installed on entry, and removed on exit.

Strongly support NO.
It would apply to clients, middle-layer services, etc. etc. where this is not wanted and not expected.
Tying it to the IOC shell sounds right.

dirk-zimoch:
> Tying it to the IOC shell sounds right.

So I will put it into iocsh.cpp rather than epicsExit.c.

modules/libcom/src/iocsh/iocsh.cpp
```diff
@@ -1587,8 +1593,34 @@ static void exitCallFunc(const iocshArgBuf *)
...
 }
 
+#ifdef HANDLE_SIGNALS
+static void exitOnSignal(void* arg)
+{
+    sigset_t* psigset = reinterpret_cast<sigset_t*>(arg);
+    int sig;
+
+    sigwait(psigset, &sig);
+    epicsExit(128+sig);
```
dirk-zimoch: Should I unblock the signals here? Test with a busy-blocking atExit handler.

dirk-zimoch: What works:
* Busy-blocking shell command interrupted with Ctrl-C runs exit handlers and terminates.
* Busy-blocking exit handler, terminate with exit, then Ctrl-C to interrupt blocking handler. Does not run any further exit handlers.

What does  not work:
* Busy-blocking exit handler. Ctrl-C. Running exit handlers stalls, we still have a shell, and further Ctrl-C does nothing. `exit` still works but does not run any more handlers.

anjohnson:
I don't understand why this should be implemented in iocsh, instead of providing an API which iocsh calls to enable it. Other portable programs that don't use iocsh may also want to do a graceful shutdown when they receive a `SIGINT` or `SIGTERM`. If this code were only required to handle a Ctrl-C from a terminal window I would agree with @mdavidsaver's opinion that this should be part of epicsReadline, but that's not the only use-case.

It's very common for a `[service]` entry in a systemd unit file to set `ExecStop=kill $MAINPID` which running `systemctl stop middle-layer.service` causes first a `SIGTERM`, optional `SIGHUP` and then a `SIGKILL` to be sent to the process. I'd like to allow EPICS middle-layer services to be shut down gracefully just like IOCs, by calling the same API as iocsh to enable the `SIGTERM` handler.

The `epicsExitLater()` routine replaced a call to `epicsExit()` inside the sub record routine in softIocExit.db, because that triggered an error from the CA client which wrote to that PV when the IOC disappeared before replying to the `ca_put()` command. The idea was to allow the IOC to go quiescent before it got taken down, hence the use of a low priority thread.

On the question of thread priorities, I'm not trying to make this bullet-proof. Nowadays it seems that only VxWorks and RTEMS IOCs ever let a high priority thread completely lock out threads running at lower priorities. Other OS schedulers seem to reserve and distribute _some_ CPU time to other threads — has anyone ever seen RSRV become unresponsive when another thread gets into a tight loop or gets very busy on other OSs? Thus I wouldn't have a problem running the thread at a lower priority as long as a second Ctrl-C unleashes a bigger hammer.

mdavidsaver:
> I don't understand why this should be implemented in iocsh, instead of providing an API which iocsh calls to enable it.

The specific behavior of `close(0)` is about interrupting `readline()` and `getc(stdin)` specifically.  For an example of what a more general purpose API looks like, see [`pvxs::SigInt`](https://github.com/epics-base/pvxs/blob/2509525b60841ea03d6d29a545bcd3be4c177c0f/tools/monitor.cpp#L123-L126).

mdavidsaver:
> ... shut down gracefully ...

All of the daemons I have looked into, or written myself, perform a graceful shutdown by returning from `main()`.  While I think it was the best way to resolve the situation with iocStats calling `epicsExit()` from a scan thread, I do not consider `epicsExitLater()` to be a graceful shutdown in a general sense.

dirk-zimoch:
> providing an API which iocsh calls to enable it

It needs to be installed early, I think. New threads inherit their parent's signal mask. Thus, I want it installed before any other thread is created. Calling something from iocsh like an `installSignalHandling` command to be called from a startup script may be too late.

anjohnson:
@mdavidsaver Since calling `close(0)` only works if the shell is waiting for input at the time, @dirk-zimoch was working on a signal handler approach for other situations which are more likely to occur in IOCs that run third-party code. You apparently already developed code in pvxs which might be able to solve this, so I withdraw from trying to offer further suggestions.

mdavidsaver:
> Since calling `close(0)` only works if the shell is waiting for input at the time

Ah, so you are thinking about a race condition?

I think there is one, although perhaps not the one you (or I) was thinking of.  On Linux, closing will both interrupt a concurrent `select()` or `read()` (what readline does), and fail any later calls (with `EBADF`).  So all well and good... unless FD 0 is reused quickly.  eg. by an inconveniently timed `accept()`.

My core take away from this is a slight headache, and a reminder that sane *NIX signal handling in a multi-threaded environment is nearly impossible.

> @dirk-zimoch was working on a signal handler ... likely to occur in IOCs ...
> You apparently already developed code in pvxs which might be able to solve this

`SigInt` is a utility to call from `main()`, not from an IOC hook.  it handles (scoped) add/remove of a signal handler, and the mechanics of executing the provided callback from a worker thread (which can be interrupted and joined on normal exit).

There may well be enough use cases to merit an equivalent API in Base.  I used `SigInt` for `pvxmonitor`, so `camonitor` would be one candidate.

> ... I withdraw from trying to offer further suggestions.

Just as I am coming around to your earlier point.

> > The direction this PR is taking doesn't seem right to me.

I am going to look at updating procServ to close the child stdin.

mdavidsaver:
> It needs to be installed early

@dirk-zimoch This is due to usage of `sigwait()`?  I have never used this function myself.  I would be concerned about depending on a specific signal being masked by every thread in a process.  This is notoriously hard to arrange, at least on Linux.

anjohnson:
> you are thinking about a race condition?

No, anything like the user running a registered iocsh command provided by a badly-written support module which doesn't return until the network device it expects to talk to actually responds to the connection request, i.e. never if they typed a bad IP address. Dirk was using a long `epicsThreadSleep` command during testing to simulate that.

I just want Ctrl-C (twice if necessary) on the IOC console to be able to stop the IOC process in any foreseeable situation where the main thread could hang or die but leave other parts of the IOC running. Hopefully our on-call personnel would ask systemd to stop or restart the IOC if it happened here, but that might not be thought of until after the 3am phone call.

> updating procServ to close the child stdin.

We only run a small subset of our IOCs under procServ, so that won't be a universal solution. My Perl `siocLaunch` wrapper which runs the other IOCs does pass a `SIGINT` down to the IOC, although systemd actually sees the GNU Screen process which runs that as `$MAINPID` so that feature isn't really being used yet.

mdavidsaver:
> No, anything like the user running a registered iocsh command provided by a badly-written support module ...

Installing our SIGINT/TERM handlers only when iocsh() is waiting for interactive input moots this concern.  It should not interfere with some support module messing with stdin and/or signal handlers to eg. implement iocsh command interruption.  I am not endorsing such games, just pointing out that what I show with #673 would not break them.

> I just want Ctrl-C (twice if necessary) on the IOC console to be able to stop the IOC process in any foreseeable situation  ...

You and I are in violent agreement on this point.

> We only run a small subset of our IOCs under procServ ...

I understand Andrew.  Any process manager would have to implement close-wait-KILL, or TERM-wait-KILL, logic to take advantage of this PR.  procServ is the one I would look at.  It sounds like your script runs with systemd, which could be deputized to do this since it will prune any lingering children when a service is stopped.

dirk-zimoch:
I have added an exit handler timeout environment variable. This allows procServ to send only one SIGTERM and still reliably terminate an ioc with hanging exit handlers. If the variable is undefined (or <= 0), no timeout is used. That is the default.

I have moved the os specific code to the pre-existing libcom/osi/os/posix/osdSignal.cpp file.
I noticed the identical ("do nothing") versions for "default", "WIN32", "vxWorks" and "RTEMS-score". Is this necessary?
 * VxWorks uses the "posix" version if no "vxWorks" version is provided. That looks wrong to me. It should use "default".
 * Without a "WIN32" version, Windows uses "default". A separate "WIN32" version seems unnecessary.
 * RTEMS 5.1 seems to compile fine with the "posix" version. Is a "RTEMS-score" version necessary? (What is "score"?)

jerzyjamroz:
Just to clarify, from my point of view, the important signal was only `SIGTERM`. This is because supervising software uses `SIGTERM` to stop/restart processes, so [I suggested controlling](https://github.com/jerzyjamroz/epics-base/commit/69bc9a11eda52b8cb4f4614630a73bad87759b8a) the universal (without threads) [SA_RESTART](https://www.gnu.org/software/libc/manual/html_node/Flags-for-Sigaction.html) flag to make sure readline closes and the process exits cleanly. The supervisor will send `SIGKILL` later if the process does not stop in time, so there is no risk of the process hanging.

I don’t think handling `SIGINT` is very important. If someone is working locally, they can just use "exit" to quit normally, or press `CTRL+C` if something goes wrong, this usually happens during development, and people accept that behavior.

Of course, it’s up to you what you want to include in epics-base. I just want to make it clear that I am not asking for features I don’t consider important.

jerzyjamroz:
I tested it and the terminal is left in a broken state.

Steps to reproduce:

```console
$ ./softIocPVA
epics> var atExitDebug 1
epics> ^C
atExit rlExit(0x0)
atExit syncShutdown(0x0)
atExit decShutdown(0x0)
atExit dbndShutdown(0x0)
atExit arrShutdown(0x0)
atExit done.
atExit errlogExitHandler(0x6000038f1a40)
atExit done.
```
After returning to the shell, the terminal no longer behaves normally. (click enter few times).

Issue #670, opened 2025-06-25
Controlled behaviour of the IOC shutdown (follow-up PR #558)


jerzyjamroz:
The idea of this ticket is to find out the best solution for handling hardware registry/config modifications upon IOC shutdown in a standardized way, whereas the IOC process is managed by e.g. systemd.

It is a follow-up to the PR #558 design conclusions.

jerzyjamroz:
@anjohnson @mdavidsaver , [this line](https://github.com/epics-modules/mrfioc2/blob/e71217d79588ea4dad7d83368af79f0dfae59c45/evrMrmApp/src/drvemSetup.cpp#L981) is never executed upon `SIGTERM`.

jerzyjamroz:
@anjohnson @mdavidsaver , I did tests after your comment, and this handler is never executed after `SIGTERM`: [epicsAtExit](https://github.com/epics-modules/mrfioc2/blob/e71217d79588ea4dad7d83368af79f0dfae59c45/evrMrmApp/src/drvemSetup.cpp#L981), so `evrShutdown`. But one can make it work only if one registers a `SIGTERM` handler (which is not done by default) e.g. `std::signal(SIGTERM, handle_sigterm);` without `exit` in it, which allows then to execute `evrShutdown`.

mdavidsaver:
fyi. the `epicsAtExit()` handlers are run by the `exit` iocsh command, and when `stdin` is closed (`ctrl+d` by default for most terminal emulators).  In both cases this happens by causing `iocsh(NULL)` to return.  In the default IOC template this is followed by a call to `epicsExit()`.

eg.

```
epics> var atExitDebug 1
epics> exit
atExit rlExit((nil))
atExit exitDatabase((nil))
atExit twdShutdown((nil))
atExit errlogExitHandler(0x5626542c0aa0)
atExit syncShutdown((nil))
atExit decShutdown((nil))
atExit dbndShutdown((nil))
atExit arrShutdown((nil))
```

https://github.com/epics-base/epics-base/blob/f911f8ca3eb94e0a7be3e3ed26166cc63ad8dac4/modules/database/src/template/top/iocApp/src/_APPNAME_Main.cpp#L24-L25

anjohnson:
It's not just the iocsh `exit` command; in the osi/posix/osdThread.c::`once()` routine we do this, which catches Ctrl-C and other exit paths:

[...]

PR #673, opened 2025-07-06
WIP: readline: ctrl+c close stdin

mdavidsaver:
My attempt at #670.  Provokes `readline()` and `getc()` to return immediately, and so also interrupts interactive `iocsh(NULL) `.  Only interrupts `iocsh(NULL)`, just as an `exit` iocsh command currently does.  Depends on IOC skeleton to 

Consequential differences with #671.

- Does not call `epicsExit()`.
- Instead of `close(0)`, uses `dup2()` to atomically replace stdin with `/dev/null` to avoid potential race with re-use of FD zero.
- Avoids a new worker thread
- Signal handler only installed while waiting for input.  Same as libreadline, so no new scope for conflicts.
- Uses `sigaction()` to also save/restore handler flags.
- Set `SA_RESETHAND` handler flag so that a repeated signal will trigger the default handler.
- No changes to thread signal masks.  (does not matter which thread runs this handler)

Unnecessary differences, which could/should be removed.

- ~~A bunch of calls to `fprintf()` for troubleshooting, which is not signal-safe.~~
- Clones, and later replaces, original stdin after interruption.  Theoretically allows calling `iocsh(NULL)` in a loop with multiple interruptions.
  - Second `dup2()` to restore stdin loses FD flags.  (could save/restore flags as well)
- Set `SA_RESTART` flag to approximate BSD behavior.  Maybe reduces testing matrix.
- Does not handle `SIGTERM`.

### DIFF
--- a/modules/libcom/src/osi/epicsReadline.c
+++ b/modules/libcom/src/osi/epicsReadline.c
@@ -11,9 +11,16 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <errno.h>
+#include <string.h>
 
 #include "envDefs.h"
+#include "epicsAssert.h"
 #include "epicsReadlinePvt.h"
+
+#ifndef SA_RESTART
+// request BSD compatible EINTR handling from Linux
+#  define SA_RESTART (0)
+#endif
 
 static void osdReadlineBegin(struct readlineContext *);
 static char * osdReadline(const char *prompt, struct readlineContext *);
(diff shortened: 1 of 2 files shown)


mdavidsaver:
@dirk-zimoch  I wrote this code as I do not think I was explaining myself properly in the #671 thread.  I am happy to leave this PR as an example.

mdavidsaver:
fyi. testing with modified ralphlange/procServ#70 with `--grace-period` argument.

```sh
./procServ -d -f --killsig 2 --grace-period 3 5678 ~/epics/base-git/bin/linux-x86_64/softIoc
...
epics> var atExitDebug 1
```

Concurrent `killall -TERM procServ` triggers:

```
epics> SigTerm received
Sending signal 2 to pid 802292
Start child cleanup timer 3 sec.
atExit errlogExitHandler(0x55e14e855930)
atExit rlExit((nil))
atExit syncShutdown((nil))
atExit decShutdown((nil))
atExit dbndShutdown((nil))
atExit arrShutdown((nil))
processItem: Got error reading input connection: Input/output error

@@@ @@@ @@@ @@@ @@@
@@@ Received a sigChild for process 802292. Normal exit status = 0
Deleting connection 0x55a192657c90
@@@ Current time: Sun Jul  6 17:02:01 2025
@@@ Child process is shutting down, oneshot mode: server will exit
~connectionItem()
child exits
Close sockets
~clientItem(); handle 0 closed
~connectionItem()
~acceptItem()
~connectionItem()
Cleanup pid and info files
```

mdavidsaver:
Rebased with no changes.  My casual testing over the past year has not turned up any misbehavior.

jerzyjamroz:
It works with SIGINT (kill -2 90763), but it does not work with SIGTERM (kill -15),so systemd default config, as described above.

Conclusion:
The systemd service can be configured to use SIGINT for shutdowns by adding the following directive:
```
[Service]
KillSignal=SIGINT
```

jerzyjamroz:
Another test with `ctrl-d` - it works fine.
```
./softIocPVA
epics> var atExitDebug 1
epics> epicsThreadSleep 10
^D
atExit rlExit(0x0)
atExit syncShutdown(0x0)
atExit decShutdown(0x0)
atExit dbndShutdown(0x0)
atExit arrShutdown(0x0)
atExit errlogExitHandler(0x6000033540c0)
```

jerzyjamroz:
SIGINT also works fine while the IOC is busy.
```
./softIocPVA
epics> var atExitDebug 1
epics> epicsThreadSleep 100000
```
Sending SIGINT stops it correctly.

PR #675, opened 2025-07-08
feat: support building linux target from darwin host

jerzyjamroz:
* darwin-x86 host - linux-x86_64 target builds.
* darwin-aarch64 host - linux-x86_64 target builds.
* If all is ok, I will prepare the `.md` release notes.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -195,6 +195,12 @@ jobs:
             name: "Cross linux-arm gnueabihf"
             cross: linux-arm@arm-linux-gnueabihf
 
+          - os: macos-latest
+            cmp: gcc
+            configuration: default
+            name: "Cross macOS x86_64-linux-gnu"
+            cross: linux-x86_64
+
     steps:
     - uses: actions/checkout@v4
       with:
@@ -206,6 +212,11 @@ jobs:
         sudo apt-get update
         sudo apt-get -y install qemu-system-x86 g++-mingw-w64-x86-64 gdb
       if: runner.os == 'Linux'
+    - name: Install cross toolchain for macOS
+      if: runner.os == 'macOS' && matrix.cross
+      run: |
+        brew tap messense/macos-cross-toolchains
+        brew install x86_64-unknown-linux-gnu
     - name: Prepare and compile dependencies
       run: python .ci/cue.py prepare
     - name: Build main module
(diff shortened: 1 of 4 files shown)


anjohnson:
@jerzyjamroz Do you have a real need for this cross-build combination? I don't particularly object to adding it if you do, but we don't like including builds that are just done as an academic exercise since they break if not regularly exercised. If there is a need, should we be setting up a GHA CI build to test it as well?

jerzyjamroz:
@anjohnson , this will not be used for production purposes, but it helps during development if one wants fast to test the Linux build on mac OS (or to track the whole cross compilation process). I noticed that this was missing in `epics-base` so I share it. I do not see any drawback of it as well, except benefits for EPICS macOS users.

jerzyjamroz:
@anjohnson , I added GHA, but it had to be for aarch64 (I use intel) as it is the only thing supported by `macos-latest` ci - but it is fine as well.
The problem I found is that this [cross](https://github.com/epics-base/epics-base/pull/675/files#diff-f734265827ede1e179c7638080c2ea333d346a46206938d495dfb12db36a6afaR202) variable does not assign `CROSS_COMPILER_TARGET_ARCHS = linux-x86_64`. Is there any special syntax for that? If I hardcode that in `CONFIG_SITE` then it works, see [here](https://github.com/jerzyjamroz/epics-base/actions/runs/16167285422/job/45632062736#step:7:38). How to use properly this `cross` variable?
Please, also check if the structure I did for that is correct as the same result can be achieved in many ways at the end.

anjohnson:
It looks like the ci-scripts module doesn't support cross-builds from MacOS, it only works on Intel-based Linux hosts. One obvious problem is that the `prepare_linux_cross()` function in cue.py can only edit `configure/os/CONFIG_SITE_` files where the host architecture is linux-x86_64. This is why `CROSS_COMPILER_TARGET_ARCHS` isn't being set automatically for you. There might be other changes needed too, I'm not as familiar with the code there.

A related design issue in ci-scripts which I also see is that in the job configuration files such as `.github/workflows/ci-scripts-build.yml` the `CMP` environment variable should be set to `clang` for your host and `gcc` for the cross-target, but there's only one variable here. Since Clang does imitate GCC this apparently isn't necessary in your case, but I'd prefer to have separate `HOST_CMP` and `CROSS_CMP` settings.

The ci-scripts module should also be responsible for installing the cross-compiler and not the GHA build script, that way downstream support modules could also test this combination.

ralphlange:
Could you please create tickets for these issues in the ci-scripts tracker?

anjohnson:
- epics-base/ci-scripts#100
- epics-base/ci-scripts#101
- epics-base/ci-scripts#102

PR #677, opened 2025-07-14
gethostbyname -> getaddrinfo

mdavidsaver:
Prefer reentrant DNS API to avoid global locking.

RTEMS classic network stack provides the declarations (inherited from newlib), but provides no implementation.

Windows also has getaddrinfo.  However their gethostbyname is reentrant, so there is no benefit to switching.

### DIFF
--- a/modules/libcom/src/osi/os/posix/osdSock.c
+++ b/modules/libcom/src/osi/os/posix/osdSock.c
@@ -41,6 +41,13 @@
 #  define SOCK_CLOEXEC (0)
 #endif
 
+#if defined(AI_PASSIVE) && !defined(__rtems__)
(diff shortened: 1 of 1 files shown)


mdavidsaver:
fyi. this change was motivated by an IOC I worked on with 6 asyn IP ports connecting by hostname to (on my laptop) non-existent hostnames.  This led to painfully slow IOC startup and shutdown.  So slow that I suspected a deadlock.  However, it was just 6 threads looping around to hostToIPAddr().  Allowing concurrency had a dramatic effect.

dirk-zimoch:
Compiles on all my target architecture, including old Linux versions, where it automatically falls back to `gethostbyname()`.

PR #678, opened 2025-07-15
allow hex and octal strings in dbPut and dbGet

dirk-zimoch:
It is now possible to convert hex and octal strings writing to integer fields with `dbPut()`, `dbGet()` and related functions like the iocsh command `dbpf` or through database links. This affects `caput` only if a `DBR_STRING` type is used to write integers.

Possible incompatibility: Up to now, leading `0`s have been ignored, now they switch to octal mode.

For backward compatibility, this behavior can be switched off, returning to the old decimal only conversions, by setting the environment variable `EPICS_DB_CONVERT_DECIMAL_ONLY` to `YES` (case insensitive) before `iocInit`.

### DIFF
--- a/documentation/new-notes/PR-678.md
+++ b/documentation/new-notes/PR-678.md
@@ -9,5 +9,4 @@ now they switch to octal mode.
 
 For backward compatibility, this behavior can be switched off, returning
 to the old decimal only conversions, by setting the environment variable
-`EPICS_DB_CONVERT_DECIMAL_ONLY` to anything except the empty string or
-strings starting with `n`, `N`, `f`, `F` or `0` before `iocInit`.
+`EPICS_DB_CONVERT_DECIMAL_ONLY` to `YES` (case insensitive) before `iocInit`.
--- a/documentation/new-notes/PR-678.md
+++ b/documentation/new-notes/PR-678.md
@@ -1,0 +1,7 @@
+### Allow hex and octal stings in dbPut
+
+It is now possible to write hex and octal strings to integer
+fields with `dbPut()` and related functions.
+
+Possible incompatibility: Up to now, leading `0`s have been ignored,
+now they switch to octal mode.
--- a/documentation/new-notes/PR-678.md
+++ b/documentation/new-notes/PR-678.md
@@ -1,7 +1,13 @@
-### Allow hex and octal stings in dbPut
+### Allow hex and octal strings in dbPut and dbGet
 
-It is now possible to write hex and octal strings to integer
-fields with `dbPut()` and related functions.
+It is now possible to convert hex and octal strings to integer fields
+with `dbPut()`, `dbGet()` and related functions like the iocsh command
+`dbpf` or through database links.
 
 Possible incompatibility: Up to now, leading `0`s have been ignored,
 now they switch to octal mode.
+
+For backward compatibility, this behavior can be switched off, returning
+to the old decimal only conversions, by setting the environment variable
+`EPICS_DB_CONVERT_DECIMAL_ONLY` to anything except the empty string or
+strings starting with `n`, `N`, `f`, `F` or `0` before `iocInit`.
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -59,6 +59,7 @@
 #include "devSup.h"
 #include "drvSup.h"
 #include "epicsRelease.h"
+#include "epicsConvert.h"
 #include "initHooks.h"
 #include "iocInit.h"
 #include "link.h"
@@ -129,6 +130,16 @@ static int iocBuild_1(void)
         errlogPrintf("iocBuild: " ERL_ERROR " Aborting, bad database definition (DBD)!\n");
         return -1;
     }
+
+    {
+        /* fall back to "old" decimal-only conversion if EPICS_DB_CONVERT_DECIMAL_ONLY
+           is set and does not look like meaning no/false/0 */
+        const char* dec_only = getenv("EPICS_DB_CONVERT_DECIMAL_ONLY");
+        if (dec_only && *dec_only && !strchr("nNfF0", *dec_only)) {
+            dbConvertBase = 10;
+        }
+    }
+
     epicsSignalInstallSigHupIgnore();
     initHookAnnounce(initHookAtBeginning);
 
--- a/modules/database/src/ioc/misc/iocInit.c
+++ b/modules/database/src/ioc/misc/iocInit.c
@@ -31,6 +31,7 @@
 #include "epicsPrint.h"
 #include "epicsSignal.h"
 #include "epicsThread.h"
+#include "epicsString.h"
 #include "errMdef.h"
 #include "iocsh.h"
 #include "taskwd.h"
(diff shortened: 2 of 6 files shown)


documentation/new-notes/PR-678.md
```diff
@@ -0,0 +1,7 @@
+### Allow hex and octal stings in dbPut
+
+It is now possible to write hex and octal strings to integer
+fields with `dbPut()` and related functions.
+
+Possible incompatibility: Up to now, leading `0`s have been ignored,
+now they switch to octal mode.
```
anjohnson, line 8: The 3.15 branch still has the code before we switched to using the `epicsParse*()` routines in dbConvert.c. That did the conversions using `sscanf()`:

```C
        if (sscanf(psrc,"%hd",&value) == 1) {
                *pbuffer = (char)value;
                return 0;
        } else if (strlen(psrc) == 0) {
                *pbuffer = 0;
                return 0;
        } else {
                return -1;
        }
```

The `"%d"` decimal specifier was deliberately used instead of `"%i"` integer, which is why I used base 10 in the `epicsParse*()` calls when converting them. (BTW, check a recent Linux man-page for `sscanf(3)`; all of the numeric conversion specifiers have been marked as **deprecated** since C11 because of the possibility of UB, see the BUGS section for more).

The thing that concerns me with your change is that an existing IOC application that reads a stringin from say a network device which includes leading zeros in data values — not uncommon for some instruments such as GPIB. A DB or CA link reading that string into an AI record would break when upgraded.

We could make the change optional though by adding another global variable, named say `dbConvertStringBase` to dbCore.dbd and dbConvert.c, which would be used as the number base for all those conversions. That allows an IOC which relied on the old behavior to switch back to that with a `var dbConvertStringBase 10` or `dbConvertStringBase=10` for VxWorks in the st.cmd file.

dirk-zimoch: I think the undefined behavior relates to the situation where a format, e.g. `%i` does not match the argument type, e.g. `long*` or `double*`.

dirk-zimoch: I could make the change so that only decimal and hex is recognized. That would not change the behavior of previously valid input. It would only make some previously invalid input valid. But that is a bit slower as I would need to find the first non-space character, test it for `0` and test the next character for `x`.

Or first try with 10 and if that returns 0 and the next char is `x` or `X`, retry with 16.

dirk-zimoch: Why did the UInt64 conversion already use `0` ?
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbConvert.c#L355
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbFastLinkConv.c#L215

anjohnson: The `UInt64` conversions used base=0 because they were a new data type, so no existing IOC could have been relying on the "old" behavior. Changing the record type would require re-testing, so that becomes an application issue.

modules/libcom/src/misc/epicsStdlib.h
```diff
@@ -50,6 +50,9 @@ extern "C" {
 /** Return code for `Number base not supported` */
 #define S_stdlib_badBase      (M_stdlib | 5) /* Number base not supported */
 
+/** Base for epicsParse* to accept hexadecimal or decimal strings */
+#define HEX_OR_DEC (-1)
+
```
anjohnson: I really don't like this and the changes to the `epicsParse*()` functions, sorry. Please drop your latest commit.

What was wrong with my suggestion that you just replace the literal `10` base arguments in all the calls with a single global variable which sites or IOCs needing the old behavior can configure? I actually liked the change that string-to-int conversions could accept values in any base (that's why I used `0` for the new UInt64 conversion cases), we just need to provide an escape hatch for old IOC applications which can't use that.

dirk-zimoch: > I really don't like this and the changes to the `epicsParse*()` functions

No? What bothers you?

anjohnson: The original design of those functions deliberately used the number base argument from the C standard `strtol()`, `strtoul()` and family, which is a beautiful and clean solution IMHO.

This change also affects the `epicsScan*()` macros without mentioning that, someone looking at the [generated documentation](https://epics--678.org.readthedocs.build/projects/base/en/678/epicsStdlib_h.html#c.HEX_OR_DEC) for them won't see that without a little digging.

I actually would like to change the conversion behavior to what your first commit was doing, just add the escape hatch for old IOCs that might need the original behavior. Escape hatches don't have to be beautiful, just functional.

dirk-zimoch: Would you prefer the escape hatch default to the new or to the old behavior?

dirk-zimoch: Compile time or run time switch?

dirk-zimoch: > a global variable, named say `dbConvertStringBase` [...], which would be used as the number base for all those conversions

What I don't like with that approach is that it would accept not only 0 and 10 but any number, e.g. 13 or -5, which could yield confusing results for the innocent user. I rather have a choice "old (base10)" or "new (base 0)", nothing else.

anjohnson: I agree that we should protect against that possibility, but I don't want an iocsh command to set it since this is only provided for backwards compatibility and I don't want it to appear in the `help` output. This is a run-time (pre-init) switch, and IOCs should default to using base=0.

There are a couple of obvious possibilities for switching back to base=10:
1. We could add code that runs at iocInit which checks the global variable and sets a local variable is uses for the number base.
2. The iocInit code could look for an environment variable, say `EPICS_DB_CONVERT_BASE` (but not a full-blown `envParam` though) instead of the global variable and use the value (0 or 10) from that if it's set.

dirk-zimoch: I would prefer an environment variable like `EPICS_DB_CONVERT_DECIMAL_ONLY` to be read once early during `iocInit`. That allows users to set it in the environment or change it in the startup script. Any run time change after `iocInit` will be ignored as that can be too confusing.

modules/database/src/ioc/misc/iocInit.c
```diff
@@ -129,6 +130,16 @@ static int iocBuild_1(void)
         errlogPrintf("iocBuild: " ERL_ERROR " Aborting, bad database definition (DBD)!\n");
         return -1;
     }
+
+    {
+        /* fall back to "old" decimal-only conversion if EPICS_DB_CONVERT_DECIMAL_ONLY
+           is set and does not look like meaning no/false/0 */
+        const char* dec_only = getenv("EPICS_DB_CONVERT_DECIMAL_ONLY");
+        if (dec_only && *dec_only && !strchr("nNfF0", *dec_only)) {
```
anjohnson: Core Group: Accept `YES` case-independent only, otherwise leave as 0;

dirk-zimoch: Done.

dirk-zimoch: Changed to: `if (dec_only && epicsStrCaseCmp(dec_only, "YES") == 0)`

modules/database/src/ioc/misc/iocInit.c
```diff
@@ -129,6 +130,16 @@
         errlogPrintf("iocBuild: " ERL_ERROR " Aborting, bad database definition (DBD)!\n");
         return -1;
     }
+
+    {
+        /* fall back to "old" decimal-only conversion if
+           EPICS_DB_CONVERT_DECIMAL_ONLY is YES (case insensitive). */
+        const char* dec_only = getenv("EPICS_DB_CONVERT_DECIMAL_ONLY");
+        if (dec_only && epicsStrCaseCmp(dec_only, "YES") == 0 {
```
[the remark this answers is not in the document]
dirk-zimoch: Missing include added.

PR #679, opened 2025-07-15
Multiline errlog

dirk-zimoch:
If multi-line ioc log messages are sent with `errlogPrintf()`, apply `logClientPrefix` (e.g. set by `iocLogPrefix`) to each line in order to help browsing the log files.
Also add newline to end of message if missing and strip off any leading newlines which some calls have.

Extend `errlog` iocsh command to interpret escape chars (just like `echo`).

Also fix a potential problem with message length > UINT_MAX and related Windows compiler warnings.

### DIFF
--- a/modules/libcom/src/log/logClient.c
+++ b/modules/libcom/src/log/logClient.c
@@ -49,8 +49,8 @@ typedef struct {
     epicsEventId        stateChangeNotify;
     epicsEventId        shutdownNotify;
     unsigned            connectCount;
-    unsigned            nextMsgIndex;
-    unsigned            backlog;
+    size_t              nextMsgIndex;
+    size_t              backlog;
     unsigned            connected;
     unsigned            shutdown;
     unsigned            shutdownConfirm;
@@ -244,8 +244,15 @@ void epicsStdCall logClientSend ( logClientId id, const char * message )
 
 void epicsStdCall logClientFlush ( logClientId id )
 {
-    unsigned nSent;
+    size_t nSent;
+#ifdef WIN32
+    /* Windows send() expects and returns int */
     int status = 0;
+#define LIMITED_LEN(len) (int)(len <= INT_MAX ? len : INT_MAX)
+#else
+    ssize_t status = 0;
+#define LIMITED_LEN(len) len
+#endif
 
     logClient * pClient = ( logClient * ) id;
 
@@ -258,7 +265,7 @@ void epicsStdCall logClientFlush ( logClientId id )
     nSent = pClient->backlog;
     while ( nSent < pClient->nextMsgIndex && pClient->connected ) {
         status = send ( pClient->sock, pClient->msgBuf + nSent,
-            pClient->nextMsgIndex - nSent, 0 );
+            LIMITED_LEN(pClient->nextMsgIndex - nSent), 0 );
         if ( status < 0 ) break;
         nSent += status;
     }
@@ -561,7 +568,7 @@ void epicsStdCall logClientShow (logClientId id, unsigned level)
             pClient->connectCount);
     }
     if (level>1) {
-        printf ("log client: %u bytes in buffer\n", pClient->nextMsgIndex);
+        printf ("log client: %zu bytes in buffer\n", pClient->nextMsgIndex);
         if (pClient->nextMsgIndex)
             printf("-------------------------\n"
                 "%.*s-------------------------\n",
@@ -590,7 +597,7 @@ void epicsStdCall iocLogPrefix(const char * prefix)
     }
 
     if (prefix) {
-        unsigned prefixLen = strlen(prefix);
+        size_t prefixLen = strlen(prefix);
         if (prefixLen > 0) {
             char * localCopy = malloc(prefixLen+1);
             strcpy(localCopy, prefix);
--- a/modules/libcom/src/log/logClient.c
+++ b/modules/libcom/src/log/logClient.c
@@ -168,12 +168,9 @@ static void logClientDestroy (logClientId id)
 /*
  * This method requires the pClient->mutex be owned already.
  */
-static void sendMessageChunk(logClient * pClient, const char * message) {
-    unsigned strSize;
-
-    strSize = strlen ( message );
+static void sendMessageChunk(logClient * pClient, const char * message, size_t strSize) {
     while ( strSize ) {
-        unsigned msgBufBytesLeft =
+        size_t msgBufBytesLeft =
             sizeof ( pClient->msgBuf ) - pClient->nextMsgIndex;
 
         if ( msgBufBytesLeft < strSize && pClient->nextMsgIndex != 0u && pClient->connected)
@@ -202,17 +199,44 @@ static void sendMessageChunk(logClient * pClient, const char * message) {
 void epicsStdCall logClientSend ( logClientId id, const char * message )
 {
     logClient * pClient = ( logClient * ) id;
+    size_t prefixLength;
 
     if ( ! pClient || ! message ) {
+        return;
+    }
+    while (*message == '\n') {
+        message++; /* skip initial newlines */
+    }
+    if (!*message) {
         return;
     }
 
     epicsMutexMustLock ( pClient->mutex );
 
     if (logClientPrefix) {
-        sendMessageChunk(pClient, logClientPrefix);
+        prefixLength = strlen(logClientPrefix);
+    } else {
+        prefixLength = 0;
     }
-    sendMessageChunk(pClient, message);
+    do {
+        /* apply logClientPrefix to each line */
+        size_t lineLength;
+        char* lineEnd = strchr(message, '\n');
+        if (lineEnd) {
+            lineLength = lineEnd + 1 - message;
+        } else {
+            lineLength = strlen(message);
+        }
+        if (prefixLength) {
+            sendMessageChunk(pClient, logClientPrefix, prefixLength);
+        }
+        sendMessageChunk(pClient, message, lineLength);
+        if (!lineEnd) {
+            /* terminate last line if necessary */
+            sendMessageChunk(pClient, "\n", 1);
+        }
+        message += lineLength;
+    } while (*message);
 
     epicsMutexUnlock (pClient->mutex);
 }
--- a/documentation/new-notes/PR-679.md
+++ b/documentation/new-notes/PR-679.md
@@ -1,0 +1,8 @@
+### Apply log prefix to each line
+
+If multi-line ioc log messages are sent with `errlogPrintf()`, apply
+`logClientPrefix` (e.g. set by `iocLogPrefix`) to each line in order to help
+browsing the log files.
+Also strip off any leading newlines which some calls have.
+
+Extend `errlog` iocsh command to interpret escape chars (just like `echo`).
--- a/modules/libcom/src/iocsh/libComRegister.c
+++ b/modules/libcom/src/iocsh/libComRegister.c
@@ -291,8 +291,12 @@ static void errlogInit2CallFunc(const iocshArgBuf *args)
 }
 
 /* errlog */
-IOCSH_STATIC_FUNC void errlog(const char *message)
+IOCSH_STATIC_FUNC void errlog(char *message)
 {
+    if (message)
+        dbTranslateEscape(message, message); /* in-place is safe */
+    else
+        message = "";
     errlogPrintfNoConsole("%s\n", message);
 }
 
--- a/modules/libcom/test/epicsErrlogTest.c
+++ b/modules/libcom/test/epicsErrlogTest.c
@@ -98,14 +98,16 @@ static SOCKET sock;
 static SOCKET insock;
 
 static const char* prefixactualmsg[]= {
-                     "A message without prefix",
-                     "A message with prefix",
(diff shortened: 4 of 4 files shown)


modules/libcom/src/log/logClient.c
```diff
@@ -537,7 +568,7 @@ void epicsStdCall logClientShow (logClientId id, unsigned level)
             pClient->connectCount);
     }
     if (level>1) {
-        printf ("log client: %u bytes in buffer\n", pClient->nextMsgIndex);
+        printf ("log client: %zu bytes in buffer\n", pClient->nextMsgIndex);
```
anjohnson, line 571: We've been removing the `%z` (`size_t`) prefix from our `printf()` formats as it isn't (wasn't?) portable. I forget which targets don't/didn't handle it, I'm pretty sure Windows was one of them though; have all those targets now been dropped? That's why the existing code uses `unsigned`, which is fine for log message lengths which could never be as large as 2**31 bytes.

anjohnson (commented):
How many multi-line errlog messages are generated by Base itself? This change makes me nervous, it could break error logging configurations at many existing sites, so we might want the multi-line changes to be runtime-configurable.

Could this break the [caPutLog module](https://github.com/epics-modules/caPutLog), especially its JSON output format? We rely on that at APS, with an iocLogPrefix that our log-server needs.

modules/libcom/src/log/logClient.c
dirk-zimoch: [Windows supports it at least since VS 2015](https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-140#size). For MinGW I have PR #674 pending. All Linux targets support it. I think RTEMS too. VxWorks is the only problem. I wrote a fix once. Maybe we can add that fix to EPICS.

Such long messages are unlikely but not impossible.

dirk-zimoch:
> How many multi-line errlog messages are generated by Base itself?

Some. For example when a scan task starves.

https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbScan.c#L828-L833

That‘s how I noticed. It breaks the logs with lines that our log server cannot relate to an ioc. I use `"$(IOCNAME) "` as the prefix. The server then assumes that the first word of the next line is the ioc the line came from. Even worse when the message starts with a newline. In that case we get an empty line that can be related to the ioc but the message itself cannot. And messages without terminating newline merge on the server with the next message.

I have not tested the JSON format, though.

modules/libcom/src/log/logClient.c
dirk-zimoch: Alternatively, I can cast to `long long` and use `%llu` here.

anjohnson:
If we're still using the same logstash input configuration here that I created a few years ago then we have our log-server set up to look for multi-line messages, meaning that any line which starts with whitespace (space or tab at least) is considered to be an continuation of the previous log message. That is supposed to be a common format for log messages, and the example you quoted conforms to that. I think I tried to make all the logs in Base do that, but I might have missed some.

The only disadvantage is that the log-server adds a delay before it recognizes the end of each message (unless another new log arrives from that IOC before the timeout). I do prefer that the log-server recognize all the lines as a being from single log message, which using our approach does.

dirk-zimoch:
### Multi-line messages *without* leading space/tab on continuation line
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbContext.cpp#L110-L112
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/std/rec/stateRecord.c#L85-L90
With leading newlines:
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/default/osdAssert.c#L32-L35

### More multi-line *with* leading space/tab on continuation line
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/as/asCa.c#L79-L87
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbLink.c#L136-L138

### Missing terminating newline

A number of `operator delete (void*)` methods, added in commits 8dc8be9ad7, 3035e9393c and 4359938ed6,  print a warning without terminating `\n`. The code should be unreachable, though.

More missing terminating newline:
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/ca/src/client/tcpiiu.cpp#L2011-L2012
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbConvertJSON.c#L174
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/misc/ipAddrToAsciiAsynchronous.cpp#L431
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/RTEMS-score/osdThread.c#L144
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/RTEMS/osdInterrupt.c#L92
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdThread.c#L512
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdThread.c#L517
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/vxWorks/osdThread.c#L143
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/vxWorks/osdThread.c#L148
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdMutex.c#L47

Indirectly:
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/RTEMS/posix/rtems_init.c#L312
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/pool/threadPool.c#L175
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/vxWorks/osdThread.c#L333
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/vxWorks/osdSpin.c#L56
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/vxWorks/osdSpin.c#L75
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdTime.cpp#L115
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdThread.c#L107
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdThread.c#L773
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/posix/osdSpin.c#L180
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/default/osdSpin.c#L51
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/RTEMS-score/osdThread.c#L290
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/RTEMS-score/osdSpin.c#L51
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/os/RTEMS-score/osdSpin.c#L69
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/epicsEvent.cpp#L113
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/epicsEvent.cpp#L121
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/epicsEvent.cpp#L128
Plus typo: https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/test/ioc/db/devx.c#L63
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/test/ioc/db/devx.c#L94
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/rsrv/online_notify.c#L133
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/dbStatic/dbStaticLib.c#L2511
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/dbStatic/dbStaticLib.c#L2517
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbNotify.c#L472
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbLock.c#L393
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/db/dbLock.c#L516
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/as/asCa.c#L230

Several `cantProcees()` calls in caservertask.c, for example:
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/rsrv/caservertask.c#L125

Several `send_err()` calls in camessage.c, for example:
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/database/src/ioc/rsrv/camessage.c#L317-L318

### Multiple calls to errlogPrintf for one line
This code calls `errlogPrintf` repeatedly to construct a single line. Each chunk will (probably?) get the prefix!
https://github.com/epics-base/epics-base/blob/def6cdb61f055b5b887900c20a8e5fc6c307e47c/modules/libcom/src/osi/epicsStackTrace.c#L46-L57

mdavidsaver:
> Missing terminating newline

Adding newline, and ensuring that error messages actually say "error" (or `ERL_ERROR`) seems to me straightforward and uncontentious.

> meaning that any line which starts with whitespace (space or tab at least) is considered to be an continuation of the previous log message

I'm inclined towards this convention, although not strongly.  I wish though that we could get away from this simple line-based protocol (eg. [systemd](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) has a datagram based protocol).  I realize though that this is a pipe dream.

> ... It breaks the logs with lines that our log server cannot relate to an ioc.

Isn't this a bug with your log server?  I receives the entries for each IOC independently, and in order.

anjohnson:
Messages passed to `cantProceed()` don't have to include a newline, the code in that routine follows them with another string that has one, but from a separate call to `errlogPrintf()`. I apparently made that less clear in the commit 3162d195517ce076841b9c14f198853d1ea7c1c2 and it could be improved, but the original code made that obvious.

We have long had the ability to generate log messages from multiple errlog calls, and those aren't all from source code within Base. That's why I really don't like adding this prefix to every call, and would much prefer that we formally adopt and police the "continuation lines start with whitespace" convention. I'd be happy to accept changes which add whitespace after newlines in the middle of log messages.

anjohnson:
Core Group: We need to document the necessary configuration of OSS loggers to allow the newline-space combination for combining messages. Some rsyslog builds don't seem to work.

mdavidsaver (changes requested):
Looking at this PR again, I don't think that prepending to all messages is the best approach.  Rather, I think this would be better done by a log client, or syslog aggregator (for stdout/err).

PR #682, opened 2025-07-18
fix: ts{} {} warning: missing initializer for members

jerzyjamroz:
- warning: missing initializer for member ‘epicsTimeStamp::secPastEpoch’ [-Wmissing-field-initializers]
- warning: missing initializer for member ‘epicsTimeStamp::nsec’ [-Wmissing-field-initializers]

### DIFF
--- a/modules/libcom/src/osi/epicsTime.h
+++ b/modules/libcom/src/osi/epicsTime.h
@@ -329,7 +329,7 @@ public:
 
     /** \brief The default constructor sets the time to the EPICS epoch. */
 #if __cplusplus>=201103L
-    constexpr epicsTime() :ts{} {}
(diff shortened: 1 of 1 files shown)


mdavidsaver:
No objection.

Which compiler version / language version complains?

jerzyjamroz:
I noticed this on Centos7
```
/usr/bin/g++ --version
g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
```
with `-std=c++11` flag.

mdavidsaver:
> I noticed this on Centos7

Ah, so this warning is from an older compiler.  fyi. I think this is a case where older GCC emitted this warning erroneously for what is a correct c++11 initialization.  Still, the `{0,0}` initializer is also correct.

PR #687, opened 2025-07-22
feat: add support for .tpp and .ipp within INC+=

jerzyjamroz:
### It is a proposal to consider `include` support for
* `.tpp` | Implementation (sometimes declaration) of template functions
* `.ipp` | Implementation of inline/template code

### Usage
It is used when you cannot write the implementation in a `.cpp` file because it's a template. E.g. [boost](https://beta.boost.org/doc/libs/1_82_0/boost/test/impl/).
One can use `.ipp` for `.tpp`, moreover, `.hpp` can cover usage of `.tpp` as well, so it is just a taste thing.
Depending on the evaluation, I will adjust this PR.


anjohnson:
I don't have a huge objection to this change, but given the simplicity of if it an application which needs to install `.tpp` and/or `.ipp` files could always include the necessary `vpath` lines in the Makefile which installs them. That way the application doesn't have to wait for them to get added to Base (it can also compile against older Base versions), and maintainers in the future won't wonder what these are used for and if they can be deleted yet.

jerzyjamroz:
I realised that iso standard does not specify it - there are proposals only, e.g. [5 Source Naming](https://isocpp.org/files/papers/p1204r0.html), so `file .?pp, header .hpp, module .mpp, inline .ipp, template .tpp, source .cpp`. Moreover, [wikipedia](https://en.wikipedia.org/wiki/C%2B%2B) points to those: `.C, .cc, .cpp, .cxx, .c++, .h, .H, .hh, .hpp, .hxx, .h++ .cppm, .ixx`, which are defined by [Microsoft](https://learn.microsoft.com/en-us/cpp/cpp/modules-cpp?view=msvc-170).

Now I am not sure what to think about that, but it would be good to add one standard for the `template implementations`.

mdavidsaver (changes requested):
Obligatory... https://xkcd.com/927/

I don't think it is worth adding file extensions which are not found in this repository.

jerzyjamroz:
@mdavidsaver , ok, I will close this PR. Also, If you do not plan to use it within epics-base code then it is not so useful.

PR #676, opened 2025-07-10
add IVOA to dfanout

dirk-zimoch:
The dfanout record now has invalid output handling with the usual fields
`IVOA` and `IVOV` just like other output records.

PR #688, opened 2025-07-23
improve dfanout record

dirk-zimoch:
The dfanout record now has invalid output handling with the usual fields IVOA and IVOV just like other output records.

As SELN already has 16 bits, allow to make use of them. Number of output links increased from 8 (OUTA-OUTH) to 16 (OUTA-OUTP).

### DIFF
--- a/documentation/new-notes/PR-688.md
+++ b/documentation/new-notes/PR-688.md
@@ -1,0 +1,6 @@
+### `dfanout` improvements
+
+The dfanout record now has invalid output handling with the usual fields
+`IVOA` and `IVOV` just like other output records.
+
+The number of output links has been increased from 8 to 16.
--- a/modules/database/src/ioc/db/dbUnitTest.c
+++ b/modules/database/src/ioc/db/dbUnitTest.c
@@ -35,6 +35,8 @@
 #include "errSymTbl.h"
 #include "iocshRegisterCommon.h"
 
+#define DBR_NAME(dbrType) (VALID_DB_REQ(dbrType) ? pamapdbfType[dbrType].strvalue+3 : "???")
+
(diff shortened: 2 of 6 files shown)


dirk-zimoch:
Maybe merge together with #676.

mdavidsaver (approved):
Looks OK to me.

dirk-zimoch:
Question to all: Would it make sense to add DRVL/DRVH fields?

ralphlange:
I wouldn't.
Changing the value that is distributed seems like going a step too far.

dirk-zimoch:
> Changing the value that is distributed seems like going a step too far.

That's exactly what IVOV is doing. This is how I got the idea.

dirk-zimoch:
Rebased and tests extended.
For the modifications in dbUnitTest.c see PR #697.

ralphlange:
> > Changing the value that is distributed seems like going a step too far.
> 
> That's exactly what IVOV is doing. This is how I got the idea.

But IVOA/IVOV is out-of-band/out-of-limits operation. "Switch to safe mode if things are invalid."
DRVH/DRVL is part of regular operation. Drive limits to protect hardware from unwanted values. Does that really apply to fanout?
But the dfanout also has alarms, so ... phhhh ... whatever.

Be aware of complexity bias.
My feeling is that we are adding complexity/functionality to lots of record types, leaving behind the traditional EPICS (and Unix) philosophy of using small simple building blocks and combine them if needed.

dirk-zimoch:
Hi Ralph
Thanks for your opinion. That‘s why I asked. To see your and other people’s point of view on these matters. I try to improve usability but feature creep was not my intention.

PR #689, opened 2025-07-23
better guesses for wrong field names

dirk-zimoch:
Sometimes mistakes are not just typos.
One of my most frequent field name errors is to use `INP` instead of `OUT` and vice versa.
In that case, I find the message
```
ERROR: ao record RECORDNAME doesn't have a field 'INP'
    Did you mean "PINI"?  (Process at iocInit)
```
much less helpful than
```
    Did you mean "OUT"?  (Output Specification)
```

The same approach may be extended for `ZNAM` vs `ZRST` and similar.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbLexRoutines.c
+++ b/modules/database/src/ioc/dbStatic/dbLexRoutines.c
@@ -1190,6 +1190,17 @@ static void dbRecordHead(char *recordType, char *name, int visible)
         dbVisibleRecord(pdbentry);
 }
 
+static const char* dbFieldConfusionMap [] = {
+    "INP","OUT",
+    "DOL","INP",
+    "ZNAM","ZRST",
+    "ONAM","ONST",
+    "INPA-J","DOL0-9",
+    "INPK-P","DOLA-F",
+    "INP0-9","INPA-J",
+    NULL
+};
+
 static void dbRecordField(char *name,char *value)
 {
     DBENTRY *pdbentry;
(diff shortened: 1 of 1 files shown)


simon-ess:
I can see the value of this, but this seems to maybe wander into very spaghetti-ish special case code to do nicely. Would it be worth adding a quasi-dictionary of options, e.g
```c
const char * guess_remappings[2][5] = {
    {"INP", "OUT"},
    {"OUT, "INP"}
};
```
or somesuch thing? This would at least make it a bit more maintainable...

anjohnson:
@simon-ess's idea has merit if doing the `ZRST` <=> `ZNAM` fields as well if those are really necessary — what proportion of them does it guess wrong?. However as both strings on a row are real field names these are bidirectional suggestions, so don't make the table twice as long as necessary.

Performance isn't a concern since this is only used when there's an error that the user needs to fix.

dirk-zimoch:
> However as both strings on a row are real field names these are bidirectional suggestions, so don't make the table twice as long as necessary.  Performance isn't a concern

Thought the same.

dirk-zimoch:
Currently guesses are:
`ai.OUT`->`PUTF`
`ao.INP`->`PINI`
`bi.ZRST`->`RSET`
`bi.ONST`->`NSTA`
`mbbi.ZNAM`->`NAME`
`mbbi.ONAM`->`NAME`

Many of them make no sense at all as they are not even configurable.

dirk-zimoch:
Removing all `SPC_NOMOD` and `SPC_DBADDR` fields from the candidate list gives:
`ai.OUT`->`EVNT`
`ao.INP`->`PINI`
`bi.ZRST`->`ZSV`
`bi.ONST`->`OSV`
`mbbi.ZNAM`->`SCAN`
`mbbi.ONAM`->`ONVL` (Near miss)

mdavidsaver:
> Removing all `SPC_NOMOD` and `SPC_DBADDR` fields

This seems like a straightforward improvement, and in retrospect obvious...

Maybe also weight by the `dbFldDes::interest` somehow?  Or absence of `dbFldDes::promptgroup`?  (would exclude `PROC` and `TPRO`)

> Would it be worth adding a quasi-dictionary of options, e.g

I can understand INP vs. OUT as this is a common mixup where there is no lexical relation.  Still, let us not go overboard with this.

Ultimately, with so few characters to work with, there is only so much that a suggestion algorithm can do.

dirk-zimoch:
> Ultimately, with so few characters to work with, there is only so much that a suggestion algorithm can do.

Correct. The chances for similarity are pretty small with such short words. A simple 1 char typo may be identified correctly. But totally wrong field names get suggestions where 1 char matches like `OUT`->`PUTF` or ->`EVNT`. There is not much a similarity based algorithm can do here. One needs more "insider knowledge". A map is one, the meaning of `SPC_NOMOD`, `interest` etc are others. Even more specific but also more complicated would be a type check. It makes no sense to suggest a `DBF_SHORT` field for a non-numeric value. But for that the algorithm starts becoming complicated.

simon-ess:
Type checks could work, but even that requires a lot of presumed knowledge since in the case of a missing field, you don't actually know the type of that field... since it doesn't exist. We can assume that INP is an input link even if it doesn't exist, but ZQRF... what type is that? How can we compare it to one DBF or another?

dirk-zimoch:
Check the value of ZQRF against the types of all field candidates.

simon-ess:
But at this point, isn't the value just a string? We would have to parse it to determine its type, but we do that based on the field type, which we don't have....

dirk-zimoch:
We check all fields. Parse according to each field type and decrease similarity for this field if it does not match.

dirk-zimoch:
Codacy does not like `strlen()` or `strncpy()` even if it could see that any possible string given in the program is no problem. Codacy it lazy. I don't like it.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1190,6 +1190,17 @@ static void dbRecordHead(char *recordType, char *name, int visible)
         dbVisibleRecord(pdbentry);
 }
 
+static const char* dbFieldConfusionMap [] = {
+    "INP","OUT",
+    "DOL","INP",
+    "ZNAM","ZRST",
```
anjohnson: Core Group: Extra `const`; please add a comment that only fields found in Base record types should be included here.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1208,16 +1219,105 @@ static void dbRecordField(char *name,char *value)
...
+                const char* guess = NULL;
+                size_t l = strlen(fld);
+                if (l >= 3 && fld[l-2] == '-' &&
+                    strncmp(name, fld, l-3) == 0 &&
+                    name[l-3] >= fld[l-3] &&
+                    name[l-3] <= fld[l-1])
+                {
+                    /* range map */
+                    size_t l2 = strlen(dbFieldConfusionMap[i^1]);
+                    strncpy(buf, dbFieldConfusionMap[i^1], sizeof(buf)-1);
```
anjohnson: Core: Add comment to explain the xor magic here.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
```diff
@@ -1190,6 +1190,17 @@ static void dbRecordHead(char *recordType, char *name, int visible)
...
+    "INP","OUT",
+    "DOL","INP",
+    "ZNAM","ZRST",
+    "ONAM","ONST",
+    "INPA-J","DOL0-9",
+    "INPK-P","DOLA-F",
+    "INP0-9","INPA-J",
+    NULL
+};
+
```
anjohnson, line 1224: Core: Static assertion that the number of elements be odd?

anjohnson (changes requested):
Core Group: Please paste some recent error messages from real .db errors. Some other requests below.

dirk-zimoch:
> Please paste some recent error messages from real .db errors.

Old:
```
ERROR: ai Record "ai" does not have a field "OUT"
    Did you mean "PUTF"?  (dbPutField process)
ERROR:
 7 |     field(OUT, "")

ERROR: ao record 'ao' doesn't have a field 'INP'
    Did you mean "PINI"?  (Process at iocInit)
ERROR:
 10 |     field(INP, "")

ERROR: bi record 'bi' doesn't have a field 'ZRST'
    Did you mean "RSET"?  (Address of RSET)
ERROR:
 13 |     field(ZRST, "")

ERROR: bi record 'bi' doesn't have a field 'ONST'
    Did you mean "NSTA"?  (New Alarm Status)
ERROR:
 14 |     field(ONST, "")
```
New:
```
ERROR: ai record 'ai' doesn't have a field 'OUT'
    Did you mean "INP"?  (Input Specification)
ERROR:
 7 |     field(OUT, "")

ERROR: ao record 'ao' doesn't have a field 'INP'
    Did you mean "OUT"?  (Output Specification)
ERROR:
 10 |     field(INP, "")

ERROR: bi record 'bi' doesn't have a field 'ZRST'
    Did you mean "ZNAM"?  (Zero Name)
ERROR:
 13 |     field(ZRST, "")

ERROR: bi record 'bi' doesn't have a field 'ONST'
    Did you mean "ONAM"?  (One Name)
ERROR:
 14 |     field(ONST, "")
```

modules/database/src/ioc/dbStatic/dbLexRoutines.c
dirk-zimoch: Done.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
dirk-zimoch: Done.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
dirk-zimoch: Done.

modules/database/src/ioc/dbStatic/dbLexRoutines.c
dirk-zimoch: Does not help with Codacy warnings.

dirk-zimoch:
Weighting the lexical similarity with the plausibility of the value is now implemented for numeric fields as well as for menus and device support. When checking all existing fields as candidates for substitution, a non-matching value type, e.g. a non-numeric string for a numeric field candidate or a string that does not match any of the enum choices gives a penalty to the similarity. A matching enum choice gives a boost.

PR #697, opened 2025-07-31
test improvements: NAN, DBR strings, floating point format

dirk-zimoch:
* Allow testdbGetFieldEqual to check for NAN double and float values
 * Print DBR_xxx names instead of numbers
 * use %g instead of %e to format floating point values to improve readability

I need the NAN capability for IVOA tests in PR #688. The rest is for better readability.


mdavidsaver (approved):
Looks ok to me.

anjohnson:
Merged as part of #688 which was rebased.

PR #698, opened 2025-07-31
Convert xxxRecord to use typed dset

### LINKED ISSUE #574: Update exampleApp to use type safe dset
Type-safe dset was added in EPICS 3.16.2. However, when one uses makeBaseApp.pl to generate an IOC template, the example code (exampleApp, and devXxxSoft.c in particular) still uses the old interface. Could be a good codeathon project to update the example to use the type-safe interface.

anjohnson:
Fixes #574

### DIFF
--- a/modules/database/src/template/top/exampleApp/src/devXxxSoft.c
+++ b/modules/database/src/template/top/exampleApp/src/devXxxSoft.c
@@ -21,42 +21,30 @@
 #include "xxxRecord.h"
 #include "epicsExport.h"
 
+static long init_record(dbCommon *prec)
+{
+    struct xxxRecord *pxxx = (struct xxxRecord *) prec;
+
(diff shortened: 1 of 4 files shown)


modules/database/src/template/top/exampleApp/src/devXxxSoft.c
```diff
@@ -21,42 +21,30 @@
...
-    read_xxx,
-};
-epicsExportAddress(dset,devXxxSoft);
-

-
-static long init_record(pxxx)
-    struct xxxRecord    *pxxx;
+static long init_record(dbCommon *prec)
 {
```
mdavidsaver, line 25: I wonder, 16 years after 39eaa0e335ae6d84db02e1e0d8e26d4a19758893, if this is the last of the K&R prototypes in Base?

mdavidsaver (approved):
After filtering out whitespace changes, the diff looks straightforwardly correct.

PR #700, opened 2025-08-01
Increase timer timeouts to avoid Appveyor failures

anjohnson:
See #587 discussion, trying to fix a flapping test on Appveyor.

### DIFF
--- a/modules/libcom/test/fdManagerTest.cpp
+++ b/modules/libcom/test/fdManagerTest.cpp
@@ -203,8 +203,10 @@ struct OneShot final : public fdReg {
 void testEmpty()
 {
     fdManager empty;
-    empty.process(0.1); // ca-gateway always passes 0.01
-    testPass("Did nothing");
+    epicsTime start(epicsTime::getCurrent());
+    empty.process(1.0);
+    epicsTime now(epicsTime::getCurrent());
+    testDiag("empty.process(1.0) took %.1f seconds", now - start);
 }
 
 void testOnlyTimer()
@@ -213,12 +215,14 @@ void testOnlyTimer()
     Expire trig, never;
     ScopedTimer trig_timer(mgr.createTimer()),
                 never_timer(mgr.createTimer());
+    epicsTime start(epicsTime::getCurrent());
+    trig_timer.timer.start(trig, start+0.1);
+    never_timer.timer.start(never, start+9999999.0);
+    mgr.process(5.0);
     epicsTime now(epicsTime::getCurrent());
-    trig_timer.timer.start(trig, now+0.1);
-    never_timer.timer.start(never, now+9999999.0);
-    mgr.process(0.2);
     testOk1(trig.expired);
     testOk1(!never.expired);
+    testDiag("mgr.process(5.0) took %.1f seconds", now - start);
 }
 
 void testSockIO()
@@ -283,7 +287,7 @@ void testSockIO()
         ScopedTimer timer(mgr.createTimer());
         timer.timer.start(tmo, epicsTime::getCurrent()); // immediate
 
-        mgr.process(1.0);
+        mgr.process(5.0);
 
         testOk1(!readable);
         testOk1(tmo.expired);
@@ -296,7 +300,7 @@ void testSockIO()
         shutdown(client.sd, SHUT_RDWR);
         //Socket().swap(client);
 
-        mgr.process(1.0);
+        mgr.process(5.0);
 
         testOk1(readable);
     }
@@ -306,7 +310,7 @@ void testSockIO()
 
 MAIN(fdManagerTest)
 {
-    testPlan(13);
+    testPlan(12);
     osiSockAttach();
     testEmpty();
     testOnlyTimer();


anjohnson:
Is `fdManager::process(double delay)` meant to return early?

I changed the tests to measure how long `process()` runs for. `testEmpty()` wasn't actually testing anything (so I changed the `testPass()` call to `testDiag()`) but it did delay for the whole requested delay period. In `testOnlyTimer()` the `process()` call is returning early on macOS, immediately after `trig_timer` expires, but long before `never_timer` does.

I wonder if we actually have a bug in fdManager…

mdavidsaver:
> Is `fdManager::process(double delay)` meant to return early?

I have a few such questions about intent, which I suspect will remain unanswered.  ~~eg. why does the timer-only path check `pTimerQueue` a second time while the I/O path does not?~~ (my misreading...)

> I wonder if we actually have a bug in fdManager…

imo. likely.  However, I suspect this is not a regression from #587, which did not change the timer-only code path.

Since epicsTimer is involved, #106 could come into play.

mdavidsaver:
> In `testOnlyTimer()` the `process()` call is returning early on macOS

I'm finding that this hard to think about considering how many times `epicsTimeGetCurrent()` is called for changing between absolute and relative time.  In `testOnlyTimer()` I think three times.  Once explicitly when the timers are stared.  Then twice within `process()`.

When writing `testOnlyTimer`(), my intent was that `trig_timer` would expire on the second test of the timer queue.

anjohnson:
I was thinking that `fdManager::process()` was something like `ca_pend_event()` which always waits for the full length of the delay given to it even if no events happen. That might not make as much sense in this case though, which would normally have I/O going on; the test program's `testOnlyTimer()` may just be an artificial situation.

I'm thinking that if my latest commit doesn't trigger any failures from this test on Appveyor I'll redo the full build there once or twice more and maybe then declare the flapper fixed. Of course now I've written this there will probably be be several more test failures, including at least one from some other test program...

mdavidsaver:
> ... which would normally have I/O going on ...

An important point to make.  I added these timer-only tests specifically to test this case, however weirdness here may not be relevant.  So at some point maybe we just mark these tests TODO and forget about them?

> I was thinking that `fdManager::process()` was something like `ca_pend_event()` which always waits for the full length of the delay given to it even if no events happen ...

Oh, the opposite...

```c
LIBCOM_API void fdManager::process(double delay)
{
...
    double minDelay = priv->pTimerQueue->process(epicsTime::getCurrent());

    if (minDelay >= delay) {
        minDelay = delay;
...
        epicsThreadSleep(minDelay);
```

The `delay` argument is ignored unless it is __shorter__ than the next timer to expire.  So increasing `delay` by itself seems unlikely to change anything.

fyi. I think the intent of `delay` is to compensate for there being no way to asynchronously interrupt `process()` to add a new timer or I/O event.  So I think of it like a maximum latency to introduce changes to reactor configuration.

anjohnson:
> increasing delay by itself seems unlikely to change anything

So the changes in this PR to date aren't likely to stop the flapping, I think that's what I was realizing too, thanks.

So what could have been causing the failures? Since `testOk1(trig.expired)` was failing (see .tap in my #587 comment) the `mgr.process()` call in `testOnlyTimer()` must have been returning _before_ the `trig_timer` expiry. I don't see what could have been causing that, if that was indeed the case.

I just triggered a rebuild of this PR on Appveyor, hoping to catch a failure of `testOnlyTimer()` after that last commit since my changes may provide some clue what's happening in the .tap file.

@ Core Group: PLEASE DON'T RERUN THE FAILED TESTS IN THIS NEXT BUILD OF #700 unless they all time out at the 1-hour maximum. If/when the fdManagerTest fails I want to see the contents of the fdManagerTest.tap artifact. This is the output from a successful run:
```
1..12
# empty.process(1.0) took 1.0 seconds
ok  1 - expired
ok  2 - trig.expired
ok  3 - !never.expired
# mgr.process(5.0) took 0.1 seconds
ok  4 - connect() 0 0
ok  5 - readable
ok  6 - OneShot event mask 3
ok  7 - readable
ok  8 - testing == testing
ok  9 - expired
ok 10 - !readable
ok 11 - tmo.expired
ok 12 - readable
```

anjohnson:
And here’s a failed test:
```
1..12
# empty.process(1.0) took 1.0 seconds
not ok  1 - trig.expired
ok  2 - !never.expired
# mgr.process(5.0) took 0.1 seconds
ok  3 - connect() 0 0
ok  4 - readable
ok  5 - OneShot event mask 3
ok  6 - readable
ok  7 - testing == testing
ok  8 - expired
ok  9 - !readable
ok 10 - tmo.expired
ok 11 - readable
```

mdavidsaver:
> And here’s a failed test:

Windows?  Only?

Also, it might be worth trying to insert a ~long fixed sleep to make sure the timer is found to be expired on the first pass, hopefully regardless of any OSD specific double->integer roundoff error.

```c

    trig_timer.timer.start(trig, now+0.1);
    never_timer.timer.start(never, now+9999999.0);
    epicsThreadSleep(1.0); // <----
    mgr.process(0.2);
```

fyi. My best theory at the moment is that the interval returned by `epicsTimerQueuePassive::process()`, which should be `<= 0.1 - quantum/2`, is being rounded down in `epicsThreadSleep()`, so that the second test of the queue finds that `trig_timer` is just short of expiring.

mdavidsaver:
or maybe just re-try like actual API user code does...

```
    never_timer.timer.start(never, now+9999999.0);
    for(unsigned i=0; i<10 && !trig.expired; i++)
        mgr.process(0.2);
    testOk1(trig.expired);
```

mdavidsaver:
> ```c
>     trig_timer.timer.start(trig, now+0.1);
>     never_timer.timer.start(never, now+9999999.0);
>     epicsThreadSleep(1.0); // <----
>     mgr.process(0.2);
> ```

Based on ~5 complete appveyor runs, this change avoids a spurious fail.  (of course, every one of those runs had at least one job hang, or fail a pvAccessCPP test)

anjohnson:
@mdavidsaver Looks good, if you have your fix ready to merge feel free to go ahead and do that. I'll get to it eventually if you don't, but I am busy working on an ICALEPCS paper (always a painfully slow process for me).

mdavidsaver:
I have pushed 9ce921f4fdc954205fbafdd2173345c2b8b090a2, which I hope moots this PR.

PR #703, opened 2025-08-14
Fix MSI tests when INSTALL_LOCATION is set

### LINKED ISSUE #622: msi tests fail when using INSTALL_LOCATION
The tests "t13" and "t14" added to module/database/test/ioc/dbtemplate/msi.plt on Nov 14 2024 in commit 9fb820b46e fail due to wrong search paths if base is installed to a different `INSTALL_LOCATION`. @simon-ess please fix.

The used path is the install location `@TOP@` but the test files are not installed. The source location must be used.

simon-ess:
If INSTALL_LOCATION is set, then `@TOP@` is set to it when using expandVars.pl; this means that the msi tests that look for source files do not correctly find them.

Fixes #622

### DIFF
--- a/modules/database/test/ioc/dbtemplate/msi.plt
+++ b/modules/database/test/ioc/dbtemplate/msi.plt
@@ -57,10 +57,10 @@ ok(msi('-I. -I.. -S ../t12-substitute.txt'), slurp('../t12-result.txt'));
 delete @ENV{ keys %envs };  # Not really needed
 
 # Substitution file, relative path includes
-ok(msi('-I @TOP@/modules -S ../t13-substitute.txt'), slurp('../t13-result.txt'));
+ok(msi('-I ../.. -S ../t13-substitute.txt'), slurp('../t13-result.txt'));
 
 # Template file, relative path includes
-ok(msi('-I @TOP@/modules ../t14-template.txt'), slurp('../t14-result.txt'));
+ok(msi('-I ../.. ../t14-template.txt'), slurp('../t14-result.txt'));
 
 # Test support routines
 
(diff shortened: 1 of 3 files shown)


modules/database/test/ioc/dbtemplate/msi.plt
```diff
@@ -57,10 +57,10 @@ ok(msi('-I. -I.. -S ../t12-substitute.txt'), slurp('../t12-result.txt'));
 delete @ENV{ keys %envs };  # Not really needed
 
 # Substitution file, relative path includes
-ok(msi('-I @TOP@/modules -S ../t13-substitute.txt'), slurp('../t13-result.txt'));
+ok(msi('-I @SRC_TOP@/modules -S ../t13-substitute.txt'), slurp('../t13-result.txt'));
 
 # Template file, relative path includes
-ok(msi('-I @TOP@/modules ../t14-template.txt'), slurp('../t14-result.txt'));
+ok(msi('-I @SRC_TOP@/modules ../t14-template.txt'), slurp('../t14-result.txt'));
```
anjohnson: Wouldn't it be simpler to change the `.plt` and `.substitutions` files to not use `$(TOP)` at all? That seems like the actual bug to me; maybe use `-I ../..` and `dbtemplate/t13-template.txt` instead, which only depends on the nearby structure of the source tree. I don't like embedding long internal paths inside sources like the t13-substitute.txt and t14-template.txt files do here because they have to be updated if we want to restructure the code in the future.

simon-ess: That's a cleaner solution.

I still sort of like adding EXPANDVARS to the .plt expansion, since other cases do use that... but it becomes unnecessary in this case, so I will remove it as well.

anjohnson (approved):
Thanks, LGTM now.

PR #704, opened 2025-08-19
iocsh.h: import documentation from AppDevGuide, minor tweaks

minijackson:
This imports the function-specifics that were documented in the Application Developer's Guide, section "IOC Shell".

Other changes include:

- Removing unneeded `@typedef` and `@enum`
- Adding `()` and `@ref` for interlinking
- Removing duplicated documentation between `iocsh` & `iocshLoad`, and `iocshCmd` & `iocshRun`

Supports the PR:

https://github.com/epics-docs/epics-docs/pull/141

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.h
+++ b/modules/libcom/src/iocsh/iocsh.h
@@ -12,23 +12,23 @@
 
 /**
  * @file iocsh.h
- * 
+ *
  * @brief C and C++ definitions of functions for IOC shell programming.
- * 
+ *
  * @details
  * The iocsh API provides an interface for running commands in the shell
  * of the IOC, as well as registering commands and variables for use in the shell.
  * It consists of 4 functions for the former and 2 functions for the latter.
- * 
+ *
  * @par Command functions:
- *     int iocsh (const char *pathname)@n 
- *     int iocshLoad (const char *pathname, const char *macros)@n 
- *     int iocshCmd (const char *cmd)@n 
- *     int iocshRun (const char *cmd, const char *macros)
- * 
+ *     - iocsh()
+ *     - iocshLoad()
+ *     - iocshCmd()
+ *     - iocshRun()
+ *
  * @par Registration functions:
- *     void iocshRegister (const iocshFuncDef * piocshFuncDef, iocshCallFunc func)@n 
- *     void epicsStdCall iocshRegisterVariable (const iocshVarDef *piocshVarDef)
+ *     - iocshRegister()
+ *     - iocshRegisterVariable()
  */
 
 #ifndef INCiocshH
@@ -49,22 +49,47 @@ extern "C" {
 #endif
 
 /**
- * @enum iocshArgType
- *
  * This typedef lists the values that can be used as argument data types
- * when building the piocshFuncDef parameter of iocshRegister().
- * 
- * @code {.cpp}
- * static const iocshArg AsynGenericConfigArg0 = {"Port Name", iocshArgString};
- * static const iocshArg AsynGenericConfigArg1 = {"Number Devices", iocshArgInt};
- * @endcode
+ * when building the first parameter of iocshRegister().
+ *
+ * ```
+ * static const iocshArg AsynGenericConfigArg0 = {
+ *     // name
+ *     "Port Name",
+ *     // type
+ *     iocshArgString,
+ * };
+ * static const iocshArg AsynGenericConfigArg1 = {
+ *     // name
+ *     "Number Devices",
+ *     // type
+ *     iocshArgInt,
+ * };
+ * ```
  */
 typedef enum {
+	/** The argument is converted to an integer value. */
     iocshArgInt,
+	/** The argument is converted to a double-precision floating point value. */
     iocshArgDouble,
+	/** The argument is left as a string.
+	 *
+	 * The memory used to hold the string is "owned" by iocsh
+	 * and will be reused once the handler function returns.
+	 */
     iocshArgString,
+	/** The argument must be pdbbase. */
     iocshArgPdbbase,
+	/** An arbitrary number of arguments is expected.
+	 *
+	 * Subsequent iocshArg structures will be ignored.
+	 */
     iocshArgArgv,
+	/** A copy of the argument will be made and a pointer to the copy will be passed to the handler.
+	 *
+	 * The called function should eventually release this copy
+	 * by using the pointer as an argument to free().
+	 */
     iocshArgPersistentString,
     /**
      * Equivalent to iocshArgString with a hint for tab completion that the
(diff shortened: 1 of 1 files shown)


modules/libcom/src/iocsh/iocsh.h
```diff
@@ -49,22 +49,42 @@ extern "C" {
...
+	/** The argument must be pdbbase. */
     iocshArgPdbbase,
+	/** An arbitrary number of arguments is expected.
+	 *
+	 * Subsequent iocshArg structures will be ignored.
+	 */
     iocshArgArgv,
+	/** A copy of the argument will be made and a pointer to the copy will be passed to the handler.
+	 *
+	 * The called function can release this copy by using the pointer as an argument to free().
```
anjohnson: s/can/should eventually/

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -81,26 +101,46 @@ typedef enum {
...
- * 
- * This union is used when building the func paramter of iocshRegister().
- * Each use should match the parameter type of the parameters of the
+ * This union is used when building the func parameter of iocshRegister().
+ * Each use should match the parameter iocshArgType of the parameters of the
  * function being registered
- * 
- * @code {.cpp}
+ *
+ * ```
```
anjohnson, line 113: What is the advantage/difference of replacing the Doxygen `@code` directive with the Markdown markers here? Does the Doxygen-generated HTML still look the same?

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -119,67 +159,77 @@ typedef struct iocshVarDef {
...
-    = { &AsynAXEConfigArg0, &AsynAXEConfigArg1 };
-
- * @endcode
+ * Data types of function parameters for iocshRegister().
+ *
+ * @par Example:
+ * ```
+ * static const iocshArg AsynGenericConfigArg0 = {
+ *     .name = "Port Name",
+ *     .type = iocshArgString,
```
anjohnson: Do all our C compilers now support this struct constructor syntax now? If not I would prefer that we not use them in our documentation yet, or at least leave a note about them.

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -119,67 +159,77 @@ typedef struct iocshVarDef {
...
+ *     .usage = "Helpful message describing the command",
+ * };
+ * ```
  */
 typedef struct iocshFuncDef {
+	/** Name of the command or function. */
     const char *name;
+ 	/** Number of entries in the array of pointers to argument descriptions.
+	 *
+	 * If 0, `arg` can be `NULL`.
```
anjohnson, line 218: Again these are markdown markers, does Doxygen display them the same way as it would show `@par` or `@ref` or whatever the appropriate syntax would be?

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -243,46 +294,73 @@ LIBCOM_API const iocshVarDef * epicsStdCall iocshFindVariable(
...
- * @param pathname A string that represents the path to a file from which commands are read.
- * @param macros NULL or a comma separated list of macro definitions.  eg. "VAR1=x,VAR2=y"
- * @return 0 on success, non-zero on error
+
+/**
+ * @brief Read and evaluate IOC shell commands from the given file.
+ *
+ * Commands are read from the file
+ * until and `exit` command is encountered
+ * or the end-of-file character is reached.
```
anjohnson: Typo s/and/an/.

Strictly speaking eof-of-file isn't a character in Unix (it was in CP/M and/or MS-DOS, Ctrl-Z if I remember), its a state of the file descriptor/pointer. Just say "until end-of-file is reached" here.

anjohnson (commented):
I appreciate your shortening the text in many of these descriptions, and also shortening the lines. This style is better, so thanks for that.

If the Sphinx output is the same when using Doxygen keywords such as `@code` … `@endcode` I would prefer that we stick with them since the markdown markers provide less information about what's being annotated, even if the output generated is the same at the moment.

minijackson:
I thought the Markdown markers helped the readability of the doc comment (I think `` `arg` `` is more readable than `@c arg`). Those Markdown markers are actually [parsed by Doxygen](https://www.doxygen.nl/manual/markdown.html) (the Doxyfile has `MARKDOWN_SUPPORT` enabled), so from Doxygen's point of view, `` `arg` `` and `@c arg` are the same.

I technically doesn't change anything in the Sphinx output, nor should it change anything in the Doxygen HTML output, but I can still remove them if you want.

modules/libcom/src/iocsh/iocsh.h
minijackson: I added them as a hint to the reader, what argument corresponds to what field. But in retrospect, I can add that information as comments.

modules/libcom/src/iocsh/iocsh.h
mdavidsaver: Seconded.  eg. Language aware highlighting.

minijackson: @mdavidsaver when I tested this, Doxygen was aware that it should highlight the syntax as C++. I guess Doxygen picks the language of the file by default.

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -49,22 +49,47 @@ extern "C" {
...
  */
 typedef enum {
+	/** The argument is converted to an integer value. */
     iocshArgInt,
+	/** The argument is converted to a double-precision floating point value. */
     iocshArgDouble,
+	/** The argument is left as a string.
+	 *
+	 * The memory used to hold the string is "owned" by iocsh
+	 * and will be reused once the handler function returns.
```
anjohnson, line 78: These second paragraphs are all being formatted as code by Sphinx.

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -81,26 +106,46 @@ typedef enum {
...
  * {
  *     AsynGenericConfig (args[0].sval, args[1].ival);
  * }
- * @endcode
+ * ```
  */
 typedef union iocshArgBuf {
+	/** The value as an integer.
+	 *
+	 * Corresponds to the @ref iocshArgInt type.
```
anjohnson, line 123: Whereas these are formatted as expected. Weird.

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -12,23 +12,23 @@
...
  * @par Command functions:
- *     int iocsh (const char *pathname)@n 
- *     int iocshLoad (const char *pathname, const char *macros)@n 
- *     int iocshCmd (const char *cmd)@n 
- *     int iocshRun (const char *cmd, const char *macros)
- * 
+ *     - iocsh()
+ *     - iocshLoad()
+ *     - iocshCmd()
+ *     - iocshRun()
```
anjohnson, line 27: These get turned into links, but they only link to the header file, not to the functions inside it. This whole header might not be particularly useful as a result.

anjohnson (commented):
Core group: Some minor issues, we'd accept this as is if you don't have time to resolve the items below but it would be nice to resolve them too.

modules/libcom/src/iocsh/iocsh.h
minijackson: That is indeed weird, it doesn't show as code on my machine, but it does in the ReadTheDocs CI build.

After a bit of investigation, it seems like a Doxygen bug, that was fixed between 1.10.0 and 1.12.0 included. I'll see if I can update the RTD CI to upgrade our dependencies.

minijackson: Welp, according to [Repology](https://repology.org/project/doxygen/versions), the latest version of Ubuntu only has Doxygen 1.9.8, which should still have this bug…

ReadTheDocs doesn't seem to have [anything other than Ubuntu](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-os) either, so I'm not sure if I can fix it…

modules/libcom/src/iocsh/iocsh.h
minijackson: These links seem to point directly to the functions on the RtD CI build, and on my local Doxygen build (1.13.2)

anjohnson: That's weird, I didn't think they did for me yesterday, but I might have been looking at the wrong place — my mistake!

modules/libcom/src/iocsh/iocsh.h
```diff
@@ -49,22 +49,47 @@ extern "C" {
...
- * @endcode
+ * when building the first parameter of iocshRegister().
+ *
+ * ```
+ * static const iocshArg AsynGenericConfigArg0 = {
+ *     // name
+ *     "Port Name",
+ *     // type
+ *     iocshArgString,
+ * };
```
anjohnson, line 61: You could save some vertical space here with something like this:
```C
 * static const iocshArg AsynGenericConfigArg0 = {
 *     // name, type
 *     "Port Name", iocshArgString,
 * };
```
although to be honest I don't see that comment line adding much clarity. I do prefer that the struct member data appears on a separate line though, the original is harder to read.

anjohnson:
Core Group: This branch now has conflicts so we can't merge right now.

minijackson:
@anjohnson rebase done!

PR #707, opened 2025-08-28
Fetch the custom cfg/CONFIG_* files early

simon-ess:
If you overwrite INSTALL_LOCATION in CONFIG_SITE, then it would be the case that TOP_CFG_CONFIGS would point to those from EPICS base and not from the module that you are loading. Fetching these earlier fixes that issue.

This should fix the issue described https://github.com/epics-base/pvxs/pull/124.

### DIFF
--- a/configure/CONFIG
+++ b/configure/CONFIG
@@ -50,6 +50,9 @@ endif
 include $(CONFIG)/CONFIG_COMMON
 include $(CONFIG)/CONFIG_FILE_TYPE
 
+# Fetch the module-specific CFGS before overwriting INSTALL_LOCATION in order to load them later
+TOP_CFG_CONFIGS := $(wildcard $(INSTALL_CFG)/CONFIG*)
+
 #  Base-specific build options
 #
 include $(CONFIG)/CONFIG_BASE
@@ -117,7 +120,6 @@ endif
 
 # Include $(INSTALL_CFG)/CONFIG* definitions
 #
-TOP_CFG_CONFIGS = $(wildcard $(INSTALL_CFG)/CONFIG*)
 ifneq ($(TOP_CFG_CONFIGS),)
   include $(TOP_CFG_CONFIGS)
 endif


anjohnson:
This looks like a simple change, but I'd like a little more detail about what is happening in the e3 builds that require it since I still don't fully understand the circumstances which cause the problem. Could you describe what that layout looks like, does your build install multiple (all?) support modules into one location for example?

It might be helpful to have a timeline of the relevant files being included and the GNUmake statements involved from them showing when the build goes wrong to clarify that.

ralphlange:
I don't think this is related to e3, other than e3 setting INSTALL_LOCATION in CONFIG_SITE, as described above.

simon-ess:
Correct. This isn't about e3, it's actually about PVXS. The problem is that if you set INSTALL_LOCATION in EPICS base in CONFIG_SITE (and not via the command line), then it overwrites the build for every module while the EPICS base config files are loaded. As such, you do not find any custom `CONFIG_*` files (unless, of course, the INSTALL_LOCATION matches that of EPICS base i.e. if you are a submodule of EPICS base).

If you look at where we define the custom config files though, we can define them at any point after we include CONFIG_FILE_TYPE, which happens _before_ we redefine INSTALL_LOCATION via CONFIG_SITE. So it makes sense to define those at this time instead of before you actually include them.

anjohnson:
Core Group: Approved.

mdavidsaver:
> If you overwrite INSTALL_LOCATION in CONFIG_SITE, then it would be the case that TOP_CFG_CONFIGS would point to those from EPICS base and not from the module that you are loading.

I do not think this statement is complete.  Changing `INSTALL_LOCATION` changes both where `cfg/CONFIG*` fragments are read from, and where they will be installed to.

I think the root problem which needs addressing is what to do for a module (Base or eg. PVXS) which wants to install cfg/ fragments, and then use them in later sub-directory builds.  For modules referenced from `configure/RELEASE`, I do not think there is an issue.  Although situations where eg. `INSTALL_LOCATION = $(EPICS_BASE)` for a support module build might need some consideration.

With https://github.com/epics-base/pvxs/pull/124 I worked around any `$(wildcard` issues from Base configure/CONFIG by explicitly including the fragments.  However, this might be seen as a hack?

PR #711, opened 2025-09-11
Allow CA clients to determine the server protocol version

bsbevins:
Adds a call to the CA client API that allows a client to determine the server's protocol minor version number. This is needed to allow the ca-nameserver to report a server's protocol version correctly to a client.

### DIFF
--- a/modules/ca/src/client/cac.h
+++ b/modules/ca/src/client/cac.h
@@ -203,6 +203,7 @@ public:
     void destroyIIU ( tcpiiu & iiu );
 
     const char * pLocalHostName ();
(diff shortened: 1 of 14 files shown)


mdavidsaver:
How should these calls behave when called on a disconnected channel?

bsbevins:
> How should these calls behave when called on a disconnected channel?

I have added a commit that makes returning CA_UKN_MINOR_VERSION explicit. No change in behavior.

mdavidsaver (approved):
I think this PR looks ok now.

PR #712, opened 2025-09-20
Use OP_SYS_LDFLAGS for -Wl,--gc-sections

JJL772:
OP_SYS_LDLIBS ends up on the GeSys object link command line, which ends up breaking things. When linking GeSys objects, a symbol root is not defined using -u or -e, which is required for --gc-sections to work.

I'm not sure why `-Wl,--gc-sections` was added to `OP_SYS_LDLIBS` in the first place, but switching that over to `OP_SYS_LDFLAGS` seems to work fine. I've only tested with the RTEMS-pc686 target though.

### DIFF
--- a/configure/os/CONFIG.Common.RTEMS-beagleboneblack
+++ b/configure/os/CONFIG.Common.RTEMS-beagleboneblack
@@ -9,7 +9,7 @@ RTEMS_BSP = beagleboneblack
 RTEMS_TARGET_CPU = arm
 GNU_TARGET = arm-rtems
 
-OP_SYS_LDLIBS += -Wl,--gc-sections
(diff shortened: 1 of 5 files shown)


mdavidsaver:
I agree that `-Wl,--gc-sections` belongs in LDFLAGS.

anjohnson (approved):
Agreed, this looks more correct to me. Why hasn’t it caused a problem for other users, is the linking only being done for GeSys builds?

JJL772:
GeSys builds do not define a symbol root with `-u` during the linking step, but the normal linking process does (either `POSIX_Init` or `Init`). So it only becomes an issue when you're building one of these RTEMS 6+ targets with GeSys objects enabled.

PR #713, opened 2025-09-20
dbCa: iocInit wait for local CA links to connect

mdavidsaver:
During `iocInit()`, wait for local CA links to connect.  With this change, database authors can be sure that any local CA link is connected prior to `initHookAfterIocRunning` and `field(PINI, RUNNING)`.

### DIFF
--- a/modules/database/src/ioc/db/dbCa.c
+++ b/modules/database/src/ioc/db/dbCa.c
@@ -40,6 +40,7 @@
 /* We can't include dbStaticLib.h here */
 #define dbCalloc(nobj,size) callocMustSucceed(nobj,size,"dbCalloc")
 
+#include <epicsAtomic.h>
 #include "db_access_routines.h"
 #include "dbCa.h"
 #include "dbCaPvt.h"
@@ -64,6 +65,7 @@ extern int dbServiceIsolate;
 static ELLLIST workList = ELLLIST_INIT;    /* Work list for dbCaTask */
 static epicsMutexId workListLock; /*Mutual exclusions semaphores for workList*/
 static epicsEventId workListEvent; /*wakeup event for dbCaTask*/
+static size_t initOutstanding;
 static int removesOutstanding = 0;
 #define removesOutstandingWarning 10000
 
@@ -340,7 +342,7 @@ static void dbCaLinkInitImpl(int isolate)
     dbCaCtl = ctlPause;
 
     dbCaWorker = epicsThreadCreateOpt("dbCaLink", dbCaTask, NULL, &opts);
-    /* wait for worker to startup and initialize dbCaClientContext */
+    /* wait for worker to startup, initialize dbCaClientContext, and connect local CA */
     epicsEventMustWait(startStopEvent);
 }
 
@@ -369,11 +371,12 @@ void dbCaPause(void)
         epicsEventSignal(workListEvent);
     }
 }
-
-void dbCaAddLinkCallback(struct link *plink,
-    dbCaCallback connect, dbCaCallback monitor, void *userPvt)
+void dbCaAddLinkCallbackOpt(struct dbLocker *locker, struct link *plink,
+                            dbCaCallback connect, dbCaCallback monitor, void *userPvt,
+                            unsigned flags)
 {
     caLink *pca;
+    (void)locker; /* Passed for symmetry with dbDbAddLink().  So far unused. */
 
     assert(!plink->value.pv_link.pvt);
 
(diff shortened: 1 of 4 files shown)


modules/database/src/ioc/db/dbCa.c
```diff
@@ -394,9 +400,15 @@ void dbCaAddLinkCallback(struct link *plink,
...
+    dbCaCallback connect, dbCaCallback monitor, void *userPvt)
+{
+    dbCaAddLinkCallbackOpt(NULL, plink, connect, monitor, userPvt, 0);
+}
+
 long dbCaAddLink(struct dbLocker *locker, struct link *plink, short dbfType)
 {
-    dbCaAddLinkCallback(plink, 0, 0, NULL);
+    dbCaAddLinkCallbackOpt(NULL, plink, 0, 0, NULL, 0);
     return 0;
```
simon-ess, line 413: Should `locker` be passed to `dbCaAddLinnkCallbackOpt`?

modules/database/src/ioc/db/dbCa.c
```diff
@@ -369,9 +371,9 @@ void dbCaPause(void)
         epicsEventSignal(workListEvent);
     }
 }
-
-void dbCaAddLinkCallback(struct link *plink,
-    dbCaCallback connect, dbCaCallback monitor, void *userPvt)
+void dbCaAddLinkCallbackOpt(struct dbLocker *locker, struct link *plink,
```
simon-ess, line 374: Re: previous comment, why are we taking the argument `locker` here as it is not used?

modules/database/src/ioc/db/dbLink.c
```diff
@@ -125,7 +127,9 @@ void dbInitLink(struct link *plink, short dbfType)
     if (dbfType == DBF_INLINK)
         plink->value.pv_link.pvlMask |= pvlOptInpNative;
 
-    dbCaAddLink(NULL, plink, dbfType);
```
simon-ess, line 128: This is probably a side point, but I note that `dbfType` isn't actually used here. Is this an old API remnant that cannot easily be removed?

mdavidsaver:
> ... why are we taking the argument locker here as it is not used?

An attempt at symmetry / future compatibility with `dbDbAddLink()`.  It will probably not be sufficient, and has so far not been used.  Still, I should be consistent.

modules/database/src/ioc/db/dbLink.c
mdavidsaver: > ... Is this an old API remnant that cannot easily be removed?

Yup.

mdavidsaver:
> ... why are we taking the argument locker here as it is not used?

I have made the passing of the locker object consistent, and added a comment noting that it is intentionally unused.

I also shuffled the declaration for the new `dbCaAddLinkCallbackOpt()` to `dbCaPvt.h` since imo. there is no reasonable use case from external code.  This leaves `dbCa.h` unchanged.

anjohnson:
I can't see the locker object ever being needed by CA links, since it's purely related to the requirement to handle lockset modifications, which only apply to DB links (where the locker is needed). I'm not sure why dbCaAddLink() was given that argument, it isn't there in 3.15 (when that was a macro), I guess I just didn't understand the locker stuff at the time that I introduced LSETs (1d99ad6cfca53ae5fec138bedd835ab4aefcc258).

It would be nice to be able to create a `{db: {…}}` JSON link type in the future, but I don't think that's likely to be happen given that the lockset calculations have to look at all the DB links in the record, and allowing JSON child DB links could severely complicate that.

I have been wanting to write a `{ca: {…}}` JSON link type that calls dbCa.c internally for its implementation, but haven't looked at it in detail. I hope these changes won't prevent that.

mdavidsaver:
> ... I hope these changes won't prevent that.

I do not think they would.

anjohnson:
Core Group: Accepted, with a release note.

mdavidsaver:
Applied as a46bd5ae88087bbadab0d547141146ebf4d46dc8

mdavidsaver:
I have realized that I did not look closely enough at the iocInit order, so this PR as merged is a no-op.  I also did not update the CI tests to demonstrate correct behavior.  eg. removing:

https://github.com/epics-base/epics-base/blob/a46bd5ae88087bbadab0d547141146ebf4d46dc8/modules/database/test/std/rec/regressTest.c#L196-L198

which should now be unnecessary, but still is.

I see a way forward, but it will be different enough that I think it will need re-review.  So 3f382f6b68521c686b078aad6782fb0c693fbc09 reverts this PR.

Issue #705, opened 2025-08-20
Delete Solaris target config files, code, docs and comments


anjohnson:
Recent versions of EPICS have not been built or tested on Sun Solaris targets; no sites or users have shown interest in maintaining or supporting this OS any more.

Delete:
- Files for solaris targets under `configure/os`
- directories `modules/libcom/src/osi/os/solaris` and `modules/libcom/src/osi/compiler/solStudio`
- code for solaris targets in `src/tools/EpicsHostArch.pl`
- variables with names ending `_solaris` in all `Makefile`s
- Section on building for Solaris in `modules/ca/src/client/CAref.html`
- comments mentioning solaris in `configure/CONFIG` and elsewhere
- Anything else I've missed (`git grep -i solaris` should help)

irierr:
I will take on this issue

dirk-zimoch:
What about the mentions of "HP-UX" and "HPUX" ?

PR #714, opened 2025-09-23
windows cl does not accept designated initializers before c++20

dirk-zimoch:
If and only if the Windows `cl` compiler is used with the flag `/Zc:__cplusplus`, it defines the macro `__cplusplus` correctly according to the used C++ standard. (https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170)
Normally (as used in the standard EPICS config files) it will always define `__cplusplus` as `199711L`, even though the default standard is c++14, which disables all `#if __cplusplus>=201103L` code blocks in EPICS.

fdManager uses such a code block but when using  `/Zc:__cplusplus` Windows complains with the error message:
```
fdManager.cpp(192): error C7555: use of designated initializers requires at least '/std:c++20'
```
Both, gcc and clang, understand the used syntax from C++11 on, but I cannot be sure if any other c++ compiler (not only cl) does.
Solution: Remove the preprocessor branch and leave optimizations to the compiler.

### DIFF
--- a/modules/libcom/src/fdmgr/fdManager.cpp
+++ b/modules/libcom/src/fdmgr/fdManager.cpp
@@ -187,18 +187,11 @@ LIBCOM_API void fdManager::process(double delay)
         ++ioPending;
 
 #ifdef FDMGR_USE_POLL
-#if __cplusplus >= 201100L
-        priv->pollfds.emplace_back(pollfd{
-            .fd = iter->getFD(),
-            .events = WIN_POLLEVENT_FILTER(PollEvents[iter->getType()])
-        });
-#else
         struct pollfd pollfd;
         pollfd.fd = iter->getFD();
         pollfd.events = WIN_POLLEVENT_FILTER(PollEvents[iter->getType()]);
         pollfd.revents = 0;
         priv->pollfds.push_back(pollfd);
-#endif
 #endif
 
 #ifdef FDMGR_USE_SELECT


anjohnson:
I could see this issue arising elsewhere too. Instead of copy/pasting that preprocessor logic everywhere in the future could we define a macro say `EPICS_HAS_DESIGNATED_INIT` (name discussion welcome) in the compiler specific headers and use that?

dirk-zimoch:
> could we define a macro say `EPICS_HAS_DESIGNATED_INIT` (name discussion welcome)

We already have macros `CXX_PLACEMENT_DELETE` and `CXX_THROW_SPECIFICATION`. Thus for consistency, I propose `CXX_DESIGATED_INIT`.

I found that gcc actually supports this already from c++98 on.

I can add this to the default (`>=202002L`), windows (`>=202002L`), gcc (`>=199711L`) and clang (`>=201103L`) version of compilerSpecific.h, but what about the solStudio version? I guess `>=202002L`, or not at all as [c++14 seems to be the latest supported standard](https://docs.oracle.com/cd/E77782_01/html/E77789/bkabe.html#OSSCPgpxxt), but can someone test this?

anjohnson:
Nobody is using EPICS on Solaris that I know of now and we can't actually support it — it probably hasn't been tested in a _long_ time. Issue #705 (tagged Codeathon) is to delete those OS targets.

mdavidsaver:
> Thus for consistency, I propose `CXX_DESIGATED_INIT`.

No.  c++20 specifies standard feature test macro names.  IMO. these should be used if at all possible.  In this case `__cpp_designated_initializers`.

https://en.cppreference.com/w/cpp/experimental/feature_test.html

mdavidsaver:
In this specific instance, would it be simpler to avoid this compatibility question by removing the `#ifdef` and using the "compatibility" code in all cases?

dirk-zimoch:
> c++20 specifies standard feature test macro names.

`g++ -std=c++20` defines `__cpp_designated_initializers` but `g++ -std=c++11` does not and still accepts them. Would you suggest we define `__cpp_designated_initializers` manually in those cases?

> would it be simpler [...] using the "compatibility" code in all cases?

Yes. It costs one more structure copy because I cannot emplace the structure.... Maybe I could but construct it not with designated initializers but just listing the field values. My idea was that with designated initializers I do not rely on the internal order of the structure fields, but I have recently learned that while this is true for C, C++ still requires to initialize the fields in-order. Also, C initializes the not mentioned fields with 0 while C++ does not initialize them.

mdavidsaver:
@dirk-zimoch I think such micro-optimizations are premature in this case.  Analyzing a simplified [example](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAtydwAZAiZsADk/ACNsUikAZnIAB3QlYhcmL19/AzSM5yEwiOi2OISAdntsRyKmESIWUiIcvwCgh2wnLMbmohKo2Pik%2ByaWtryeMf7BsoqQSoBKe3QfUlROLktE8NRfHABqM0SPADduklIT3DMtAEE7%2B6UiUh8nQ4AxdHRjypsHocgYdwkRDixyIcYicAY9KgARGFPJ5ndAETCHFI%2BJRICAvTAgEAXJxkE4eb7oG6WABsh0E6UhoPBS2R/yewLpPyUADpsGwUkZNgB9GIsVAAawgFLMbMBHOB3JYx0S8PB5HZ8qB3JiytVLFs0w1QJl8JZiVhJq4K3o3AArPwAlwdOR0NwPLZbIclGsNthjhZEnxyERtFaVuKQLatIZuNJ%2BGxI9HHc7XVx%2BEoQNGQ06reQ4LAUBgcPhiGRKNQ6IxWBxuEHBMIxBJODI5MJlGpNDnyPogkYTGgPdYanUsm4mJ5vO0DKFwkNyiMWwVMkJJgFpkv6nNhgkW10ekI%2BhNJ1Nh5deuMBrP5gvZC9%2BquDHeWlv5zvpCtvetNtwdnsDn6yWJK4qQeJ4XjeD4KT%2BWEOSZCEoSRUCEUQx4HlRdFMWxXF8UJIDSVOaVEluCxaXpJRGWEZlWRg4EoKIPkUhQjl6P5RVdXBJjgRYlJtXY/VrB4TigTI7ksRxEUxUlbizQtZCHlzG0uHtchk34VN3WsT1P19f1Egsfhsx0JYViQbAWBwBIIGtWN40TFTQxdbh00zYNQ2M8gI2kW1uWpABOHhfK0aktC0W0LAsALkkUxIHQc1MDLcvNEAgQt0H5Bh4nLCAMHSxgEn2YxgB4RItGmSt6NIDMIBiByYnCZoAE9a34OrWFIBqAHkYl0S5mvIHKOGEDqmHoJquxwGIfGADwJHoDNeH4HA2EKyRxoIUgzwuebnWwdRuh8ei%2BtBWoHPoAgYlIRqvBwBzXgIBMFvIC5SBidJsHhPkVvCUAcxWGgjGAJQADUCGwAB3DqUmYPr61EcRJBbWH2w0Bz9GmPtTEHOwzpiDNIBWdAUnqeaAFoOpil1ntIdFsDxqzT33AIIHcB9phnUpt3ydJl2yY811SbnNyvTnpj3epD1aPnH1qM8DwvF8FlFi9WZmZ9hdfKQPx9b8eGspTYq7VNDnUAAOakSepaRDgKkxDmK7ktG5HhDggEsrl03WEt%2BvNUtyzKqGytKUgy/KMeK0qBAYCqqpqrtWsavr4/arqeqcPqBuYIhhtGhyJqmmb6DmvqlpWrZnUIDaei2hzdv2w7HuOxTnRxy72uusuDOph6g2e16VA%2B5aTDO/tvf%2BlhAZB8HIehx7YcbBHZCRlQUa7HtDEKgdNKHHG6YJomslJ8m1Kpmnd4Z%2BoxwnXJ%2BfZudFYFwoshVjcsgVhdz/Pe8pdFmXGYlt%2Bdyq0ltfR88t1aKy1l%2BTgFg9bKVUo5LgxszYWytjbYAdtEgOydi7N2ZBdIwK9kZH2IBCA0BoFlOe8NmyL3kMjTszckAZnRow0hNAiANShi5egjDdqvBYITC%2BaZyCkEYYEYRLCCBkPYZw6MIiVDqD4QI0caZYEGxTNweEkiaCHEnmDeISDzaW2thjDBWDnZk0SIcM4SgDEoOMYVUxjtzFH1cr9EyZkLLUHDJGaQ3ISomy0L5RIvlfLSEqJUAJlRqQxi4HGcgCYeAhXsobJy9gxGGTDDE/SyT1FCIye5Z6GRXDSCAA) shows that clang 13 with `-O3` produces identical machine code for both variants.

```c++
#include <vector>

struct Foo {
    int a, b;
};

void push(std::vector<Foo>& foos, int a) // construct then emplace_back
{
    foos.emplace_back(Foo{
        .a = a,
        .b = a+1,
    });
}

void push(std::vector<Foo>& foos, int a) // construct, then push_back
{
    Foo temp;
    temp.a = a;
    temp.b = a+1;
    foos.push_back(temp);
}
```

A [few instructions](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAtydwAZAiZsADk/ACNsUikAZnIAB3QlYhcmL19/AzSM5yEwiOi2OISAdntsRyKmESIWUiIcvwCgh2wnLMbmohKo2Pik%2ByaWtryeMf7BsoqQSoBKe3QfUlROLktE8NRfHABqM0SPADduklIT3DMtAEE7%2B6UiUh8nQ4AxdHRjypsHocgYdwkRDixyIcYicAY9KgARGFPJ5ndAETCHFI%2BJRICAvTAgEAXJxkE4eb7oG6WABsh0E6UhoPBS2R/yewLpPyUADpsGwUkZNgB9GIsVAAawgFLMbMBHOB3JYx0S8PB5HZ8qB3JiytVLFs0w1QJl8JZiVhJq4K3o3AArPwAlwdOR0NwPLZbIclGsNthjhZEnxyERtFaVuKQLatIZuNJ%2BGxI9HHc7XVx%2BEoQNGQ06reQ4LAUBgcPhiGRKNQ6IxWBxuEHBMIxBJODI5MJlGpNDnyPogkYTGgPdYanUsm4mJ5vO1AsFx3NhgkeMkCpkhJMAtNl/U5%2BURtUuj0hH0JpO8p1apdeuMBuEhjuF8kXv01wZHy1twtlqt1ptuDs9gc/TJYkripB4nheN4PgpP5YQ5JkIShJEwIRJDHgeVF0UxbFcXxQlgNJU5pUSW4LFpeklEZYRmVZWDgXI7ksRxEUxUlCkICWM1aKBFgfBIGlDnUXVOXSbUWPY1COXURUhP1c0jQE7UZINVDLWtO0HVDF03UHKwvR9TZ/USCx%2BGzHQOPIJBsBYHAEnYmMuDjcgEyjchk34VN00zYNQ3MiNpFtblqQATh4IKtGpLQtFtCwLFC5IbS4RINK7DzvJzFZ82QNB0H5Bh4nLCAMFyxgEn2YxgEXLRpkrIh4gzCAYk0mJwmaABPWt%2BGa1hSFagB5GJdEuDryCKjhhF6ph6HarscBiHxgA8CR6AzXh%2BBwNhyskGaCFIC8LhW51sHUbpeK2INQVqTT6AIGJSDarwcE014CATVbyAuUgYnSbB4T5TbwlAdKBCMYAlAANQIbAAHdepSZhhvrURxEkFtEfbDRNP0aY%2B1MHTDBujNIBWdAUnqFaAFpeqSl0PtIdFsEJuz93qMcJ1ydcZ0wd8Rg3dIV2yE8Oc3LJuYXYcL0PK9n2mZnL1mG95h5mZj3Zl8r1FqQVm9b9mzUrh7VczTUwEgAOalyepaRDjKkxDkXbktG5HhDggEsrkMnglhMny80LHKUjyigqEK/3A7QHHKuqhhatIerGq7Lq2uGxOev6wanGG0bmCICaps02b5sW%2BhluG9bNq2Z1CF2np9s0o6Ttq4aLoS51rtu%2B6sArky6deoMPq%2BlRfo2kxrv7IGaBB8HIZhuHHTreRGxR2Q0ZUDGux7QxyoHaxbHxmJGeJ0msgpqn3Np%2BmD/Fg8Aggdxpc5jXecKLJ7%2BF4oFfnF9z2vo9WkFr%2BRyS3lqUT%2BMspb/zAcA28CxPZfl9IEPWBs3JaS4Kbc2ltrY4ztokB2TsXZuzIIZCwXs0pmV9iAQgNAaAFURovZsy95Do07K3JAGZsasMoTQIgrU4ZeXoKwo6rwWAkxZmmcgpBWHTgkUoTh3DeHRmkYIu6IjRxpkQclFM3B4QECoYcCG0N4hoItlbG2wBsG4OdpTRIhwzhKCMRg0x5jHaWNPqQsMFkrI2WoOGSM0huSJC0CbLQQVEhBSCtISolQgmVGpPZRyCYeCRUNilbgnkgimXcQlYyyTNFiIyeZD6GRXDSCAA) can be saved by avoiding constructing a temporary on the stack.

```c++
void push(std::vector<Foo>& foos, int a) // push_back, then assign
{
    foos.push_back(Foo());
    auto& x = foos.back();
    x.a = a;
    x.b = a+1;
}
```

However, to my mind this reduction is likely to be insignificant compared to the simpler, and more impactful, optimization of reusing `fdManagerPrivate::pollfds` between calls to avoid a heap allocation.  Which you already do :)

dirk-zimoch:
Removed the emplace with designated initializers.

PR #715, opened 2025-09-23
Implement epicsMessageQueueSend/epicsMessageQueueReceive to resolve link error

JJL772:
These were defined as macros in the RTEMS `osdMessageQueue.h`, but declared as function prototypes in the `epicsMessageQueue.h` public header. We need to implement them as C functions for these to actually be usable.


anjohnson (approved):
No objections from me.

PR #716, opened 2025-09-24
Add a variable for handling the C/C++ standard

simon-ess:
There is no standard location to specify a C/C++ standard in a way that doesn't cause some potential conflict; ARCH_DEP_C*FLAGS seems the most logical, but we pass the CFLAGS in that case to the CXXFLAGS which cases warnings. POSIX is the next best, but if we are not in the posix case it would be morally incorrect.

Note that we are running into this because builds with quite modern compilers are assuming c23 which causes failed builds (see https://github.com/epics-modules/iocStats/issues/65#issuecomment-3321532900 for example, but this hits other support modules such as asyn and streamdevice); this is the first time that we are needing to specify a C standard and not a C++ standard that I know of.

This feature adds a new compiler flag for this purpose: `LANG_C*FLAGS`. This is then specified by setting one of the following variables:
* `DEFAULT_LANG_C*`
* `USR_LANG_C*`
* `$(basename file)_LANG_C*`

depending on scope. For example, if you would the default c++ language standard to be `c++17` but for the file `foo.cpp` that it should be compile with `g++23` then you woul set in `CONFIG_SITE`:
```make
DEFAULT_LANG_CXX = c++17
```
and in the makefile that includes `foo.cpp` you would have
```make
SRCS += foo.cpp
foo_LANG_CXX = g++23
```

Acceptible options for C standards using the gcc/clang compilers are
```
c89 c90 c99 c11 c17 c18 c23
gnu89 gnu90 gnu99 gnu11 gnu17 gnu18 gnu23
```
and for c++ standards
```
c++98 c++03 c++11 c++14 c++17 c++20 c++23
gnu++98 gnu++03 gnu++11 gnu++14 gnu++17 gnu++20 gnu++23
```
while for msvc we have for c standardss
```
c11 c17
```
and for c++ standards
```
c++14 c++17 c++20 c++23
```
You can also select `latest` to choose the highest version.


simon-ess:
I am not sure why the RTEMS-5 job is failing. while this code does technically touch on compilation configuration for RTEMS 5, it should only do so for c++ files. Moreover, comparing the last successful run I found to this one, the exact same command is run.

Was there an updated release of an RTEMS toolchain recently?

anjohnson:
There might be more to this question that we should think about: If I’ve understood the PR it assumes a single C/C++ standard for all the code being compiled for these targets. That doesn’t work when a downstream application wants to set the language standard used for some group of its sources, possibly even on a per-file basis. AFAIK GCC and Clang at least can link together object files that have been compiled with different standards, I would expect MSVC to handle that too but haven’t checked.

The latter is relative easy to do in Make, you use a target-specific variable, something like this:
```makefile
myOldCode.$(OBJ): STD_CXXFLAGS=-std=c++98
```

simon-ess:
My thought here was that we put the standard as the first flag passed, so then it can be overridden in later flags (e.g. USR_C[XX]FLAGS). But it allows you to set a default that can be used.

We are seeing this with our conda builds, mostly because conda-forge is assuming a much newer gcc than we are using locally, see https://github.com/conda-forge/epics-base-feedstock/pull/45/files.

mdavidsaver:
How would I use `STD_CXXFLAGS` in a portable application supporting both gcc/clang and MSVC?

As @dirk-zimoch lays out in #714, MSVC has a different, and annoyingly complex, scheme for selecting c++ standard version.

eg. with PVXS I currently select c++11 for the implementation standard.  The installed headers and ABI have to support a wider range of standards, however the implementation could internally use syntax/stdc++ bits which are not present with other standard versions.

dirk-zimoch:
> MSVC has a different, and annoyingly complex, scheme for selecting c++ standard version.

[Not really too complex](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170): `/std:c++17` (or `-std:c++17)` is not too different from `-std=c++17`.
The annoying part is that MSVC does not update `__cplusplus` unless [explicitly told to do](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170). And some other "[backward compatibility features](https://learn.microsoft.com/en-us/cpp/build/reference/zc-conformance?view=msvc-170)".

What about a user variable like `CXX_STD=17` that expands to `$(CXX_STD:%=-std=c++%)` for gcc and to `$(CXX_STD:%=/std:c++%)` for MSVC ?

Or, if using `STD_CXXFLAGS` as proposed, the command line expansion for MSVC could be `$(subst =,:,$(STD_CXXFLAGS))`.

dirk-zimoch:
BTW, I am using `STD_CXXFLAGS` already for many years and define a default in CONFIG.Common.\<arch\>, adding it to `USR_CXXFLAGS` higher up in the config file hierarchy. Some archs get `STD_CXXFLAGS=-std=gnu++20`, some get `STD_CXXFLAGS=-std=c++0x` and MSVC may get for example `STD_CXXFLAGS=-std:c++17`.
However to make redefining easier in a module Makefile, I will probably change the expansion to:
`$(if $(filter msvc,$(CMPLR_CLASS)),$(subst =,:,$(STD_CXXFLAGS)),$(STD_CXXFLAGS))`

And I have added `-Zc:__cplusplus` to the MSVC C++ options, which is necessary from `-std:c++17` on because MSVC removes `auto_ptr` from c++17 on and without `__cplusplus` having the correct value, EPICS does not select the `unique_ptr` implementation.

mdavidsaver:
> What about a user variable like `CXX_STD=17` ...

This is closer to what I had in mind from my experience with cmake's [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html).  eg.

```cmake
set_property(TARGET tgt PROPERTY CXX_STANDARD 11)
```

anjohnson:
Thanks for the discussion, it sounds like we agree that Makefiles need to be able to set the language standard, and that it can be done on all our supported compilers. I hope we agree about that for both C and C++ code.

Another question that @mdavidsaver and @dirk-zimoch might like to think about: Might the build system want or need need to know what the compiler's language standard defaults to? If so, can we add some preprocessor tests to the configure/toolchain.c file to detect that?

I still want to be able to use target-specific variables as I mentioned above; these examples look suitable and not too verbose:

```Makefile
# Defaults for this Makefile
C_STD=11
CXX_STD=17

# Settings for specific source files
oldC.$(OBJ): C_STD = 99
oldCxx.$(OBJ): CXX_STD = 98
newC.$(OBJ): C_STD = 23
newCxx.$(OBJ): CXX_STD = 23
```

I suspect supporting the gnu standards (on GCC & Clang compilers only) should be trivial, and I hope we'll be able to error-out the build if someone tries to use those on MSVC.

mdavidsaver:
> ... need to know what the compiler's language standard defaults to?

The default does not seem interesting to me.  Knowing which standard versions are supported would be more interesting.

Ultimately though, if I care about language standard for a project, then I specify it.  If the compiler does not support that version at all, then the build will fail regardless.  If I care about being conditional, then I want the standard identification macros like `__cplusplus` to accurately indicate the supported language version.  None of this MSVC opt in to standards compliance for me!

> I still want to be able to use target-specific variables

Agreed, compile unit granularity seems desirable to me.

mdavidsaver:
Maybe something like:

```make
# definitions for for clang/gcc

LANG_CXX_CXX11 = -std=c++11
... for all supported for CMPLR_CLASS

LANG_C_C99 = -std=c99
```

```make
# definitions for impl msvc

# what to do about msvc having a much limited set of versions?  next one higher?
LANG_CXX_CXX11 = /Zc:__cplusplus /std:c++17

LANG_C_C99 = /std:c11
```

```make
# common definitions

LANG_CFLAGS = $(LANG_C_$(USR_LANG_C)) $(LANG_C_$($(basename $@)_LANG_C))

COMPILE.c = ... $(LANG_CFLAGS) ...

LANG_CXXFLAGS = $(LANG_CXX_$(USR_LANG_CXX)) $(LANG_CXX_$($(basename $@)_LANG_CXX))

COMPILE.cpp = ... $(LANG_CXXFLAGS) ...
```

```make
# usage

ifdef LANG_C_C99
# test if Makefiles support some language version
endif

USR_LANG_CXX = CXX11

PROD_SRCS += someFile.c

someFile_LANG_C = C99

```

mdavidsaver:
fyi. it looks like msvc version which understand `/std:c11` will update [`__STDC_VERSION__`](https://github.com/cpredef/predef/blob/master/Standards.md) without additional prompting.

dirk-zimoch:
> Knowing which standard versions are supported would be more interesting.

According to https://stackoverflow.com/questions/34836775/compiler-standards-support-c11-c14-c17, you can find out the supported standards of gcc with:
`gcc -v --help 2>/dev/null | sed -n '/^ *-std=\([^<][^ ]\+\).*/ {s//\1/p}'`

For MSVC I found that `cl /nologo /help` shows the required information:
```
[...]
/std:<c++14|c++17|c++latest> C++ standard version
    c++14 - ISO/IEC 14882:2014 (default)
    c++17 - ISO/IEC 14882:2017
    c++latest - latest draft standard (feature set subject to change)
[...]
/std:<c11|c17> C standard version
    c11 - ISO/IEC 9899:2011
    c17 - ISO/IEC 9899:2018
[...]
```
The necessary `sed` filter is left as an exercise for the reader.
The C standard options are available in MSVC 2019 but not in MSVC 2017. The listed C++ standard options are the the same for both versions but MSVC 2019 accepts `/std:c++20` as well without listing it in the `/help` output. MSVC 2022 finally lists `/std:c++20` in the `/help` output.

Other compilers may have different ways to find out, or none at all.

simon-ess:
So after setting this aside for several months, I've come back to pick it up along the lines of what @mdavidsaver suggested.

This currently adds a new configuration file (`CONFIG_LANG`, which could be debated) which defines all the variables on a per-std basis.

I preferred to allow both `SITE_LANG_*` as well as `USR_LANG_*`, since that seems to be more consistent with how most other configuration is handled. The msvc configuration is just a special case, since that seemed easiest.

simon-ess:
The choice of standards to include were from https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html and https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170.

configure/os/CONFIG.Common.solaris-sparc-gnu
```diff
@@ -9,7 +9,8 @@ include $(CONFIG)/os/CONFIG.Common.solaris-sparc
 
 COMPILER_CPPFLAGS = -D_REENTRANT
 
-POSIX_CFLAGS = -std=gnu99 -D_POSIX_C_SOURCE=200112L
+SITE_LANG_C = gnu99
+POSIX_CFLAGS = -D_POSIX_C_SOURCE=200112L
```
anjohnson: Please don't modify the config files for Solaris, to avoid conflicts if a Codeathon PR deletes them.

configure/os/CONFIG.Common.solaris-x86-gnu
```diff
@@ -9,7 +9,8 @@ include $(CONFIG)/os/CONFIG.Common.solaris-x86
 
 COMPILER_CPPFLAGS = -D_REENTRANT
 
-POSIX_CFLAGS = -std=gnu99 -D_POSIX_C_SOURCE=200112L
+SITE_LANG_C = gnu99
+POSIX_CFLAGS = -D_POSIX_C_SOURCE=200112L
```
anjohnson: Ditto…

configure/CONFIG_LANG
```diff
@@ -0,0 +1,32 @@
+ifneq ($(strip $(CMPLR_CLASS)),msvc)
+
```
anjohnson: The existing EPICS build rules get all their target-specific settings through their `$(CONFIG)/os/CONFIG.Common.$(T_A)` or `$(CONFIG)/os/CONFIG.$(EPICS_HOST_ARCH).$(T_A)` files, which can include other `$(CONFIG)/os/CONFIG.*` files for settings common to multiple targets. That was a deliberate design choice to simplify adding ports for new targets or removing old ones; please don't add a large file with conditionals like this for settings.

I suggest putting the MSVC settings in `$(CONFIG)/os/CONFIG.win32-x86.win32-x86` which is a catch-all for MS compiler flags and gets included by the other MSVC targets as well. If you wanted to split that monolith up into multiple files more like the Unixy targets do I would have no objection, but a split should probably go into a separate PR (I have some recollection that @dirk-zimoch may have proposed something like that at one point, but I don't remember what happened with it).

The non-MSVC settings may belong in `$(CONFIG)/CONFIG.gnuCommon` but that may cause complications with `CONFIG_SITE` overrides, so it might make more sense to put these into a new `$(CONFIG)/os/CONFIG.*` file instead.

configure/os/CONFIG.Common.RTEMS-uC5282
```diff
@@ -10,8 +10,8 @@ RTEMS_TARGET_CPU = m68k
 ARCH_DEP_CFLAGS += -DMY_DO_BOOTP=NULL
 
 # Hopefully a temporary fix:
-ARCH_DEP_CXXFLAGS_5 = -std=c++98
-ARCH_DEP_CXXFLAGS += $(ARCH_DEP_CXXFLAGS_$(RTEMS_VERSION))
+SITE_LANG_CXX_5 = c++98
+SITE_LANG_CXX = $(SITE_LANG_CXX_$(RTEMS_VERSION))
```
anjohnson: Given that this isn't a site-specific setting, the `SITE_LANG_` prefixes here don't seem right to me, how about `ARCH_LANG_` instead?

Given conflicting `SITE_LANG` and `ARCH_LANG` settings, which should win?

configure/CONFIG_COMMON
```diff
@@ -194,6 +194,10 @@ LIBRARY_LD_RESS = $(TARGET_RESS) $(LIBRARY_RESS)
 #--------------------------------------------------
 # C preprocessor, compiler, and linker flag defaults
 
+# C/C++ standard specification
+LANG_CFLAGS = $(LANG_C_$(SITE_LANG_C)) $(LANG_C_$(USR_LANG_C)) $(LANG_C_$($(basname $@)_LANG_C))
+LANG_CXXFLAGS = $(LANG_CXX_$(SITE_LANG_CXX)) $(LANG_CXX_$(USR_LANG_CXX)) $(LANG_CXX_$($(basname $@)_LANG_CXX))
```
anjohnson: Two typo's here, should be `$(basename $@)` instead of `$(basname $@)`.

I'm trying to work out what variable names should appear in a Makefile to set the language standards for this source directory and for specific object files, does the above code expect these, or have I misread it?

```makefile
# Defaults for this Makefile
USR_LANG_C = c11
USR_LANG_CXX = c++17

# File-specific:
older-C_LANG_C = c99
older-Cxx_LANG_CXX = c++98
newer-C_LANG_C = c23
newer-Cxx_LANG_CXX = c++23
```

I'm not convinced that the `c` and `c++` prefixes are useful in those values, but I won't push dropping them if others disagree.

In an earlier comment in this PR I encouraged using the target-specific variable assignment syntax built into GNUmake. I still prefer these, and my variable names:

```makefile
# Defaults for this Makefile
C_STD = c11
CXX_STD = c++17

# File-specific:
o = $(OBJ)
older-C$o: C_STD = c99
older-Cxx$o: CXX_STD = c++98
newer-C$o: C_STD = c23
newer-Cxx$o: CXX_STD = c++23
```

This also teaches users about the target-specific assignment syntax, which I find useful.

Your assignments above would shorten to this:

```makefile
LANG_CFLAGS = $(LANG_C_$(SITE_LANG_C)) $(LANG_C_$(C_STD))
LANG_CXXFLAGS = $(LANG_CXX_$(SITE_LANG_CXX)) $(LANG_CXX_$(CXX_STD))
```

configure/CONFIG_LANG
```diff
@@ -0,0 +1,32 @@
...
+# From the latest version of gcc
+SUPPORTED_CXX_STDS = 98 03 11 14 17 20 23
+SUPPORTED_C_STDS = 89 90 99 11 17 18 23
```
anjohnson: Should we aim to stop the build if a non-standard year is given?

I think it very likely that someone on Linux will set the standards to c11 and c++11 say, not realizing that MSVC didn't support c++11 so a Windows build of that code would do something weird. Similarly the gnu standards obviously aren't available on MSVC, but there's nothing to stop a build that requests them.

configure/CONFIG_SITE
```diff
@@ -164,6 +164,13 @@ EPICS_SITE_VERSION =
 #  Must be either YES or NO
 GCC_PIPE = NO
 
+# Set site-specific language defaults. Specify for example
+#  SITE_LANG_C = c23
+#  SITE_LANG_CXX = c++23
+# in order to set default compiler flags.
```
anjohnson: Could this say "language standards", or maybe "default language standards"?

I would like to see a warning here that (I believe) the defaults will apply to all targets built in this copy of Base and all support modules built using it, unless the standards are overridden more locally. It's taking a while for many support modules to build with the latest gcc.

I am also nervous about mixing code compiled using different language standards. I thought the g++ `std::string` changed it internal representation at one point, and I'm not sure if they were able to maintain the ability to link both.

anjohnson (changes requested):
Some comments, questions and requests.

configure/CONFIG_LANG
simon-ess: I had considered this, and in particular given that these numbers are really just passed through semi-directly to the compiler.

There is also the question of maintenance of these standards; as compilers are updated this is one of those things that would necessarily be changed constantly, and it would be nice if we could fetch them from the compiler instead of manually listing them.

I think this is why my original thought to implement this was just a matter of allowing users to specify the flags themselves. While it seems nice to implement something smart like this it does seem somewhat complex to try and implement.

That said, I'll try move the components into the more generic configure/os directories and see how that looks...

configure/os/CONFIG.Common.RTEMS-uC5282
simon-ess: ARCH_DEP might be better. My point with this PR is that I would like a natural way to set defaults _somehow_, but that this seems to be broadly a site-specific thing.

configure/CONFIG_SITE
simon-ess: Part of the challenge here is that we are sort of trying to support that  already, with the idea of setting it on a per-object file basis. Maybe `std=c++11` vs `std=c++17` is a smaller difference than `std=c++11` and `std=gnu++11` though.

It seems to me that all of these options make this into a bit of a rabbit-hole. In some sense I think we have to trust the users if they are specifying different standards for whatever reason they feel is necessary. To repeat my statement above, maybe it is better to just allow for a setting and let folks use it however they see fit...

configure/os/CONFIG.Common.RTEMS-uC5282
simon-ess: One downside though of ARCH_DEP is that we have the following priority:
```make
CFLAGS = $(LANG_CFLAGS) $($(BUILD_CLASS)_CFLAGS) $(POSIX_CFLAGS) $(OPT_CFLAGS) \
    $(DEBUG_CFLAGS) $(PIPE_CFLAGS) $(WARN_CFLAGS) $(TARGET_CFLAGS) \
    $(USR_CFLAGS) $(CMD_CFLAGS) $(ARCH_DEP_CFLAGS) $(CODE_CFLAGS) \
    $(STATIC_CFLAGS) $(OP_SYS_CFLAGS) $(LIBRARY_SRC_CFLAGS)
```
which puts `ARCH_DEP_CFLAGS` (and similarly for `CXXFLAGS`) at the top. This makes it a bit harder to override on an object-file basis... In fact, these are even higher than the target-specific ones!

configure/CONFIG_LANG
simon-ess: I've split this up as you suggested, which is roughly equivalent to how I had set it up in the first place. It doesn't conflict any further with the CONFIG_SITE overrides, since the order of loading is
* `CONFIG_SITE`
* `CONFIG.$(EPICS_HOST_ARCH).Common`

so this is the same as it was a moment ago.

configure/CONFIG_COMMON
simon-ess: The file-specific ones should be
```make
# Defaults for this Makefile
USR_LANG_C = c11
USR_LANG_CXX = c++17

# File-specific:
older_LANG_C = c99
older_LANG_CXX = c++98
newer_LANG_C = c23
newer_LANG_CXX = c++23
```
without the prefixes. But maybe your target-specific override is better. The downside a bit is that you need to know more make to remember this (not inherrently a bad idea!), but also that it doesn't match how other file-specific overrides works. But I could go either way. There is elegance in your solution.

anjohnson: Sorry I should have been clearer, I was assuming source files called `older-C.c`, `older-Cxx.cpp`, `newer-C.c` and `newer-Cxx.cpp` so the `-C` and `-Cxx` bits that you removed were actually parts of my filenames. Without them both `older.o` and `newer.o` might be created twice, or the C or C++ sources wouldn't all get compiled.

Anyway, I think my originals above were correct for my example.

simon-ess: Ah, I understand. I misunderstood which prefix you meant.

I'm putting those prefixes there more to allow for the distinction of `gnu++17` vs `c++17`. I suppose we could use more generically `gnu17` and `17` for the CXX standards and update the macros...

anjohnson: And I forgot that some files might need to specify `gnu` or `gnu++` standards, like this:

```makefile
older-C$o: C_STD = gnu99
older-Cxx$o: CXX_STD = gnu++98
```

I hereby withdraw this sentence from my earlier comment:

> I'm not convinced that the c and c++ prefixes are useful in those values, but I won't push dropping them if others disagree.

simon-ess:
Following discussion with @dirk-zimoch and @mdavidsaver I have added a default definition for msvc that will ensure that `__cplusplus` is defined correctly.

configure/CONFIG_COMMON
anjohnson: Looking again at the user Makefile variable names, these look best:

```makefile
USR_LANG_C = c11
USR_LANG_CXX = c++14

older-C_LANG_C = c99
older-Cxx_LANG_CXX = c++03
newer-C_LANG_C = c23
newer-Cxx_LANG_CXX = c++23
```

configure/CONFIG_SITE
anjohnson: After a conversation with @mdavidsaver the other day I want to reiterate my request to change the comment and names to this:

```Makefile
# Set the compiler language standard for code that doesn't set one
# and the compiler default doesn't work. If no standard is set or
# the variable is empty, the compiler default is used.
# 
LANG_C_STD = 
LANG_CXX_STD =
```

simon-ess:
> I am not sure why the RTEMS-5 job is failing. while this code does technically touch on compilation configuration for RTEMS 5, it should only do so for c++ files. Moreover, comparing the last successful run I found to this one, the exact same command is run.

So after waaaaaay to much time trying to figure this out I realised that the RTEMS config files redefine the CFLAGS and CXXFLAGS variables on their own 🙃

mdavidsaver:
@simon-ess Since the discussion for this PR has gotten so long, could you edit the initial message with a summary?  Something which might later end up as a release note.  Also, any open questions.  eg. what to do about `c++11` and msvc?

For my part, I think I would apply this to PVXS with logic like the following:

```
ifdef LANG_CXX_c++11
# Can select c++11 directly
DEFAULT_LANG_CXX = c++11

else # Base too old, or toolchain does can not select c++11
ifdef LANG_CXX_c++14
# Fall-up to 14 (MSVC)
DEFAULT_LANG_CXX = c++14

else # Base too old
# detect gcc/clang
USR_CXXFLAGS += -std=c++11
endif
endif
```

Needing to encode the "fall-up" to c++14 for MSVC is a bit awkward.  However, I am of two minds about whether the MSVC case should automatically "fall-up" to c++14.  eg.

```
LANG_CXX_c++11    = /std:c++14
```

simon-ess:
Updated comment above. To be clear regarding the fallup for msvc, since they officially don't support c++11: should I add 
```
LANG_CXX_c++11    = /std:c++14
```
to the msvc configuration?

anjohnson:
Core Group: Some more ideas to look at but maybe this works well enough. Will delay at least one more meeting.

PR #717, opened 2025-09-24
Docs: Mention that only pva is supported for JSON links

eddybl:
The documentation on JSON links: https://docs.epics-controls.org/projects/base/en/latest/links.html#id1 does not mention, that if using JSON `calc` links, only links to PVA records work right now, but not ca or database links.

### DIFF
--- a/modules/database/src/std/link/links.dbd.pod
+++ b/modules/database/src/std/link/links.dbd.pod
@@ -130,8 +130,9 @@ An optional expression that returns non-zero to raise a minor alarm.
 
 A JSON list of up to 24 input arguments for the expression, which are assigned
 to the inputs C<A>, C<B>, C<C>, ... C<U>. Each input argument may be either a
-numeric literal or an embedded JSON link inside C<{}> braces. The same input
-values are provided to the two alarm expressions as to the primary expression.
+numeric literal or an embedded JSON link inside C<{}> braces. Only C<pva> links
+are currently supported. The same input values are provided to the two alarm
+expressions as to the primary expression.
 
 =item out
 


modules/database/src/std/link/links.dbd.pod
```diff
@@ -130,8 +130,9 @@ An optional expression that returns non-zero to raise a minor alarm.
 
 A JSON list of up to 24 input arguments for the expression, which are assigned
 to the inputs C<A>, C<B>, C<C>, ... C<U>. Each input argument may be either a
-numeric literal or an embedded JSON link inside C<{}> braces. The same input
-values are provided to the two alarm expressions as to the primary expression.
+numeric literal or an embedded JSON link inside C<{}> braces. Only C<pva> links
+are currently supported. The same input values are provided to the two alarm
```
anjohnson, line 134: I'd like this to be worded in a way that doesn't require editing when other JSON link types such as a future "ca" link get added. That's harder to do, and is why I didn't put that in the original description. Any suggestions?

eddybl: I assume this issue is a more general limitations of JSON links? Maybe the "general limitations" of JSON links could be listed elsewhere?

anjohnson: It's not really a limitation of JSON links, the older non-JSON link implementations don't provide the necessary function table to be used as a child link. That can be done to create a "ca" link type, but won't be possible for "db" links because of the additional local locking requirements of those links.

eddybl: ok, so we should then probably close this PR, as I currently do not have a better suggestion of how to document the fact that this currently only works for pva, ca could be implemented but will not work for dblinks.

Maybe create an issue to track the ca support (possible codathon project?)?

Or.. can you force a link to be a CA link similar to normal input links with the `<LINK> CA` flag?
I could see some scenarios where  the JSON `calc` feature would replace internal calc records, but that would mostly be inside an IOC...

anjohnson:
Core group: Maybe use the phrase used at the top of the document "older link types that do not use JSON" to describe the link types that can't be used in the `args` list?

anjohnson:
Core: ANJ to review and fix.

anjohnson:
Applied with edit as 7d6ef325025a0c91f715f7943d007023742a8947

Issue #718, opened 2025-10-01
callbackParallelThreads() negative count handling?


mdavidsaver:
While writing doc. comments for callback.h I noticed something odd about the handling of negative count values for `callbackParallelThreads()`.

https://github.com/epics-base/epics-base/blob/7444936a22c00a8776f3bba9bed84acf91801b1c/modules/database/src/ioc/db/callback.c#L163-L164

eg. `-10` on an 8 core system would result in a count of `8 + -10 == -2`.  This would later be silently be corrected to `1`, so no harm done.  

@anjohnson @ralphlange Do you remember if the intended behavior for a negative number was to select a number of workers in excess of the core count?

This would be:

```c
    if (count < 0)
        count = epicsThreadGetCPUs() - count;
```

ralphlange:
I guess that might be missing a limit? (Or maybe that's what the silent correction to 1 is.)

In any case: The intended behavior was to use `count` _less_ workers than CPU cores, but at least one.

anjohnson:
I agree with Ralph's edited comment, the existing code does exactly what was originally intended — negative numbers mean start a worker thread for most CPU cores, but reserve this many CPUs for other work. This code won't lock the workers to a specific CPU, but presumably that's something you would do with MCoreUtils.

mdavidsaver:
I guess I have demonstrated that documentation is necessary!  I will continue with #719.

PR #719, opened 2025-10-01
doc for callback.h and callbackSetQueueSize sanity check

mdavidsaver:
Add doc. comments to `callback.h`.  The only functional change is an added sanity check for `callbackSetQueueSize()` to reject non-positive sizes.

~~I'm making this a draft PR pending an answer to #718.~~

### DIFF
--- a/modules/database/src/ioc/db/callback.c
+++ b/modules/database/src/ioc/db/callback.c
@@ -100,6 +100,10 @@ static int priorityValue[NUM_CALLBACK_PRIORITIES] = {0, 1, 2};
 
 int callbackSetQueueSize(int size)
 {
+    if (size<=0) {
+        fprintf(stderr, "Queue size must be positive\n");
+        return -1;
+    }
     if (epicsAtomicGetIntT(&cbState)!=cbInit) {
         fprintf(stderr, "Callback system already initialized\n");
         return -1;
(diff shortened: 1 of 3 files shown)


anjohnson:
Most of the newer Doxygen annotations in `modules/database/src/ioc/db` use `@` instead of `\` to introduce keywords; I think in one Codeathon we suggested that because `@` is easier to grep for, although both work so this isn't a blocker.

This file isn't appearing in the RTD documentation yet. The header needs an `@file ` annotation adding near the top for Doxygen to process it, and then Sphinx needs `documentation/Makefile.headers` to add this to the "IOC Database C/C++ APIs" index:
```Makefile
database_HEADERS += callback
```
I want to automate the latter (it should be reasonably straightforward), but we just haven't got around to doing it yet.

anjohnson:
Core: Approved after modification to `documentation/Makefile.headers` as above.

mdavidsaver:
Merged as of c2140ad0f32cd510d8a41809213a016c9b53ceea

I changed the doxygen tag style from `\` to `@` (which qtcreator wants to do anyway), and added some additional discussion of timer allocation to `callbackRequestDelayed()`.

I tested a local `make -C documentation doxygen` and see that a page for `callback.h` is created.

Issue #720, opened 2025-10-03
Unknown MSVC option '-analysis'


dirk-zimoch:
.github/workflows/ci-scripts-build.yml contains `CMD_CXXFLAGS=-analysis` for some Windows builds.
However, MSVC does not know such an option:
```
 cl : Command line warning D9002 : ignoring unknown option '-analysis'
```
There is an option with a similar name: [`-analyze`](https://learn.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis?view=msvc-170)
As the builds succeed without that option, I am not sure if `-analysis` should be replaced with `-analyze` or simply removed.
The option had been added in commit b878295d06.

mdavidsaver:
This typo is probably my doing, carried over from epics-base/pvxs@230fbc19b7df8fd90c11a42da60c5c1086b5a4c6.

> I am not sure if `-analysis` should be replaced with `-analyze` or simply removed.

I think replaced unless the default has changed to "on".

PR #730, opened 2025-10-27
Ci win cplusplus

dirk-zimoch:
New ci builds for Windows using option [`-Zc:__cplusplus`](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170):
  * -std:c++14 (the default)
  * -std:c++17
  * [currently commented out](https://github.com/epics-base/pvDataCPP/issues/99): -std:c++20
  * currently commented out: [-std:c++latest](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170#stdclatest)

Additionally, I have changed the [unknown (probably misspelled) command line option `-analysis`](https://github.com/epics-base/epics-base/issues/720) to [`-analyze`](https://learn.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis?view=msvc-170).

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -171,19 +171,6 @@ jobs:
             name: "Win-22 MSC-22 c++17"
             extra: "CMD_CXXFLAGS='-analyze -Zc:__cplusplus -std:c++17'"
 
-# pvData does not build with MVSC c++20
-#          - os: windows-2022
-#            cmp: vs2022
-#            configuration: debug
-#            name: "Win-22 MSC-22 c++20"
-#            extra: "CMD_CXXFLAGS='-analyze -Zc:__cplusplus -std:c++20'"
-#
-#          - os: windows-2022
-#            cmp: vs2022
-#            configuration: debug
-#            name: "Win-22 MSC-22 c++latest"
-#            extra: "CMD_CXXFLAGS='-analyze -Zc:__cplusplus -std:c++latest'"
-
           - os: windows-2022
             cmp: vs2022
             configuration: static-debug
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -157,13 +157,13 @@ jobs:
             cmp: vs2022
             configuration: debug
             name: "Win-22 MSC-22"
-            extra: "CMD_CXXFLAGS=-analysis"
+            extra: "CMD_CXXFLAGS=-analyze"
 
           - os: windows-2022
             cmp: vs2022
             configuration: static-debug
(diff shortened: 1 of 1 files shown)


mdavidsaver:
> the [unknown (probably misspelled) command line option `-analysis`](https://github.com/epics-base/epics-base/issues/720) to [`-analyze`](https://learn.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis?view=msvc-170).

This is my misspelling, originating from PVXS.  I have corrected it there with https://github.com/epics-base/pvxs/commit/2a3db0c9b25a57cdfeeae0343a3760657d75ef9c

Since the new jobs pass, I think it is reasonable to merge.

Going through all of the new warnings though...  that will be a larger task.

```
   D:\a\epics-base\epics-base\modules\libcom\src\fdmgr\fdManager.cpp(244) : warning C6328: Size mismatch: 'unsigned __int64' passed as _Param_(2) when 'int' is required in call to 'errlogPrintf'. 
```

dirk-zimoch:
> Going through all of the new warnings though... that will be a larger task.

It had been a larger task already to go though the Windows warnings before `-analyze`.

Many of the Windows warnings are totally justified. In particular those about not matching size. Passing an `__int64` for an `int` may lead to unexpected information loss. At least the Windows compiler warns about that.

Gcc warns about passing an `int64_t` value for a `"%d"` format as well (with `-Wall`) but not about assigning an `int64_t` value to an `int` variable, which is basically the same problem. Neither does clang, even though it generally has better (i.e. more helpful) warnings than gcc  But Windows does. That is why the Windows builds have always shown much more warnings than the Linux builds. Using explicit casts remove those warnings, telling the compiler "I know what I am doing." (which may be wrong if only done to silence the annoying warnings).

> `Size mismatch: 'unsigned __int64' passed as _Param_(2) when 'int' is required in call to 'errlogPrintf'.`

Windows using `unsigned __int64` instead of `int` for file descriptors and sockets is a lot of fun. That would require a lot of compiler branches (or casting, which is kind of "safe" at the moment, because those `unsigned __int64` file descriptors only actually hold small numbers -- in the current Windows versions. Subject to change without notice, of course).
Unfortunately, inttypes.h does not define a `PRI*` macro for file descriptor types (probably assuming that is it `int` on all systems anyway). Using C++ `std::cerr << fd` does not nave this problem, of course.

Affects every debug message where file descriptors or sockets are printed. On the other hand, does it make much sense to print those? For whom may that information be useful? The really useful information would be the file name or the socket port (and peer host:port) associated with that fd. But that is difficult to get.

anjohnson:
Core: Please remove the commented-out lines, then it's approved.

PR #731, opened 2025-10-28
docs: clarify Drive Limits behavior in aoRecord

jerzyjamroz:
Clarification added:
`Note: These limits are not enforced if they are not set or if DRVH E<lt>= DRVL.`

It is specified in `longoutRecord`, but `aoRecord` lacks it.

### DIFF
--- a/modules/database/src/std/rec/aoRecord.dbd.pod
+++ b/modules/database/src/std/rec/aoRecord.dbd.pod
@@ -69,9 +69,13 @@ output value PVAL is added to it.
 
 =head4 Drive Limits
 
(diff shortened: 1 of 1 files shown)


simon-ess (commented):
Seems fine, but if you want to harmonise these it might be good to include the more explicit
```
The VAL field's value will be clipped within limits specified in the fields DRVH
and DRVL if these have been configured by the database designer:

  DRVL <= VAL <= DRVH
```
from the longoutRecord.dbd.pod file.

jerzyjamroz:
@simon-ess , done. 
The important info addition here is the condition when it is not used even if you configure the fields.

PR #735, opened 2025-11-06
Draft: Always include EPICS base "module" configuration files

### LINKED ISSUE #724: INSTALL_LOCATION broken
Commit e4fbd83 broke installations with `INSTALL_LOCATION` set.

Building asLib.c fails with:
```
In file included from ../as/asLib.y:205:
../as/asLib_lex.l: In function ‘yyreset’:
../as/asLib_lex.l:31:15: error: ‘INITIAL’ undeclared (first use in this function)
         BEGIN INITIAL;
               ^~~~~~~
```
I suggest we add builds with `INSTALL_LOCATION` to our CI.

simon-ess:
EPICS base offers a path to install custom configuration files using the `CFG` variable. This is also used to install a few files from the EPICS build system. However, this setup provides a source of conflict between EPICS base and any modules that want to do this.

Since the EPICS cfg files are always included (EPICS_BASE must be defined in RELEASE), it makes sense to simply install these as regular configuration files instead of cfg/ files.

Resolves #724

### DIFF
--- a/configure/CONFIG
+++ b/configure/CONFIG
@@ -76,7 +76,7 @@ RELEASE_TOPS := $(shell $(CONVERTRELEASE) -T $(TOP) releaseTops)
 ifdef T_A
   # Information from the target's compiler
   #
-  -include $(EPICS_BASE)/cfg/TOOLCHAIN.$(EPICS_HOST_ARCH).$(T_A)
+  -include $(CONFIG)/TOOLCHAIN.$(EPICS_HOST_ARCH).$(T_A)
 
   #  Cross compile specific definitions
   #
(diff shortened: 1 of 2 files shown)


configure/CONFIG
```diff
@@ -118,6 +118,15 @@ ifneq ($(CONFIG),$(TOP)/configure)
   endif
 endif
 
+include $(CONFIG)/CONFIG_LIBCOM_MODULE
```
simon-ess, line 121: Not that this is a reasonable compromise between the previous behaviour and the proposed behaviour:

For EPICS base we used to include these files under TOP_CFG_CONFIGS; for other modules we would include them via RELEASE_CFG_CONFIGS. This middle ground means that we will include them in every case at "the same time" (or close enough).

configure/Makefile
```diff
@@ -21,16 +21,7 @@ CONFIGS += $(subst ../,,$(wildcard ../os/CONFIG_SITE.*))
 CONFIGS += $(subst ../,,$(wildcard ../RELEASE*))
 CONFIGS += $(subst ../,,$(wildcard ../RULES*))
 
-CFG += CONFIG_LIBCOM_MODULE
```
simon-ess, line 24: These files were already being installed by the lines above,
```make
CONFIGS += $(subst ../,,$(wildcard ../CONFIG*))
```
This just means we only install them into a single location.

simon-ess:
I just realised that this doesn't work with the v4 submodules; while I think that morally an approach such as this is plausibly correct, it may be thornier than I expected.

PR #736, opened 2025-11-07
added __attribute__((__malloc__)) for gcc and clang

dirk-zimoch:
The macro `EPICS_MALLOC` expands to `__attribute__((__malloc__))` for gcc and clang.
It is used to mark `mallocMustSucceed()` and `callocMustSucceed()`.

### DIFF
--- a/modules/libcom/src/osi/compiler/gcc/compilerSpecific.h
+++ b/modules/libcom/src/osi/compiler/gcc/compilerSpecific.h
@@ -69,4 +69,13 @@
 #define EPICS_NORETURN __attribute__((noreturn))
 #endif
 
+/*
+ * malloc marker takes 1 or 2 args: (index of size) or (index of count, index of element size)
+ */
+#if __GNUC__ * 100 + __GNUC_MINOR__  >= 403
+#define EPICS_MALLOC(...) __attribute__((__malloc__, __alloc_size__(__VA_ARGS__)))
+#else
+#define EPICS_MALLOC(...) __attribute__((__malloc__))
+#endif
+
 #endif  /* ifndef compilerSpecific_h */
--- a/modules/libcom/src/misc/cantProceed.h
+++ b/modules/libcom/src/misc/cantProceed.h
@@ -67,7 +67,7 @@ void cantProceed(
  * Will never return NULL otherwise.
  */
 LIBCOM_API void * callocMustSucceed(size_t count, size_t size,
-    const char *errorMessage);
+    const char *errorMessage) EPICS_MALLOC(1,2);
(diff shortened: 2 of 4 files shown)


modules/libcom/src/osi/compiler/gcc/compilerSpecific.h
```diff
@@ -69,4 +69,9 @@
 #define EPICS_NORETURN __attribute__((noreturn))
 #endif
 
+/*
+ * malloc marker
+ */
+#define EPICS_MALLOC __attribute__((__malloc__))
```
mdavidsaver: Looking at the declaration of `malloc()` from glibc, both the `__malloc__` and `__alloc_size__(...)` [attributes](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) are given.  This combines a malloc aliasing hint with an object size hint.  With `-D_FORTIFY_SOURCE=3`, size hints have teeth.

```c
#define EPICS_MALLOC(...) __attribute__(__malloc__, __alloc_size__(__VA_ARGS__))
```

Since there would either 1 or two arguments, I don't think this would cause problems with the MSVC "traditional" preprocesser.

dirk-zimoch: `__alloc_size__` only exists since gcc 4.3  --  fixed for gcc; what about clang?

dirk-zimoch: Some pre-C99 compilers may be offended by a variadic macro. :scream:

dirk-zimoch:
Rebased and `__alloc_size__` added where possible.

dirk-zimoch:
Several Appveyor jobs (vs2013-vs2017) fail with
>  ERROR: Package 'ci-core-dumper' requires a different Python: 2.7.17 not in '>=3.8'
WARNING: You are using pip version 19.3.1; however, version 20.3.4 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
Command exited with code 1

The vs2019 jobs succeed.

There was a deprecation warning for a while:

>DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support

mdavidsaver:
> Several Appveyor jobs (vs2013-vs2017) fail with

https://github.com/epics-base/epics-base/pull/770

PR #737, opened 2025-11-10
libCom/posix: add env var to opt out of FIFO scheduling

### LINKED ISSUE #734: Add env variable to allow *not* using real-time scheduling
The mechanism in the OSI lib switches priority scheduling on when it finds EPICS was compiled with USE_POSIX_THREAD_PRIORITY_SCHEDULING and the application is running on a system that allows it to do so. 

There is no way to have the application *not* switch to priority scheduling in such a case.

Use case:
Processes that are part of the ITER Real-Time-Framework need to have the necessary roles/authorizations. However, the EPICS PVA client they are starting should *not* use priority scheduling, as it is run on a specific set of cores for the non real-time stuff.

Suggest: Add an environment variable that acts similar to setting USE_POSIX_THREAD_PRIORITY_SCHEDULING = NO
(We use that as a workaround.)

ralphlange:
- add EPICS_ALLOW_POSIX_THREAD_PRIORITY_SCHEDULING (default: YES)
- start EPICS threads with default policy when set to NO

fixes #734


documentation/new-notes/PR-737.md
```diff
@@ -0,0 +1,5 @@
+### Add environment variable to opt out of POSIX Real-Time scheduling
+
+On POSIX systems, processes with real-time capabilities can opt out of using
+Posix thread priority scheduling and memory locking.
+Set EPICS_USE_POSIX_THREAD_PRIORITY_SCHEDULING=NO to achieve this.
```
anjohnson: Strict Markdown rules require literal underscore characters to be escaped, and in this case we want RTD to format that setting as code so please use back-ticks, like this:
```markdown
Set `EPICS_THREAD_ALLOW_PRIORITY_SCHEDULING=NO` in the environment to achieve this.
```

configure/CONFIG_ENV
```diff
@@ -53,3 +53,5 @@ EPICS_IOC_IGNORE_SERVERS=""
 # EPICS_IOC_LOG_PORT Log server port number etc.
 EPICS_IOC_LOG_PORT=7004
 
+# Posix priority scheduling
+EPICS_USE_POSIX_THREAD_PRIORITY_SCHEDULING=YES
```
anjohnson: With my naming Czar hat on, the name of this setting doesn't quite match what it does.

The build variable `USE_POSIX_THREAD_PRIORITY_SCHEDULING` that is set in `configure/CONFIG_SITE` controls whether the epicsThreads implementation should be built to allow suitably privileged IOCs to use priority scheduling. I question whether anyone actually disables that now (I suspect it was needed for some early Linux libc versions, we should check the core-talk archives), but that's a different topic and might become a Codeathon project to rip it out completely.

This new environment variable provides the ability to deny the use of priorities for code built on epicsThreads, but if the build was configured to disable them this can't force them to be used (so the name shouldn't say "use").

The comment should be something like "Allow/prevent POSIX priority scheduling of EPICS threads", and the variable name should be more like `EPICS_THREAD_ALLOW_POSIX_PRIORITY_SCHEDULING`

anjohnson (changes requested):
Name and doc changes

anjohnson (approved):
Thanks, looks good now.

ralphlange:
Confirmed as working in the original setup by @anzez.

anjohnson:
The long environment variable name used here coupled with a recent change to iocStats unfortunately triggered [this issue](https://github.com/epics-modules/iocStats/issues/75).

PR #738, opened 2025-11-10
Fix gcc warning 'stringop-truncation'

dirk-zimoch:
* When using `strncpy()`, always use full `length` (not `length-1`) and set `target_string[length-1] = 0;` immediately afterwards.
* Do not use `sizeof(source)`
* Use `memcpy()` instead if possible, e.g. if `strlen()` had been used before to calculate `length`.

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -743,8 +743,8 @@ static long dbAddOnePath (DBBASE *pdbbase, const char *path, unsigned length)
     ppathList = (ELLLIST *)pdbbase->pathPvt;
 
     pdbPathNode = (dbPathNode *)dbCalloc(1, sizeof(dbPathNode));
-    pdbPathNode->directory = (char *)dbCalloc(length+1, sizeof(char));
-    strncpy(pdbPathNode->directory, path, length);
+    pdbPathNode->directory = (char *)dbMalloc(length+1);
+    memcpy(pdbPathNode->directory, path, length);
     pdbPathNode->directory[length] = '\0';
     ellAdd(ppathList, &pdbPathNode->node);
     return 0;
--- a/modules/ca/src/perl/Cap5.xs
+++ b/modules/ca/src/perl/Cap5.xs
@@ -613,6 +613,7 @@ void CA_put(SV *ca_ref, SV *val, ...) {
                 break;
             case DBF_STRING:
                 strncpy(data.dbr_string, SvPV_nolen(val), MAX_STRING_SIZE);
+                data.dbr_string[MAX_STRING_SIZE-1] = 0;
                 break;
             }
             status = ca_put(type, pch->chan, &data);
@@ -705,6 +706,7 @@ void CA_put_callback(SV *ca_ref, SV *sub, SV *val, ...) {
                 break;
             case DBF_STRING:
                 strncpy(data.dbr_string, SvPV_nolen(val), MAX_STRING_SIZE);
+                data.dbr_string[MAX_STRING_SIZE-1] = 0;
                 break;
             }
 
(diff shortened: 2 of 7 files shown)


modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -743,8 +743,8 @@ static long dbAddOnePath (DBBASE *pdbbase, const char *path, unsigned length)
     ppathList = (ELLLIST *)pdbbase->pathPvt;
 
     pdbPathNode = (dbPathNode *)dbCalloc(1, sizeof(dbPathNode));
-    pdbPathNode->directory = (char *)dbCalloc(length+1, sizeof(char));
-    strncpy(pdbPathNode->directory, path, length);
+    pdbPathNode->directory = (char *)dbMalloc(length+1);
+    strncpy(pdbPathNode->directory, path, length+1);
```
mdavidsaver: Since the length is already known exactly, this could become a `memcpy()`, which would also silence this warning.

modules/database/src/std/rec/biRecord.c
```diff
@@ -181,10 +181,10 @@ static long get_enum_str(const DBADDR *paddr, char *pstring)
     if(index!=biRecordVAL) {
         strcpy(pstring,"Illegal_Value");
     } else if(*pfield==0) {
-        strncpy(pstring,prec->znam,sizeof(prec->znam));
+        memcpy(pstring,prec->znam,sizeof(prec->znam));
```
mdavidsaver: Since the length is not known here, the `strncpy()` is more appropriate.  The `memcpy()` may copy additional undefined bytes over the network.  In other circumstances this would be seen as an information disclosure vulnerability.  In this case, it is probably innocuous since these leftover bytes would usually have been sent to some client.

I am uncertain whether to ask for a change.

anjohnson:
Core Group: See comments, still missing a `STATIC_ASSERT()`.

modules/database/src/std/rec/biRecord.c
mdavidsaver: > I am uncertain whether to ask for a change.

After further discussion, please restore `strncpy()`.  Instead of `sizeof()`, substitute the appropriate macro, and add a `STATIC_ASSERT()` of equality between that macro and `sizeof()`.

dirk-zimoch: This has turned out to be a bit tricky. The `pstring` argument always points to an EPICS string size 40 in the caller's space. But of course `get_enum_str()` does not know that. The field itself is only 26 chars long. It makes sense to use the source size instead of the technically unknown target size because according to a comment in [dbAccess.c](https://github.com/epics-base/epics-base/blob/419a704ad9c1c653d15a05c800b1e47aa5589e1f/modules/database/src/ioc/db/dbAccess.c#L194), the enum string fields were, at least originally, not expected to be 0-terminated. The correct way to handle them under this assumption is indeed to copy all `sizeof(field)` (26) characters and then add a 0 to index `[sizeof(field)]` (*not* overwrite index `[sizeof(field)-1]` !) , as the current implementation does. Only modern compilers complain because they assume unterminated strings are a dangerous mistake (and in most cases they are right).

The only suitable macro here would be `MAX_ENUM_STRING_SIZE` (26) but that requires `#include "db_access.h"`, leading to other problems because `dbAccess.h` is included too. I think, `sizeof(field)` is the better solution here. Without the macro and without information about the target buffer size, `STATIC_ASSERT` no longer makes any sense here. There is nothing left to assert.

In the current implementation, it is no longer possible for the enum string fields to be unterminated. Thus we could simply replace the call with `strcpy()` later (but then possibly leave old bytes in the buffer after the string end).

I think, for the moment, I will simply remove the changes in the record implementations from this PR, in order to proceed with  the non-controversial part. This leaves the compiler still warning about all records with enums.

PR #740, opened 2025-11-10
gcc < 5 does not know #pragma GCC diagnostic push

dirk-zimoch:
But it also does not know warning -Wformat-security.

### DIFF
--- a/modules/database/src/std/rec/printfRecord.c
+++ b/modules/database/src/std/rec/printfRecord.c
@@ -57,7 +57,7 @@
     else \
         flags |= F_BADLNK
 
-#ifdef __GNUC__
+#if __GNUC__ >= 5 || defined(__clang__)
 #  pragma GCC diagnostic push
 #  pragma GCC diagnostic ignored "-Wformat-security"
 /* Intentionally passing non-const format string to epicsSnprintf() below.
@@ -322,7 +322,7 @@ static void doPrintf(printfRecord *prec)
     prec->len = pval - prec->val;
 }
 
-#ifdef __GNUC__
+#if __GNUC__ >= 5 || defined(__clang__)
 #  pragma GCC diagnostic pop
 #endif
 


modules/database/src/std/rec/printfRecord.c
```diff
@@ -57,7 +57,7 @@
     else \
         flags |= F_BADLNK
 
-#ifdef __GNUC__
+#if __GNUC__ >= 5 || defined(__clang__)
```
anjohnson, line 60: Clang sets `__GNUC__` anyway, what does it set it to?

dirk-zimoch: My clang sets it to 4.

dirk-zimoch:
```
$ echo | clang -E -dM - | grep GNU
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
$ clang --version
clang version 19.1.7 ( 19.1.7-2.module+el8.10.0+23045+e1f8e80e)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/x86_64-redhat-linux-gnu-clang.cfg
```

anjohnson: Thanks, this condition makes sense then.

```bash
$ echo | clang -E -dM - | grep GNU
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
$ clang --version
Apple clang version 17.0.0 (clang-1700.4.4.1)
Target: x86_64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

anjohnson (approved):
LGTM

PR #741, opened 2025-11-11
fix format warning on mingw

dirk-zimoch:
Since we use `gnu_printf` format on mingw, the windows `I` modifier is no longer recognized by the format checker. Use `ll` instead. Meanwhile, MSVC understands that as well.

### DIFF
--- a/modules/libcom/src/fdmgr/fdManager.cpp
+++ b/modules/libcom/src/fdmgr/fdManager.cpp
@@ -19,6 +19,8 @@
 // 1) This library is not thread safe
 //
 
+#include <iostream>
+
 #define instantiateRecourceLib
 #include "epicsAssert.h"
 #include "epicsThread.h"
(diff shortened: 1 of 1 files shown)


modules/libcom/src/fdmgr/fdManager.cpp
```diff
@@ -382,7 +382,7 @@ void fdRegId::show(unsigned level) const
     if (level > 1u) {
         printf("\tfd = %"
 #if defined(_WIN32)
-            "I"
+            "ll"
 #endif
         "d, type = %d\n",
             fd, type);
```
mdavidsaver: Could these lines become one, and avoid an `#if`?

```c
printf("\tfd = %lld, type = %d\n", (long long)fd, type);
```

dirk-zimoch: It would probably be easier in C++ style (`cout << `)

mdavidsaver: No doubt simpler source code.  Personally I do no think it is worth changing to `std::cout` in this case, but I have no objection provided that all of the `printf()`s in fdManager.cpp are changed.  It is good practice to avoid mixing stdout and std::cout due to differences in buffering.

I do not think that `epicsStdio.h` is included, so iocsh redirection is not relevant.  The buffering behavior could be different unless `std::endl` or `std::cout.flush()` happens.

dirk-zimoch:
Using `std::cout <<` instead of `printf()` and enhancing output of enums from numbers to meaningful strings.
(Of course printing the raw `this` pointer is still quite meaningless.)

PR #742, opened 2025-11-11
fix(aao, aai): change VAL field type from DOUBLE[] to FTVL[]

jerzyjamroz:
I realized that `aao` and `aai` records are not actually `analog` records, since they follow the `FTVL` field type.
Out of curiosity - why not simply name them `array records`?

### DIFF
--- a/modules/database/src/std/rec/aaiRecord.dbd.pod
+++ b/modules/database/src/std/rec/aaiRecord.dbd.pod
@@ -7,7 +7,7 @@
 # in file LICENSE that is included with this distribution.
 #*************************************************************************
 
-=title Array Analog Input (aai)
+=title Array Input (aai)
 
 The array input record type is used to read array data. The array data can
 contain any of the supported data types. The record is in many ways similar to the
--- a/modules/database/src/std/rec/aaiRecord.dbd.pod
+++ b/modules/database/src/std/rec/aaiRecord.dbd.pod
@@ -9,7 +9,7 @@
 
 =title Array Analog Input (aai)
 
-The array analog input record type is used to read array data. The array data can
+The array input record type is used to read array data. The array data can
 contain any of the supported data types. The record is in many ways similar to the
 waveform record. It allows, however, the device support to allocate the array
 storage.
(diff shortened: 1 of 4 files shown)


modules/database/src/std/rec/aaiRecord.dbd.pod
```diff
@@ -311,7 +311,7 @@ Scan forward link if necessary, set PACT FALSE, and return.
 		special(SPC_DBADDR)
 		pp(TRUE)
 		extra("void *		val")
-		#=type DOUBLE[NELM]
+		#=type Set by FTVL[NELM]
```
anjohnson, line 314: I agree with replacing `DOUBLE` here and for the aao; the `[NELM]` suffix is interesting. The aSub and waveform records don't currently use it, but the lsi, lso and printf records do use `[SIZV]`, so maybe the aSub and waveform records should. If you don't mind fixing them too the aSub.A field that would be `Set by FTA[NOA]`, aSub.VALA would be `Set by FTVA[NOVA]`.

anjohnson (approved):
Well spotted, this is a definite bug in the generated documentation.

I asked for a few more similar changes in the same vein below, they're optional but would be good if you have time to make them too.

Thanks!

anjohnson:
> why not simply name them array records?

Good question. My guess is that they started out only supporting `DOUBLE`s, but got extended later on. @dirk-zimoch, was that something you added when you took them over?

ralphlange:
Unless we add aliases for record types, renaming breaks every single application that uses the record type, that's why they never get renamed.
(Many record types have weird names that don't really match. E.g.,  mbb[io], mbb[io]Direct, waveform, compress, ...)

dirk-zimoch:
> dirk-zimoch, was that something you added

The records already existed in base. I only added them to the list of records to build and link, added a Soft Channel support and fixed the init-record function to allocate the buffer if the device support didn‘t.

modules/database/src/std/rec/aaiRecord.dbd.pod
jerzyjamroz: Done

jerzyjamroz:
> Unless we add aliases for record types, renaming breaks every single application that uses the record type, that's why they never get renamed. (Many record types have weird names that don't really match. E.g., mbb[io], mbb[io]Direct, waveform, compress, ...)

No problem, I removed the word `analog` from aai and aao documentation (so no missleading) and all works. `aa` can be an acronym of `array` at the end if we agree.

PR #743, opened 2025-11-11
add HOST_BUILD option to CONFIG_SITE

JJL772:
## Description

If HOST_BUILD=YES, the previous behavior of the EPICS build system is preserved. Cross targets will depend on the $(EPICS_HOST_ARCH) targets.

If HOST_BUILD=NO, cross targets will build independently of $(EPICS_HOST_ARCH), except for EPICS base where the option is forcefully enabled.

## Rationale 

This is useful for situations where you only want to cross compile instead of building for the host architecture too. For example, Yocto/BitBake recipes require some fanangling to work correctly with EPICS packages since they're always producing and installing binaries for the target _and_ build host. For areaDetector packages, where lots of dependencies start to get pulled in, building for the host architecture starts to become a real problem. The build machine ends up needing a bunch of packages installed just so that it can cross compile.

I'm aware of some caveats with this idea, but I don't think I have the full picture. Feedback would be greatly appreciated.

### DIFF
--- a/configure/CONFIG_SITE
+++ b/configure/CONFIG_SITE
@@ -139,6 +139,11 @@ HOST_WARN=YES
 #  Must be either YES or NO.
 CROSS_WARN=YES
 
+# Always build for the host architecture in addition to cross targets.
+# This will be forced on for EPICS base to ensure that host tools (i.e. msi) are available.
+#  Must be either YES or NO.
+HOST_BUILD=YES
+
 # Installation directory, if you want Base to be installed into a
 #  different location then uncomment and set this.
 #INSTALL_LOCATION=<fullpathname>
--- a/configure/RULES_ARCHS
+++ b/configure/RULES_ARCHS
@@ -16,8 +16,21 @@ ACTIONS += install
 ACTIONS += buildInstall
 ACTIONS += runtests tapfiles clean-tests test-results junitfiles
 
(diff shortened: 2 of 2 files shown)


anjohnson:
Take a look at PR #665 which was my start to allow minimizing the code that gets built for the host to just the parts that are essential for cross-building. It works for the epics-base code, but needs changes to the PVA sub-modules and I didn’t get very far through them IIRC.

Base isn’t the only place where host software must be built to allow cross-targets to be compiled, the SNL compiler is similar to the e_flex and antelope programs in Base in that it needs to be built for use on the host at run-time. Most other support modules would need to be looked at as well, but only to make sure they’re using the right Makefile variables.

JJL772:
That definitely seems like a cleaner solution, especially with the relatively fine-grained control.

My idea with this PR was to allow individual modules to explicitly set `HOST_BUILD=YES` in their CONFIG_SITE to override the user setting in EPICS base. That way packages like SNL could force host builds. But after thinking about it more, it's not great for modules that build host tools and IOCs/device support, since it's an all-or-nothing option.

I'm going to close this in favor of #665. Looking forward to using that when it's merged. Cheers!

anjohnson:
Okay, thanks. Feel free to review or test it out and add comments that you support it, I'm not sure how many of the other core dev's think it's worth doing.

PR #744, opened 2025-11-13
Avoid early expiration of timers on !vxWorks

mdavidsaver:
Address #106 on non-RTOS targets.

### DIFF
--- a/modules/libcom/src/timer/timerPrivate.h
+++ b/modules/libcom/src/timer/timerPrivate.h
@@ -35,6 +35,15 @@
 #   define debugPrintf(ARGSINPAREN)
 #endif
 
+/* For historical reasons, round down expiration time on vxWorks and RTEMS.
+ * This allows timers which expire on every tick.
+ * Otherwise, the fastest timer period is 2x the tick interval.
+ * This results in timers expiring early.
+ */
+#if defined(vxWorks) || defined(__rtems__)
+#  define TIMER_QUANTUM_BIAS 1
+#endif
+
 template < class T > class epicsGuard;
 
 class timer : public epicsTimer, public tsDLNode < timer > {
--- a/modules/libcom/src/timer/timer.cpp
+++ b/modules/libcom/src/timer/timer.cpp
@@ -65,7 +65,11 @@ void timer::start ( epicsTimerNotify & notify, const epicsTime & expire )
 void timer::privateStart ( epicsTimerNotify & notify, const epicsTime & expire )
 {
     this->pNotify = & notify;
-    this->exp = expire - ( this->queue.notify.quantum () / 2.0 );
+    this->exp = expire
+#ifdef TIMER_QUANTUM_BIAS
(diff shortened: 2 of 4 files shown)


mdavidsaver:
If there are concerns about windows/osx, I could flip the logic and to only apply to Linux pending further testing.

anjohnson:
Core: Try changing the test case to see what the RTEMS 4.x tests do; disable there if they fail. Generally acceptable.

mdavidsaver:
The following change to epicsTimerTest passes except for `Ub-22 gcc + RT-5.1 pc686` (expected as it still has the `-quantum/2` bias).  I will try to formulate a lower bound which takes into account `epicsThreadSleepQuantuam()` on RTEMS/vxWorks.


```diff
--- a/modules/libcom/test/epicsTimerTest.cpp
+++ b/modules/libcom/test/epicsTimerTest.cpp
@@ -113,9 +113,8 @@ double delayVerify::checkError () const
     const double minError = testImpreciseTiming() ? 0.25 : 0.05;
     double measuredDelay = this->expireStamp - this->beginStamp;
     double measuredError = measuredDelay - this->expectedDelay;
-    double absoluteError = fabs(measuredError);
     double percentError = 100.0 * measuredError / this->expectedDelay;
-    testOk(absoluteError < minError,
+    testOk(measuredError>=0 && measuredError < minError,
            "Delay %.3f s, error = %+.6f ms (%+.3f %%)",
            this->expectedDelay, measuredError * 1000, percentError);
     return measuredError;
```

```
  epicsTimerTest.tap ........... 
  not ok  3 - Delay 1.000 s, error = -0.412573 ms (-0.041 %)
  not ok  4 - Delay 1.100 s, error = -0.779566 ms (-0.071 %)
...
```

mdavidsaver:
Updated.  I opted to move the `#if` condition into `timerPrivate.h`, and include it both in implementation and test code.

I will re-run the GHA jobs a few times to see if epicsTimerTest burps.

mdavidsaver:
I've re-run the GHA jobs 6 times.  No errors from `epicsTimerTest` (and only a couple overall).

modules/libcom/src/timer/timerPrivate.h
```diff
@@ -35,6 +35,15 @@
 #   define debugPrintf(ARGSINPAREN)
 #endif
 
+/* For historical reasons, round down expiration time on vxWorks and RTEMS.
+ * This allows timers which expire on every tick.
+ * Otherwise, the fastest timer period is 2x the tick interval.
+ * This results in timers expiring early.
```
anjohnson, line 41: It's unclear to me what "This" in line 41 is actually referring to; for the sanity of future maintainers it would be good to clarify that wording.
Also "For historical reasons" could become "On OSs with lo-res timers".

anjohnson:
Core group: Approved with Release Note entry.

mdavidsaver:
Merged as of 7ae3e4c2dfc258ea3a872013aad1dcfc3f6e1c35

PR #745, opened 2025-11-13
posix: correct race and mis-count in epicsThreadCreateOpt()

mdavidsaver:
Corrects two issues.

012139638d6a212ba347c11cbedaedaf8875ab6f added a dec-ref in the `status==EPERM` path, but I neglected to add a matching inc-ref after the second `pthreadInfo` alloc.  So if a joinable `pthread_create()` actually errors with `EPERM`, then no reference is left for the joiner.

Secondly, and far more subtle, is a data race if `pthread_create()` succeeds in starting a new thread, which ends and `free()`'s `pthreadInfo` before `pthread_create()` writes `pthreadInfo->tid`.

### DIFF
--- a/modules/libcom/src/osi/epicsThread.h
+++ b/modules/libcom/src/osi/epicsThread.h
@@ -179,6 +179,15 @@ typedef struct epicsThreadOpts {
  * \param parm Passed to thread main function.
  * \param opts Modifiers for the new thread, or NULL to use target specific defaults.
  * \return NULL on error
+ *
+ * The newly created thread will start running immediately, and may end immediately too!
+ * If joinable, the caller is responsible to arrange for a later call to epicsThreadMustJoin().
+ * The returned epicsThreadId remains valid until joined.
+ * For a non-joinable (detached) thread, the lifetime of the returned epicsThreadId
+ * ends when that thread returns, and so it must not be used unless the caller has
+ * external knowledge that the thread is still running.
+ *
+ * \since 7.0.2
  */
 LIBCOM_API epicsThreadId epicsThreadCreateOpt (
     const char * name,
@@ -197,7 +206,10 @@ LIBCOM_API epicsThreadId epicsStdCall epicsThreadMustCreate (
 
 /* This gets undefined in osdThread.h on VxWorks < 6.9 */
 #define EPICS_THREAD_CAN_JOIN
-/** Wait for a joinable thread to exit (return from its main function) */
+/** Wait for a joinable thread to exit (return from its main function).
+ *  Thread must have been created as joinable.
+ * \since 7.0.2 Test EPICS_THREAD_CAN_JOIN macro for BSP support
+ */
 LIBCOM_API void epicsThreadMustJoin(epicsThreadId id);
 /** Block the current thread until epicsThreadResume(). */
 LIBCOM_API void epicsStdCall epicsThreadSuspendSelf(void);
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -425,6 +425,10 @@ static void * start_routine(void *arg)
     int status;
     sigset_t blockAllSig;
 
+    // concurrently written from creator thread with same value
+    pthreadInfo->tid = pthread_self();
+    epicsAtomicWriteMemoryBarrier();
+
     sigfillset(&blockAllSig);
     pthread_sigmask(SIG_SETMASK,&blockAllSig,NULL);
     status = pthread_setspecific(getpthreadInfo,arg);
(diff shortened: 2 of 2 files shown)


mdavidsaver:
Due credit for my noticing these bugs goes to [https://github.com/pizlonator/fil-c](Fil-c).  cf. this [LWN](https://lwn.net/Articles/1042938/) article.

Running `epicsAtomicTest`, which starts a lot of short lived threads, triggers:

```
... hundreds of apparently successful pthread_create() calls
XXX pthread_create(0x7fbab4eba728, 0x7fbab4eba738, 0x7fbac80f5cb0, 0x7fbab4eba710)
filc safety error: cannot write pointer to free object.
    pointer: 0x7fbab4eba728,0x7fbab4eba710,0x7fbab4eba710,aux=0x7fbab618fd40,free
    expected 8 writable bytes with ptr aligned to 8 bytes.
semantic origin:
    src/thread/pthread_create.c:363:7: __pthread_create
check scheduled at:
    src/thread/pthread_create.c:363:7: __pthread_create
    ../osi/os/posix/osdThread.c:616:14: epicsThreadCreateOpt
    ../osi/epicsThread.cpp:44:12: epicsThreadCreate (inlined)
    ../osi/epicsThread.cpp:346:28: epicsThreadMustCreate
    ../epicsAtomicTest.cpp:193:9: testAddSub<int> (inlined)
    ../epicsAtomicTest.cpp:423:5: main
    src/env/__libc_start_main.c:79:7: __libc_start_main
    <runtime>: start_program
[15180] filc panic: thwarted a futile attempt to violate memory safety.
```

The "XXX" comes from an added print:

```c
index 8c3d8c8..34c2356 100644
--- a/modules/libcom/src/osi/os/posix/osdThread.c
+++ b/modules/libcom/src/osi/os/posix/osdThread.c
@@ -611,6 +611,8 @@ epicsThreadCreateOpt(const char * name,
         epicsAtomicIncrIntT(&pthreadInfo->refcnt);
     }
 
+fprintf(stderr, "XXX pthread_create(%p, %p, %p, %p)\n", &pthreadInfo->tid, &pthreadInfo->attr,
+        start_routine, pthreadInfo);
     status = pthread_create(&pthreadInfo->tid, &pthreadInfo->attr,
         start_routine, pthreadInfo);
     if (status==EPERM) {
```

So the access violation is of the first pointer (`0x7fbab4eba728`).  Which as I thought about it, could already be free'd by the short lived thread before `pthread_create()` has written its result.

mdavidsaver:
> if a joinable `pthread_create()` actually errors with `EPERM`

I think this is unlikely in practice since `find_pri_range()` has already run.

mdavidsaver:
@anjohnson @ralphlange @simon-ess @dirk-zimoch  During this morning's meeting I was distracted by TLS, ACF and other TLAs, so I forgot about this PR.  This is not the first time I have realized that epicsThreadOSD ref. counting was incorrect.  So I would like to request some special attention for inclusion in a December release.

anjohnson (approved):
The additional comments here make the ref-counting easier to follow, thanks for them.

The concurrent assignments to `pthreadInfo->tid` is the one part that feels "off" and might even be UB, but looking at the functions that use `id->tid` there are some that are legal for the new thread to call. The POSIX documentation doesn't require that `*thread` in `pthread_create()` be written before the thread starts running, so I think this is reasonable. Does valgrind have anything to say about it?

You are adding single-line `//` comments to a .c file which some older C compilers used to complain about, but since all the CI builds have succeeded we've apparently dropped all those targets from our officially supported list now.

mdavidsaver:
> The concurrent assignments to `pthreadInfo->tid` is the one part that feels "off" and might even be UB, but looking at the functions that use `id->tid` there are some that are legal for the new thread to call.

I agree that this seems sketchy.  What would be the alternatives?  Switch to an atomic set?

> You are adding single-line `//` comments to a .c file which some older C compilers used to complain about

I thought that we had explicitly decided to allow this when we dropped support for old MSVC?

anjohnson:
> What would be the alternatives? Switch to an atomic set?

That's why I asked whether valgrind complains about it; if not I say stick with what you've proposed until we see evidence that it's problematic. Just going with one atomic set might not work since `pthread_create()` isn't going to be using that to write through its `*thread` pointer.

> I thought that we had explicitly decided to allow this when we dropped support for old MSVC?

That is very possible/likely, it's just taking me a while to get used to _not_ pointing out some things that _used to_ be problematic. My editor highlighting patterns still show `//` comments in red when in a .c file.  🦕

mdavidsaver:
> That's why I asked whether valgrind complains about it

Somewhat surprisingly, [no](https://github.com/user-attachments/files/23664923/valgrind.txt).  I guess valgrind 3.24 is better about false positives.

> `pthread_create()` isn't going to be using that

This can be easily avoided by having `pthread_create()` assign to a local variable first.  The problem I find is that `pthread_t` is not one of the types which `epicsAtomic` knows about.

mdavidsaver:
Last thoughts for the evening.

> The problem I find is that `pthread_t` is not one of the types which `epicsAtomic` knows about.

Add `epicsAtomicWriteMemoryBarrier()` after both assignments?  Or move the assignments under `listLock`?

The former sees sufficient.  The later more strictly correct, but would likely need unnecessary locking around all reads or cause valgrind to complain.

mdavidsaver:
> Add `epicsAtomicWriteMemoryBarrier()` after both assignments?

I opted for this change to have something to discuss.

I also expanded the doc. comments for `epicsThreadCreateOpt()` with a discussion of returned `epicsThreadId` lifetime.

PR #746, opened 2025-11-17
fix parallel build link problem

dirk-zimoch:
When running parallel `make` with more than 2 jobs (I often use `make -j4`) I regularly see linker errors in the tests like this:
```
modules/database/test/ioc/db/O.linux-x86_64/libdbTestIoc.so: file not recognized: File truncated
```
Re-running `make`usually fixes the problem because meanwhile the shared test library contains data.
It is always a `TESTLIBRARY` that is incomplete.
This PR seems to fix that problem.

### DIFF
--- a/configure/RULES_BUILD
+++ b/configure/RULES_BUILD
@@ -197,7 +197,7 @@ $(sort $(DIRECTORY_TARGETS)):
 	$(MKDIR) $@
 
 # Install LIB_INSTALLS libraries before linking executables
-$(TESTPRODNAME) $(PRODNAME): | $(INSTALL_LIB_INSTALLS)
+$(TESTPRODNAME) $(PRODNAME): | $(INSTALL_LIB_INSTALLS) $(TESTSHRLIBNAME)
 
 # Install built libraries too, unless Makefile says to wait
 ifneq ($(DELAY_INSTALL_LIBS),YES)


dirk-zimoch:
I suggest we add test builds with more than 2 parallel jobs, if possible.

mdavidsaver:
I do not recall seeing `File truncated` before, although I try to avoid building executables and shared libraries in the same directory.  That said, adding an order-only prerequisite seems a reasonable change.

Can (and should) it go further?  Also `$(TESTLIBNAME)`?  Or beyond test libraries to `$(LIBNAME)` and `$(SHRLIBNAME)`?

mdavidsaver:
> I suggest we add test builds with more than 2 parallel jobs, if possible.

Semi-related.  It looks like [GHA](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) changed their build environment at some point.  As I read it, runners now get 4 cores.

https://github.com/epics-base/ci-scripts/issues/105

anjohnson:
I almost always do `make -sj` on hosts that have up to 27 CPUs, and I've never seen this issue (`make -j4` doesn't trigger it here either, on MacOS or RHEL-9). I would be interested to see your build output when it happens, can you trigger it just by cleaning and then building that database/test/ioc/db directory?

What is your `make --version` output?

dirk-zimoch:
```
$ make --version
GNU Make 4.2.1
Built for x86_64-redhat-linux-gnu
```

simon-ess:
I've also never seen this problem, and I regularly use `make -j 16` (builds fast!). I use variously make 3.81 (Mac OS) or 4.3 (ubuntu).

dirk-zimoch:
> I would be interested to see your build output when it happens, can you trigger it just by cleaning and then building that database/test/ioc/db directory?

Yes, it happens in that case, too. This is the output on my 4 core laptop, building on a local disk:
```
$ nproc
4
$ df .
Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/sda1      984304956 76844472 857433940   9% /home
$ make host -j4
[... compiles all sources without problem ...]
/bin/ar -rc    libdbTestIoc.a    arrRecord.o xRecord.o dbLinkdset.o xLink.o devx.o jlinkz.o 
/bin/ranlib libdbTestIoc.a
/bin/g++ -o libdbTestIoc.so  -shared -fPIC -Wl,-hlibdbTestIoc.so -L/home/zimoch/epics/epics-base-7.0/lib/linux-x86_64 -Wl,-rpath,/home/zimoch/epics/epics-base-7.0/lib/linux-x86_64 -Wl,-rpath-link,/home/zimoch/epics/epics-base-7.0/lib/linux-x86_64           -rdynamic -m64           arrRecord.o xRecord.o dbLinkdset.o xLink.o devx.o jlinkz.o    -ldbCore -lca -lCom   -lpthread   -lreadline -lm -lrt -ldl -lgcc
/bin/g++ -o dbScanTest  -L/home/zimoch/epics/epics-base-7.0/lib/linux-x86_64 -L/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64 -Wl,-rpath,/home/zimoch/epics/epics-base-7.0/lib/linux-x86_64 -Wl,-rpath,/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64 -Wl,-rpath-link,/home/zimoch/epics/epics-base-7.0/lib/linux-x86_64           -rdynamic -m64           dbScanTest.o dbTestIoc_registerRecordDeviceDriver.o    -ldbTestIoc -ldbCore -lca -lCom   
/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64/libdbTestIoc.so: file not recognized: File truncated
collect2: error: ld returned 1 exit status
make[1]: *** [../../../../../../configure/RULES_BUILD:249: dbScanTest] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64'
make: *** [../../../../../configure/RULES_ARCHS:58: install.linux-x86_64] Error 2
```
Running `make -j4` again completes the build.

Here is the output of `make clean; make host -js`:
```
/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64/libdbTestIoc.so: file not recognized: File truncated
collect2: error: ld returned 1 exit status
make[1]: *** [../../../../../../configure/RULES_BUILD:248: dbScanTest] Error 1
make[1]: *** Waiting for unfinished jobs....
/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64/libdbTestIoc.so: file not recognized: File truncated
collect2: error: ld returned 1 exit status
make[1]: *** [../../../../../../configure/RULES_BUILD:249: dbShutdownTest] Error 1
/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64/libdbTestIoc.so: file not recognized: File truncated
collect2: error: ld returned 1 exit status
make[1]: *** [../../../../../../configure/RULES_BUILD:249: testdbConvert] Error 1
/home/zimoch/epics/epics-base-7.0/modules/database/test/ioc/db/O.linux-x86_64/libdbTestIoc.so: file not recognized: File truncated
collect2: error: ld returned 1 exit status
make[1]: *** [../../../../../../configure/RULES_BUILD:249: callbackTest] Error 1
make: *** [../../../../../configure/RULES_ARCHS:58: install.linux-x86_64] Error 2
```

With this PR applied, `make clean; make host -js` runs silently and successfully.

dirk-zimoch:
I have cloned a fresh copy of the 7.0 branch without any of my changes to configure for supporting our cross architectures. And it builds. On my 4 core laptop as well as on a 16 core VM. On NFS.

Checking all differences in configure now...

dirk-zimoch:
I found that the origin of my build problem seems to be this line in my CONFIG_SITE.Common.linuxCommon:
```
# backward compatibility: Keep unversioned libs
# in order to prevent dynamic linking problems
# when minor changes increment *_MAINTENANCE_VERSION
SHRLIB_SUFFIX = $(SHRLIB_SUFFIX_BASE)
```
The same happens if I change `SHRLIB_SUFFIX` directly in CONFIG.Common.UnixCommon.

dirk-zimoch:
Whenever `SHRLIB_SUFFIX` is not the same as `SHRLIB_SUFFIX_BASE`, then everything works. If it is the same, I need the additional ordering prerequisite.
If you think `SHRLIB_SUFFIX = $(SHRLIB_SUFFIX_BASE)` is nothing we should support, I can just withdraw this PR and keep it locally in our PSI configuration.

anjohnson:
If you could replicate the failure with an unmodified BASE for an application that's building a shared library where `SHRLIB_VERSION` is empty then we would probably want to fix it. Given that the problem only seems to occur with your local patch though, I think that's where the solution belongs.

I suspect your local patch might be missing something to do with the `TARGET2_template` and/or `LIBRARY_template` definitions in `configure/RULES_TARGET` but I'm not sure what. When looking at the `$$(wildcard … *$(SHRLIB_SUFFIX_BASE)*, …)` there bear in mind that Linux puts the version number after the `.so` but Darwin puts it before the `.dylib`, hence the two `*` wildcards in the pattern, I think…

dirk-zimoch:
> an application that's building a shared library where `SHRLIB_VERSION` is empty

That would need to be an application that builds a test library. I know of no such application.

Anyway, closing now.

Issue #749, opened 2025-11-23
Incorrect file/line in generated lexer/parser


mdavidsaver:
Bundled `flex` and `antelope` produce C code files with incorrect/incomplete `#line` directives.  Not a functional problem, but this confuses debuggers, coverage analysis, and other tools which need to map program code back to source line.

- [ ] antelope emits `#line` mentioning `*.tab.c`, while our Make rules rename to `.c`.  https://github.com/epics-base/epics-base/pull/750
- [ ] ~flex seems to miss a trailing `#line *.c`, leaving `lcov` to complain about references to line numbers beyond the end of the `.l` input file.~  Not worth the effort

mdavidsaver:
> * flex seems to miss a trailing `#line *.c`

Actually, flex only emits `#line` mentioning the .y file.  Never the .c file.  It does not look like it tracks output line number at all.  Changing this issue seems too complex for the benefit.

PR #750, opened 2025-11-23
flex/antelope: fix #line directives

mdavidsaver:
Partially address #749.  Change antelope/flex output file suffix to remote `.tab`.


simon-ess:
What was the reasoning behind using the intermediate .tab.* files? This changes seems _technically_ fine, but I assume there was a reason we did it that way...

mdavidsaver:
> What was the reasoning behind using the intermediate .tab.* files?

The only reason I can think of would be a desire to avoid unnecessary changes to this bundled code, which would complicate accepting a new upstream version.  Of course, that ship sailed several decades ago.

PR #753, opened 2025-11-25
remove needless pointer casts

dirk-zimoch:
Unnecessary casts to `(void *)` and `(char *)` are dangerous because they undermine the type checker of the compiler and can thus hide severe programming errors. None found though, so this is more a "cosmetic" change.

No new warnings, even on Windows.

### DIFF
--- a/modules/libcom/src/flex/misc.c
+++ b/modules/libcom/src/flex/misc.c
@@ -76,7 +76,7 @@ void *allocate_array(int size, int element_size)
     if ( element_size * size <= 0 )
         flexfatal( "request for < 1 byte in allocate_array()" );
 
-    mem = malloc( element_size * size );
+    mem = malloc( (size_t) element_size * size );
 
     if ( mem == NULL )
         flexfatal( "memory allocation failed in allocate_array()" );
@@ -705,7 +705,7 @@ void *reallocate_array(void *array, int size, int element_size)
         flexfatal( "attempt to increase array size by less than 1 byte" );
 
     new_array =
-        realloc( array, size * element_size );
+        realloc( array, (size_t) size * element_size );
 
     if ( new_array == NULL )
         flexfatal( "attempt to increase array size failed" );
--- a/modules/libcom/src/flex/misc.c
+++ b/modules/libcom/src/flex/misc.c
@@ -76,7 +76,7 @@ void *allocate_array(int size, int element_size)
     if ( element_size * size <= 0 )
         flexfatal( "request for < 1 byte in allocate_array()" );
 
-    mem = (void *) malloc( (unsigned) (element_size * size) );
+    mem = malloc( element_size * size );
 
     if ( mem == NULL )
         flexfatal( "memory allocation failed in allocate_array()" );
@@ -705,7 +705,7 @@ void *reallocate_array(void *array, int size, int element_size)
         flexfatal( "attempt to increase array size by less than 1 byte" );
 
     new_array =
-        (void *) realloc( (char *)array, (unsigned) (size * element_size ));
+        realloc( array, size * element_size );
 
     if ( new_array == NULL )
         flexfatal( "attempt to increase array size failed" );
--- a/modules/libcom/src/macLib/macUtil.c
+++ b/modules/libcom/src/macLib/macUtil.c
@@ -226,7 +226,7 @@ epicsStdCall macParseDefns(
     /* free workspace */
     free( ( void * ) ptr );
     free( ( void * ) end );
-    free( ( char * ) del );
+    free( del );
 
     /* debug output */
     if ( handle != NULL && handle->debug & 1 )
@@ -240,7 +240,7 @@ error:
     errlogPrintf( "macParseDefns: failed to allocate memory\n" );
     if ( ptr != NULL ) free( ( void * ) ptr );
     if ( end != NULL ) free( ( void * ) end );
-    if ( del != NULL ) free( ( char * ) del );
+    if ( del != NULL ) free( del );
     *pairs = NULL;
     return -1;
 }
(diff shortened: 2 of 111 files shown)


modules/libcom/src/macLib/macUtil.c
```diff
@@ -226,7 +226,7 @@ epicsStdCall macParseDefns(
     /* free workspace */
     free( ( void * ) ptr );
     free( ( void * ) end );
-    free( ( char * ) del );
+    free( del );
```
anjohnson: Why fix `del` but not `ptr` and `end`? Similar changes below.

modules/libcom/src/flex/misc.c
```diff
@@ -76,7 +76,7 @@ void *allocate_array(int size, int element_size)
     if ( element_size * size <= 0 )
         flexfatal( "request for < 1 byte in allocate_array()" );
```
anjohnson, line 78: Given that CodeQL is complaining about the new line 79 below, the above comment about overflow is no longer appropriate, maybe you could fix that too?

modules/libcom/src/flex/misc.c
```diff
@@ -705,7 +705,7 @@ void *reallocate_array(void *array, int size, int element_size)
         flexfatal( "attempt to increase array size by less than 1 byte" );
```
anjohnson, line 705: Similar point as above. The "fatal" wording here is wrong too, it's an attempt to make the array size be less than 1 byte, not increase it.

anjohnson (changes requested):
I approve this PR in principle, even though it touches lots of code; some suggestions here.

modules/libcom/src/macLib/macUtil.c
dirk-zimoch: Because those cast away `const`.

modules/libcom/src/flex/misc.c
dirk-zimoch: The cast to `(unsigned)` was useless anyway as it was outside the multiplication. A cast to `(size_t)` of at least one of the two values before the multiplication would be the right thing to do. I can do that, even though it seems slightly beyond the scope of this PR. But I don‘t want to trigger new warnings either and a code line is owned by whoever touched it last 😉

modules/libcom/src/flex/misc.c
dirk-zimoch: Who uses `int` for object sizes anyway? The whole flex thing could need an overhaul.

modules/libcom/src/flex/misc.c
dirk-zimoch: The comment worries about 16 bit implementations. I think we can safely assume nobody runs EPICS 7 on 16 bit Windows 3.11. Linux always required at least 32 bit architectures, I think. And nobody develops (i.e. runs flex) on a tiny embedded system. Maybe we can safely remove the whole check.

anjohnson: The check is still comparing against zero though, which is protecting against some bugs.

modules/libcom/src/flex/misc.c
anjohnson: It's ancient code, `int` was just a machine word that you used by default as long as you weren't expecting any values too big to fit in one.

mdavidsaver (approved):
The void...  it does not end.

No objection in principle.  Although, I am inclined to defer this PR until just after 7.0.10.  It should be a no-op.  No change to build output (functional or documentation).  However, can we prove this?

dirk-zimoch:
> No change to build output

Andrew requested some changes that would change the build output.

modules/libcom/src/macLib/macUtil.c
dirk-zimoch: Some const-away-casts I found and kept have a comment. `/* cast away const */` Should I add such a comment? Some remaining casts also cast away `volatile`.

modules/libcom/src/flex/misc.c
dirk-zimoch: Changing that code/wording here would cause the output to change. Michael seems concerned about such functional changes. Maybe defer that to a later PR.

dirk-zimoch:
> However, can we prove this?

I have run `objdump -d` on all binaries in bin/ and lib/ with and without this PR and compared the results with `cmp`. No differences.

dirk-zimoch:
CodeQL now complains about lines and files I didn't even touch. I'm going to ignore that.

mdavidsaver:
> CodeQL now complains about lines and files I didn't even touch. I'm going to ignore that.

Agreed.  There may be one real issue hiding in there (#758), but it has nothing to do with the exclusively whitespace change of this PR.

dirk-zimoch:
I wanted to keep any functional change out of this PR. I have checked that the generated assembler code is unmodified and I would prefer to keep it like that. Functional changes should go into a separate PR.

Maybe we should run CodeQL on the whole projects and not only on lines someone touched for totally different reasons. But I first would like to complete my "housekeeping" and remove all warnings.

anjohnson:
No objections from me to separating out the changes.

There is now a (probably trivial) conflict in this PR in callback.h which will need resolving before it can be merged.

Should this PR be added to the "Post 7.0.10" Milestone?

dirk-zimoch:
Rebased and trivial conflict resolved.

> Should this PR be added to the "Post 7.0.10" Milestone?

Touching many files makes this PR very prone to merge conflicts. Thus, I would like to have it merged rather sooner than later, either before 7.0.10 or immediately after.

mdavidsaver:
Since 7.0.10 has been tagged, and I think that @dirk-zimoch has address previous comments, I think this PR is ready to merge.

PR #755, opened 2025-12-02
Doc epicsSnprintf()

mdavidsaver:
Clearly document that `epicsSnprintf()` will output at least a nil on success.  

I  see some ambiguities, and inconsistencies.  `size==0` not handled on WIN32.  No specification behavior on error for Linux and vxWorks.  Not surprising as many format string / vararg issues are UB.  I'm not going to try to fix these before 7.0.10, only to document the least common denominator.

I have left the long description of OS specific return values, although I wonder if it is still accurate.

### DIFF
--- a/modules/libcom/src/osi/epicsStdio.h
+++ b/modules/libcom/src/osi/epicsStdio.h
@@ -102,8 +102,16 @@ extern "C" {
 #endif
 
 /**
- * \brief epicsSnprintf() is meant to have the same semantics as the C99
- * function snprintf()
+ * \brief Wrapper around OS snprintf()
+ *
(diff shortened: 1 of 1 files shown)


mdavidsaver:
A note for the future use of [`snprintf`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170) on WIN32.

> Beginning with the UCRT in Visual Studio 2015 and Windows 10, snprintf is no longer identical to _snprintf. The snprintf behavior is now C99 standard conformant.  ...

Issue #761, opened 2025-12-10
Appveyor: Can't Proceed in CAC-event caused by dbca.c changes — BLOCKER for 7.0.10?


anjohnson:
I looked at some of the recent Windows Static build failures on Appveyor, these aren't the 1hr timeouts, but and ones I looked at all seemed to be `cantProceed()` suspensions of a CAC-event thread in dbca.c, from Windows static build test runs (both 32 and 64-bit). I would link to them, but restarting the job destroys the log so those links aren't really useful. Stack-traces from the log are below.

These all come through testdbCaWaitForEventCB() which was added in #210 commit [e9e576f4bb](https://github.com/epics-base/epics-base/pull/210/changes/e9e576f4bb66b1e57ec0327789e5650187fb3005) although the function was named testdbCaWaitForUpdateCountCB() back then.

I’m concerned that we might need to fix this before 7.0.10.

```
make[5]: Entering directory 'C:/projects/epics-base/modules/database/test/ioc/db/O.win32-x86-static'
perl -CSD dbCaLinkTest.t -tap > dbCaLinkTest.tap
epicsEventMustTriggerCRITICAL ERROR Thread CAC-event (009311E8) can't proceed, suspending.
Dumping a stack trace of thread 'CAC-event':
[ 0021C695]: C:\projects\epics-base\modules\libcom\src\osi\os\WIN32\osdBackTrace.cpp[line 23](epicsBackTrace+0x25)
[ 00210420]: C:\projects\epics-base\modules\libcom\src\osi\epicsStackTrace.c[line 80](epicsStackTrace+0x70)
[ 001E722C]: C:\projects\epics-base\modules\libcom\src\misc\cantProceed.c[line 67](cantProceed+0x4c)
[ 001E769A]: C:\projects\epics-base\modules\libcom\src\osi\epicsEvent.cpp[line 122](epicsEventMustTrigger+0x2a)u
[ 001510A8]: C:\projects\epics-base\modules\database\src\ioc\db\dbCa.c[line 213](testdbCaWaitForEventCB+0x58)
[ 0015291A]: C:\projects\epics-base\modules\database\src\ioc\db\dbCa.c[line 966](eventCallback+0x35a)
[ 001BC74F]: C:\projects\epics-base\modules\ca\src\client\oldSubscription.cpp[line 67](oldSubscription::current+0xaf)
[...]

Issue #445, opened 2023-11-26
regressTest crash


mdavidsaver:
I suspect this is due to a bug in `testdbCaWaitForUpdateCount()`.  The setup looks like a race between `epicsEventMustTrigger()` and `epicsEventDestroy()`, however both calls are guarded by `caLink::lock` so I don't think the situation is so straightforward.

```
...
  Thread 7 (Thread 0x7fe3772eb1c0 (LWP 16840)):
  #0  0x00007fe377388a24 in __pthread_mutex_lock_full () from /lib/x86_64-linux-gnu/libpthread.so.0
  #1  0x00007fe37793664c in mutexLock (id=0x21421a0) at ../osi/os/posix/osdMutex.c:116
  #2  epicsMutexOsdLock (pmutex=0x21421a0) at ../osi/os/posix/osdMutex.c:163
  #3  0x00007fe377a1d1d8 in doLocked (plink=0x21e21a0, rtn=0x7fe377a98920 <readLocked>, priv=0x0) at ../db/dbCa.c:781
  #4  0x00007fe377a988a0 in read_stringin (prec=0x21e1ea0) at ../dev/devSiSoft.c:66
  #5  0x00007fe377a93dea in readValue (prec=0x21e1ea0) at ../rec/stringinRecord.c:202
  #6  process (pcommon=0x21e1ea0) at ../rec/stringinRecord.c:136
  #7  0x00007fe3779f7812 in dbProcess (precord=precord@entry=0x21e1ea0) at ../db/dbAccess.c:610
  #8  0x00007fe377a0b132 in processTarget (psrc=0x21e1a60, pdst=0x21e1ea0) at ../db/dbDbLink.c:498
  #9  0x00007fe377a18c90 in recGblFwdLink (precord=precord@entry=0x21e1a60) at ../db/recGbl.c:286
  #10 0x00007fe377a88e7d in process (pcommon=0x21e1a60) at ../rec/longinRecord.c:155
  #11 0x00007fe3779f7812 in dbProcess (precord=precord@entry=0x21e1a60) at ../db/dbAccess.c:610
  #12 0x00007fe377a0b132 in processTarget (psrc=0x21e1610, pdst=0x21e1a60) at ../db/dbDbLink.c:498
  #13 0x00007fe377a18c90 in recGblFwdLink (precord=precord@entry=0x21e1610) at ../db/recGbl.c:286
[...]

PR #762, opened 2025-12-10
fix testdbCaWaitForEvent() again...

mdavidsaver:
Sync with CA context thread(s) as well as dbCa worker.

Possible fix for https://github.com/epics-base/epics-base/issues/445 and https://github.com/epics-base/epics-base/issues/761

### DIFF
--- a/modules/ca/src/client/access.cpp
+++ b/modules/ca/src/client/access.cpp
@@ -717,6 +717,18 @@ int epicsStdCall ca_context_status ( ca_client_context * pcac, unsigned level )
     return ECA_NORMAL;
 }
 
+extern "C"
+LIBCA_API
+void dbCaSyncLocal(void);
+
+void dbCaSyncLocal(void)
+{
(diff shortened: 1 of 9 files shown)


mdavidsaver:
@anjohnson So far the only appveyor failure is one 1hr timeout.  4 jobs remaining.

anjohnson (commented):
These results are very promising, I'd like to see runs of these builds a few more times on Appveyor to get more evidence that this fixes that problem. Have we seen any similar failures on GHA? I don't remember any. If we don't get any other `cantProceed()` failures with this change I would definitely want to merge it.

I've taken a look at the 1 hr failure logs to try and work out what's hanging there, but it's tricky because the build order changes so much. Interestingly the build that hung was running on a somewhat faster CPU than the re-run that passed: It built libcom in 3'46", the slower one took 4'23". 🐰🐢

Thanks!

mdavidsaver:
With 7.0.10 out the door, I am comfortable merging this attempted fix.

PR #763, opened 2025-12-15
rtems getNtpTime once on rtems_init  udp socket funtions with timeout

hjunkes:
Currently, posix/rtems_init.c is called when the IOC starts under RTEMS 6/7. 
If there is no RTC on the board, an attempt is made to read the current time from 
a time server (from the DHCP request or hard coded, or NVRAM) 
before the NTP synchronisation mechanism 
([Chris Johns](https://github.com/kiwichris/epics-base.git): rtems-ntpd-7_0-with-5-and-6-v2 feature branch) 
gets started/activated.
A timeout of 2 seconds has now been implemented. Unfortunately, this was not previously provided for, and the call waited for the TCP timeout.

### DIFF
--- a/modules/libcom/RTEMS/epicsNtp.c
+++ b/modules/libcom/RTEMS/epicsNtp.c
@@ -2,6 +2,8 @@
 * (C) 2014 David Lettier.
 * http://www.lettier.com/
 * SPDX-License-Identifier: BSD-3-Clause
+* 
+* use of UDP funktions and implement timeout, HPJ 15.12.2025
 \*************************************************************************/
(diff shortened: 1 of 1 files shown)


kiwichris:
I think this approach should be depreciated and we move to using NTP update to get an initial time. This means the NTP time server (optionally) makes a call to NTP update (or something) to set an initial time as part of its initialization.

The code should be stable with and without this call being made and I have not seen a reason this call is needed?

mdavidsaver (approved):
Looks ok to me.

PR #764, opened 2025-12-15
Ensure that the necessary /etc/dhcpcd.conf is written.

hjunkes:
add some useful dhcp options to conf file.

### DIFF
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -760,6 +760,11 @@ dhcpcd_hook_handler(rtems_dhcpcd_hook *hook, char *const *env)
         DEFVAR("new_tftp_server_name", bootp_server_name_init),
         DEFVAR("new_bootfile_name", bootp_boot_file_name_init),
         DEFVAR("new_rtems_cmdline", bootp_cmdline_init),
+  /* Our new dhcp-server does not react correctly to client requests
+   and option 129 (rtems_cmd_line) is not transmitted. So as a
+   workaround I now use option 77 (user_class) to transmit the cmd string
+   */
+        DEFVAR("new_user_class", bootp_cmdline_init),
 #undef DEFVAR
         {NULL}
     };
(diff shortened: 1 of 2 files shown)


modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -760,6 +760,11 @@ dhcpcd_hook_handler(rtems_dhcpcd_hook *hook, char *const *env)
         DEFVAR("new_tftp_server_name", bootp_server_name_init),
         DEFVAR("new_bootfile_name", bootp_boot_file_name_init),
         DEFVAR("new_rtems_cmdline", bootp_cmdline_init),
+  /* Our new dhcp-server does not react correctly to client requests
+   and option 129 (rtems_cmd_line) is not transmitted. So as a
+   workaround I now use option 77 (user_class) to transmit the cmd string
```
mdavidsaver, line 765: imo. some concrete reason is needed before changing this long standing convention.  eg. does 129 now collide with some other common, but also non-standard option?

hjunkes:
Closing this. The change was right, but the reason given in the description
was not, and the real explanation belongs with the new init framework in #905.

@mdavidsaver was right to ask for something concrete before changing a long
standing convention. Here it is.

rtems-libbsd bundles dhcpcd **6.2.1**, and in that version an option defined
in `/etc/dhcpcd.conf` can never reach the parameter request list:

* `read_config()` parses dhcpcd's *embedded* definitions first and moves them
  into the global `dhcp_opts[]`, then parses `/etc/dhcpcd.conf` into a
  separate per-interface `ifo->dhcp_override[]`.
* `option <name>` resolves through `set_option_space()` → `make_option_mask()`
  (`dhcp-common.c`), which searches `dhcp_opts[]` only. `option rtems_cmdline`
  therefore fails with "unknown option", and a numeric `option 129` has
  nothing to match either.
* `make_message()` in `dhcp.c` then builds option 55 by walking `dhcp_opts[]`,
  so 129 is unreachable even if the mask bit could somehow be set.

Decoding is different: `dhcp_env()` *does* consult `ifo->dhcp_override[]`.
That is why option 129 has always seemed to work — it does, but only when the
server volunteers it unsolicited. Our DHCP server was configured to return
everything it has, which is exactly what masked this. Against a server that
honours the client's request list, option 129 never arrives.

This is not a case of an outdated libbsd. On every rtems-libbsd branch —
4.11, 5, 5-freebsd-12, 6-freebsd-12, 6-freebsd-13, 6-freebsd-14,
**7-freebsd-14**, main, freebsd-current — `dhcpcd/defs.h` reads
`VERSION "6.2.1"`. The FreeBSD side is current; the bundled dhcpcd has simply
never been updated.

Option 77 `user_class` works because it *is* in dhcpcd's built-in table and
can therefore be requested. So the switch in this PR was the right workaround
for the wrong stated reason.

Two further defects in the `dhcpcd.conf` this PR writes, for the record:
the option names were hyphenated (`ntp-servers`, `tftp-server-name`,
`bootfile-name`) where dhcpcd matches with `strcmp()` against underscored
names, so none of them was ever actually requested; and `clientid EPICS boot`
makes every board send the same option 61, so a server cannot tell two boards
apart.

Verified on a BeagleBone Black with the server's "return all defined options"
switched off: option 129 is absent, the startup script comes from option 77,
NFS mounts, NTP is set and the IOC reaches
`iocRun: All initialization complete`.

Continuing in #905, where the boot-configuration convention for the new init
framework is being decided.

Issue #765, opened 2025-12-26
Make pattern rule for dbd causes confusion


mdavidsaver:
The rules:

https://github.com/epics-base/epics-base/blob/12c56ffc954e4cde72c43174ebd11ccd0d523c1e/configure/RULES.Db#L264-L267

introduce confusion, and perhaps actual problems.  The `$($*_DBD)` always expands to an empty string since automatic variables can not appear in a prerequisite list.  (see https://github.com/epics-base/pvxs/pull/145#issuecomment-3690583947 and 4640f0a8ae1780163797c9e918df21cde3b00975)

With an empty prerequisite list, there pattern rules will be considered to build __all__ DBD files.  Even those with other rules, pattern or explicit.

```make
%.dbd$(DEP): $($*_DBD)
$(COMMON_DIR)/%.dbd: $($*_DBD)
```

Having a unrestricted pattern rule with an empty prerequisite list seems to me an invitation for confusion.  Is there some way author intent could be more clearly captured?  eg. how to express the distinction between a .dbd file to be generated via `dbdExpand`, and one generated through some other means.

mdavidsaver:
Also, having rules using shell redirection `> $@` is problematic as failure will leave an empty or incomplete output file with a current timestamp.  The preferred scheme would be `> $@.tmp` followed by a copy/move.

```make
 %.dbd$(DEP):
 	$(DBEXPAND) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $($*_DBD) > $@.tmp
 	echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.tmp
 	$(MV) $@.tmp $@
```

PR #766, opened 2025-12-27
configure: ensure output file only updated on success

mdavidsaver:
Ensure that Make rule failure does not erroneously update the timestamp on `$@`.  (see https://github.com/epics-base/epics-base/issues/765#issuecomment-3693491270)

eg. After running `false > $@`, the output file modification time will be updated.  So despite failure, a repeated "make" will consider the target up to date.

Handle this by accumulating output to `$@.tmp`, then move to `$@` as a final step.

Most cases are straightforward substitutions local to one recipe.  A few others are command snippets defined in variables.  All but one can be handled as a [multi-line variable](https://www.gnu.org/software/make/manual/make.html#Multi_002dLine).

This last one is only relevant on solaris, where `HDEPENDS_COMPFLAGS` contains `> $@`.

https://github.com/epics-base/epics-base/blob/12c56ffc954e4cde72c43174ebd11ccd0d523c1e/configure/os/CONFIG.solarisCommon.solarisCommon#L58-L59

On all other targets `HDEPENDS_COMPFLAGS` is either no used, or takes the GCC form.

https://github.com/epics-base/epics-base/blob/12c56ffc954e4cde72c43174ebd11ccd0d523c1e/configure/CONFIG.gnuCommon#L69-L70

Switching to a multi-line variable is not compatible with the usage:

https://github.com/epics-base/epics-base/blob/12c56ffc954e4cde72c43174ebd11ccd0d523c1e/configure/RULES_BUILD#L239-L241

I am not inclined to spend time on a solaris specific enhancement.

### DIFF
--- a/configure/RULES.Db
+++ b/configure/RULES.Db
@@ -124,7 +124,10 @@ ACF_CPPFLAGS = $(ACF_CPPFLAGS_$(GNU))
 ACF_INCLUDES = -I. $(TARGET_INCLUDES) $(USR_INCLUDES)\
                 $(SRC_INCLUDES) -I$(INSTALL_DB)
 ACFDEPENDS_CMD = $(MKMF) -m $@ $(ACF_INCLUDES) $(COMMONDEP_TARGET) $<
-ACF_CMD = $(CPP) $(ACF_CPPFLAGS) $(ACF_INCLUDES) $< > $@
+define ACF_CMD
+$(CPP) $(ACF_CPPFLAGS) $(ACF_INCLUDES) $< > $@.stdout
+$(MV) $@.stdout $@
+endef
 
 #---------------------------------------------------------------
 # dependencies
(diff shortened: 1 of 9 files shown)


configure/RULES.Db
```diff
@@ -219,68 +222,82 @@ realclean: clean
 
 %Record.h$(DEP): $(COMMON_DIR)/%Record.dbd $(DBDTORECTYPEH_dep)
 	@$(RM) $@
```
anjohnson, line 224: This and several following rules explicitly delete `$@` before generating the new file. I don’t understand the `$(MV)` is necessary or useful for those rules.

mdavidsaver:
@anjohnson Consider the following `x.make`.

```make
all: x.d

x.d:
        rm -f $@
        false > $@
        echo "other" >> $@
```

Run it once:

```
$ make -f x.make 
rm -f x.d
false > x.d
make: *** [x.make:6: x.d] Error 1
```

And again:

```
$ make -f x.make 
make: Nothing to be done for 'all'.
```

So despite the earlier failure, this repeat has **erroneously succeeded**.

configure/RULES.Db
mdavidsaver: `make` will not execute the recipe if `$@` exists with a new enough timestamp.  And the following `... > $@` will either create, or open and truncate. So `> $@` is equivalent to `fopen($@, "w")`, and `>> $@` to `fopen($@, "a")`.  So I do not think an initial `$(RM) $@` ever had any useful effect with `SHELL = sh`.  I think the same is true with `SHELL=cmd` as well, but I am less certain.

mdavidsaver:
... contrast the previous with:

```make
all: y.d

y.d:
        false > $@.tmp
        echo "other" >> $@.tmp
        mv $@.tmp $@
```

Now the first iteration fails:

```
$ make -f y.make 
false > y.d.tmp
make: *** [y.make:5: y.d] Error 1
```

and so does the repeat

```
$ make -f y.make 
false > y.d.tmp
make: *** [y.make:5: y.d] Error 1
```

anjohnson:
Rename `$@.tmp` to `$@.stdout` to clarify the purpose of the rename, then I'm good with it.

PR #767, opened 2025-12-28
Fix generation of RELEASE_NOTES.md at release

anjohnson:
My previous rules didn't create the RELEASE_NOTES.md file for the release-tagged commit, so the RTD build for 7.0.10 didn't publish this file. This PR fixes the rules, and also tweaks a Markdown header level for the pvaClient entry in that release note.

How can we get RTD to publish this as the documentation for 7.0.10? In my fork which this PR is from I used a branch named 7.0.10 which I started from the R7.0.10 tagged commit. The RTD output can be seen [here](https://anjohnson-epics-base.readthedocs.io/stable/) and contains what I was expecting, but RTD named it "Stable", not 7.0.10.

RTD experts @minijackson and @ttkorhonen please provide any suggestions/instructions that you think might work.

### DIFF
--- a/documentation/Makefile
+++ b/documentation/Makefile
@@ -21,107 +21,131 @@ RTD_SRC = $(COMMON_DIR)/rtd-src
 
 DOCS += README.md
 DOCS += ca-cli.md
+DOCS += RELEASE_NOTES.md
 DOCS += ReleaseChecklist.md
 
 OLD_NOTES = $(wildcard ../RELEASE-*.md)
 DOCS += $(OLD_NOTES:../%=%)
 
-ifneq ($(EPICS_DEV_SNAPSHOT),)
-  # Between releases the release target is disabled
-  DOCS += RELEASE_NOTES.md
-  REASON = EPICS_DEV_SNAPSHOT not empty
+NOTES_FILE = RELEASE-$(EPICS_SHORT_VERSION).md
+# NOTES_PATH works in both source and build directories
+ifdef T_A
+  NOTES_PATH = ../$(NOTES_FILE)
 else
-  # Not a snapshot, user may be creating a new release
-  ifdef T_A
-    DOC = ../
-  endif
-  # Could we be in the middle of anything?
-  NOTES_FILE = RELEASE-$(EPICS_SHORT_VERSION).md
-  NOTES_PATH = $(DOC)$(NOTES_FILE)
-  NOTES_GIT := $(firstword $(shell git status --porcelain $(NOTES_PATH)))
-  ifeq ($(NOTES_GIT),A)
+  NOTES_PATH = $(NOTES_FILE)
+endif
+
+ifneq ($(EPICS_DEV_SNAPSHOT),)
+  # Between releases, our special release targets are disabled
+  REASON = EPICS_DEV_SNAPSHOT not empty
+  NOTES_PATH =
+else
+  # Not a snapshot. User could be creating a new release,
+  # or this is a build of an already-tagged release.
+
+  # Are we in the process of releasing?
+  NOTES_STAT := $(firstword $(shell git status --porcelain $(NOTES_PATH)))
+  ifeq ($(NOTES_STAT),A)
     # New file was added to the Git index; allow unrelease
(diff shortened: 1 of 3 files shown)


ralphlange:
Did I mention this in another ticket? (Only have my phone this week.)

For versioned documentation on RTD, the versions (choices of the drop-down menu) are configured _manually_ on the configuration page of the project in the RTD web interface. To add a version, press the "+ New Version" button, existing versions can be edited or deleted from that page.

anjohnson:
> Did I mention this in another ticket?

Thanks, I vaguely remember something like that, but I wasn't focussed on this problem at the time and didn't recognize it as a something I needed to remember.

So I experimented in my area, adding a new RTD version for my branch 7.0.10 which is now [here](https://anjohnson-epics-base.readthedocs.io/7.0.10/) and I think that looks right. Should I push this branch to the epics-base project to publish these notes properly there (as well as merging the PR)? We need the branch since other commits have been merged onto the 7.0 branch since the R7.0.10 tag. We can lock that branch afterwards to prevent subsequent modifications.

minijackson:
> For versioned documentation on RTD, the versions (choices of the drop-down menu) are configured manually on the configuration page of the project in the RTD web interface. To add a version, press the "+ New Version" button, existing versions can be edited or deleted from that page.

You can actually add those versions automatically, using automation rules. I actually researched this a bit for Phoebus (see issue https://github.com/ControlSystemStudio/phoebus/issues/3556).

This can be done by going into the RtD project settings -> Automation Rules -> Add Rule, then make sure the regex matches the epics-base version scheme, that it's run on tags (not branches), and that the Action is "Activate version".

I think it doesn't retroactively add previous tags, only new ones.

anjohnson:
Core Group: Looks good, comment from documenters would be welcome before this is implemented.

anjohnson:
After committing Ralph's parallel build fix I pushed my 7.0.10 branch to this project, and created a new RTD Version for it using b7.0.10 as the slug, which got built [here](https://docs.epics-controls.org/projects/base/en/b7.0.10/). Hiding the old 7.0.10 version doesn't seem to have removed it from the "versions" window though, I'm hoping it just takes a while for some old cached data to time out.

anjohnson:
The old cache has expired now and I'm seeing the b7.0.10 link as expected. I will regenerate and sign the tar file next week, and merge this PR.

ralphlange:
> I will regenerate and sign the tar file next week, and merge this PR.

Please don't forget to also upload the tar to epics-controls.org - thank you!!

anjohnson:
Should I tag this, and if so how? I don't want to move the R7.0.10 tag, but I think a release on GitHub may require a tag. Is this R7.0.10.1 now?

ralphlange:
I think it would be appropriate to tag this as R7.0.10.1

I'm not sure if it would be appropriate to silently update the 7.0.10 tar - or do the regular thing and push another tar with the new name. (The Principle of Least Surprise would suggest the latter.)

mdavidsaver:
> I think it would be appropriate to tag this as R7.0.10.1

I thought we were done with 4 part version numbers?

ralphlange:
Ah, were we?
We had 7.0.4.1, 7.0.6.1 and 7.0.8.1 in the 7 series. 7.0.10.1 would be very regular.
My assumption was always: 7 - never changes - plus a more or less semantic three number scheme.

Well, if not: How would you handle this situation, then? 7.0.11 without any code changes?

anjohnson:
New tarfile created, signed and uploaded to GitHub (draft release page), epics-controls and APS websites. I will publish the doc's, merge this PR and announce the release soon.

anjohnson:
I'm not going to recreate the tarfile again, but I did just modify the `Makefile` and `make-notes.pl` in `documentation` to restore the inclusion of the older release entries into the generated `RELEASE_NOTES.md` file. I have unlocked the 7.0.10 branch and will push this change to it to regenerate the `b7.0.10` docs. I've also reconstructed the `10.php` release webpage on the APS site to point to the files on rtd, and when the update is done I will link to the release everywhere, send out an announcement email, and merge this PR.

PR #768, opened 2026-01-04
dbCa: iocInit wait for local CA links to connect

mdavidsaver:
A second attempt at #713.  This time with a proper test case.  Also, the notion of "unconnected" is expanded to include subsequent GET or MONITOR operation setup.  Meaning that any local `CP` link will have its value cache populated by the initial update.

During iocInit(), wait for local CA links to connect.  With this change, database authors can be certain that any local CA link is connected prior to `initHookAfterIocRunning`
and `field(PINI, RUNNING)`.

The process adds a global counter `initOutstanding` counting the number of unconnected local CA links.  Locality is checked in `dbInitLink()`, and the local counter is incremented if necessary.

A new private function `dbCaAddLinkCallbackOpt()` is added, which adds a `flags` argument which is propagated through to `caLink::flags` (guarded by `caLink:lock`).  Currently the only flag is `DBCA_CALLBACK_INIT_WAIT`, which is set when `initOutstanding` needs to be decremented on connect.

A new dbCa action `CA_INIT_READY` is added.  The process of synchronization is that the callback on local dbEvent worker does not directly decrement `initOutstanding`, but instead wakes up the dbCa worker to do so.  This ensures completion of initial processing of records with local `CP` links.

### DIFF
--- a/documentation/new-notes/PR-768.md
+++ b/documentation/new-notes/PR-768.md
@@ -1,0 +1,8 @@
+### iocInit wait for local CA links to connect
+
+PR [768](https://github.com/epics-base/epics-base/pull/768)
+
+During iocInit(), wait for local CA links to connect.
+With this change, database authors can be certain that
+any local CA link is connected prior to `initHookAfterIocRunning`
+and `field(PINI, RUNNING)`.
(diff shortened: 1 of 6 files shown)


anjohnson:
Core group: Fix PR # in RELEASE entry, then approved.

PR #772, opened 2026-01-08
fix dir dependency

dirk-zimoch:
`pvData` depends on `database` because the tests include `iocInit.h` via `epicsUnitTest.h`.

### DIFF
--- a/modules/Makefile
+++ b/modules/Makefile
@@ -17,7 +17,7 @@ database_DEPEND_DIRS = ca
 
 # Submodules for bundle build
 SUBMODULES += pvData
-pvData_DEPEND_DIRS = libcom
+pvData_DEPEND_DIRS = libcom database
 
 SUBMODULES += pvAccess
 pvAccess_DEPEND_DIRS = pvData database


anjohnson:
I don't think the code actually supports this as being an issue. There _is_ a line `#include "iocInit.h"` in the epicsUnitTest.h file, but it's inside a comment block, part of an example showing how to use the unit test API for tests that need to run an IOC.

```
 * iocRun(). The suggested call sequence in a test program that needs to run the
 * IOC without Channel Access is:
\code
#include "iocInit.h"

MAIN(iocTest)
{
```

mdavidsaver:
> There _is_ a line `#include "iocInit.h"` in the epicsUnitTest.h

imo.  The content of this sub-section in `epicsUnitTest.h` should be folded into `dbUnitTest.h`, and replaced with a `@ref dbunittest`.

anjohnson:
> imo. The content of this sub-section should …

No objections from me to someone doing that. Create an issue labelled "documentation" and "codeathon" for that?

mdavidsaver:
> No objections from me to someone doing that. Create an issue labelled "documentation" and "codeathon" for that?

The situation seems to me a bit too confused to be a good intro project.  Neither the code block which confused Dirk, nor the code block I intended, actually appear in the readthedocs generated output.  I have made at attempt to clear this up with #773  Also, it looks like attempts at markdown style sub-sections are lost. I will leave this to someone else #774

dirk-zimoch:
I actually got a compiler error looking for iocInit.h which was not installed yet when building in parallel. And that went away with this change. Investigating in more detail now…

dirk-zimoch:
I get this during the windows cross build:
```
make[4]: *** No rule to make target '/usr/local/epics/base-7.0.10/include/iocInit.h', needed by 'pvUnitTest.obj'.  Stop.
[...]
make[3]: *** [/usr/local/epics/base-7.0.10/configure/RULES_ARCHS:58: install.windows-x64] Error 2
make[3]: *** Waiting for unfinished jobs....
[...]
make[4]: Leaving directory '/home/zimoch/git/BASE/epics-base-7.0/modules/pvData/src/O.windows-x64-mingw'
make[3]: Leaving directory '/home/zimoch/git/BASE/epics-base-7.0/modules/pvData/src'
make[2]: *** [/usr/local/epics/base-7.0.10/configure/RULES_DIRS:85: src.install] Error 2
make[2]: Leaving directory '/home/zimoch/git/BASE/epics-base-7.0/modules/pvData'
make[1]: *** [../configure/RULES_DIRS:85: pvData.install] Error 2
make[1]: *** Waiting for unfinished jobs....
[...]
make: *** [configure/RULES_DIRS:85: modules.install] Error 2
```

Looks like when trying to build `pvUnitTest.obj` (cross for Windows on RHEL9), `iocInit.h` is not installed yet to my `INSTALL_LOCATION=/usr/local/epics/base-7.0.10/`.
`pvUnitTest.obj` is built by `modules/pvData/src/misc/Makefile`.

Here is pvUnitTest.d:
```
# DO NOT EDIT: This file created by mkmf.pl

pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsUnitTest.h
pvUnitTest.obj : ../../src/misc/pv/pvUnitTest.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/iocInit.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/testMain.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/compilerDependencies.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/libComAPI.h
pvUnitTest.obj : ../../src/misc/pv/sharedPtr.h
pvUnitTest.obj : ../../src/misc/pv/epicsException.h
pvUnitTest.obj : ../pv/pvData.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/dbCoreAPI.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/compiler/msvc/compilerSpecific.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/shareLib.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsAssert.h
pvUnitTest.obj : ../pv/pvIntrospect.h
pvUnitTest.obj : ../../src/misc/pv/typeCast.h
pvUnitTest.obj : ../../src/misc/pv/anyscalar.h
pvUnitTest.obj : ../../src/misc/pv/sharedVector.h
pvUnitTest.obj : ../../src/misc/pv/lock.h
pvUnitTest.obj : ../../src/misc/pv/noDefaultMethods.h
pvUnitTest.obj : ../pv/pvType.h
pvUnitTest.obj : ../../src/misc/pv/byteBuffer.h
pvUnitTest.obj : ../../src/misc/pv/serialize.h
pvUnitTest.obj : ../pv/pvdVersion.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsConvert.h
pvUnitTest.obj : ../../src/misc/pv/templateMeta.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsMutex.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsTypes.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsEndian.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsVersion.h
pvUnitTest.obj : ../O.Common/pv/pvdVersionNum.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/epicsGuard.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/os/WIN32/osdMutex.h
pvUnitTest.obj : /usr/local/epics/base-7.0.10/include/os/WIN32/osdWireConfig.h
```
You can see that there is `iocInit.h`. 

Even though the line `#include "iocInit.h"` in `epicsUnitTest.h` is in a block comment, `mkmf.pl` adds it to the dependencies. So that line confused not only me but `mkmf.pl` as well.

dirk-zimoch:
According to its own comment header, `mkmf.pl` has a lot of limitations that make it not do "the right thing". And that list of limitations does not even mention that it cannot recognize comments.
In my opinion, `HDEPENDS` should rather run the preprocessor (with all `CFLAGS`/`CXXFLAGS` to get all necessary command line macros right) and then parse the output for `/^#(line)?\s*[0-9]+\s+"([^"]+)"/`.

dirk-zimoch:
For MSVC, the compiler option [`/showIncludes`](https://learn.microsoft.com/en-us/cpp/build/reference/showincludes-list-include-files?view=msvc-170) may be helpful to generate a list of include files.

anjohnson:
Ahh, `mkmf.pl` — that makes more sense now. This was a script that Janet developed at a time when many C compilers didn't have the ability to collect and list dependencies. It is currently used on these targets:

- vxWorks targets built using GCC 2.x (so not actually used for a long time).
- darwin and ios: These are trivial to switch to COMP, I just tested darwin-aarch64.

```patch
diff --git a/configure/os/CONFIG.Common.iosCommon b/configure/os/CONFIG.Common.iosCommon
index 03d3cab8e..3f6dada42 100644
--- a/configure/os/CONFIG.Common.iosCommon
+++ b/configure/os/CONFIG.Common.iosCommon
@@ -91,11 +91,6 @@ vpath %.m $(USR_VPATH) $(ALL_SRC_DIRS)
 %.o: %.m
        $(COMPILE.c) -c $<
 
-#--------------------------------------------------
-# Header dependency file generation
-#
-HDEPENDS_METHOD = MKMF
-
 # Allow site overrides
 -include $(CONFIG)/os/CONFIG_SITE.Common.iosCommon
diff --git a/configure/os/CONFIG.Common.vxWorksCommon b/configure/os/CONFIG.Common.vxWorksCommon
index f28f0e327..8f562abb6 100644
--- a/configure/os/CONFIG.Common.vxWorksCommon
+++ b/configure/os/CONFIG.Common.vxWorksCommon
@@ -145,12 +145,6 @@ CODE_CXXFLAGS = $(CODE_CXXFLAGS_$(VXWORKS_MAJOR_VERSION))
 SHRLIB_CFLAGS =
 SHRLIB_LDFLAGS =
 
-#--------------------------------------------------
-# Don't use gcc 2.x for dependency generation
-
-HDEPENDS_METHOD_2 = MKMF
-HDEPENDS_METHOD = $(firstword $(HDEPENDS_METHOD_$(VX_GNU_MAJOR_VERSION)) COMP)
-
 # osithead use default stack, YES or NO override
 OSITHREAD_USE_DEFAULT_STACK = NO
diff --git a/configure/os/CONFIG.darwinCommon.darwinCommon b/configure/os/CONFIG.darwinCommon.darwinCommon
index 2c12b2fbb..d6a7ff11a 100644
--- a/configure/os/CONFIG.darwinCommon.darwinCommon
+++ b/configure/os/CONFIG.darwinCommon.darwinCommon
@@ -82,11 +82,6 @@ vpath %.m $(USR_VPATH) $(ALL_SRC_DIRS)
 %.o: %.m
        $(COMPILE.c) -c $<
 
-#
-# Header dependency file generation
-#
-HDEPENDS_METHOD = MKMF
-
 # Allow site overrides
 -include $(CONFIG)/os/CONFIG_SITE.darwinCommon.darwinCommon
```

- Windows — needs work.

[`/sourceDependencies`](https://learn.microsoft.com/en-us/cpp/build/reference/sourcedependencies?view=msvc-170) might be better than `/showIncludes` but was only added in VS-2019 version 16.7. `/showIncludes` may have come from VS-2015 (unclear), if it was neither would work for our VS-2013 builds. In both cases the output seems to be generated during the compile process too, which might have some knock-on effects on other parts of the EPICS build rules. There are [stack overflow questions](https://stackoverflow.com/questions/37685069/generate-makefile-dependency-from-msvc-output) about this, and the Google Gemini AI also suggested using a script to process the `/showIncludes` output.

I don't particularly like the idea of examining the CPP output for `#line` entries, that seems like a brittle solution that could take a lot of post-processing and be slow — do all the MSVC compiler versions generate the same output? I would prefer to fix the `mkmf.pl` script if the previous ideas won't work.

mdavidsaver:
> According to its own comment header, `mkmf.pl` has a lot of limitations that make it not do "the right thing". And that list of limitations does not even mention that it cannot recognize comments.

ick.  I can recall some past discussion about whether lines within `@code` blocks should have leading `*`.  Until `mkmf.pl` is history, I think that answer is now clear.

dirk-zimoch:
> I don't particularly like the idea of examining the CPP output for `#line` entries, that seems like a brittle solution that could take a lot of post-processing and be slow — do all the MSVC compiler versions generate the same output?

I noticed that MSVC writes `#line 123 "filename"` while gcc simply writes `# 123 "filename"` hence the `#(line)?` in the regex.

I think doing the full preprocessor run is the only way to get the correct list of `#include` files because they may be wrapped in `#if`/`#ifdef` and those may even depend on predefined compiler macros. Reimplementing all that in `mkmf.pl` will be a pain.

anjohnson:
Michael's PR #773 apparently fixes this dependency bug by replacing the sample code that has the `#include "iocInit.h"` line and adding a new "dbunittest" section.

#773 doesn't fix the problem that `mkmf.pl` generates dependencies on `#include` lines that appear in comments. Currently the modules/pvData/src/pvUnitTest.cpp has 3 extra dependencies in the output from mkmf.pl on darwin compared to using clang: dbCoreAPI.h iocInit.h and testMain.h

I agree with @mdavidsaver that for now the solution to that is to indent all Doxygen header `@code` and `\code` blocks that contain a `#include` line for files other than themself with ` * `. The files in libCom that would need this fix are (drum-roll please … 🥁🥁🥁🥁): only epicsUnitTest.h with one additional change to what's currently in #773:

```patch
diff --git a/modules/libcom/src/misc/epicsUnitTest.h b/modules/libcom/src/misc/epicsUnitTest.h
index 1f51eeeb2..8cb8da1b8 100644
--- a/modules/libcom/src/misc/epicsUnitTest.h
+++ b/modules/libcom/src/misc/epicsUnitTest.h
@@ -116,22 +116,22 @@ MAIN(iocTest)
  *
  * The following is a simple example of a test program using the epicsUnitTest
  * routines:
-\code
-#include <math.h>
-#include "epicsUnitTest.h"
-#include "testMain.h"
-
-MAIN(mathTest)
-{
-    testPlan(3);
-    testOk(sin(0.0) == 0.0, "Sine starts");
-    testOk(cos(0.0) == 1.0, "Cosine continues");
-    if (!testOk1(M_PI == 4.0*atan(1.0)))
-        testDiag("4 * atan(1) = %g", 4.0 * atan(1.0));
-    return testDone();
-}
-\endcode
-
+ * \code
+ * #include <math.h>
+ * #include "epicsUnitTest.h"
+ * #include "testMain.h"
+ *
+ * MAIN(mathTest)
+ * {
+ *     testPlan(3);
+ *     testOk(sin(0.0) == 0.0, "Sine starts");
+ *     testOk(cos(0.0) == 1.0, "Cosine continues");
+ *     if (!testOk1(M_PI == 4.0*atan(1.0)))
+ *         testDiag("4 * atan(1) = %g", 4.0 * atan(1.0));
+ *     return testDone();
+ * }
+ * \endcode
+ *
  * The output from running the above program looks like this:
 \code
 1..3
```

I will commit my changes above that switches the Apple targets to use clang to generate dependency files and remove the VxWorks support for gcc 2.x. I can also commit the above patch if @mdavidsaver agrees, it may conflict with #773.

Do we want an issue asking a Windows developer to look at using `cl /showIncludes` to generate dependency lists for MSVC targets?

The build also calls `mkmf.pl` in `ACFDEPENDS_CMD` to generate dependencies from `.acs` files. They can be run through `$(CPP)` to create `.acf` files, allowing the use of macros etc. Not sure if anyone uses these rules, some sites might.

mdavidsaver:
> only epicsUnitTest.h with

Ha.  I'm not sure how I missed that.  I think around that time I let myself be distracted by disappearing sections on readthedocs.  Anyway, now indented.

mdavidsaver:
> process the `/showIncludes` output.

Asking myself the question "what does cmake do?" led to [`cmcldeps`](https://github.com/Kitware/CMake/blob/master/Source/cmcldeps.cxx), which claims to be a wrapper around `cl.exe` adding `/showIncludes` and processing the output into a Makefile style `.d` file.

anjohnson:
Not sure that cmcldeps actually handles dependencies for C/C++ files. [This test](https://github.com/Kitware/CMake/blob/538926c5325ab04aab5203e38912621c6bd79f47/Source/cmcldeps.cxx#L251) implies that it's only expecting RC (resource compiler) files, which get do pre-processed.

mdavidsaver:
Since this PR is turning into a wider discussion, let's move to an issue #777.

PR #773, opened 2026-01-08
No doxygen @page

mdavidsaver:
The readthedocs build does not find the `@page` directives.  Since there are only two, just switch them to `@file` directives.  Also, move the dbUnitTest discussion in `epicsUnitTest.h` to `dbUnitTest.h`.  Some updated for b2ed8bb98d948584eccfcd600a7fe2a56761ee06.

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.h
+++ b/modules/database/src/ioc/db/dbEvent.h
@@ -85,14 +85,12 @@ DBCORE_API int db_start_events (
  * @param ctx Context
  *
  * @pre Call after initHookAfterInitDatabase and before initHookBeforeFree.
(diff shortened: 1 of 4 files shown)


mdavidsaver:
It looks like readthedocs does not understand markdown section markers, and truncates when it encounters one.  eg. the `IOC Testing` section does not appear...

https://github.com/epics-base/epics-base/blob/59b595f8cf1800c91ba6eb572ecf1eb50f87ac77/modules/libcom/src/misc/epicsUnitTest.h#L72-L78

image

anjohnson:
See my latest comment in #772.

PR #776, opened 2026-01-12
Soft Time Part device support

mdavidsaver:
Adds an `ai` DTYP `Soft Time Part` which wrap a call to `epicsTimeToTM()` or `epicsTimeToGMTM()`.  My motivation was wanting to trigger processing at a certain time of day and/or time of week (see below).  Likes *nix `cron` daemon.  I could almost accomplish this with `ai` `Soft Timestamp` and a modulo operation, except for those pesky timezone offsets.

Doing a separate support module seems excessive for such a small piece of code, and I see some similarity with the other bits of `devTimestamp.c`, so I propose adding to Base.

INP spec is: `@[<ref>.]<part>`, where `<ref>` is either `local` or `gm`.  `<part>` is one of the parts of `struct tm`: `sec`, `min`, `hour`, `wday`, `mday`, `mon`, `year`, or `isdst`.

So `@local.wday` is the fractional number of days.  `@gm.hour` gives the fractional hours since GMT midnight.  These are according to (I think) the system locale.  eg. for me, `@local.wday` at Monday 12:00 is `1.5`.

```
record(ai, "TST:PHAS") {
    field(DTYP, "Soft Time Part")
    field(SCAN, "1 second")
    field(INP , "@local.wday")
    field(FLNK, "TST:COMP")
}
record(ao, "TST:TGT") {
    field(VAL, "$(TGT=)")
}
record(calcout, "TST:COMP") {
    field(INPA, "TST:PHAS")
    field(INPB, "TST:TGT")
    field(CALC, "A>=B")
    field(OUT , "TST:SEQ_ PP")
    field(OOPT, "Transition To Non-zero")
}
record(seq, "TST:SEQ_") {
    field(TPRO, "1")
}
```

### DIFF
--- a/modules/database/src/std/dev/devSoft.dbd
+++ b/modules/database/src/std/dev/devSoft.dbd
@@ -54,6 +54,7 @@ device(stringout,CONSTANT,devSoSoftCallback,"Async Soft Channel")
 
 device(ai,      INST_IO,devTimestampAI,"Soft Timestamp")
 device(stringin,INST_IO,devTimestampSI,"Soft Timestamp")
+device(ai,      INST_IO,devTimePartAI ,"Soft Time Part")
 
 device(ai,	INST_IO,devAiGeneralTime,"General Time")
 device(bo,	INST_IO,devBoGeneralTime,"General Time")
--- a/modules/database/src/std/dev/devTimestamp.c
+++ b/modules/database/src/std/dev/devTimestamp.c
@@ -12,11 +12,15 @@
  *   Original Author:   Eric Norum
  */
 
+#include <string.h>
+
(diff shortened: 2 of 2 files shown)


anjohnson:
I know this idea isn't terribly efficient because of the string conversions, but for the local timezone you should be able to make the `TST:PHAS` record a stringin using the existing "Soft Timestamp" device support and `field(INP, "@%w")`. I don't have an existing solution if you need to support GMT though.

However, given that all the fields of `struct tm` are small integers that fit into an `epicsInt32`, why not change this to support the longin record type, with DTYP "Soft Timestamp" to match the others? I'd be happy to accept that into Base.

PR #778, opened 2026-01-16
appveyor: remove global email notification

mdavidsaver:
Do not pester core-talk with build failures from forks.  Remove core-talk from global appveyor notifications.  If core-talk notifications are still desired, instead switch to global or per-project notification [configured](https://www.appveyor.com/docs/notifications/#global-email-notifications) through the web ui.

I think `GitHubPullRequest` is still reasonable in the versioned `appveyor.yml`s, although this too could be moved to the per-project configuration.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -122,10 +122,4 @@ on_failure:
 #---------------------------------#
 
 notifications:
-
-  - provider: Email
-    to:
-      - core-talk@aps.anl.gov
-    on_build_success: false
-
   - provider: GitHubPullRequest
--- a/.appveyor/epics-base-7.yml
+++ b/.appveyor/epics-base-7.yml
@@ -129,10 +129,4 @@ on_failure:
 #---------------------------------#
 
 notifications:
-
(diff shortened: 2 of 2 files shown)


ralphlange:
+1

I would also suggest to leave the link to GitHub PRs. It's much more convenient to see it there than having to find the right build manually on the Appveyor web interface - which would probably often be skipped.

Also reminds me of my open AI to establish CI bridges to EPICS Chat, so that specific chat rooms could subscribe to AppVeyor results.

anjohnson:
Merged, thanks!

Does anyone else actually read emails to core-talk now? I have occasionally replied to Appveyor build failures with investigative results, but nobody else seemed to see them. I do see every message in my RSS reader, and like Ralph I find that helps me check the PR failures.

mdavidsaver:
> Does anyone else actually read emails to core-talk now?

Sometimes.  Mostly not though.  I pay much more attention to the GHA results, which are more focused, and considerably more timely.  This PR is about improving the "focused" part.

PR #781, opened 2026-01-24
Small cleanup and logging changes

JJL772:
Upstreaming some small changes from our EPICS base fork to keep things more in sync.

### DIFF
--- a/modules/database/src/ioc/rsrv/caservertask.c
+++ b/modules/database/src/ioc/rsrv/caservertask.c
@@ -312,8 +312,7 @@ void rsrv_build_addr_lists(void)
                 char sockErrBuf[64];
                 epicsSocketConvertErrnoToString (
                             sockErrBuf, sizeof ( sockErrBuf ) );
-                errlogPrintf("CAS: failed to set mcast loopback: %s\n",
-                    sockErrBuf);
+                errlogPrintf("rsrv: failed to set mcast loopback: %s (%d)\n", sockErrBuf, errno);
             }
         }
 #endif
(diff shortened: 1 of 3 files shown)


mdavidsaver (approved):
These all look ok to me.

ralphlange:
LGTM

modules/database/src/ioc/rsrv/caservertask.c
```diff
@@ -312,8 +312,7 @@ void rsrv_build_addr_lists(void)
                 char sockErrBuf[64];
                 epicsSocketConvertErrnoToString (
                             sockErrBuf, sizeof ( sockErrBuf ) );
-                errlogPrintf("CAS: failed to set mcast loopback: %s\n",
-                    sockErrBuf);
+                errlogPrintf("rsrv: failed to set mcast loopback: %s (%d)\n", sockErrBuf, errno);
```
anjohnson, line 315: Other errors from this code don’t seem to print the value of errno, why should this message be different?

mdavidsaver: > ... why should this message be different?

I would be curious to know if this message was seen in the wild.  I do (did?) not think that `IP_MULTICAST_LOOP` or `IP_MULTICAST_TTL` can fail due to system configuration.

While it would be helpful to have the other prints updated to match, I don't see this as a blocker for this PR.

JJL772: I'm not sure of the context either, the commit was from 2017. I'll update the other printfs to make them consistent though

mdavidsaver: > ... was from 2017 ...

So it was likely a result of: https://epics.anl.gov/core-talk/2017/msg00157.php .  This was a logic error specific to the RTEMS "legacy" network stack, which was corrected by a62c357e99b75d75d611848a49a7f4252ea8a539.

anjohnson:
Core Group: DZ to generate issue to have the errno to string converter always include the errno value, (and undo the then unnecessary change here.

Approved.

PR #788, opened 2026-02-06
avoid overreporting available CPUs

dirk-zimoch:
If the IOC has been started with a limited set of CPU cores (e.g. using `taskset`), `callbackParallelThreadsDefault` should reflect this. Otherwise, `callbackParallelThreads` will create more callback threads than available CPU cores, resulting in possible performance degradation.

On Linux at least, `sysconf(_SC_NPROCESSORS_ONLN)` does not take the cpu affinity mask into account.
I do not know how MacOS and RTEMS handle this. They do not have `sched_getaffinity()` (neither do older Linux systems).

### DIFF
--- a/documentation/new-notes/PR-788.md
+++ b/documentation/new-notes/PR-788.md
@@ -6,3 +6,16 @@ That fixes the problem that an IOC which was started with a limited affinity
 mask (e.g. using `taskset`) and that called `callbackParallelThreads` with an
 argument of 0 in the startup script, had started more parallel callback
 threads than it has CPU cores available.
+
+The iocsh function `callbackParallelThreads` now also accepts positive or
+negative percentages for `count`, referring to the availabple number of cores
+and always rounding the number of used cores down with a minimum of 1.
+
+#### Examples
+
+```text
+# use up to 25% of available cpu cores for callback threads
+callbackParallelThreads 25%
+# keep at least 10% of available cpu cores unused for callback threads
+callbackParallelThreads -10%
+```
--- a/documentation/new-notes/PR-788.md
+++ b/documentation/new-notes/PR-788.md
@@ -1,0 +1,8 @@
+### epicsThreadGetCPUs and callbackParallelThreads
+
+On systems that support it (e.g. recent Linux systems), `epicsThreadGetCPUs()`
+now takes the current cpu affinity mask of the calling thread into acount.
+That fixes the problem that an IOC which was started with a limited affinity
+mask (e.g. using `taskset`) and that called `callbackParallelThreads` with an
+argument of 0 in the startup script, had started more parallel callback
+threads than it has CPU cores available.
--- a/modules/database/src/ioc/db/callback.h
+++ b/modules/database/src/ioc/db/callback.h
@@ -199,7 +199,10 @@ DBCORE_API void callbackQueueShow(const int reset);
  * Calling with count<0 computes the count based on the number of CPU cores
  * currently available to the calling thread,
  * eg. Passing -2 on an 8 core system will start 6 worker threads.
- * In all cases, at least one worker thread will always run.
+ * The iocsh wrapper also accepts positive or negative percentages for count,
+ * which allows to refer to a fraction of the currently available CPU cores,
+ * always rounding the number of created threads down.
+ * In any case, at least one worker thread will always run.
  *
  * An empty prio name or the special value "*" will modify all priorities.
  * Otherwise, only the named priority is modified.
--- a/modules/database/src/ioc/db/callback.h
+++ b/modules/database/src/ioc/db/callback.h
@@ -193,18 +193,21 @@ DBCORE_API void callbackQueueShow(const int reset);
  * By default, only one thread is run for each priority (3 in total).
  *
  * Calling with count==0 will take the count from the callbackParallelThreadsDefault
- * global variable (default default is 2).
+ * global variable (initialized to the number of CPU cores initially available to
+ * the main thread).
  * Calling with count>0 sets the number of worker threads directly.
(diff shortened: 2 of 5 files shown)


anjohnson:
Good catch, this looks appropriate to me.

As a bug fix, should it have a Release Notes entry to let users know about the possible behavior change that it introduces?

mdavidsaver:
The situation on Linux is more complicated as the CPU mask is __per thread__.  So making use of `sched_getaffinity(0, ...)` can give a different answer depending on the calling thread.

mdavidsaver:
> ... can give a different answer depending on the calling thread.

With tools like mcoreutils, child threads could have either a wider or narrowing mask.  This mask can also be changed with time.

As I think about it, this seems like a limitation in the concept of `epicsThreadGetCPUs()` which can't be worked around directly.  Rather, we need to make sure that every usage of `epicsThreadGetCPUs()` can be overridden somehow.  I think this is already the case?

As for whether to accept this PR.  I am slightly inclined towards "no" as it makes the behavior of `epicsThreadGetCPUs()` more complex.

If it were to be accepted, the documentation of `epicsThreadGetCPUs()` would need to be updated to reflect this change in behavior.

anjohnson:
> … the CPU mask is **per thread**.

Is it possible to set (or query) the CPU affinity mask of a thread before it has been created? My guess would be no, and in that case it seems like an impossible problem to get right if the affinity settings are going to be different for the main and callback threads. The st.cmd script can always override the value anyway when creating callback threads.

Exclusing tests and the threadPool, the `epicsThreadGetCPUs()` routine is called in 2 places inside IOC code:
```
Cox$ git grep epicsThreadGetCPUs
modules/database/src/ioc/db/callback.c:        count = epicsThreadGetCPUs() + count;
modules/database/src/ioc/db/dbIocRegister.c:    callbackParallelThreadsDefault = epicsThreadGetCPUs();
```

The [first match](https://docs.epics-controls.org/projects/base/en/latest/callback_h.html#_CPPv423callbackParallelThreadsiPKc) is used when a `callbackParallelThreads` IOCSH command is given a negative count, meaning "reserve this many CPUs to _not_ run parallel callback threads for this priority". The second match sets how many parallel callback threads should be started for any `callbackParallelThreads` command that is given a count of zero; it is run by the the `dbIocRegister()` registrar routine. That value can be changed afterwards with a `var callbackParallelThreadsDefault` command.

Given that overrides are possible for both in the st.cmd file, IMO Dirk's change makes the default IOC behavior correct for the simplest case that users might set up (one CPU affinity mask for all IOC threads). I prefer this, which follows the principle of least surprise. However I'm not seeing any documentation about parallel threads on RTD.

This discussion indicates that a Release Note explaining the change is appropriate.

mdavidsaver:
> a negative count, meaning "reserve this many CPUs to _not_ run parallel callback threads for this priority".

This is a way better description then what I put in! (baa4cb5495cf103b4276c61b1e2a4e0077bb0d9b)

mdavidsaver:
> Is it possible to set (or query) the CPU affinity mask of a thread before it has been created? My guess would be no,

There is a non-standard `pthread_attr_setaffinity_np()` for Linux at least.

> IMO Dirk's change makes the default IOC behavior correct for the simplest case that users might set up (one CPU affinity mask for all IOC threads).

Fair enough.

ralphlange:
The original requirement was to
1. have a reasonable default for `callbackParallelThreads` and
2. allow calling it in `st.cmd` in a way that doesn't have to be updated when you start the IOC on a different machine with a different number of cores.

In another context, I recently saw a floating point number F argument, so that the number of threads would be 'F * no_of_cores', which is also interesting.

anjohnson:
> 'F * no_of_cores'

That could be implemented by making the iocsh parameter a string and looking for a leading `-` or `*` character to decide how to interpret and use the rest of the string. However the underlying C function might have to change too.

I would guess that could make sense for big machines that have large numbers of cores.

ralphlange:
> However I'm not seeing any documentation about parallel threads on RTD.

The AppDevelopers Guide (PDF) has a paragraph in the chapter on "EPICS General Purpose Tasks".

**16.2.5 Parallel Callback Tasks**

To enable multiple parallel callback tasks, and set the number of tasks to be started for each priority level, call `callbackParallelThreads` before iocInit in the startup file. The syntax is:
```
int callbackParallelThreads(int count, const char *prio)
```
The _count_ argument is the number of tasks to start, with 0 indicating to use the default (number of CPUs), and negative numbers indicating to use the number of CPUs minus the specified amount.
The _prio_ argument specifies the priority level, with ”” (empty string), ”*”, or NULL indicating to apply the definition to all priority levels.
The default value is stored in the variable `callbackParallelThreadsDefault` (initialized to the number of CPUs), which can be changed using the iocShell’s var command.

dirk-zimoch:
> 'F * no_of_cores'

```
callbackParallelThreads "50%"
```
?

dirk-zimoch:
> The AppDevelopers Guide (PDF) has a paragraph in the chapter on...

Where is the source for that PDF? Not in base it seems. How  to update documentation in that file?

ralphlange:
> Where is the source for that PDF? Not in base it seems. How to update documentation in that file?

https://github.com/epics-docs/appdev/tree/master/tex

Updates by Pull Request, obviously. Or - preferably - by converting to the new format/structure.

dirk-zimoch:
That repo builds "EPICS Application Developer’s Guide EPICS Base Release 3.16.2". My change is for EPICS 7. Thus, the pdf is no longer relevant.

> There is no 7.0 branch of this repository as the text here is being converted into other documentation formats and moved elsewhere.

PR #789, opened 2026-02-07
Fixing minor memory leak when dbPutConvertJSON call fails

### LINKED ISSUE #701: Minor memory leak in dbpf()
https://github.com/epics-base/epics-base/blob/007df69726ae6c7711985a59491f078b7b55dab5/modules/database/src/ioc/db/dbTest.c#L419-L426

A `dbpf` command to an array field that provides an invalid JSON value string will leak the `array` buffer at the return statement in line 426.

jwlodek:
Resolves #701


mdavidsaver (approved):
This looks like a straightforward fix to me.

PR #792, opened 2026-02-19
add indent_size to .editorconfig

dirk-zimoch:
Without `indent_size`, `eclint` ignores `indent_style` (which is a [bug](https://github.com/greut/eclint/issues/11)).
We prefer 4 spaces but many comment blocks are indented 1 space like this:
```
/*
 * Some comment
 */
```
These would all trigger `eclint`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,4 +15,5 @@ insert_final_newline = false
 # and remove all trailing spaces
 [*.{c,cc,cpp,h,hpp,pl,py,yml}]
 indent_style = space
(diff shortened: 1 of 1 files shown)


dirk-zimoch:
Or maybe better use [editorconfig_checker](https://editorconfig-checker.github.io/)?
That one can handle `indent_style` without `indent_size`.

dirk-zimoch:
Rather switch to editorconfig-checker (PR #793).

PR #794, opened 2026-02-20
Add contribution guidelines

anderslindho:
It would be useful to have clear contribution guidelines for this repository. GitHub surfaces `CONTRIBUTING.md` in several places - the repository overview, the sidebar, and whenever someone opens a PR or issue - making it easy for contributors to find guidance before submitting work. See https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors.

This file is adapted from the EPICS documentation repository's contribution guidelines, trimmed to remove standard Git/GitHub workflow instructions and content specific to older EPICS Base versions (3.x), which are maintained on separate branches.

This should at least be something to build on.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 For bug fixes, please create new pull requests against the default (7.0)
 branch, unless the bug only affects an older release branch.
-Fixes will be forward-ported to newer EPICS versions by the core
+These fixes will then be forward-ported to newer EPICS versions by the core
 developer team.
 
 For new features, please open an issue or discuss the idea on mailing lists
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,43 +1,5 @@
 # Contribution guidelines
 
-```{tags} beginner, user, developer, advanced
-```
-
-This document aims to show to software developers
-how to get the current EPICS Base code, modify it
-and create pull request with the changes.
-
-## Organization of the EPICS Git Repository
-
-The main EPICS repositories are hosted on Github: [github.com/epics-base](https://github.com/epics-base/). 
-The source code of EPICS Base is at [github.com/epics-base/epics-base](https://github.com/epics-base/epics-base).
-
-A mirror of the repository is available on Launchpad: <https://git.launchpad.net/epics-base>.
-
-All current and past EPICS Base versions are in the same repository on different branches.
-
-### EPICS 7
-
-The current development branch is `7.0`.
-However this branch only acts as a kind of envelope.
-The actual EPICS 7 code is divided into modules,
-each included as a Git submodule.
-
-### Older EPICS Versions
-
-There is a separate branch for each of the older EPICS Base versions:
-3.13, 3.14, 3.15, and 3.16. (However there is no 3.12 branch.)
-
-Use these to check out the latest developments of one of the older versions,
-for example to fix a bug in one of those versions.
-
-### Specific Releases
-
-Individual releases as well as pre-releases and
-release candidates are tagged like R3.16.1, R7.0.1-pre or R7.0.1-rc1.
-
-## Making Changes
-
 Changes should always be made against the head of the relevant branch,
 not against the release tags.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
+# Contribution guidelines
+
+Changes should always be made against the head of the relevant branch,
+not against the release tags.
+
```
jacomago: Is this necessary?

ralphlange:
Nice!

Are you aware of https://github.com/epics-modules/gtest ?
It's not part of Base and only applies to C++11, so it really doesn't belong here.
But with Base 7.1 and C++11 getting closer...

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Your test should run (and succeed) when you execute
+
+```bash
+make runtests
+```
+
+## Review Process
+
```
jacomago, line 39: I would add here something about joining the epics core weekly meeting if you'd like to discuss a pr in a call. Or something like this.

CONTRIBUTING.md
anderslindho: I agree it is not - removed.

CONTRIBUTING.md
anderslindho: Seems out of scope for CONTRIBUTING to discuss communication channels. I instead added a generic reference.

Where is the weekly EPICS core meeting documented?

anderslindho:
FYI I left out guidance on code style (formatting, naming conventions), commit messages, and the review process, as I'm not informed - this can be filled in by the core team later.

CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+EPICS Base comes with a testing framework which allows to run IOCs,
+set and read/compare values and more.
+
+To add a test, you will typically create a xxxTest.c
+and probably some records in a xxxTest.db file.
+(Choose a suitable name.)
+Also you need to edit the Makefile in the test directory
+as well as a file with a name like "epicsRun*Tests.c" to include your new test.
+
+Here is a basic example of a test code (xxxTest.c):
```
simon-ess: I'm not really sure that this example code adds a lot. There are a lot of test files that already exist.

CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Testing
+
+All new features must come with automated tests to prove their functionality.
+This also helps to find out if future changes break existing features.
+
+There are several “test” directories.
+Choose the one appropriate for the test.
+Keep in mind that some tests may run before all parts of Base are built.
+Details vary depending on the EPICS Base version.
```
simon-ess: This seems a bit odd. The general structure of this repository is roughly the same as most repositories; parallel src/ and test/ directories. Choosing the appropriate one should be a matter of adding a test in the "same" test directory as the source files, not an "appropriate" one.

CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
+# Contribution guidelines
+
+For bug fixes check out the branch where the bug appears first.
```
simon-ess: This seems surprising advice. It seems to suggest that you should fork off of and add commits to an older commit rather than HEAD.

I suppose that this is trying to convey that if the bug is on a release branch, put it in that release branch?

CONTRIBUTING.md
anderslindho: Can you suggest an improvement that would be accepted for all your 3 review threads? As the commit message shows, this was copied from existing guidelines under epics-docs.

simon-ess: Fair enough. In this case I think I would delete this code and say something akin to "See any of the example tests within the repository for ideas on how to structure your tests including any necessary dependencies."

CONTRIBUTING.md
simon-ess: "Add the introduced tests to the appropriate test directory, and ensure that they are added to the test `Makefile` so that they are built and run on every commit"

CONTRIBUTING.md
simon-ess: "Please create new pull requests off of the head of the default (`7.0`) branch, unless the bug in question only affects an existing release and must be released as a patch to that release"

The second part of the sentence could perhaps be dropped tbh.

CONTRIBUTING.md
anderslindho: Might leave out `so that they are built and run on every commit` as I have a mention of `runtests` being executed in CI later on - fine?

simon-ess: Sounds fine to me.

CONTRIBUTING.md
```diff
@@ -1,9 +1,11 @@
-# Contribution guidelines
+# Contribution Guidelines
 
-For bug fixes check out the branch where the bug appears first.
-The fix will be merged into newer EPICS versions by the core developer team.
+For bug fixes, please create new pull requests against the default (7.0)
+branch, unless the bug only affects an older release branch.
+Fixes will be forward-ported to newer EPICS versions by the core
```
simon-ess: Very minor, but I think this should say something more like "These fixes will then be forward-ported".

(I say this as I was confused for a second that this should say back-ported/older until I realised that you were referring in the second sentence to the _second_ half of the first sentence.)

anderslindho: Adjusted after discussion

simon-ess (approved):
LGTM

CONTRIBUTING.md
jacomago: couldn't find it anywhere :(

mdavidsaver:
Should this new page be added under `documentation/`, and made to appear under https://docs.epics-controls.org/en/latest/ ?

anderslindho:
FYI @mdavidsaver the initial document (first commit) stems from there.

FYI2 I argue in a PR (or issue - don't recall) at epics-docs that contribution guidelines are more suitably hosted in the repositories they apply for.

anderslindho:
If the document is renamed or moved, I think you lose out on the GH integrations/features listed in the initial comment in this PR.

mdavidsaver:
Ok.  Then how about moving the content into a separate file under `documentation/`, then linking from a stub `CONTRIBUTING.md`?

anjohnson:
The content looks good to me, thanks.

The [GitHub docs say](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) the site looks for a CONTRIBUTING.* file in:

> your project repository's root, `docs`, or `.github` folder.

We don't have a `docs` directory and there's no precedent for us having documents under `.github` (which would make them invisible in git worktrees), so I'm fine with it staying in the root.

Yes it should be added to the General Information section of the installed `index.html` file on docs.epics-controls.org, but that is a bit complicated to do (involving adding it to the `documentation/index.rst` file but also requiring rule changes in the `documentation/Makefile`). I'd be happy to add those to your PR repo if you want.

anderslindho:
My PR in epics-docs (https://github.com/epics-docs/epics-docs/pull/163) should probably be evaluated before any such addition so as to not duplicate contents with slightly deviating instructions.

It should be noted that epics-docs is decoupled from Base version and its branches - where my submission is tailored for 7.0 - so the guidelines might want further refinement before any such integration. The content would also need to be adjusted so as to clarify that the contribution guidelines actually apply to epics-base and related repos, and not to the docs source themselves. As this increases the scope of this task, it might be good to see that as a later improvement and defer it for now.

I would prefer that people do not push to my branches, but I am happy to continue discussions here.

Since I already have an approval from @simon-ess, could someone clarify what the merge criteria are? If a specific number of approvals is required, I'm happy to document that in the guidelines.

lucmaga:
Coming from a non EPICS background, is quite useful for a new contributor to have this file on the root directory, as it is the common place for most projects. I can clone the repo and have the instructions locally in a easy to find place.

anjohnson (approved):
This is a good start for this document. After this has been merged I will arrange to include this in the published web pages.

PR #796, opened 2026-02-24
feat(macLib): add arithmetic to macro expansion

jerzyjamroz:
It’s a feature that optimizes writing .template/.db files and reduces the risk of offset-related mistakes.

After normal macro expansion, msi evaluates the pattern as well:
```
  <number>~(+<number>)
  <number>~(-<number>)
```
Examples (after macro expansion):
```
  99~(+1)  -> 100
  99~(-1)  -> 98
  -5~(+2)  -> -3
```
Scope:
- Integer arithmetic only.
- Only +/- offsets for now.

Feedback welcome on the chosen syntax and whether we should extend this later (e.g. allow spaces, add other operators, formatting, etc.). For now I kept it intentionally minimal (unless it is fine).

Check `msiArithTest.template` within this PR for the example.

It’s implemented in a way that doesn’t change the existing behavior, it only adds a calculation step on top.


dirk-zimoch:
At PSI, I have solved this problem with calculating macros. Those expand for example `$(100+1)` to `101`, where the arguments may of course be macros themselves. My calculating macros support `*`  too, e.g. `$(100+10*$(N))`.
Unfortunately they do not work well with `()` in the mathematical expression if `$()` is used for the macro. In that case, `${}` must be used for the macro. And some things may be confusing like `${3*${N}}` with `N="2+1"` gives 7, not 9.

jerzyjamroz:
@dirk-zimoch ok so I understand you did an in-house implementation of `PSI calculating macros`. Here, I have tried to find a solution that always works and does not "fight" against the existing `msi`.
I can add multiplication. At this stage, the question is whether we go in this direction, and what syntax?
I plan to use it with `offsets` the most. 
Do you have examples where multiplication would be useful?

jerzyjamroz:
@dirk-zimoch, notice that my solution also supports default values e.g. `$(NUM=0)~(+1)` will give 1. You can even write `$(NUM=0)~($(OFF=+1))` - then `msi` resolves that and after `msiArith` will result to 1.

jerzyjamroz:
And probably this implementation will be moved to `macLib` but let's collect the feedback.

dirk-zimoch:
Multiplication may be useful when an address of a device depends on the number of the device multiplied by the size of the device address space.

mdavidsaver:
> And probably this implementation will be moved to `macLib`

I think macLib is definitely the correct place to add this syntax extension.  I have certainly wished for basic arithmetic in the past.  However, you should be aware that this will significantly increase the complexity of this project as macLib has a hand-coded and "optimized" parser for a syntax with more complexity than is usually seen (see `macLibTest.c`).  I would very much like to see someone take the time to understand this parser to the point of being able to make changes to it.  eg. the `MAC_SIZE` limit has stalled https://github.com/epics-base/epics-base/pull/102#issuecomment-796958504.

jerzyjamroz:
@mdavidsaver , ok I will add this arithmetic post process function to `macCore.c` somewhere after `trans( handle, &entry, 0, "", &s, &d, d + capacity - 1 );` therefore I will not need to analyse how it exactly works in details as it will just do its job after.
The question is about the syntax:
```
  <number>~(+<number>)
  <number>~(-<number>)
```
is this ok?
Could you write me examples of the syntax that would be the best?

jerzyjamroz:
Or just copy the idea above and e.g. `"$($(TESTNUM)+100+$(OFFSET))"` (the actual code does not support that).

dirk-zimoch:
For comparison, I have created a Draft PR #832 of how I had implemented calculating macros in macLib back in 2019. Have a look if you like.

jerzyjamroz:
@dirk-zimoch , I target a bit different architecture than in https://github.com/epics-base/epics-base/pull/832
You did that within `static MAC_ENTRY *lookup( MAC_HANDLE *handle, const char *name, int special )`.

I consider doing this within `epicsStdCall macExpandString(`
```
    trans( handle, &entry, 0, "", &s, &d, d + capacity - 1 );
    d = macArithPostprocess(dest, dest, dest + capacity - 1); <-- HERE
```
It means I support everything that is there, so `macExpandString 1:1`, plus extend with a certain syntax that we agree on [this idea evolved after the @mdavidsaver comment but I have not updated the code as I wait for the feedback].

P.S.
If I knew about your implementation before, probably I would not spend time, but just use it :)

dirk-zimoch:
> It means I support everything that is there, so `macExpandString 1:1`,

True, my version is not 100% backward compatible. If before, someone had defined a macro with the name `6*9` and the value `42`, he would now get a different result for `$(6*9)`.

modules/database/src/ioc/dbtemplate/msiArith.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+    if (!buffer) return;
+
+    size_t r = 0, w = 0;
+
+    while (buffer[r]) {
+
+        if (isInteger(buffer + r)) {
+
+            char* end1 = nullptr;
```
dirk-zimoch: Do not rely on C++ 11 in EPICS 7.0 yet!
Should be `NULL` instead.

modules/database/src/ioc/dbtemplate/msiArith.cpp
```diff
@@ -0,0 +1,62 @@
...
+            long long base = strtoll(buffer + r, &end1, 0);
+            size_t j = (size_t)(end1 - buffer);
+
+            if (buffer[j] == '~' &&
+                buffer[j + 1] == '(' &&
+                (buffer[j + 2] == '+' || buffer[j + 2] == '-'))
+            {
+                char sign = buffer[j + 2];
+
+                char* end2 = nullptr;
```
dirk-zimoch: Same as above.

jerzyjamroz:
@dirk-zimoch , Thank you, I will upload the corrected version soon, so I understand you prefer that in `.c`. No problem.

dirk-zimoch:
> I understand you prefer that in `.c`. No problem.

No, you can use C++. But not C++ 11. Stay C++ 98 compatible. VxWorks 6 still has no C++ 11 and several older Linux versions have the same problem. See the failed Docker CentOS-7 build.

`nullptr` is a C++ 11 feature and did not exist in C++ 03 or earlier.

mdavidsaver:
> ... someone had defined a macro with the name `6*9` ...

I'm inclined to think that someone who has done this should expect a problem!

jerzyjamroz:
> I'm inclined to think that someone who has done this should expect a problem!

Well, I have a feeling PSI’s EPICS will keep surprising us.

jerzyjamroz:
### Summary of how it works:
```
macExpandString()
     -> trans() [text]
     -> macArithPostprocess() [arithmetic]
     -> result string
```

### Design assumptions:
1. modules/libcom/src/macLib/macArith.h within the function description
2. modules/libcom/src/macLib/macCore.c -> `d = macArithPostprocess(dest, capacity);`

### Examples:
* modules/libcom/test/macArithTest.template

jerzyjamroz:
@mdavidsaver , @dirk-zimoch I also consider another architecture, so to add `macArithPostprocess` to the public API and use it directly within `msi` so `msi.cpp -> after: n = macExpandString(macPvt, input, buffer, MAX_BUFFER_SIZE - 1);`.

jerzyjamroz:
Also I think I found a bug in `trans` function.
This:
```
{(1+2} == $(1+2 == $(1+2
{((1+2)} == $((1+2) == $((1+2)
```
expands into:
```
{(1+2} == = $(1+2
){((1+2)} == $((1+2) == $((1+2)
```
The test done without `macArithPostprocess` so with the vanilla epics.

jerzyjamroz:
So let me know your opinion, the design is modular, so if we want comparisons, then it is just to add a function `parseComparison()` the same way as other functions.

But, do you have a real usage example for:
* comparisons
* floating point calculations?

Could you share those examples?

dirk-zimoch:
I am on vacation now. I can only have a look after the epics collaboration meeting.

modules/libcom/src/macLib/Makefile
```diff
@@ -2,15 +2,16 @@
...
 # EPICS BASE is distributed subject to a Software License Agreement found
-# in file LICENSE that is included with this distribution. 
+# in file LICENSE that is included with this distribution.
 #*************************************************************************
 
 # This is a Makefile fragment, see src/libCom/Makefile.
 
 SRC_DIRS += $(LIBCOM)/macLib
-INC += macLib.h
+INC += macLib.h macArith.h
```
mdavidsaver: Does `macArith.h` need to be added as a installed/public header and API?  Could `macArithPostprocess()` be only an internal API?

mdavidsaver:
> Also I think I found a bug in `trans` function.

You may well have found a hole in that parser.  Looking through `macLibTest.c`, I do not see any test cases with unbalanced `{}` or `()`.  Can you add a failing case which demonstrates?

Note that the `check()` function in `macLibTest.c` uses a special leading exclamation mark (`"!...`) in the `expect` string to indicate an expected failure.

mdavidsaver:
> Could you share those examples?

The two use cases which come immediately to my mind are `x+1` and `1<<x`.

Shift to compute a bit mask from a bit index.

```
dbLoadRecords("my.db", "P=TST:,N=0")
```

```
record(bi, "$(P)unit$(N)") {
    field(MASK, "${1<<N}")
}
```

Or in a MSI substitutions file where I want both 0-index and 1-index counters.

```
global {
    IDX="${N-1}"
}
file "some.db" {
{N=1}
{N=2}
...
}
```

mdavidsaver:
... a third example being left hand zero padding to a certain width.  eg. so that the record name is eg. `TST:unit01`.

```
record(ai, "$(P)unit${%02s N}") {
    field(INP, "@unit=${N}")
}
```

A real (ugly) [example](https://github.com/osprey-dcs/atf-acq-ioc/blob/ae9f966732318e0bd3321e1dcde35bb7fea9aebc/nasaAcqApp/Db/nasaAcqSup.substitutions#L73-L76) of all three cases at once...

```
file "perChannel.template" {
  pattern { P,   R,      PORT,           N,  IDX, MASK }
  { \$(P),"",\$(PORT), 01,   0, 0x00000001 }
  { \$(P),"",\$(PORT), 02,   1, 0x00000002 }
  { \$(P),"",\$(PORT), 03,   2, 0x00000004 }
...
```

I wish that both `N` and `MASK` could be computed from `IDX`.

mdavidsaver:
> ... I do not see any test cases with unbalanced `{}` or `()`. Can you add a failing case which demonstrates?

https://github.com/epics-base/epics-base/pull/834 demonstrated how to add tests, and also one minor bug.

modules/libcom/src/macLib/Makefile
jerzyjamroz: Ok, I removed that from the public API, also I understand [this option](https://github.com/epics-base/epics-base/pull/796#issuecomment-4023474755) is not considered.
Moreover, I changed the name of this function to `macArithExpandExpressions` (it will be clear later why).

jerzyjamroz:
@mdavidsaver , [comment](https://github.com/epics-base/epics-base/pull/796#issuecomment-4026036343)

* Conditions, 
* and bit shifting (it can be done by the actual multiplication as well) 
follow the actual design.

Comment about the .db file:
```
record(bi, "$(P)unit$(N)") {
    field(MASK, "${1<<N}")
}
```
requires this syntax:
```
record(bi, "$(P)unit$(N)") {
    field(MASK, "$(1<<$(N))")
}
```
because all the variables have to be expanded by the `trans`. The function I add is just the `calc` using numbers within `$(...)`. It means you need to use the actual EPICS syntax to get the numbers and then `macArithExpandExpressions` manages newly created `$(...)` numbers.

Also this:
```
global {
    IDX="${N-1}"
}
```
it would need to be:
```
global {
    IDX="$($(N)-1)"
}
```
so if `N=5` - the line will result `IDX="4"` otherwise it will stay `${N-1}`.

jerzyjamroz:
@mdavidsaver , [this comment](https://github.com/epics-base/epics-base/pull/796#issuecomment-4026072381)

it requires the architecture expansion, so to do:
```
macExpandString()
    -> trans() [exists, and stays the same as always]
    -> macArithExpandExpressions()   // e.g. ${%02 $(1<<IDX)} -> ${%02 16} or $(%02 $(1<<IDX)) -> $(%02 16) [it does not exist but can be added]
    -> macArithFormat()        // e.g. ${%02 16} -> 16, ${%02 1} -> 01 [it does not exist, architecture update]
    -> result string
```
This feature is basically formatting but not arithmetic, so it cannot be done just with one function as in other cases. Also a question appears if the prefix `macArith` is correct for this module, if there is an idea to go for such versatility.

I think, it is better to talk about this subject during the meeting.
My initial idea was related mostly to the arithmetic as it solves relations between numbers within the project.

modules/libcom/src/macLib/macArith.cpp
```diff
@@ -0,0 +1,201 @@
...
+    if (!buf || capacity == 0)
+        return buf;
+
+    char * const srcCopy = epicsStrDup(buf);
+    char * const wend = buf + capacity - 1;
+    r = srcCopy;
+    w = buf;
+
+    while (*r && w < wend) {
+        if (r[0] == '$' && r[1] == '(') {
```
anjohnson: This only supports expressions in parentheses `$(…)`, and not those in braces `${…}`. That's reasonable but should be explicitly noted in the documentation, since substitutions in macLib have always allowed the use of either (it does check that the closing character matches the opener).

However, would you consider using square brackets `$[…]` instead, to make expressions visibly different from substitutions? I suspect that would make them easier for humans to distinguish.

modules/libcom/src/macLib/macArith.cpp
```diff
@@ -0,0 +1,201 @@
...
+static int parseAddSub(const char *&p, long &out);
+
+/* Parse an integer literal from the current parse position using strtol() base auto-detection. */
+static int parseNumber(const char *&p, long &out)
+{
+    char *end = NULL;
+    long value;
+
+    skipBlankSpaces(p);
+    value = strtol(p, &end, 0);
```
anjohnson: Please look at the `epicsParseLong()` function declared in epicsStdlib.h. The implementation (in libcom/src/misc/epicsStdlib.c) looks very similar to this and should be usable here instead. The `units` argument can return a pointer to the next non-space after the number.

modules/libcom/src/macLib/macArith.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * \brief Replace valid $(...) arithmetic expressions in a string buffer.
+ * \return Pointer to the terminating NUL character in the output buffer.
+ *
+ * Scans the buffer for fragments of the form $(...) and evaluates each
+ * fragment as an integer arithmetic expression. Supported syntax includes
+ * whitespace, parentheses, leading '+' and '-', and the binary operators
+ * '+', '-', '*', '/', and '%'.
+ *
```
anjohnson, line 18: It looks like you're implementing operator precedence in expressions, that should be mentioned here. Your examples also show the use of parentheses `()` for sub-expressions, they should also be documented.

Are you planning to add bit-wise operators later? @mdavidsaver has already mentioned using `<<` and of course `~`, `&` and `|` would also be useful, but these could all be added later.

modules/libcom/src/macLib/macArith.cpp
```diff
@@ -0,0 +1,201 @@
+#include "macArith.h"
+
+#include <epicsStdio.h>
+#include <epicsStdlib.h>
+#include <epicsString.h>
+
+/* Skip spaces and tabs at the current parse position. */
+static void skipBlankSpaces(const char *&p)
+{
+    while (*p && (*p == ' ' || *p == '\t'))
```
anjohnson: How about
```c++
int c;
while ((c = *p) && isspace(c))
```

anjohnson (commented):
I haven't had time to review this until today, it looks like a clean addition. Some suggestions below.

Since the arithmetic only runs after the macro expansions have been finished, there's no way to use arithmetic to determine the name of a macro to be expanded. That might not be missed by many people though.

The main question/suggestion I have is whether enclosing expressions in `$[…]` might make it easier for humans to recognize and distinguish math expressions from macro names than re-using one of the two `$(…)` and `${…}` wrappers that macLib supports.

jerzyjamroz:
> Since the arithmetic only runs after the macro expansions have been finished, there's no way to use arithmetic to determine the name of a macro to be expanded. That might not be missed by many people though.

So you suggest something like:
`$(DEV$(N+1))` -> then it had to be written `$(DEV$($(N)+1))` -> `trans` `N=3` -> `$(DEV$(3+1))` -> `$(DEV4)` -> then it is the point you refer (it cannot be done), it means we would need to do `trans` once again -> `trans` `DEV4 = TESTDEV` -> `TESTDEV` (result).

It would mean to expand the architecture like that:
```
macExpandString()
    -> trans() [text]
    -> macArithExpandExpressions()  [arithmetic]
    -> macArithFormat()   [formatting]
+   -> trans() [text again]
    -> result string
```
and it would be possible. 

P.S.
I have not done any tests like that.

jerzyjamroz:
> The main question/suggestion I have is whether enclosing expressions in `$[…]` might make it easier for humans to recognize and distinguish math expressions from macro names than re-using one of the two `$(…)` and `${…}` wrappers that macLib supports.

No problem, any syntax can be implemented, so I will change to `$[…]`.

modules/libcom/src/macLib/macArith.cpp
jerzyjamroz: For me, any syntax is ok that everyone feels comfortable.

modules/libcom/src/macLib/macArith.cpp
jerzyjamroz: Ok, I realised I was not checking errors. There is also `epicsParseDouble`, others were asking about `floats`, so there is a question, do we want floats?

modules/libcom/src/macLib/macArith.h
jerzyjamroz: This is a small change in regards to e.g. `floats` it can be done.

modules/libcom/src/macLib/macArith.cpp
jerzyjamroz: I did this `while (*p && (*p == ' ' || *p == '\t'))` on purpose to lock to `tabs` and `spaces` only, so to detect `'\n', '\v', '\f', '\r'` as a syntax error. I will change it to `isspace`.

jerzyjamroz:
In summary, we would like to support more, so:
1) floating-point arithmetic
2) `<< , ~ , & , |`

jerzyjamroz:
I started working on the floating-point arithmetic and it seems I have a good idea, clearly it requires a new back-end function so `epicsParseDouble` but the code probably will be the same. I consider the following syntax:
- `$[...]` - integer expression
- `$f[...]` - floating point expression
so to split the processing.

modules/libcom/src/macLib/macArith.h
jerzyjamroz: I will prepare the documentation and tests once we finish with the functionality.

simon-ess:
I would also be curious about the use case for floating point operations. Integer ones are clear, floating point less so.

anjohnson:
Personally I'm not convinced there's a significant case to support floating-point here. Until someone asks for it I would hold off on implementing that, I think YAGNI applies.

You're currently using (signed) `long` values for your arithmetic. The `long` type is still 32-bit on some EPICS systems, not always 64-bit. It can simplify testing if you used one of the `epicsInt32` or `epicsInt64` types, but then you have to be careful about `printf()` format strings and standard functions like `strtol()`. The epicsStdlib.h header provides help with parsing these though.

When considering the shift operators be aware that left-shifting a negative signed value was UB until C++20, see [here](https://en.cppreference.com/w/cpp/language/operator_arithmetic.html#Built-in_bitwise_shift_operators), and right-shifting a negative value is implementation-defined and seems to give different results on ARM and Intel CPUs.

jerzyjamroz:
@anjohnson , so the preference is to use `epicsParseLLong` (`epicsParseInt64`) and adjust the code for that.

Ok, so for the future evaluation:
1. `$f[...]` - floating point expression
1. `$[...]` - integer expression
1. `$<format>[...]`, `format = 'Common Format Specifier'` e.g. `$10d[...]` - `%10d` will be inserted into the formating, so the back-end function `epicsSnprintf`.
1. `<format>` can be used to detect which processing to select, decimal or float, so as in the point 1.
1. If we do not need the floating point, then the concept with the `$<format>[...]` still can be there, e.g. for the hex `%04x`.
1. Do we want the recursive `trans`? So before and after `macArith`?

P.S.
Either way, the algorithm will stay the same, but adjusted to the agreements.

simon-ess:
I think preference for now is drop floating point stuff. Format strings seems useful (see e.g. Michael's use case above).

I don't think that recursive would be needed IMO. Isn't it enough to do successive `epicsEnvSet`s and evaluate them that way? e.g.
```
epicsEnvSet(val, $[1+2])
epicsEnvSet(foo$(val), bar)
```

jerzyjamroz:
@simon-ess , ok, so you select points 2 and 3 (with 5) without any changes to them. I will wait for the @dirk-zimoch feedback as he is on holidays.

jerzyjamroz:
Also, one thing worth noting is that this `skipWhitespace` function is used quite often throughout epics-base, so it may be worth considering adding it to the public API, for example as `epicsSkipWhitespace`.

jerzyjamroz:
> Ok, so for the future evaluation:
> 
> 1. `$f[...]` - floating point expression
> 2. `$[...]` - integer expression
> 3. `$<format>[...]`, `format = 'Common Format Specifier'` e.g. `$10d[...]` - `%10d` will be inserted into the formating, so the back-end function `epicsSnprintf`.
> 4. `<format>` can be used to detect which processing to select, decimal or float, so as in the point 1.
> 5. If we do not need the floating point, then the concept with the `$<format>[...]` still can be there, e.g. for the hex `%04x`.
> 6. Do we want the recursive `trans`? So before and after `macArith`?

@dirk-zimoch , it waits for your feedback as agreed in this PR.

jerzyjamroz:
# Formatting
Before implementing the next changes to this PR, I would like to discuss the expression and formatting syntax.
```
1) $10d[expression]      - printf-style format before the expression
2) $[expression:10d]     - Python-style format specification after the expression
3) ${expression:10d}     - Python-style but collides with macro tool `${...}` so: `,undefined)`
```
Also we resign from the floating point operations.

# How it would work:
```
epicsEnvSet("P",      "DEV:")
epicsEnvSet("IDX",    "7")
epicsEnvSet("BASE",   "$[0x800 * 2:#x]")
epicsEnvSet("STRIDE", "$[0x10 * 2:#x]")

epicsEnvSet("CHANNEL", "$[$(IDX) + 1:d]")
epicsEnvSet("CHANNEL_TEXT", "$[$(CHANNEL):02d]")
epicsEnvSet("PV", "$(P)CH$[$(CHANNEL):02d]")
epicsEnvSet("ADDRESS", "$[$(BASE) + $(IDX) * $(STRIDE):#06x]")
epicsEnvSet("LABEL", "CH$[$(CHANNEL):02d]-ADDR$[$(BASE) + $(IDX) * $(STRIDE):#06x]")
```
converted into:
```
epicsEnvSet("P",      "DEV:")
epicsEnvSet("IDX",    "7")
epicsEnvSet("BASE",   "0x1000")
epicsEnvSet("STRIDE", "0x20")

epicsEnvSet("CHANNEL",      "8")
epicsEnvSet("CHANNEL_TEXT", "08")
epicsEnvSet("PV",           "DEV:CH08")
epicsEnvSet("ADDRESS",      "0x10e0")
epicsEnvSet("LABEL",        "CH08-ADDR0x10e0")
```
example:
```
"$(P)CH$[$(CHANNEL):02d]"
    → "DEV:CH$[8:02d]"
    → "DEV:CH08"
```

# Summary
1) Calculations: integers only.
2) Formatting: under evaluation - Python’s format_spec or printf-style format
3) Integration: in macLib, after normal macro expansion, preserving existing macro syntax.
4) Evaluation order: macro expansion -> calculation -> formatting.

PR #797, opened 2026-02-24
Add dbAccess.h doxygen comments

irierr:


simon-ess (approved):
LGTM

anjohnson (approved):
Thanks!

ralphlange:
Please prefer rebase over back-merge of the target branch to keep the branch tree clean. GitHub handles rebase/force-push usually fine.

anjohnson (changes requested):
Having looked at the RTD output it would be nice if this could provide cross-links to the other header files that it includes. I think the Doxygen HTML output may provide those links automatically, but our Sphinx plugin doesn't, and without them users would have to read the header file to see what else gets pulled in by this. If you are able to add those the result would be much better.

Thanks!

PR #800, opened 2026-02-24
Added Doxygen annotations to dbAddr.h

ReBarsotti-HZB:


modules/database/src/ioc/db/dbAddr.h
```diff
@@ -8,22 +8,30 @@
...
-        struct dbCommon *precord;   /* address of record                     */
-        void    *pfield;            /* address of field                      */
-        struct dbFldDes *pfldDes;   /* address of struct fldDes              */
-        long    no_elements;        /* number of elements (arrays)           */
-        short   field_type;         /* type of database field                */
-        short   field_size;         /* size of the field being accessed      */
-        short   special;            /* special processing                    */
-        short   dbr_field_type;     /* field type as seen by database request*/
-                                    /* DBR_STRING,...,DBR_ENUM,DBR_NOACCESS  */
+        struct dbCommon *precord;   /*!< address of record                     */
```
simon-ess: Out of curiosity, why use the annotation `/*!<`? This is not how (at the moment) any of the annotations in EPICS base are done.

ReBarsotti-HZB: Ah, just a habit! But I can certainly change it if there's a preferred convention

mdavidsaver: `/**!<` is how doxygen notates a suffix comment.  Note the double `**`.  cf. https://www.doxygen.nl/manual/docblocks.html

mdavidsaver: Ah. doxygen recognizes both `/**<` and `/*!<`.

simon-ess: That was more my question. Every usage _so far_ in EPICS base is of the former, not the latter. That doesn't mean it's correct, but as someone not super used to doxygen I was wondering if it meant something different.

Honestly, `/*!<` is what the documentation says; I wonder if it's worth changing them all to match _that_, although that seems to be a completely separate task.

modules/database/src/ioc/db/dbAddr.h
```diff
@@ -8,22 +8,30 @@
...
+        void    *pfield;            /*!< address of field                      */
+        struct dbFldDes *pfldDes;   /*!< address of struct fldDes              */
```
anjohnson: While you're submitting edits to this file, could you change the language `address of …` to be `pointer to …` since the former isn't strictly correct in modern C parlance.

modules/database/src/ioc/db/dbAddr.h
```diff
@@ -8,22 +8,30 @@
 * in file LICENSE that is included with this distribution.
 \*************************************************************************/
 
+
+/**
+ * @file dbAddr.h 
+ * @brief defines structure describing a field
```
anjohnson: Another language change, a `struct dbAddr` doesn't really _describe_ a field (that's what the `struct dbFldDes` does), it's for _referencing_ one.

modules/database/src/ioc/db/dbAddr.h
```diff
@@ -8,22 +8,30 @@
...
-                                    /* DBR_STRING,...,DBR_ENUM,DBR_NOACCESS  */
+        struct dbCommon *precord;   /*!< address of record                     */
+        void    *pfield;            /*!< address of field                      */
+        struct dbFldDes *pfldDes;   /*!< address of struct fldDes              */
+        long    no_elements;        /*!< number of elements (arrays)           */
+        short   field_type;         /*!< type of database field                */
+        short   field_size;         /*!< size of the field being accessed      */
+        short   special;            /*!< special processing                    */
+        short   dbr_field_type;     /*!< field type as seen by database request \n
+                                    * DBR_STRING,...,DBR_ENUM,DBR_NOACCESS  */
```
anjohnson: There is a literal string `\n` (i.e. not just a newline character) between these two lines, causing them to be set as 2 separate `<p>…</p>` tags in the HTML output. Please remove the `\n` characters, which I think should let the browser decide where any line-break should go.

modules/database/src/ioc/db/dbAddr.h
dirk-zimoch: There are several other uses of `address of` meaning `pointer to` in several comments (happily mixed with (IP) `address of` a network device and `address of` a device on some bus, e.g. VME). But that may be worth a separate cleanup task.

modules/database/src/ioc/db/dbAddr.h
```diff
@@ -8,22 +8,31 @@
...
+        struct dbCommon *precord;   /*!< pointer to record                     */
+        void    *pfield;            /*!< pointer to field                      */
+        struct dbFldDes *pfldDes;   /*!< pointer to struct fldDes              */
+        long    no_elements;        /*!< number of elements (arrays)           */
+        short   field_type;         /*!< type of database field                */
+        short   field_size;         /*!< size of the field being accessed      */
+        short   special;            /*!< special processing                    */
+        short   dbr_field_type;     /*!< field type as seen by database request
+
+                                    * DBR_STRING,...,DBR_ENUM,DBR_NOACCESS  */
```
dirk-zimoch: Why is there a blank line now?

PR #801, opened 2026-02-24
Improvements to errorlog

### LINKED ISSUE #747: Enhancements to the errlog subsystem
The `errlogInit` and `errlogInit2` commands are currently silent when run after the errlog subsystem (implemented in `modules/libcom/src/errlog`) has been initialized, even though doing that too late doesn't actually set the buffer and message sizes.

This task is to look into the possibility of being able to set those parameters to larger values after the subsystem is already running, or at least to make the `errlogInit2` command print an error if it's called too late.

Note that `errlogInit(0)` gets called by most of the errlog API routines to implement lazy-init; this should never trigger an error message.

Add a new command `errlogShow` to display the current status and configuration of the errlog subsystem, showing the settings in the `pvt` structure, how many listeners are registered, and whether the errlog thread is still running.

simon-ess:
Fixes #747. This does three things:
* Increases the default size of the error log (and makes clear what is a default vs. minimum message size)
* Adds an IOC shell function to allow for resizing the error log buffer sizes
* Adds an IOC shell function to display information about the error log buffer

### DIFF
--- a/documentation/new-notes/PR-801.md
+++ b/documentation/new-notes/PR-801.md
@@ -1,0 +1,7 @@
+### errlogInit can be run more than once
+
+Previously, running `errlogInit` after initialisation would silently fail. It is
+now possible to call `errlogInit` or `errlogInit2` multiple times in order to
+increase the buffer size.
+
+Note that it is not possible to _shrink_ the buffer size.
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -639,19 +639,22 @@ void errlogBufResize(struct initArgs config)
     }
 
     epicsMutexMustLock(pvt.msgQueueLock);
+    char *oldlogbuf = pvt.log->base;
+    char *oldprintbuf = pvt.print->base;
 
     memcpy(logbuf, pvt.log->base, pvt.bufSize);
-    free(pvt.log->base);
     pvt.log->base = logbuf;
 
     memcpy(printbuf, &pvt.print->base, pvt.bufSize);
-    free(pvt.print->base);
     pvt.print->base = printbuf;
 
     pvt.bufSize = config.bufsize;
     pvt.maxMsgSize = config.maxMsgSize;
 
     epicsMutexUnlock(pvt.msgQueueLock);
+
+    free(oldlogbuf);
+    free(oldprintbuf);
 }
 
 int errlogInit2(int bufsize, int maxMsgSize)
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -77,6 +77,7 @@ typedef struct {
 } buffer_t;
 
 static struct {
+    epicsThreadId tid;
     epicsMutexId bufSizeLock;
     /* guarded by bufSizeLock, allowing them to be resized after initialisation */
     size_t maxMsgSize;
@@ -563,7 +564,6 @@ static void errlogInitPvt(void *arg)
 {
     struct initArgs *pconfig = (struct initArgs *) arg;
 
-    epicsThreadId tid = NULL;
     epicsThreadOpts topts = EPICS_THREAD_OPTS_INIT;
 
     topts.priority = epicsThreadPriorityLow;
@@ -574,6 +574,7 @@ static void errlogInitPvt(void *arg)
      * cantProceed() calls us.
      */
 
+    pvt.tid = NULL;
     pvt.errlogInitFailed = TRUE;
     pvt.bufSizeLock = epicsMutexCreate();
     pvt.bufSize = pconfig->bufsize;
@@ -601,11 +602,11 @@ static void errlogInitPvt(void *arg)
             && pvt.log->base
             && pvt.print->base
             ) {
-        tid = epicsThreadCreateOpt("errlog", (EPICSTHREADFUNC)errlogThread, 0, &topts);
+        pvt.tid = epicsThreadCreateOpt("errlog", (EPICSTHREADFUNC)errlogThread, 0, &topts);
     }
-    if (tid) {
+    if (pvt.tid) {
         pvt.errlogInitFailed = FALSE;
-        epicsAtExit(errlogExitHandler, tid);
+        epicsAtExit(errlogExitHandler, pvt.tid);
     }
 }
 
@@ -670,6 +671,11 @@ int errlogInit2(int bufsize, int maxMsgSize)
 
     // Don't resize the buffer for any of the calls to errlogInit(0)
     if (bufsize == 0) return 0;
+
+    if (epicsThreadGetIdSelf() == pvt.tid) {
+        fprintf(stderr, "Cannot resize buffer from callback function.\n");
+        return -1;
+    }
 
     epicsMutexMustLock(pvt.bufSizeLock);
     if (pvt.bufSize != config.bufsize || pvt.maxMsgSize != config.maxMsgSize) {
--- a/modules/libcom/src/error/errlog.c
+++ b/modules/libcom/src/error/errlog.c
@@ -583,6 +583,33 @@ static void errlogInitPvt(void *arg)
     }
 }
 
+static
+int errlogClampConfig(struct initArgs *config, int bufsize, int maxMsgSize)
+{
+    int clamped = 0;
+
+    if (bufsize < MIN_BUFFER_SIZE)
+    {
+        // This is to deal with the fact that errlogInit() is often called with arg 0
+        clamped = bufsize == 0 ? 0 : 1;
+        bufsize = MIN_BUFFER_SIZE;
+    }
+    config->bufsize = bufsize;
+
+    if (maxMsgSize > MAX_MESSAGE_SIZE(bufsize))
+    {
+        clamped = 1;
+        maxMsgSize = MAX_MESSAGE_SIZE(bufsize);
+    }
+    if (maxMsgSize < MIN_MESSAGE_SIZE)
+    {
+        clamped = 1;
+        maxMsgSize = MIN_MESSAGE_SIZE;
+    }
+    config->maxMsgSize = maxMsgSize;
+
+    return clamped;
+}
 
 int errlogInit2(int bufsize, int maxMsgSize)
 {
@@ -592,15 +619,9 @@ int errlogInit2(int bufsize, int maxMsgSize)
     if (pvt.atExit)
         return 0;
 
-    if (bufsize < MIN_BUFFER_SIZE)
-        bufsize = MIN_BUFFER_SIZE;
-    config.bufsize = bufsize;
-
-    if (maxMsgSize > MAX_MESSAGE_SIZE(bufsize))
-        maxMsgSize = MAX_MESSAGE_SIZE(bufsize);
-    if (maxMsgSize < MIN_MESSAGE_SIZE)
-        maxMsgSize = MIN_MESSAGE_SIZE;
-    config.maxMsgSize = maxMsgSize;
+    if (errlogClampConfig(&config, bufsize, maxMsgSize)) {
+        fprintf(stderr, "Warning: errlog config clamped from (%zu, %zu) to (%zu, %zu)\n", bufsize, maxMsgSize, config.bufsize, config.maxMsgSize);
+    }
 
     epicsThreadOnce(&errlogOnceFlag, errlogInitPvt, &config);
     if (pvt.errlogInitFailed) {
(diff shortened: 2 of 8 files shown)


simon-ess:
Some updates following discussion with @mdavidsaver and @anjohnson. Also added a warning if the errorlog buffer sizes are modified based on the defined macros.

simon-ess:
After further conversation with @mdavidsaver, I have removed the function `errlogBufResize` in favour of allowing `errlogInit` being called more than once.

modules/libcom/src/error/errlog.c
```diff
@@ -579,36 +608,97 @@ static void errlogInitPvt(void *arg)
...
+    memcpy(printbuf, &pvt.print->base, pvt.bufSize);
+    pvt.print->base = printbuf;
+
+    pvt.bufSize = config.bufsize;
+    pvt.maxMsgSize = config.maxMsgSize;
+
+    epicsMutexUnlock(pvt.msgQueueLock);
+
+    free(pvt.log->base);
+    free(pvt.print->base);
```
anjohnson: Freeing the new buffer pointer...

simon-ess: 🤦

modules/libcom/src/error/errlog.c
mdavidsaver, line 603: The new `pvt.bufSizeLock` needs to be added to this condition.

modules/libcom/src/error/errlog.c
```diff
@@ -579,36 +608,99 @@ static void errlogInitPvt(void *arg)
...
+{
+    if (config.bufsize < pvt.bufSize) {
+        fprintf(stderr, "Warning: Cannot shrink buffer size.\n");
+        return;
+    }
+
+    char *logbuf = calloc(1, config.bufsize);
+    char *oldlogbuf = pvt.log->base;
+    char *printbuf = calloc(1, config.bufsize);
+    char *oldprintbuf = pvt.print->base;
```
mdavidsaver: The loads of the old pointers need to be done with `msgQueueLock` locked.

mdavidsaver (changes requested):
Now that I have a chance to sit and read through `errlog.c`, I see another locking issue.  The errlogThread will access `pvt.print` with `msgQueueLock` unlocked.  So, as written, there is a risk that a concurrent resize will free that buffer while being access by errlogThread.

I think that one solution would be for the errlogThread to hold the new `bufSizeLock` when not idle (in `epicsEventMustWait(pvt.waitForWork);`).  This would mean holding `bufSizeLock` and `listenerLock` together while callbacks are being delivered.

Together, I think the locking sequences would then be:

- `bufSizeLock` before `msgQueueLock`
  - During resize
- `bufSizeLock` before `listenerLock`
  - During print/callback
  - `listenerLock` before `msgQueueLock`
    - A listener callback calls `errlogPrintf()` or resize with `errlogInit()`.

So the overall locking order would be:  `bufSizeLock` -> `listenerLock` -> `msgQueueLock`.  I think the only point where all three are held would be in the pathological case where a listener callback tries to resize.

simon-ess:
Shoot, you're right. That's annoying. I also don't think it's enough to lock the buffer resizing before the listener: there still is a gap after we release the msgQueueLock where the errlogThread reads from the print buffer.

Perhaps this is what you had in mind, but I've tried moving this to:
* errlogThread locks the buffer size lock
* errlogThread only unlocks this while waiting of messages

Thus during most circumstances a call to resize the buffer should wait until all of the messages have been handled. Overall I _think_ this is acceptable...

As you said there is a weird edge case where a listener tries to resize the error log that is... annoying.

anjohnson:
Could the "weird edge case" be detected and just ignored (do nothing)? In any case you should document that listener callbacks cannot change the buffer sizes.

modules/libcom/src/error/errlog.c
```diff
@@ -635,7 +729,14 @@ static void errlogThread(void)
             epicsMutexUnlock(pvt.msgQueueLock);
             if(wakeFlusher)
                 epicsEventMustTrigger(pvt.waitForSeq);
+
+            // Allow the buffer size to be modified only at this point
+            epicsMutexUnlock(pvt.bufSizeLock);
```
mdavidsaver: Please move this unlock before the `epicsEventMustTrigger()` call.  This general idea is a performance micro optimization to avoid unnecessary context switches.  Unlock before waking to avoid a situation where a woken thread will immediately need to wait for a mutex held by the triggering thread.  There is no problem with this logic, it just adds two avoidable context switches.

mdavidsaver (commented):
> ... there still is a gap after we release the msgQueueLock where the errlogThread reads from the print buffer.

Now that `errlogThread` must hold `bufSizeLock` while accessing the print buffer.  My thinking is that if the resizing thread holds `bufSizeLock` through `errlogInit()`, then it is certain that the errlog thread is idle or waiting for `bufSizeLock`.  So I think what you have now is safe from a concurrent resize.  That leaves the self-resize situation (from a listener callback).  This seems a niche situation, so I would be ok with a no-op/warning as Andrew suggests.

That said, it might not be too difficult to handle self-resize.  The problem I see is the `char* base = print->base + pos;` local variable potentially being invalidated after the listener callback.  Since the contents of the resized print buffer will be the same, it seems sufficient to re-fetch the base pointer each time.  (with a comment noting that such sneakiness could occur)

```diff 
-                   (*plistenerNode->listener)(plistenerNode->pPrivate, base+1u);
+                   (*plistenerNode->listener)(plistenerNode->pPrivate, print->base + pos + 1u);
```

simon-ess:
I have added a check before resizing that the listener lock is not held.

modules/libcom/src/error/errlog.c
```diff
@@ -579,36 +609,106 @@ static void errlogInitPvt(void *arg)
...
         fprintf(stderr,"errlogInit failed\n");
         exit(1);
     }
+
+    // Don't both resizing for any of the calls to errlogInit(0)
+    if (bufsize == 0) return 0;
+
+    if (epicsMutexTryLock(pvt.listenerLock) == epicsMutexLockOK) {
+        epicsMutexUnlock(pvt.listenerLock);
+    } else {
```
anjohnson: Core: Detect running in the callback thread by comparing the current thread ID, storing the errlog's thread ID on creation if we aren't already doing that.

modules/libcom/src/error/errlog.h
anjohnson: Document here and above, with Since UNRELEASED.

documentation/new-notes/PR-801.md
```diff
@@ -0,0 +1,5 @@
+### errlogInit can be run more than once
+
+Previously, running `errlogInit` after initialisation would silently fail, meaning that the fastest thread to start the error log got to select the buffer size.
```
anjohnson: Reword, remove double negatives.

modules/libcom/src/iocsh/libComRegister.c
anjohnson: Default changed...

anjohnson: Changed.

anjohnson (changes requested):
Some mod's here.

modules/libcom/src/error/errlog.h
```diff
@@ -217,9 +219,19 @@ LIBCOM_API int errlogInit(int bufsize);
...
+ * \since UNRELEASED The default buffer size is 2560 and the default maximum message size is 512.
  */
 LIBCOM_API int errlogInit2(int bufsize, int maxMsgSize);
 
+/**
+ * errlogShow is used to display the contents of the error log buffers
+ *
+ * \param level How much detail to print out
+ */
+LIBCOM_API void errlogShow(int level);
```
mdavidsaver, line 235: Should `errlogShow()` be added as public API?

Or rather placed in some private header, and only exposed via iocsh.  Maybe in a new `errlogPvt.h`?  This might later become a home for some mechanics to add logging for logClient.

wrt. adding a "show" iocsh function for logClient.  There is currently no global list of clients kept.  This could of course be usefully added, but I is a large and disjoint change that I think it should be a separate PR.

modules/libcom/src/error/errlog.c
```diff
@@ -565,50 +595,119 @@ static void errlogInitPvt(void *arg)
...
     if (pvt.errlogInitFailed) {
         fprintf(stderr,"errlogInit failed\n");
         exit(1);
     }
+
+    // Don't both resizing for any of the calls to errlogInit(0)
+    if (bufsize == 0) return 0;
+
+    if (epicsThreadGetIdSelf() == pvt.tid) {
+        fprintf(stderr, "Cannot resize buffer from callback function.\n");
```
mdavidsaver: Error messages should contain the word "error".  Secondly, provide appropriate context.

```c
fprintf(stderr, ERL_ERROR ": Cannot call errlogInit2() from an errlogListener callback function.\n");
```

Also, document new `\return` condition.

modules/libcom/src/error/errlog.c
mdavidsaver, line 766: Add a comment indicating that `bufSizeLock` is intentionally kept locked while dispatching listener callbacks.

modules/libcom/src/error/errlog.c
```diff
@@ -565,50 +595,119 @@ static void errlogInitPvt(void *arg)
...
+    }
 
     epicsThreadOnce(&errlogOnceFlag, errlogInitPvt, &config);
     if (pvt.errlogInitFailed) {
         fprintf(stderr,"errlogInit failed\n");
         exit(1);
     }
+
+    // Don't both resizing for any of the calls to errlogInit(0)
+    if (bufsize == 0) return 0;
```
mdavidsaver: The wording of this comment does not compute for me.

modules/libcom/src/error/errlog.h
```diff
@@ -217,9 +219,19 @@ LIBCOM_API int errlogInit(int bufsize);
  *
  * \param bufsize The desired buffer size
  * \param maxMsgSize The desired max message size
+ * 
+ * \since UNRELEASED It is possible to enlarge the error log buffer size after initialisation.
+ * \since UNRELEASED The default buffer size is 2560 and the default maximum message size is 512.
```
mdavidsaver: The text of a `\since` can span multiple lines if indented.

```c
 * \since UNRELEASED It is possible to enlarge the error log buffer size after initialisation.
 *                   The default buffer size is 2560 and the default maximum message size is 512.
```

mdavidsaver (commented):
Looking good.

How about some test coverage?

As a starting point, I here is [a refactor](https://github.com/user-attachments/files/26100414/errlogTestRefactor.diff.txt) of epicsErrlogTest.c to run the truncation tests twice, with the same size.

modules/libcom/src/error/errlog.c
simon-ess: If it makes you feel better it doesn't parse for me either... 🙃

simon-ess:
Resolved @mdavidsaver 's comments and added tests based on his suggestion.

mdavidsaver (approved):
This PR looks correct to me.

modules/libcom/src/error/errlog.h
simon-ess: I have moved this to a private header file as (semi-)suggested to avoid adding more public-facing API that we would need to worry about.

modules/libcom/src/error/errlog.c
```diff
@@ -565,50 +595,119 @@ static void errlogInitPvt(void *arg)
...
+static
+void errlogBufResize(struct initArgs config)
+{
+    if (config.bufsize < pvt.bufSize) {
+        fprintf(stderr, ERL_WARNING ": Cannot shrink buffer size.\n");
+        return;
+    }
+
+    char *logbuf = calloc(1, config.bufsize);
+    char *printbuf = calloc(1, config.bufsize);
```
mdavidsaver, line 627: These two allocations need a check for failure.

mdavidsaver (changes requested):

modules/libcom/src/error/errlog.c
```diff
@@ -41,9 +41,10 @@
 #include "osiUnistd.h"
 
 
-#define MIN_BUFFER_SIZE 1280
+#define MIN_BUFFER_SIZE 2560
+#define DEFAULT_MAX_MSG_SIZE 512
 #define MIN_MESSAGE_SIZE 256
```
lucmaga, line 46: I wonder if its really ok to change this default values. Otherwise LGTM

simon-ess: The general sense was that these were probably holdovers from olden times, and it's fine to update them.

lucmaga:
LGTM. Any reasons this haven't been merged?

ralphlange:
Last CI build finished a week ago. Core meeting is every two weeks. No particular reason.

PR #802, opened 2026-02-24
Delete Solaris & Sun OS target config files, code, docs, and comments

### LINKED ISSUE #705: Delete Solaris target config files, code, docs and comments

irierr:
Closes #705 
Includes all mentions of sun os, solaris, solaris studio, sun pro, HPUX


modules/libcom/src/cxxTemplates/resourceLib.h
```diff
@@ -1000,7 +1000,8 @@ inline void chronIntIdRes<ITEM>::setId (unsigned newId)
 //
 // intId::intId
 //
-// (if this is inline SUN PRO botches the template instantiation)
+// (not inline as SUN PRO would botch the template instantiation 
+// when it was supported)
 template <class T, unsigned MIN_INDEX_WIDTH, unsigned MAX_ID_WIDTH>
```
dirk-zimoch, line 1003: So can we make it `inline` now?

modules/ca/src/client/netIO.h
```diff
@@ -28,17 +28,10 @@
...
-// destructor virtual.
-#if defined ( __SUNPRO_CC ) && ( __SUNPRO_CC <= 0x540 )
-#   define NETIO_VIRTUAL_DESTRUCTOR
-#else
-#   define NETIO_VIRTUAL_DESTRUCTOR virtual
-#endif
+// we get warnings (and potential problems) if we don't 
+// make the baseNMIU destructor virtual.
+#define NETIO_VIRTUAL_DESTRUCTOR virtual
+
```
anjohnson: This header file is not installed so the only code using the `NETIO_VIRTUAL_DESTRUCTOR` macro should be inside Base, and probably inside this source directory. Since it was only added for the SUN PRO compiler, please replace all uses of the macro with the keyword `virtual`.

modules/ca/src/client/repeater.cpp
```diff
@@ -44,8 +44,8 @@
 
 /* verifyClients() Mechanism
  *
- * This is required because Solaris and HPUX have half baked versions
- * of sockets.
+ * This is required because HPUX (and previously also Solaris) has a half 
+ * baked version of sockets.
```
anjohnson: We removed support for HPUX many years ago, so these comments mentioning it and any code left which would only be used by that can also be removed. That might need a little more study in this case, it looks like the "`verifyClients()` Mechanism" described in this comment might need removing too, but I don't know how involved that might be. Please use a separate commit for that so we could restore that if something doesn't work out.

modules/ca/src/client/repeater.cpp
```diff
@@ -462,9 +462,9 @@ static void register_new_client ( osiSockAddr & from,
 
     if ( newClient ) {
         /*
-         * For HPUX and Solaris we need to verify that the clients
-         * have not gone away - because an ICMP error return does not
-         * get through to send(), which returns no error code.
+         * For HPUX we need to verify that the clients have not 
+         * gone away - because an ICMP error return does not get 
+         * through to send(), which returns no error code.
```
anjohnson: See previous comment, it looks like this whole `if () {…}` block can be deleted, along with the `verifyClients()` routine.

modules/libcom/src/cxxTemplates/resourceLib.h
```diff
@@ -1000,7 +1000,8 @@ inline void chronIntIdRes<ITEM>::setId (unsigned newId)
 //
 // intId::intId
 //
-// (if this is inline SUN PRO botches the template instantiation)
+// (not inline as SUN PRO would botch the template instantiation 
+// when it was supported)
```
anjohnson: No need to name the SUN PRO compiler here, maybe change this comment to say "could be inline", or just try making it inline?

modules/libcom/src/osi/os/posix/osdElfFindAddr.c
```diff
@@ -537,15 +536,6 @@ epicsFindAddr(void *addr, epicsSymbol *sym_p)
...
 
-    /* If the symbol is in the main executable then solaris' dladdr returns bogus info */
-#ifndef __sun
-    if ( (sym_p->s_nam = inf.dli_sname) ) {
-        sym_p->s_val = inf.dli_saddr;
-        /* Have a symbol name - just use it and be done */
-        return 0;
-    }
-#endif
-
```
anjohnson, line 548: This is `#ifndef __sun`, so the code is _not_ built on Suns and removing this block is probably wrong. Try running the `epicsStackTraceTest` program from your modules/libcom/test/O.<host> directory, it might fail with this modification.

anjohnson (changes requested):
This looks great, one removal that I don't think is right though, and an additional request detailed below.

modules/ca/src/client/repeater.cpp
irierr: Ok I will make a separate branch and PR for this and the below comment.

anjohnson: I don't think that needs a separate PR, just a separate commit within this PR.

modules/libcom/src/osi/os/posix/osdElfFindAddr.c
irierr: So on `linux-x86_64` epicsStackTraceTest passed with and without this code block - not sure how to proceed!

anjohnson: I'm reading the comment and `ifdef` as saying don't build this piece of code on Sun systems because their `dladdr()` is broken. Apparently our tests don't actually check this, which seems to be a short-cut that worked everywhere else.

Thus we want to take out the comment and preprocessor lines, but restore the C code.

modules/libcom/src/cxxTemplates/resourceLib.h
irierr: made inline in 275e0448c5626bcbaa44d049b6dec00f54eac32a

modules/libcom/src/cxxTemplates/resourceLib.h
irierr: made inline in 275e0448c5626bcbaa44d049b6dec00f54eac32a

modules/ca/src/client/repeater.cpp
irierr: Relevant commit is e60faa04cab610df99804db389e969babb92defb

irierr: Wondering if these lines referencing HPUX need looking at too though:
[modules/ca/src/client/cac.cpp#L161-L168](https://github.com/irierr/epics-base/blob/delete-solaris-target-config-files-code-docs-and-comments/modules/ca/src/client/cac.cpp#L161-L168)

modules/libcom/src/osi/epicsSignal.h
```diff
@@ -43,8 +43,6 @@ LIBCOM_API void epicsStdCall epicsSignalInstallSigPipeIgnore ( void );
  * a socket system call
  */
 LIBCOM_API void epicsStdCall epicsSignalInstallSigAlarmIgnore ( void );
-/** Raise a SIGALRM signal to a specific epicsThread */
-LIBCOM_API void epicsStdCall epicsSignalRaiseSigAlarm ( struct epicsThreadOSD * );
```
mdavidsaver, line 47: Should this public API function be removed right away?

modules/libcom/src/osi/osiSock.h
```diff
@@ -124,9 +124,6 @@ enum epicsSocketSystemCallInterruptMechanismQueryInfo {
 
     //! calling shutdown() for both read and write required to interrupt
     esscimqi_socketBothShutdownRequired,
-
-    //! NO LONGER USED/SUPPORTED
-    esscimqi_socketSigAlarmRequired
```
mdavidsaver, line 129: ... Same question for this public enum?

mdavidsaver (changes requested):
I worry that the scope of this "cleanup" PR is growing too large.  imo. it would be simpler to review and approve without public API changes which might break dependent code.

modules/libcom/src/osi/epicsSignal.h
dirk-zimoch: No. Out of scope.

modules/libcom/src/osi/osiSock.h
anjohnson: Maybe, is it ignored if requested/used now, or does it generate an error? If ignored leave it in, if it currently errors out it can be removed. IMHO.

mdavidsaver: It depends how downstream code uses this enum.  Absent a search of dependent code, I think it would be better to separate this public API removal, which may break dependent code having nothing to do with solaris.

mdavidsaver: Removing `esscimqi_socketSigAlarmRequired` breaks pvAccessCPP, which has three references.

https://github.com/search?q=repo%3Aepics-base%2FpvAccessCPP%20esscimqi_socketSigAlarmRequired&type=code

```
C:\projects\epics-base\modules\pvAccess\src\remote\blockingUDPTransport.cpp(168): error C2065: 'esscimqi_socketSigAlarmRequired': undeclared identifier
C:\projects\epics-base\modules\pvAccess\src\remote\blockingUDPTransport.cpp(168): error C2051: case expression not constant
```

irierr:
> I worry that the scope of this "cleanup" PR is growing too large. imo. it would be simpler to review and approve without public API changes which might break dependent code.

Have reverted the changes removing the public API function and enum

modules/ca/src/client/repeater.cpp
```diff
@@ -42,21 +42,6 @@
  *   to the broadcast address
  */
 
-/* verifyClients() Mechanism
- *
- * This is required because Solaris and HPUX have half baked versions
- * of sockets.
```
mdavidsaver, line 48: @anjohnson @ralphlange Are we confident that this comment is accurate?  If not, how to confirm?

mdavidsaver (changes requested):
Please remove changes to files under `modules/libcom/vxWorks/boost/`.  These files are copied verbatim from the Boost libraries, and will anyway be removed along with vxWorks support.

modules/ca/src/client/repeater.cpp
anjohnson: Good question; this comment about Solaris and HPUX was added by @ralphlange in 2002 about code and the comment below that Jeff added in 2000.

The `repeaterClient::verify()` method (at line 282 in this file) attempts to bind to this client's UDP port, which fails if the client still has that port open. The `verifyClients()` function uses that to check every client in the `client_list` and drops those that have closed their UDP socket. Since 2000 it has been run by the repeater after every new client connected to it and it has finished responding.

I suggested based on the above comment that we shouldn't need to run `verifyClients()` any more, having dropped both Solaris and HPUX.

The normal way that clients get dropped by the repeater is in the `fanOut()` function that forwards beacon messages to its clients. When the `send()` to a client fails with `SOCK_ECONNREFUSED` the `repeaterClient::sendMessage()` method returns `false`, triggering a `verify()` attempt for that client and dropping it if its UDP port can be bound to.

See PR #831 for code changes to make this visible.

ralphlange: Hard to confirm without Solaris or HP-UX systems around.

I assume I was adding the additional comment after I found that this was also necessary on HP-UX. I remember that - working on the CA Gateway - I found the the HP-UX socket handling was weird.

mdavidsaver: My concern is not about Solaris/HPUX, but rather if this behavior also applies to some currently supported OS.  I agree with Andrew's understanding about what should happen, and I like #831 to make this more observable now, and in future.

I think a test of this PR combined with #831 showing a client being added and dropped would make clear that there no regression is introduced.  I would be satisfied to see this result on one of Linux/OSX/Windows.

anjohnson:
Core Group: Conflict needs to be resolved, there is a C++20 WError job CI failure, and we also need to confirm using the PR #831 addition that `verifyClients()` is not needed.

mdavidsaver:
The `-Werror` CI job is failing with:

```
   ../repeater.cpp: In function ‘void register_new_client(osiSockAddr&, tsFreeList<repeaterClient, 32>&)’:
  make[5]: Leaving directory '/home/runner/work/epics-base/epics-base/modules/ca/src/client/O.linux-x86_64'
  ../repeater.cpp:325:10: error: variable ‘newClient’ set but not used [-Werror=unused-but-set-variable]
    325 |     bool newClient = false;
        |          ^~~~~~~~~
```

Since `newClient` is now unused, this variable can simply be removed.

PR #803, opened 2026-02-24
Implement `TOUT` (timeout) common record field

dirk-zimoch:
If `TOUT` contains a number > 0, it specifies a timeout in seconds during which a record must be processed. Otherwise it raises an `INVALID/TIMEOUT` alarm. The timeout is restarted each time the record processes and is disabled while the IOC is paused.

This is helpful in particular for records processed in irregular intervals by `"I/O Intr"`, `"Event"` or by other records and where one wants to detect when processing ceases to happen.

### DIFF
--- a/documentation/new-notes/PR-803.md
+++ b/documentation/new-notes/PR-803.md
@@ -1,0 +1,10 @@
+### New common record field `TOUT`
+
+If `TOUT` contains a number > 0, it specifies a timeout in seconds during which
+a record must be processed. Otherwise it raises an INVALID/TIMEOUT alarm.
+The timeout is restarted each time the record processes and is disabled while
+the IOC is paused.
+
+This is helpful in particular for records processed in irregular intervals by
+`"I/O Intr"`, `"Event"` or by other records and where one wants to detect when
+processing ceases to happen.
--- a/documentation/new-notes/PR-803.md
+++ b/documentation/new-notes/PR-803.md
@@ -1,10 +1,16 @@
 ### New common record field `TOUT`
 
 If `TOUT` contains a number > 0, it specifies a timeout in seconds during which
-a record must be processed. Otherwise it raises an INVALID/TIMEOUT alarm.
-The timeout is restarted each time the record processes and is disabled while
-the IOC is paused.
+record processing must have started. Otherwise the record updates its time
+stamp and then raises an INVALID/TIMEOUT alarm.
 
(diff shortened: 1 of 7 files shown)


dirk-zimoch:
Accidentally hit the close button it seems.

dirk-zimoch:
Will add tests if this gets positive feedback.

documentation/new-notes/PR-803.md
```diff
@@ -0,0 +1,10 @@
+### New common record field `TOUT`
+
+If `TOUT` contains a number > 0, it specifies a timeout in seconds during which
+a record must be processed. Otherwise it raises an INVALID/TIMEOUT alarm.
+The timeout is restarted each time the record processes and is disabled while
+the IOC is paused.
```
anjohnson: I could read this description at least 2 ways: Is the time limit value in TOUT compared to how long the record remains active (i.e PACT is true true) after a single call to `process()`, or is it the period between two adjacent calls to the `process()` routine by whatever is triggering the record? Please clarify this, I could see uses for both, although I'm not asking you to implement both.

dirk-zimoch: Updated. The time is measured between end of processing and beginning of next processing.

dirk-zimoch:
If a timeout expires but before the callback runs, a different thread cancels the timeout, will the callback still run?

mdavidsaver:
While `epicsTimerCancel()` lacks documentation, a quick look into the code behind it tries to wait for completion if the callback is in progress.

https://github.com/epics-base/epics-base/blob/8c61b02198885506ef6fbb495ab8037234aba6c6/modules/libcom/src/timer/epicsTimer.cpp#L261-L263

https://github.com/epics-base/epics-base/blob/8c61b02198885506ef6fbb495ab8037234aba6c6/modules/libcom/src/timer/timer.cpp#L139-L140

https://github.com/epics-base/epics-base/blob/8c61b02198885506ef6fbb495ab8037234aba6c6/modules/libcom/src/timer/timer.cpp#L154-L166

dirk-zimoch:
Should be safe then.

dirk-zimoch:
Actually, I fear there is a deadlock. When `timerCancel()` is called in `dbProcess()`, the record is locked and the callback function tries to lock the record and would be waiting forever because `timerCancel()`would be looping forever.

mdavidsaver:
I think you are right to worry about a deadlock.

PR #804, opened 2026-02-24
Added Doxygen annotations to iocInit.h

ReBarsotti-HZB:


documentation/Makefile.headers
```diff
@@ -130,3 +130,4 @@ menu_HEADERS += menuScan
 menu_HEADERS += menuSimm
 menu_HEADERS += menuYesNo
 
+misc_HEADERS += iocInit
```
anjohnson: This should be added to the `database_HEADERS` variable, we don't currently have any finer divisions within the indexes. Once you make that change the [IOC Header Files index](https://epics--804.org.readthedocs.build/projects/base/en/804/database-api.html) will link to your iocInit annotations.

modules/database/src/ioc/misc/iocInit.h
```diff
@@ -22,15 +23,29 @@ enum iocStateEnum {
...
  */
 DBCORE_API enum iocStateEnum getIocState(void);
+
+/** @brief Initalizes the IOC in two distinct parts - iocBuild and iocRun
+ */
 DBCORE_API int iocInit(void);
+
+/** @brief Puts the IOC into a quiescent state without allowing the various internal threads it starts to actually run */
 DBCORE_API int iocBuild(void);
+
```
anjohnson: Please annotate the `iocBuildIsolated()` routine too, which is a version of `iocBuild()` used for running IOC tests which doesn't start up the CA server. Most tests don't need that, and it can't currently be shut down cleanly and restarted. There is some discussion about this routine in the Discussion paragraph on [this page](https://epics--804.org.readthedocs.build/projects/base/en/804/dbUnitTest_h.html#db_unit_test_8h_1dbtestskel), a cross-link to that might be appropriate here.

anjohnson (changes requested):
A good start, I'll be happy to merge this after you make the fixes described below.

modules/database/src/ioc/misc/iocInit.h
mdavidsaver: @anjohnson Is it worth documenting `iocBuild*()` or `iocShutdown()` separately?  Should any user ever call these?  imo. these seem like candidates to become "private" API.

I might imagine that `iocPause()` and `iocRun()` being called from user code in some cases.  eg. some python IOC binding.

ReBarsotti-HZB: @anjohnson Ah that is very helpful, thank you!

PR #805, opened 2026-02-24
Remove some compiler warnings highlighted on windows 

FreddieAkeroyd:
Currently have "warnings as errors"  compiler flag enabled on branch, will push batches of changes as I find them when build stops locallly...


modules/libcom/src/bucketLib/bucketLib.c
```diff
@@ -167,20 +168,14 @@ static BUCKETID bucketUnsignedHash (BUCKET *pb, const void *pId)
...
-     * CPU architectures)
-     */
-    src = (size_t) *ppId;
-    hashid = src;
+    src = (uintptr_t)*ppId;
+    hashid = (BUCKETID)src;
     src = src >> pb->hashIdNBits;
     while(src){
-        hashid = hashid ^ src;
+        hashid = (BUCKETID)(hashid ^ src);
```
anjohnson:
```C
    hashid ^= (BUCKETID) src;
    src >>= pb->hashIdNBits;
```
perhaps? Is the cast even needed there?

modules/libcom/src/cvtFast/cvtFast.c
```diff
@@ -309,7 +309,7 @@ static size_t
     for (i = 0; val; i++) {
         epicsUInt64 tenth = val / 10;
 
-        digit[i] = val - tenth * 10 + '0';
+        digit[i] = (char)(val - tenth * 10 + '0');
```
anjohnson: Would this construct work without a cast?

```C
    digit[i] = '0' + (val - tenth * 10);
```

modules/libcom/src/yacc/defs.h
```diff
@@ -271,45 +271,47 @@ extern short final_state;
...
-extern void unterminated_union(int u_lineno, char *u_line, char *u_cptr) NORETURN;
-extern void over_unionized(char *u_cptr) NORETURN;
-extern void illegal_tag(int t_lineno, char *t_line, char *t_cptr) NORETURN;
-extern void illegal_character(char *c_cptr) NORETURN;
-extern void used_reserved(char *s) NORETURN;
-extern void tokenized_start(char *s) NORETURN;
+NORETURN extern void no_space(void) ;
+NORETURN extern void fatal(char *msg) ;
+NORETURN extern void open_error(char *filename) ;
+NORETURN extern void unexpected_EOF(void) ;
```
anjohnson: Please remove the spaces before the trailing semicolons in this file.

anjohnson (commented):
Thanks; some minor requests below, if you are so inclined.

modules/libcom/src/cvtFast/cvtFast.c
FreddieAkeroyd: unfortunately not i'm afraid,  it gives `C:\devel\epics-base\modules\libcom\src\cvtFast\cvtFast.c(312): warning C4244: '=': conversion from 'epicsUInt64' to 'char', possible loss of data`.  

Is `'0' + (char)(val - tenth * 10)` OK?

modules/libcom/src/bucketLib/bucketLib.c
FreddieAkeroyd: Thanks - yes changing to the combination operators removes the need for a cast

modules/libcom/src/yacc/defs.h
FreddieAkeroyd: now removed

modules/ca/src/client/access.cpp
```diff
@@ -309,7 +309,8 @@ int epicsStdCall ca_create_channel (
             }
         }
         if ( pFunc ) {
-            ( *pFunc ) ( pArg, pcac->sock, true );
+            // modify CAFDHANDLER to take SOCKET ?
+            ( *pFunc ) ( pArg, static_cast<int>(pcac->sock), true );
```
mdavidsaver: I gather that `ca_add_fd_registration()` is obscure enough that no one calls it on Windows?

@FreddieAkeroyd Do you know of other places where epics-base is still casting (truncating) `SOCKET` to `int`?

modules/libcom/src/osi/os/default/osdMessageQueue.cpp
```diff
@@ -408,7 +409,7 @@ LIBCOM_API int epicsStdCall
 epicsMessageQueuePending(epicsMessageQueueId pmsg)
 {
     char *myInPtr, *myOutPtr;
-    int nmsg;
+    int64_t nmsg;
```
mdavidsaver: This would be the first usage of `int64_t` in Base.  `epicsMessageQueueOSD::capacity` is `unsigned long`.  Or maybe `uintptr_t` to accommodate the pointer arithmetic?

```c
        nmsg = (myInPtr - myOutPtr) / pmsg->slotSize;
```

modules/libcom/src/bucketLib/bucketLib.c
```diff
@@ -167,21 +168,15 @@ static BUCKETID bucketUnsignedHash (BUCKET *pb, const void *pId)
 static BUCKETID bucketPointerHash (BUCKET *pb, const void *pId)
 {
     void * const    *ppId = (void * const *) pId;
-    size_t      src;
+    uintptr_t      src;
```
mdavidsaver: `uintptr_t` is not available on some/all vxWorks.  See the top of `registerAllRecordDeviceDrivers.cpp`.

modules/libcom/src/log/iocLogServer.c
```diff
@@ -556,7 +556,7 @@ static void readFromClient(void *pParam)
                 epicsSocketConvertErrnoToString ( sockErrBuf, sizeof ( sockErrBuf ) );
                 fprintf(stderr,
         "%s:%d socket=%d size=%d read error=%s\n",
-                    __FILE__, __LINE__, pclient->insock,
+                    __FILE__, __LINE__, (int)pclient->insock,
```
mdavidsaver: Another cast of `SOCKET` to `int`.  In this case, no functional issue.

modules/ca/src/client/access.cpp
FreddieAkeroyd: I'll review all the places with int/SOCKET and look at the call flow. Though we are truncating `socket` to `int` i believe only 24bits of the integer are used hence it will work if eventually passed to the correct win32/winsock function (but not a CRTL function)

FreddieAkeroyd:
@mdavidsaver  @anjohnson 

Tests are failing with an issue in `cvtFastTest` at https://github.com/epics-base/epics-base/blob/8c61b02198885506ef6fbb495ab8037234aba6c6/modules/libcom/test/cvtFastTest.c#L269

It warned about double truncated to float, so I replaced   `9.5000001`   with   `9.5000001F`   which caused the failure. 
```
ok 562 - cvtFloatToString(9.5000001F, 0) == 2 (2) -> "10"
ok 563 - epicsParseFloat('10') OK
not ok 564 - 9.5000001F => '10'
```

It does convert 9.5 to 10, but then fails the check after https://github.com/epics-base/epics-base/blob/8c61b02198885506ef6fbb495ab8037234aba6c6/modules/libcom/test/cvtFastTest.c#L36

I think the issue is that     `val_##typ - lit`     was being performed at double precision until I explicitly made `lit` a float then it was performed with a single precision number and failed.  If I assign `9.5000001` to a `float it` prints `9.5`. If i change `9.5000001F`  ->  `9.500001F` the test passes. Should i do this?

mdavidsaver:
> If I assign `9.5000001` to a `float it` prints `9.5`. If i change `9.5000001F`  ->  `9.500001F` the test passes. Should i do this?

If the goal is to merge this change sooner, it might be better to remove the `F` for this case.  Better a compiler warning then a runtime behavior change.

FreddieAkeroyd:
given the task size probably a good point to stop this PR at the current position, do some tidy ups as above, and continue on another PR. Regarding the float issue above i can leave out the F for now on the 9.5000001 values, should i create a separate ticket for discussing this as it is a runtime behaviour change?

mdavidsaver:
> good point to stop this PR at the current position, do some tidy ups

Agreed.

> should i create a separate ticket for discussing this as it is a runtime behavior change?

Sure.  Or maybe a more general ticket to track work on reducing warnings from MSVC?

modules/libcom/src/osi/os/default/osdMessageQueue.cpp
FreddieAkeroyd: I see we use `ptrdiff_t`  in modules/database/src/ioc/db/dbChannel.c so is that available on vxworks?

FreddieAkeroyd: looks like `ptrdiff_t` is C89 and `uintptr_t` is C99

modules/libcom/src/bucketLib/bucketLib.c
FreddieAkeroyd: have gone back to `size_t` as should be OK here

modules/libcom/src/log/iocLogServer.c
FreddieAkeroyd: cast now removed as formatters adjusted

modules/ca/src/client/access.cpp
FreddieAkeroyd: Have merged in a PR form @dirk-zimoch that changed type to SOCKET

modules/libcom/src/bucketLib/bucketLib.c
dirk-zimoch: VxWorks 6 has `uintptr_t` and VxWorks 5 is unsupported.
The more concerning part is `stdint.h`, which shows up only in VxWorks 6.9. Earlier versions define `uintptr_t` in other headers.

dirk-zimoch: To be clear: I only really care about VxWorks 6.9 (where neither `uintptr_t` nor `stdint.h` are problems) and I don't think any older versions are still in use anywhere in the world together with EPICS 7.

dirk-zimoch:
I have tested the build with VxWorks 6.9 and very old Linux versions (ELDK-4.2, Montavista 4.0) and found no incompatibility.

modules/ca/src/client/cadef.h
```diff
@@ -1479,7 +1480,7 @@ LIBCA_API unsigned epicsStdCall ca_host_minor_protocol (chid pChan);
  *  if (opened) then fd was created
  *  if (!opened) then fd was deleted
  */
-typedef void CAFDHANDLER (void *parg, int fd, int opened);
+typedef void CAFDHANDLER (void *parg, SOCKET sock, int opened);
```
mdavidsaver, line 1483: Please add a note to the doc. comment.

```
@since UNRELEASED socket argument type changed from `int` to `SOCKET`.
```

modules/libcom/src/osi/os/default/osdMessageQueue.cpp
mdavidsaver: > I see we use ptrdiff_t in modules/database/src/ioc/db/dbChannel.c so is that available on vxworks?

Apparently so.  `ptrdiff_t` seems like a reasonable answer.

Issue #791, opened 2026-02-18
Inconsistent language for string termination


dirk-zimoch:
Several comments and other documentation refer to C string terminators, but some call them "null terminated", others "NULL terminated", "nil terminated" or "zero-terminated". It would make sense to use a consistent terminology. And "NULL terminated" is absolutely wrong, in my opinion because "NULL" traditionally refers to the NULL pointer which is a totally different thing.

mdavidsaver:
I'm certain that I have contributed to this inconsistency.  Recently, I have preferred "nil terminated", although "zero terminated" is perhaps more technically accurate.  I agree that NULL is not the correct word.

anjohnson:
I asked Google's AI, and was slightly surprised when it said this:

------
> The terms "nil-terminated" and "zero-terminated" are common, but "null-terminated" seems like a mistake for describing C strings. Which is better?

While all three terms are used to describe the character-array-based strings in C, "null-terminated" is the most standard, widely understood, and technically accurate term in the context of C programming. [[1](https://en.wikipedia.org/wiki/Null-terminated_string#:~:text=In%20computer%20programming%2C%20a%20null%2Dterminated%20string%20is,article%2C%20not%20same%20as%20the%20glyph%20zero), [2](https://softwareengineering.stackexchange.com/questions/344603/are-c-strings-always-null-terminated-or-does-it-depend-on-the-platform)]  
Here is a breakdown of why, and a clarification of the common misconceptions: 

• **Null-terminated (Best/Standard)**: This refers to a character array terminated by a character with an integer value of zero, known as the NUL character (written as `'\0'`). The term "null" in this context refers to the null character, not the null pointer (`NULL`). While the terminology is often confused with `NULL`, C-style strings are officially defined as being terminated by a `NUL` character. 
• **Zero-terminated (Accurate, Less Common)**: This is technically precise because the character at the end is a numerical zero (0x00). It avoids the confusion with `NULL` pointers, but it is less standard than "null-terminated". 
[...]

PR #806, opened 2026-02-25
consistent language for string termination

ronaldomercado:
Re issue #791

### DIFF
--- a/modules/ca/src/client/CAref.html
+++ b/modules/ca/src/client/CAref.html
@@ -2751,7 +2751,7 @@ time.</p>
 <h4>Arguments</h4>
 <dl>
   <dt><code>PVNAME</code></dt>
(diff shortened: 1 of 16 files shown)


dirk-zimoch:
modules/libcom/test/epicsStringTest.c has a few `0-terminated` and I find a few `null-terminated` (with hyphen) in addition to `null terminated` (with space).
I would suggest to change those, too. Other opinions?

modules/database/src/ioc/db/chfPlugin.h mentions a "null-terminated array" for function 'chfPluginEnumString()` while the really required terminator is actually a `{NULL,0}` pair.

ronaldomercado:
Thanks @dirk-zimoch for the comments.
I have added two commits to address your observations

dirk-zimoch (approved):

anjohnson (approved):
Thanks!

PR #807, opened 2026-02-25
Doxygen annotation for dbFldTypes.h

ReBarsotti-HZB:


mdavidsaver:
When thinking about the `DBF_*` and `DBR_*`, the key connection I make is that `DBF_*` describes in-memory storage of a field of a record type (data at rest).  While the `DBR_*` codes are requests to start data in motion.  These are used by the `dbGet()` and `dbPut()` family of functions to interact with a field of a record, possibly including a value type conversion.  eg. if is valid to request `DBR_DOUBLE` of a `DBF_LONG` field.

A couple of points worth noting.

- Most `DBF_*` types concretely map to a specific C storage type.  eg `DBR_CHAR` fields will be `epicsInt8` typedef from `epicsTypes.h`.  Or `struct link` for `DBF_INLINK`.  Maybe notate the `dbfType` enum with this association to get cross-links in the generated html?
- `DBR_NOACCESS` is an exception.  This is a special case for fields using a storage type which is not mapped to one of the primitive types.  With no additional effort, these fields can not be accessed via `dbGet()` and friends.  However, the `SPC_DBADDR` modifier can be used to enable special access through the `rset::cvt_dbaddr()` record support function (eg. waveformRecord does this).
- There are more than 3 types of link.  See under `/* link types */` in `link.h`.  `PV_LINK` is a transient state used internally.

https://docs.epics-controls.org/en/latest/appdevguide/databaseDefinition.html ([source](https://github.com/epics-docs/epics-docs/blob/master/appdevguide/databaseDefinition.rst)) has more details, and may be a better place for these details.  I'm not sure if there is an easy way to link to from the dbFldTypes.h page.

PR #809, opened 2026-02-25
Added Doxygen annotation to dbFldTypes.h

ReBarsotti-HZB:


mdavidsaver:
wrt. #807 it is not necessary to start a new PR when updating.  Just do a force push to your branch.

modules/database/src/ioc/dbStatic/dbFldTypes.h
```diff
@@ -11,6 +11,43 @@
...
+ *  in a record in another IOC. A `DBF_FWDLINK` can only refer to a record 
+ *  in the same IOC. Link ﬁelds are described in a later chapter.
+ *
+ *  `DBF_INLINK` (input), `DBF_OUTLINK` (output), and `DBF_FWDLINK` 
+ *  (forward) specify that the ﬁeld is a link structure as deﬁned in `link.h`.
+ *
+ *
+ *  `DBF_NOACCESS` ﬁelds are for private use by record processing routines. 
+ *  
+ *  More information can be found at https://docs.epics-controls.org/en/latest/appdevguide/databaseDefinition.html */
```
anjohnson, line 49: A URL appearing in the text like this does get converted into a link in the Sphinx output, but it would be nicer to use a proper cross-reference link so the URL itself is hidden. @minijackson can you suggest what to use in the Doxygen annotation to get that?

I'm also aware that the particular document you're referring to is being split up slightly, but I hope the part you want will stay at the same location.

anjohnson (commented):
One request for a change if it's possible to do, but in any case this looks good, thanks!

PR #811, opened 2026-02-25
Add doxygen tags for dbAccessDefs.h

ronaldomercado:


modules/database/src/ioc/db/dbAccessDefs.h
```diff
@@ -174,35 +179,35 @@ struct dbr_alDouble     {DBRalDouble};
 
 #include "errMdef.h"
 
-#define S_db_notFound   (M_dbAccess| 1) /*Process Variable Not Found*/
```
mdavidsaver, line 177: @anjohnson Am I right in thinking that adding doxygen annotations here will be passed through by `makeStatTbl.pl`, and appear in runtime error message strings?

anjohnson: Yes, as written, the annotations added to the comment strings here would appear in the generated error strings. Please modify the block of code shown below from the `modules/libcom/src/error/makeStatTbl.pl` file:

```perl
# Extract names, values and comments from all S_ and M_ symbol definitions
while (<>) {
    chomp;
    next unless m/^ \s* \# \s* define \s+ ([SM]_[A-Za-z0-9_]+)
        \s+ (.*?) \s* \/ \* \s* (.*?) \s* \* \/ \s* $/x;
    push @syms, $1;
    $vals{$1} = $2;
    $msgs{$1} = $3;
}
```

The two `next unless …` regex lines should become this:
```perl
    next unless m/^ \s* \# \s* define \s+ ([SM]_[A-Za-z0-9_]+)
        \s+ (.*?) \s* \/ \* (?: [*!]\< \s+ [@\\]brief \s )? 
        \s* (.*?) \s* \* \/ \s* $/x;
```
I'm using a non-capturing group for the annotation part, so `$msgs{$1}` still gets the contents of match group `$3` after the change.

After editing the Perl script as above, please run `make` in your `modules/libcom/src` directory and look at the regenerated `errSymTbl.c` file in the `O.<host>` directory; the message texts in the `symbols[]` array definition should not include the annotations (if you check before running `make` they should include them).

Ref: [regex101.com](https://regex101.com/r/0mQx7R/1)

modules/database/src/ioc/db/dbAccessDefs.h
```diff
@@ -174,35 +179,35 @@ struct dbr_alDouble     {DBRalDouble};
...
+#define S_db_precision   (M_dbAccess| 9) /**< \brief get precision failed */
+#define S_db_onlyOne     (M_dbAccess|11) /**< \brief Only one element allowed*/
+#define S_db_badChoice   (M_dbAccess|13) /**< \brief Illegal choice*/
+#define S_db_badField    (M_dbAccess|15) /**< \brief Illegal field value*/
+#define S_db_lsetLogic   (M_dbAccess|17) /**< \brief Logic error generating lock sets*/
+#define S_db_noLSET      (M_dbAccess|21) /**< \brief No link support table or entry*/
+#define S_db_noRSET      (M_dbAccess|31) /**< \brief missing record support entry table*/
+#define S_db_noSupport   (M_dbAccess|33) /**< \brief RSET or DSXT routine not defined*/
+#define S_db_BadSub      (M_dbAccess|35) /**< \brief Subroutine not found*/
+/* !!!! Do not change next line without changing src/rsrv/server.h!!!!!!!! */
```
anjohnson, line 195: Please delete this comment about `src/rsrv/server.h`, the reason for that warning was removed in commit 14184c679f64b30b8f5aa73106481880ebe8f175 back in 2003 (3.14.0-beta2).

anjohnson (changes requested):
Generally looks good.

I wrote detailed instructions below (oops, above!) for a change to the Perl script that @mdavidsaver pointed out, and to remove a comment that is no longer applicable.

Thanks!

PR #812, opened 2026-02-25
iocsh command dbCreateRecord

### LINKED ISSUE #680: iocsh command dbCreateRecord
We already have an iocsh command `dbCreateAlias pdbbase record alias`. For code golfing and manual testing purposes, I would find a related command useful, this could be the help text to describe it:

```
dbCreateRecord pdbbase rectype recname

Create a new instance of the given record type.

Example: dbCreateRecord pdbbase ai my:analog
```

Unlike the older command which silently works just fine after `iocInit`, `dbCreateRecord` must do nothing and display an error if it's run after `iocInit` as any new records must be added to several data structures inside the IOC database code.

irierr:
Closes #680

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
@@ -287,7 +287,7 @@ static void dbCreateRecordCallFunc(const iocshArgBuf *args)
 
     iocState = getIocState();
     if(iocState != iocVoid) {
-        status = S_dbLib_iocNotStarted;
+        status = S_dbLib_postInitRecRegister;
     }
     else {
         dbInitEntry(*iocshPpdbbase, &ent);
--- a/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
@@ -291,10 +291,10 @@ static void dbCreateRecordCallFunc(const iocshArgBuf *args)
     }
     else {
         dbInitEntry(*iocshPpdbbase, &ent);
-        if(!args[1].sval) { // record type is required
+        if(!args[2].sval || (args[2].sval[0] == '\0')) {
+            status = S_dbLib_recordNameMissing;
+        } else if(!args[1].sval) {
             status = S_dbLib_recordTypeNotFound;
-        } else if(!args[2].sval) { // record name is required
-            status = S_dbLib_recNotFound;
         } else {
             status = dbFindRecordType(&ent, args[1].sval);
             if(!status) {
--- a/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
@@ -9,6 +9,7 @@
 \*************************************************************************/
 
 #include "iocsh.h"
+#include "iocInit.h"
 #include "errSymTbl.h"
 #include "errlog.h"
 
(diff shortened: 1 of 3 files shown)


modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
```diff
@@ -260,6 +261,54 @@ static void dbCreateAliasCallFunc(const iocshArgBuf *args)
...
+                status = dbCreateRecord(&ent, args[2].sval);
+            }
+        }
+        dbFinishEntry(&ent);
+    }
+    if(status) {
+        fprintf(stderr, ERL_ERROR ": %ld %s\n", status, errSymMsg(status));
+        iocshSetError(1);
+    }
+}
```
simon-ess, line 310: I realise that this matches the code in dbCreateAlias, but with this level of nesting I might create a separate function and `return` the statuses instead of nesting so many `ifs`.

(This is admittedly a stylistic choice of course)

modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
```diff
@@ -260,6 +261,54 @@ static void dbCreateAliasCallFunc(const iocshArgBuf *args)
...
+ */
+static void dbCreateRecordCallFunc(const iocshArgBuf *args)
+{
+    DBENTRY ent;
+    long status;
+    enum iocStateEnum iocState;
+
+    iocState = getIocState();
+    if(iocState != iocVoid) {
+        status = S_dbLib_iocNotStarted;
```
dirk-zimoch: Prints "IOC has not been started" when it should be something along "IOC already running".

modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
```diff
@@ -260,6 +261,54 @@ static void dbCreateAliasCallFunc(const iocshArgBuf *args)
...
+    }
+    else {
+        dbInitEntry(*iocshPpdbbase, &ent);
+        if(!args[1].sval) { // record type is required
+            status = S_dbLib_recordTypeNotFound;
+        } else if(!args[2].sval) { // record name is required
+            status = S_dbLib_recNotFound;
```
dirk-zimoch: When calling with a non-existing record type and missing record name, it prints "Record Not Found" instead of "Record Type does not exist".

Also when calling with existing record type and missing record name, "Record Not Found" is a bit misleading. I would prefer something like "Record name required".

anjohnson (approved):
Looks nice, thanks!

Could you also add a Release Notes entry for the PR please, to document the new command. See instructions for that in the file `documentation/new-notes/README.txt` and you could look at the entry for the `dbCreateAlias` command in `documentation/RELEASE-7.0.8.md`, although this might want a bit more wording.

dirk-zimoch (changes requested):

modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
irierr: Resolved in caffa8baeee2b32c2a525f06876ce67f353d6e93

modules/database/src/ioc/dbStatic/dbStaticIocRegister.c
irierr: Resolved in 89970f4eedc6570dee51cbe0f10af6e13780689d

dirk-zimoch (approved):
LGTM

PR #813, opened 2026-02-25
- Removing/fixing some unused initialization

hanlet:
code-a-thon fixing several codacy warnings of unused initializations

### DIFF
--- a/modules/ca/src/client/acctst.c
+++ b/modules/ca/src/client/acctst.c
@@ -1569,7 +1569,7 @@ void multiSubscriptionDeleteTest ( chid chan, unsigned interestLevel  )
  */
 void singleSubscriptionDeleteTest ( chid chan, unsigned interestLevel  )
 {
-    unsigned count = 0u;
+    unsigned count;
     evid sid;
     dbr_float_t temp, getResp;
     unsigned i;
--- a/modules/database/src/std/rec/int64outRecord.c
+++ b/modules/database/src/std/rec/int64outRecord.c
@@ -91,7 +91,7 @@ static long init_record(dbCommon *pcommon, int pass)
 {
     int64outRecord *prec = (int64outRecord*)pcommon;
     int64outdset *pdset;
-    long status=0;
+    long status;
 
     if (pass == 0) return 0;
 
--- a/modules/database/src/std/rec/mbbiRecord.c
+++ b/modules/database/src/std/rec/mbbiRecord.c
@@ -332,10 +332,10 @@ static void checkAlarms(mbbiRecord *prec, epicsTimeStamp *timeLast)
                 ((afvl > 0) ? (1.0 - alpha) : (alpha - 1.0)) * alarm;
             if (afvl - floor(afvl) > THRESHOLD)
                 afvl = -afvl;
-
-            alarm = abs((int)floor(afvl));
         }
     }
+
+    alarm = abs((int)floor(afvl));
 
     asev = alarm;
     recGblSetSevr(prec, STATE_ALARM, asev);
--- a/modules/libcom/RTEMS/score/rtems_init.c
+++ b/modules/libcom/RTEMS/score/rtems_init.c
@@ -250,7 +250,7 @@ initialize_remote_filesystem(char **argv, int hasLocalFilesystem)
     char *server_path;
     char *mount_point;
     char *cp;
-    int l = 0;
+    int l;
 
     printf ("***** Initializing NFS *****\n");
     NFS_INIT
--- a/modules/libcom/src/as/asLibRoutines.c
+++ b/modules/libcom/src/as/asLibRoutines.c
@@ -74,7 +74,7 @@ static long asAsgRuleDisable(ASGRULE *pasgrule);
 
   The version pointed to by pasbase is kept as is but locked against changes
   A new version is created and pointed to by pasbasenew
-  If anything goes wrong. The original version is kept. This results is some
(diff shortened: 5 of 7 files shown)


modules/ca/src/client/acctst.c
```diff
@@ -1569,7 +1569,7 @@ void multiSubscriptionDeleteTest ( chid chan, unsigned interestLevel  )
  */
 void singleSubscriptionDeleteTest ( chid chan, unsigned interestLevel  )
 {
-    unsigned count = 0u;
+    unsigned count;
```
mdavidsaver, line 1572: fyi. this local is only used within the following loop, and so could be moved into the loop block.

modules/database/src/std/rec/int64outRecord.c
```diff
@@ -91,7 +91,7 @@ static long init_record(dbCommon *pcommon, int pass)
 {
     int64outRecord *prec = (int64outRecord*)pcommon;
     int64outdset *pdset;
-    long status=0;
+    long status;
```
mdavidsaver, line 94: Similarly this local could be moved into the body of the `if(` below.

modules/database/src/std/rec/mbbiRecord.c
```diff
@@ -332,11 +332,11 @@ static void checkAlarms(mbbiRecord *prec, epicsTimeStamp *timeLast)
                 ((afvl > 0) ? (1.0 - alpha) : (alpha - 1.0)) * alarm;
             if (afvl - floor(afvl) > THRESHOLD)
                 afvl = -afvl;
-
-            alarm = abs((int)floor(afvl));
         }
     }
 
+    alarm = abs((int)floor(afvl));
+
```
mdavidsaver, line 339: I'm not sure about this change.  It looks like it might clobber `alarm` in the `aftc <= 0` case.

modules/libcom/RTEMS/score/rtems_init.c
```diff
@@ -250,7 +250,7 @@ initialize_remote_filesystem(char **argv, int hasLocalFilesystem)
     char *server_path;
     char *mount_point;
     char *cp;
-    int l = 0;
+    int l;
```
mdavidsaver, line 253: `initialize_remote_filesystem()` is quite convoluted.  I guess if gcc/clang does not now complain about use of uninitialized, then ok?

modules/libcom/src/as/asLibRoutines.c
```diff
@@ -159,7 +159,11 @@ long epicsStdCall asInitialize(ASINPUTFUNCPTR inputfunction)
                 pnextoldmem = (ASGMEMBER *)ellNext(&poldmem->node);
                 ellDelete(&poldasg->memberList,&poldmem->node);
                 status = asAddMemberPvt(&poldmem,poldmem->asgName);
-                poldmem = pnextoldmem;
+		if (!status) {
+		  poldmem = pnextoldmem;
+		} else {
+		  fprintf(stderr, ERL_ERROR " asInitialize: Failed asAddMemberPvt\n");
+		}
```
mdavidsaver, line 166: It certainly looks like there is missing error handling here.  I'm not sure what the correct handling is though.

also, please use consistent indentation.

modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -122,8 +122,9 @@ static void pwdCallFunc (const iocshArgBuf *args)
     char buf[1024];
     char *pwd = getcwd ( buf, sizeof(buf) );
     if ( pwd ) {
-        buf[sizeof(buf)-1u] = '\0';
-        printf ( "%s\n", pwd );
+      /* Ensure proper termination */
+      buf[sizeof(buf)-1u] = '\0';
+      printf ( "%s\n", pwd );
```
mdavidsaver, line 127: Indentation?

modules/libcom/src/log/iocLogServer.c
```diff
@@ -253,7 +253,7 @@ static int seekLatestLine (struct ioc_log_server *pserver)
             static const char *pMonths[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun",
                 "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
             static const unsigned nMonths = sizeof(pMonths)/sizeof(pMonths[0]);
-            time_t lineTime = (time_t) -1;
+            time_t lineTime;
```
mdavidsaver, line 256: Another candidate local variable to move into an inner scope.

modules/database/src/std/rec/mbbiRecord.c
anjohnson: While correct, this change will probably clash with PR #817 which performs exactly the same move to fix a bug in the mbbi record that this represents. @eddybl has other changes in there too though.

modules/libcom/src/iocsh/libComRegister.c
anjohnson: Agreed, I don't see any other changes in the functionality, the original indentation was more correct (4 spaces) than the newer one, and the new comment doesn't seem necessary.

modules/database/src/std/rec/mbbiRecord.c
eddybl: actually this will not fix the bug. It will only fix the warning of unused variable, but the actual bug fix would be to write the value back to `prec->afvl` (as done in #817 )

Issue #783, opened 2026-01-31
Local build of RTD documentation fails in hoverxref


ralphlange:
I am trying to build a local copy of the Base RTD documentation, to have the equivalent of AppDevGuide/Doxygen in a local install for systems without internet connection.

On ITER's build servers (RHEL8), we had to jump through multiple hoops to get this almost working, including packaging our own version of sphinx, RTD theme, markdown parser and all their dependencies. Some features (e.g., the 'option' keyword in CLI documentation) do not work at all. Probably because of sphinx being too old (using 5.3.0 there)?

On my laptop (WSL2/Ubuntu24), using a venv, 'make sphinx' also fails, with

```
cd ../O.Common; /home/langer/work/EPICS/7.0/base/documentation/.venv/bin/python3 -m sphinx rtd-src readthedocs
Running Sphinx v9.1.0
loading translations [en]... done
making output directory... done

Extension error (hoverxref.extension)!

Versions
========

* Platform:         linux; (Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39)
* Python version:   3.12.3 (CPython)
* Sphinx version:   9.1.0
* Docutils version: 0.22.4
* Jinja2 version:   3.1.6
* Pygments version: 2.19.2

Last Messages
=============

None.

Loaded Extensions
=================

None.

Traceback
=========

      File "/home/langer/work/EPICS/7.0/base/documentation/.venv/lib/python3.12/site-packages/sphinx/events.py", line 452, in emit
        raise ExtensionError(
    sphinx.errors.ExtensionError: Handler <function deprecated_configs_warning at 0x779c8c049e40> for event 'config-inited' threw an exception (exception: cannot unpack non-iterable _Opt object)

```
What versions of things do I need to build the sphinx/RTD docs correctly?

anjohnson:
I think we need to update our doc tool versions. Starting from the `.readthedocs.yaml` file:
```YAML
# .readthedocs.yml
# Read the Docs configuration file
# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details

# Required
version: 2

build:
  os: ubuntu-22.04
  tools:
      python: "3.9"
  apt_packages:
    - graphviz
    - rsync
  jobs:
    pre_build:
      - make inc
      - make -C documentation rtd

# Build documentation in the documentation/O.Common directory with Sphinx
sphinx:
  configuration: documentation/O.Common/rtd-src/conf.py

# …

submodules:
[...]

PR #808, opened 2026-02-25
Add more information about epics-base documentation workflow

eddybl:
This PR does two things:

- Updates the ReadTheDocs build setup to use Ubuntu 24.04, which should make it stable for a longer time, before potentially running into #783 
- Add documentation of how the EPICS Base documentation building works and how to build it locally

PR #814, opened 2026-02-25
Upgrade documentation dependencies, add build guide, switch to `pyproject.toml`

minijackson:
This should fix the Doxygen + Breathe issue mentioned in #783, by upgrading to the alpha release of Breathe `5.0.0a5`.

As requested by @anjohnson, I've added a small guide on how to build the documentation. The file is not exactly in the place I would want, but I think with the way things are set up currently, we're not allowed to have subfolders.

I've also switched the `requirements.txt` to be a `pyproject.toml`, so that anyone can use their favorite Python tool. I'm tagging @simon-ess as the setuptools/pip fan, to see if it's okay with him.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,9 +6,9 @@
 version: 2
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
-      python: "3.9"
+      python: "3.14"
   apt_packages:
     - graphviz
     - rsync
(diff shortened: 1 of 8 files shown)


eddybl:
This is probably the better PR instead of #808

.readthedocs.yaml
```diff
@@ -8,7 +8,7 @@ version: 2
 build:
   os: ubuntu-22.04
```
eddybl:
```suggestion
  os: ubuntu-24.04
```

why not use the opportunity to switch to 24.04. Ideally this should make it stable for a longer period of time

minijackson:
@eddybl oops sorry, I didn't see that you already opened a PR…

eddybl:
I updated #808 to only add some additional information about the docs building into the `documentation/README.md` section and reference to the new MD file

Maybe it makes more sense to pull the changes into this PR

PR #816, opened 2026-02-26
Change Win-22 MSC-22 from debug to default configuration

FreddieAkeroyd:
change `debug` to `default` - guessing typo as job is otherwise a duplicate build

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -179,7 +179,6 @@ jobs:
             os: windows-2022
             cmp: vs2022
             configuration: default
-            extra: "CMD_CXXFLAGS=-analyze"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-scripts-build.yml
```diff
@@ -178,7 +178,7 @@ jobs:
           - name: "Win-22 MSC-22"
             os: windows-2022
             cmp: vs2022
-            configuration: debug
+            configuration: default
             extra: "CMD_CXXFLAGS=-analyze"
```
mdavidsaver: I think the "extra" should then move to the "duplicate" below.

FreddieAkeroyd: yes - now moved

PR #817, opened 2026-02-26
Add AFTC flag for bi record and fix mbbi AFTC bug

### LINKED ISSUE #573: `AFTC` for `bi` record
The `AFTC` field is implemented for 5 different records, as far as I am aware:

- `ai`
- `longin`
- `mbbi`
- `calc`
- `int64in`

I see use cases also for the `bi` record to support this feature (imagine a record which indicates a mismatch between a setpoint and readback)

eddybl:
Closes #573 

**bi record**
- Adds ATFC feature to bi record
- Adds AFTC tests for the bi record

**mbbi record**
- Fixes bug in mbbi record which made the AFTC feature not work
- Sets up basic testing for mbbi record, including AFTC feature
- Fixes a bug for not properly updating LALM if COSV is set


modules/database/test/std/rec/mbbiTest.c
```diff
@@ -0,0 +1,201 @@
...
+    /* set ZRSV to MAJOR and verify that SEVR is now MAJOR */
+    testdbPutFieldOk("test_mbbi_rec.ZRSV", DBF_SHORT, menuAlarmSevrMAJOR);
+    testdbGetFieldEqual("test_mbbi_rec.SEVR", DBF_SHORT, menuAlarmSevrMAJOR);
+
+    /* set VAL to 1 on linked record and verify that COSV now sets the SEVR to INVALID */
+    testdbPutFieldOk("test_mbbi_link_rec.VAL", DBF_SHORT, 1);
+    testdbGetFieldEqual("test_mbbi_rec.SEVR", DBF_SHORT, menuAlarmSevrINVALID);
+
+    /* verify LAML */
+    testdbGetFieldEqual("test_mbbi_rec.LALM", DBF_SHORT, 1);
```
eddybl, line 111: this test currently fails in my local testing, but I couldn't understand why exactly.
I tried to manually reproduce it and was not successful, so I assume this is a bug in the test procedure

eddybl:
I was wrong: the failing test actually pointed out a bug in interaction between LALM and COSV fields for the mbbi record.

modules/database/test/std/rec/mbbiTest.c
```diff
@@ -0,0 +1,201 @@
...
+#include "errlog.h"
+#include "dbAccess.h"
+#include "menuAlarmSevr.h"
+#include "menuScan.h"
+
+void recTestIoc_registerRecordDeviceDriver(struct dbBase *);
+
+static void test_soft_input(void){
+    /* set soft channel */
+    testdbPutFieldOk("test_mbbi_rec.DTYP", DBF_STRING, "Soft Channel");
```
mdavidsaver, line 18: What is being tested?  "Soft Channel" is the default for `DTYP`.

modules/database/src/std/rec/biRecord.c
```diff
@@ -217,23 +224,50 @@ static long put_enum_str(const DBADDR *paddr, const char *pstring)
...
+
+    if (val == 0) {
+        asev = prec->zsv;
+    } else {
+        asev = prec->osv;
+    }
+
+    alarm = asev;  /* Initialize alarm with current state severity */
+    aftc = prec->aftc;
+    afvl = 0;
```
mdavidsaver: fyi. it is no longer required to declare variables at the top of each block.  So:

```c
unsigned short alarm = asev;  /* Initialize alarm with current state severity */
double aftc = prec->aftc;
double afvl = 0;
```

modules/database/test/std/rec/mbbiTest.c
eddybl: This test was copied from the `bi` record. I did not really question why this particular test exists, but if it is not useful I can remove it

modules/database/src/std/rec/mbbiRecord.c
```diff
@@ -336,15 +336,15 @@ static void checkAlarms(mbbiRecord *prec, epicsTimeStamp *timeLast)
...
 
     /* Check for COS alarm */
-    if (val == prec->lalm ||
-        recGblSetSevr(prec, COS_ALARM, prec->cosv))
+    if (val == prec->lalm)
         return;
-
+    recGblSetSevr(prec, COS_ALARM, prec->cosv);
     prec->lalm = val;
 }
```
anjohnson, line 349: Core Group: Needs a second look from us to confirm these.

modules/database/test/std/rec/mbbiTest.c
anjohnson: I regard the use of `testdbPutFieldOk()` to configure the record for later tests as benign, it could catch weird issues in the future.

Setting DTYP to the default value isn't necessary in this case, but it makes the tests symmetrical (testBi has checks with "Raw Soft Channel" too) and prevents problems if someone decided to re-order the test subroutines in the future.

modules/database/src/std/rec/mbbiRecord.c
anjohnson: On comparing with other record types I believe this change is correct. The old code was a bit obscure, and the original commit on the 3.15 branch that added AFTC missed this change.

modules/database/src/std/rec/biRecord.c
anjohnson: The indentation of this call to `recGblSetSevr()` is wrong, it should be only 4 spaces.

anjohnson (approved):
Looks good; one indentation fix to the original file would be nice, but the Core group can handle that if we get to merging first.

modules/database/src/std/rec/mbbiRecord.c
eddybl: The tests run by the `test_alarm()` routine pointed me to this issue. If you run the previous version against the tests, it will fail

modules/database/src/std/rec/biRecord.c
eddybl: fixed

modules/database/src/std/rec/biRecord.c
eddybl: fixed

anjohnson:
Core Group: Will merge after CI finishes successfully.

PR #819, opened 2026-02-26
Hanlet codeathon26

hanlet:
Remove cygwin in partial preparation for v7.1


modules/database/src/std/rec/mbbiRecord.c
```diff
@@ -332,11 +332,11 @@ static void checkAlarms(mbbiRecord *prec, epicsTimeStamp *timeLast)
                 ((afvl > 0) ? (1.0 - alpha) : (alpha - 1.0)) * alarm;
             if (afvl - floor(afvl) > THRESHOLD)
                 afvl = -afvl;
-
-            alarm = abs((int)floor(afvl));
         }
     }
 
+    alarm = abs((int)floor(afvl));
+
```
eddybl: I assume this change was due to  `avfl` being declared as an unsued variable (in the situation that only the if-clause is executed).

But this was actually due to a bug, which is resolved within #817 which would create a conflict with this change.

the actual issue is that `avfl` is not written back to `proc->avfl`.

mdavidsaver: This change was unintentional cross-talk with #813, which has now been removed from this PR.

configure/os/CONFIG.win32-x86.Common
```diff
@@ -18,7 +18,6 @@ OSITHREAD_USE_DEFAULT_STACK = NO
 
 HOSTEXE=.exe
 
-# Does not work if using cygwin make
 # because HOME is always defined
```
anjohnson, line 21: Not sure that leaving only the second comment line here makes sense on its own.

configure/os/CONFIG_SITE.windows-x64-mingw.windows-x64-mingw
```diff
@@ -3,5 +3,5 @@
 # Site Specific definitions for windows-x64-mingw target
 # Only the local epics system manager should modify this file
 
-# Prefix for mingw compiler from cygwin
+# Prefix for mingw compiler
 #CMPLR_PREFIX = x86_64-w64-mingw32-
```
anjohnson, line 7: I think we may want to delete this whole `os/CONFIG_SITE.…` file, not just 2 words in the comment. I read the original comment as suggesting you might need this prefix when you're running a MinGW cross-compiler that's packaged with Cygwin. I don't see that as making sense when the host architecture is `windows-x64-mingw` though.

@ralphlange could you comment on this?

modules/database/test/std/rec/netget.plt
```diff
@@ -30,7 +30,7 @@ $ENV{EPICS_PVA_BROADCAST_PORT} = 55076;
 $ENV{EPICS_PVAS_INTF_ADDR_LIST} = 'localhost';
 
 my $bin = '@TOP@/bin/@ARCH@';
-my $exe = ($^O =~ m/^(MSWin32|cygwin)$/x) ? '.exe' : '';
+my $exe = ($^O =~ m/^(MSWin32)$/x) ? '.exe' : '';
```
anjohnson, line 33: Without cygwin this could be just

```perl
my $exe = ($^O eq 'MSWin32') ? '.exe' : '';
```

anjohnson (commented):
Looks good for the most part, 3 questions/comments below.

modules/libcom/src/valgrind/valgrind.h
```diff
@@ -129,7 +129,6 @@
 #elif defined(__APPLE__) && defined(__x86_64__)
 #  define PLAT_amd64_darwin 1
 #elif (defined(__MINGW32__) && !defined(__MINGW64__)) \
-      || defined(__CYGWIN32__) \
```
anjohnson, line 132: This is externally sourced code, please remove the change to this header.

PR #832, opened 2026-03-04
add simple calculations to macros

dirk-zimoch:
This is the "calculating macros" patch I use at PSI a for quite a while.
It supports integer arithmetic `+-*/%()`, comparisons `<>`(no `==` unfortunately because of how macLib parses `=`) and conditionals `?:`. Additionally, it supports printf-style integer formatting of the result with a `%` prefix e.g. `%#04x`.

Example: `${%#010x 0x00100000+($(CARD_NO))*0x1000}`
Note the `()`around the inner macro in case the macro contains an addition like `2+3` and the outer `{}` necessary to allow inner `()` to be parsed correctly by macLib.

I had planned for a long time to modify macLib further so it can better handle embedded `=`and `()` and maybe also to add floating point arithmetic and string manipulation, but never found the time and never had the urgent need.

### DIFF
--- a/modules/libcom/src/macLib/macCore.c
+++ b/modules/libcom/src/macLib/macCore.c
@@ -568,6 +568,186 @@ static MAC_ENTRY *create( MAC_HANDLE *handle, const char *name, int special )
     return entry;
 }
 
+#define SUCCESS 0
+#define NOVALUE 1
+#define UNTERMINATED 2
+#define DIVZERO 3
+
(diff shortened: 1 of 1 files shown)


mdavidsaver:
This syntax looks more like what I had imagined.  The integration of sprintf is clever.  I'm a bit nervous about a user constructed printf format string, but that might be mitigated with some validation.  Maybe shared with printfRecord?

Test coverage would also need to be added.

PR #831, opened 2026-03-03
Add run-time debug switch to caRepeater

anjohnson:
Convert the compile-time repeater `DEBUG` macro into a command-line flag `-d`.
Only works for the stand-alone `caRepeater` program.
Added extra debug messages for active client verification.

This is to help with the review of PR #802.

With -d the repeater now shows client events, e.g.

```
  $ caRepeater -d
  CA Repeater: Attached and initialized
  New client on port 60307
  Verified 1 active clients
  New client on port 52541
  Verified 2 active clients
  Client on port 60307 refused message
  Deleted client on port 60307
  Client on port 52541 refused message
  Deleted client on port 52541
  ^C
```

Running "caRepeater -dd" also shows beacon forwarding and individual client verification:

```
  $ caRepeater -dd
  CA Repeater: Attached and initialized
  New client on port 49968
  Client on port 49968 is alive
  Verified 1 active clients
  Sent to port 49968
  Sent to port 49968
  Sent to port 49968
  New client on port 65377
  Sent to port 49968
  Client on port 49968 is alive
  Client on port 65377 is alive
  Verified 2 active clients
  Sent to port 49968
  Sent to port 65377
  Sent to port 49968
  Sent to port 65377
  Client on port 49968 refused message
  Deleted client on port 49968
  Sent to port 65377
  Sent to port 65377
  Client on port 65377 refused message
  Deleted client on port 65377
  ^C
```

PR #833, opened 2026-03-06
Fix EOL whitespace

simon-ess:
Trivial fix to whitespace at EOL breaking editorconfig job.


anjohnson:
I committed the identical fix in a last-minute addition to #831 which I just merged. I'd have merged this PR instead if I'd seen it.

PR #834, opened 2026-03-09
test: macLib unbalanced {}()

mdavidsaver:
See #796.  Test macLib handling of unbalanced `{}` or `()`.

Demonstrated one minor issue.  `${FOO)`, with mismatched open and close, expands to `$(FOO),undefined)` with unbalanced parenthesis.

### DIFF
--- a/modules/libcom/test/macLibTest.c
+++ b/modules/libcom/test/macLibTest.c
@@ -66,13 +66,19 @@ static void ovcheck(void)
 
 MAIN(macLibTest)
 {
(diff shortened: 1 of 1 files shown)


jerzyjamroz:
Ok, so the bug has been found.

anjohnson:
I agree that there's a bug in macLib, but I don't agree with the expected expansions of the test strings added here. See #837 for my fix with different tests.

mdavidsaver:
Superseded by #837.

PR #837, opened 2026-03-16
Fix 834

anjohnson:
I agree that macLib should detect mismatched macro delimiters, but it shouldn't modify the string assuming the other delimiter was intended. This PR fixes that, and adds tests for some additional cases to ensure it doesn't break legitimate cases where delimiters are used.


jerzyjamroz:
This example:
```
{(1+2} == $(1+2 == $(1+2
{((1+2)} == $((1+2) == $((1+2)
```
works fine with this patch.

PR #838, opened 2026-03-17
Fix memory overwrite in macParseDefns

FreddieAkeroyd:
With `windows-x64-debug` a corrupted memory block was detected when parsing some substitution files, this fixes the issue that was caused by the writing past end of allocated array

### DIFF
--- a/modules/libcom/src/macLib/macUtil.c
+++ b/modules/libcom/src/macLib/macUtil.c
@@ -17,12 +17,10 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <limits.h>
 
 #include "dbDefs.h"
 #include "errlog.h"
 #include "macLib.h"
-#include "epicsString.h"
 
 /*
  * Parse macros definitions in "a=xxx,b=yyy" format and convert them to
@@ -68,7 +66,7 @@ epicsStdCall macParseDefns(
        for defns="1,3,5" give 3 pairs so num will be 6 (strlen(defns)+1)
        but as num is used as an index e.g. del[num] we need
        numMax to be strlen(defns) + 2 */
-    numMax = epicsStrnLen( defns, INT_MAX ) + 2;
+    numMax = strlen( defns ) + 2;
     if ( numMax < altNumMax )
         numMax = altNumMax;
     ptr = (const char **) calloc( numMax, sizeof( char * ) );
--- a/modules/libcom/src/macLib/macUtil.c
+++ b/modules/libcom/src/macLib/macUtil.c
@@ -66,7 +66,7 @@ epicsStdCall macParseDefns(
        for defns="1,3,5" give 3 pairs so num will be 6 (strlen(defns)+1)
        but as num is used as an index e.g. del[num] we need
        numMax to be strlen(defns) + 2 */
-    numMax = strlen( defns ) + 2;
+    numMax = strlen( defns ) + 2; /* Flawfinder: ignore */
     if ( numMax < altNumMax )
         numMax = altNumMax;
     ptr = (const char **) calloc( numMax, sizeof( char * ) );
--- a/modules/libcom/src/macLib/macUtil.c
+++ b/modules/libcom/src/macLib/macUtil.c
@@ -17,10 +17,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <limits.h>
 
 #include "dbDefs.h"
 #include "errlog.h"
 #include "macLib.h"
+#include "epicsString.h"
 
 /*
  * Parse macros definitions in "a=xxx,b=yyy" format and convert them to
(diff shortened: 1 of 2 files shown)


ralphlange (approved):
LGTM

modules/libcom/src/macLib/macUtil.c
```diff
@@ -62,8 +64,11 @@ epicsStdCall macParseDefns(
         printf( "macParseDefns( %s )\n", defns );
 
     /* allocate temporary pointer arrays; in worst case they need to have
-       as many entries as the length of the defns string */
-    numMax = strlen( defns );
+       as many entries as the length of the defns string + 2
+       for defns="1,3,5" give 3 pairs so num will be 6 (strlen(defns)+1)
+       but as num is used as an index e.g. del[num] we need
+       numMax to be strlen(defns) + 2 */
+    numMax = epicsStrnLen( defns, INT_MAX ) + 2;
```
mdavidsaver: I can follow the `+2`, but why `epicsStrnLen(..., INT_MAX)`?  The caller must provide an unterminated string, or we are off into Undefined Behavior.

mdavidsaver:
A test case to add.

```diff
--- a/modules/libcom/test/macDefExpandTest.c
+++ b/modules/libcom/test/macDefExpandTest.c
@@ -61,6 +61,25 @@ static void macEnvScope(void)
     checkMac(handle, "E", "15");
     checkMac(handle, "F", NULL);
 
+    {
+        // test worst case for short definitions
+        // https://github.com/epics-base/epics-base/pull/838
+        macParseDefns(NULL, "1,3,5", &defines); // smoke
+
+#define CASE(I,V) \
+        testOk(defines[I] && strcmp(defines[I], V)==0, "defines[%u] \"%s\" == \"" V "\"", I, defines[I])
+        CASE(0,"1");
+        testOk1(defines[1]==NULL);
+        CASE(2,"3");
+        testOk1(defines[3]==NULL);
+        CASE(4,"5");
+        testOk1(defines[5]==NULL);
+        testOk1(defines[6]==NULL);
+#undef CASE
+
+        free(defines);
+        defines = NULL;
+    }
     {
         macPushScope(handle);
```

modules/libcom/src/macLib/macUtil.c
ralphlange: Isn't that the maximum length that is searched?
(IIRC, to avoid strlen's "may overread" warning.)

mdavidsaver: Yes, the second argument of `epicsStrnLen()` is that limit.  My question is why `INT_MAX`?  To avoid UB this limit should be less than or equal to the allocation size.  However, in this instance the caller allocates, so `macParseDefns()` has no knowledge beyond the presence of a null termination.

I suspect this is a holdover from the troubleshooting process, which can now be switched to `strlen(defns)+2`

FreddieAkeroyd: I got a check failure with `strlen()` so changed to `epicsStrnLen()` to clear this - is there a preferred approach to clearing such errors e.g. a hint to the checker that the string is NULL terminated already? `INT_MAX` was arbitrary - I needed some large number I could add 2 to that would still fit into a `size_t`

FreddieAkeroyd:
Thank you @mdavidsaver  - test added

modules/libcom/src/macLib/macUtil.c
FreddieAkeroyd: I agree that what i have done with `epicsStrnLen` just avoids a check failure as opposed to adds any additional security in that you can still read invalid memory after end of buffer as you mention, and it hsa no way to know what end of buffer is without us adding such an argument to  `macParseDefns`

mdavidsaver: > I got a check failure with `strlen()` so changed to `epicsStrnLen()` to clear this

So your switch from `strlen()` to `epicsStrnLen()` was in response to a static analyzer complaint?  Was this specifically about usage of `strlen()`?


> e.g. a hint to the checker that the string is NULL terminated already?

If this turns out to be another false positive from codacy, then I'm am inclined to disable these checks.

FreddieAkeroyd: Have reverted to `strlen()` and codacy is giving

image

FreddieAkeroyd: I can silence it with ` /* Flawfinder: ignore */`.   Codacy supports several linters, maybe one of the others is less false positive prone?

mdavidsaver: imo. this "critical" message is not even a false positive.  It is unactionable paranoia.

> I can silence it with ` /* Flawfinder: ignore */`. Codacy supports several linters, maybe one of the others is less false positive prone?

Please don't add these sorts of linter "command" comments.  We will just ignore codacy.

@anjohnson How do we turn down the noise level on Codacy?

anjohnson:
Core Group: @mdavidsaver to merge, without the flawfinder tag.

mdavidsaver:
Applied as of acf53d31eebfcce694c093141cf0ba6d88189263

Issue #839, opened 2026-03-18
Random calc expression is not random


daykin:
EPICS Version: 7.0.9

EPICS_HOST_ARCH: linux-x86_64

OS: Debian 13

the PRNG for the RNDM calc expression returns to a known state between IOC restarts. I would only expect this to be the case if a particular seed is specified.

 To reproduce, create the following test.db:
```
record(calc, "RND_TEST")
{
    field(CALC, "RNDM")
    field(PINI, "YES")
}

```
then load it in softIoc twice:

```
daykin@diag-cam:~/git/testioc/testApp/Db$ softIoc -D /usr/lib/epics/dbd/softIoc.dbd -d test.db
Starting iocInit
############################################################################
## EPICS R7.0.9-3+0~20250716123251.15+debian13~1.gbpc159bf
## Rev. 2025-07-16T12:58+0000
## Rev. Date build date/time: 
############################################################################
iocRun: All initialization complete
epics> dbgf RND_TEST
DBF_DOUBLE:         0.750072480354 
epics> exit

daykin@diag-cam:~/git/testioc/testApp/Db$ softIoc -D /usr/lib/epics/dbd/softIoc.dbd -d test.db
Starting iocInit
############################################################################
## EPICS R7.0.9-3+0~20250716123251.15+debian13~1.gbpc159bf
## Rev. 2025-07-16T12:58+0000
## Rev. Date build date/time: 
############################################################################
iocRun: All initialization complete
epics> dbgf RND_TEST
DBF_DOUBLE:         0.750072480354
```

mdavidsaver:
Wow.  I guess a data race would add randomness as well?

https://github.com/epics-base/epics-base/blob/2d8a6f91a902b4cc0e8d9e11accb070275dd98fe/modules/libcom/src/calc/calcPerform.c#L508-L523

More seriously, I guess that `srand()` and `rand()` were/are not available on vxworks?  The linux manpage lists these as `C11, POSIX.1-2008`.

mdavidsaver:
> More seriously, I guess that `srand()` and `rand()` were/are not available on vxworks? The linux manpage lists these as `C11, POSIX.1-2008`.

These functions are called from `epicsMessageQueueTest()`, which is part of the RTOS test harness (`epicsRunLibComTests()`).  Which I guess means that they are available on all currently supported targets.

daykin:
[...]

PR #840, opened 2026-03-18
use (s)rand() to generate RNDM calc values

daykin:
Re: #839 

all supported platforms now include `(s)rand()` for random number generation. seed with `srand()` using the system clock and generate from there using `rand()`.

### DIFF
--- a/modules/libcom/src/calc/calcPerform.c
+++ b/modules/libcom/src/calc/calcPerform.c
@@ -20,6 +20,7 @@
 #include "osiUnistd.h"
 #include "dbDefs.h"
 #include "epicsMath.h"
+#include "epicsTime.h"
 #include "epicsTypes.h"
 #include "errlog.h"
 #include "postfix.h"
@@ -505,22 +506,18 @@ calcArgUsage(const char *pinst, unsigned long *pinputs, unsigned long *pstores)
     return 0;
 }
 
-/* Generate a random number between 0 and 1 using the algorithm
- * seed = (multy * seed) + addy         Random Number Generator by Knuth
- *                                              SemiNumerical Algorithms
- *                                              Chapter 1
- * randy = seed / 65535.0          To normalize the number between 0 - 1
- */
-static unsigned short seed = 0xa3bf;
-static unsigned short multy = 191 * 8 + 5;  /* 191 % 8 == 5 */
-static unsigned short addy = 0x3141;
-
 static double calcRandom(void)
 {
-    seed = (seed * multy) + addy;
-
+    static int seeded = 0;
+    static struct epicsTimeStamp ts;
+    if(!seeded)
+    {
+        epicsTimeGetMonotonic(&ts);
+        srand((unsigned) ts.nsec);
+        seeded = 1;
+    }
     /* between 0 - 1 */
-    return (double) seed / 65535.0;
+    return (double) rand() / RAND_MAX;
 }
 
 /* Search the instruction stream for a matching operator, skipping any


modules/libcom/src/calc/calcPerform.c
```diff
@@ -505,22 +506,19 @@ calcArgUsage(const char *pinst, unsigned long *pinputs, unsigned long *pstores)
...
 
 static double calcRandom(void)
 {
-    seed = (seed * multy) + addy;
-
+    if(!seeded)
+    {
+        clock_gettime(CLOCK_REALTIME, &ts);
+        srand((unsigned) (ts.tv_nsec % RAND_MAX));
+        seeded = 1;
```
mdavidsaver, line 517: Use of `epicsTimeGetCurrent()` is preferable.

Also, as a minor point, the `seeded` flag could be added as a static local variable.  Same semantics, just with a smaller scope.

anjohnson: Using `epicsTimeGetMonotonic()` should be better than `…Currrent()`, I believe it can tick faster on embedded OSs.

modules/libcom/src/calc/calcPerform.c
```diff
@@ -505,22 +506,18 @@ calcArgUsage(const char *pinst, unsigned long *pinputs, unsigned long *pstores)
...
 static double calcRandom(void)
 {
-    seed = (seed * multy) + addy;
-
+    static int seeded = 0;
+    static struct epicsTimeStamp ts;
+    if(!seeded)
+    {
+        epicsTimeGetCurrent(&ts);
+        srand((unsigned) (ts.nsec % RAND_MAX));
```
anjohnson: Here are a series of full-resolution timestamps from the heartbeat records on 3 different VxWorks IOCs, which process in the 1 second scan thread. Compare the bottom 3 digits of the timestamps, which come from `epicsTimeGetCurrent()`:

```
ioclid1:heartbeat        2026-03-18 17:08:18.460333459 5.38474e+06  
ioclid3:heartbeat        2026-03-18 17:08:18.539125459 5.46948e+06  
ioclid2:heartbeat        2026-03-18 17:08:18.760380459 5.46974e+06  
ioclid1:heartbeat        2026-03-18 17:08:19.443702459 5.38474e+06  
ioclid3:heartbeat        2026-03-18 17:08:19.522365459 5.46948e+06  
ioclid2:heartbeat        2026-03-18 17:08:19.743754459 5.46974e+06  
ioclid1:heartbeat        2026-03-18 17:08:20.427038459 5.38474e+06  
ioclid3:heartbeat        2026-03-18 17:08:20.505749459 5.46948e+06  
ioclid2:heartbeat        2026-03-18 17:08:20.727155459 5.46974e+06  
ioclid1:heartbeat        2026-03-18 17:08:21.410446459 5.38474e+06  
ioclid3:heartbeat        2026-03-18 17:08:21.489088459 5.46948e+06  
ioclid2:heartbeat        2026-03-18 17:08:21.710474459 5.46974e+06  
ioclid1:heartbeat        2026-03-18 17:08:22.393786459 5.38474e+06  
ioclid3:heartbeat        2026-03-18 17:08:22.472643459 5.46948e+06  
ioclid2:heartbeat        2026-03-18 17:08:22.693871459 5.46974e+06
```

That's why I suggested using `epicsTimeGetMonotonic()` instead, that doesn't have this issue.

daykin: Thank you for the suggestion, I have changed it accordingly.

modules/libcom/src/calc/calcPerform.c
```diff
@@ -505,22 +506,18 @@ calcArgUsage(const char *pinst, unsigned long *pinputs, unsigned long *pstores)
...
 static double calcRandom(void)
 {
-    seed = (seed * multy) + addy;
-
+    static int seeded = 0;
+    static struct epicsTimeStamp ts;
+    if(!seeded)
+    {
+        epicsTimeGetMonotonic(&ts);
+        srand((unsigned) (ts.nsec % RAND_MAX));
```
mdavidsaver: Why `% RAND_MAX` for the input to `srand()`?  All the usages of `srand()` I have seen pass in the full range of `unsigned int`.  eg. [`man srand`](https://linux.die.net/man/3/srand).

daykin: My exact thought process:

1. POSIX `srand()` takes an unsigned
2. I was not sure if the type of `RAND_MAX`  is implementation specific
3. This section is not continuously executing, and the compiler may even optimize it out, so it isn't actively hurting much

If, however, `RAND_MAX` and EPICS timestamp nsec are always unsigned on all supported platforms, it can be removed.

anjohnson:
I knew there'd been a discussion about this somewhere recently, I just found it on the EPICS Chat server in the EPICS Random room (adding it here only for completeness):

image

modules/libcom/src/calc/calcPerform.c
mdavidsaver: > 2\. I was not sure if the type of `RAND_MAX`  is implementation specific

I think you are making a association here where none exists.  As documented, `RAND_MAX` is not relevant to the argument of `srand()`.  Only to the return value of `rand()`.

> 3\. This section is not continuously executing, and the compiler may even optimize it out, so it isn't actively hurting much

It is restricting the range of the seed, which will effect the output of the PRNG.

daykin: >It is restricting the range of the seed, which will effect the output of the PRNG.

My concern is that depending on the type of PRNG, an out-of-bounds seed could create unanticipated results. More of a case of not knowing what I don't know. But, just did some brief googling... ANSI `rand()` is just an LCG with modulo 2^31. So the effective behavior of some hypothetical seed bigger than m would just degenerate implicitly to `s mod m` regardless. I'll get rid of the `% RAND_MAX`.

anjohnson (approved):
Looks good to me, thanks!

PR #841, opened 2026-03-19
Use deterministic timezones in epicsTimeZoneTest

ericonr:
The DST characteristics of the POSIX EST5EDT and CST6CDT timezones depend on whether the implementation uses the built-in POSIX definition or reads an updated version from the tzdata files, and for POSIX timezones, there is no specified behavior [1].

To avoid any plaform-specific behavior, we can specify the desired DST ourselves. We used the timezone values available in configure/CONFIG_SITE_ENV, which already included DST information, and means the tests are consistent with our own usage.

[1] https://www.openwall.com/lists/musl/2024/03/05/2

---

This is a blocker for running tests on musl in #413

### DIFF
--- a/modules/libcom/test/epicsTimeZoneTest.c
+++ b/modules/libcom/test/epicsTimeZoneTest.c
@@ -89,10 +89,10 @@ MAIN(epicsTimeZoneTest)
      *  Mon Oct 19 13:00:16 2015 UTC
      */
     testDiag("POSIX 1445259616");
-    setTZ("EST5EDT");
+    setTZ("EST5EDT,M3.2.0/2,M11.1.0/2");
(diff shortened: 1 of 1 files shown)


ericonr:
The appVeyor build failure is unrelated to my changes, and seems to be caused by something non-deterministic in `regressTest`:

```
regressTest.tap ........... 
not ok 70 - dbGetField("li2", DBR_LONG) -> 3 == 0
Failed 1/70 subtests 
```

Should I open an issue about this?

mdavidsaver:
> Should I open an issue about this?

Better add to https://github.com/epics-base/epics-base/discussions/162#discussioncomment-1460745 .

mdavidsaver:
As long as the test continues to pass, I see no issue with this PR.  Encoding a historical definition of DST dates seems reasonable when they are only applied to historical dates.

PR #845, opened 2026-03-25
Check build arch's are known before creating build dirs

anjohnson:
Discussed in [this thread](https://epics.anl.gov/tech-talk/2026/msg00448.php).

### DIFF
--- a/configure/RULES_ARCHS
+++ b/configure/RULES_ARCHS
@@ -64,6 +64,9 @@ $(BUILD_ARCHS) : % : O.% O.Common
 $(ACTIONS) : % : $(foreach arch, $(BUILD_ARCHS), %$(DIVIDER)$(arch))
 
 $(buildDirs):
+	$(if $(wildcard $(CONFIG)/os/CONFIG.Common.$(archPart) \
+	        $(CONFIG)/os/CONFIG.$(EPICS_HOST_ARCH).$(archPart)),, \
+	    $(error Base not configured to build target $(archPart)))
 	$(PERL) $(TOOLS)/makeMakefile.pl $@ $(TOP)/..
 
 O.Common:


anjohnson:
… and it breaks our Windows MSVC CI builds, e.g.:
```
▾ Build the main module
  make -C ./configure install
  ../configure/RULES_ARCHS:67: *** Base not configured to build target windows-x64.  Stop.
  make[1]: Entering directory 'D:/a/epics-base/epics-base/configure'
  make[1]: Leaving directory 'D:/a/epics-base/epics-base/configure'
  make: *** [configure/RULES_DIRS:85: configure.install] Error 2
  Error: Process completed with exit code 1.
```

I was wrong about the file `$(CONFIG)/os/CONFIG.Common.$(T_A)`, it isn't required, only `$(CONFIG)/os/CONFIG.$(EPICS_HOST_ARCH).Common` is essential. Bummer.

Would there be objections to my creating those missing files and making them required? That would obviously require a Release Note warning sites they would need to add them for their local architectures too.

mdavidsaver:
I am in favor of typo detection.

> file `$(CONFIG)/os/CONFIG.Common.$(T_A)`

> Would there be objections to my creating those missing files and making them required?

Not from me.

mdavidsaver:
As an alternative, is there some make variable to test which is only and always set from a `configure/os/*` fragment.  eg. OS_CLASS ?

```
$ git grep 'OS_CLASS\s*='
configure/CONFIG_ADDONS:#               USR_CFLAGS_vxWorks = <defines for all OS_CLASS=vxWorks target systems>
configure/os/CONFIG.Common.RTEMS:OS_CLASS = RTEMS
configure/os/CONFIG.Common.cygwin-x86:OS_CLASS = cygwin32
configure/os/CONFIG.Common.freebsdCommon:OS_CLASS = freebsd
configure/os/CONFIG.Common.iosCommon:OS_CLASS = iOS
configure/os/CONFIG.Common.linuxCommon:OS_CLASS = Linux
configure/os/CONFIG.Common.solaris-sparc:OS_CLASS = solaris
configure/os/CONFIG.Common.solaris-x86:OS_CLASS = solaris
configure/os/CONFIG.Common.vxWorksCommon:OS_CLASS = vxWorks
configure/os/CONFIG.Common.win32-x86-mingw:OS_CLASS = WIN32
configure/os/CONFIG.darwinCommon.darwinCommon:OS_CLASS = Darwin
configure/os/CONFIG.win32-x86.win32-x86:OS_CLASS=WIN32
modules/database/src/tools/epics-base-arch.pc@:OS_CLASS=@OS_CLASS@
modules/database/src/tools/epics-base.pc@:OS_CLASS=@OS_CLASS@
```

anjohnson:
> is there some make variable to test … eg. OS_CLASS ?

Unfortunately that can't work, the `$(CONFIG)/os/CONFIG.….$(T_A)` files are only pulled in when we're building inside an `O.<arch>` build directory. The check I added here happens just before we create the build directory, and I'd much prefer that we don't create it at all for unknown architectures.

However the Windows targets do all have a `$(CONFIG)/os/CONFIG.$(EPICS_HOST_ARCH).$(T_A)` file, so this next commit which checks for that too might pass…

mdavidsaver:
> The check I added here happens just before we create the build directory, and I'd much prefer that we don't create it at all for unknown architectures.

To my mind, reporting an error is important.  Avoiding leaving "debris" which `make clean` will remove seems secondary to me.  I am not sure how "debris" should weighed against potential breakage of external targets (a loud and clear breakage at that).

anjohnson:
`make clean` doesn't delete `O.*` directories for targets that this host isn't configured to build, although `make distclean` does.

My latest commit does solve the problem properly though. The regressTest is back to failing frequently on Appveyor.

Issue #847, opened 2026-03-31
Hand-typed iocshLoad commands aren't added to the shell history record for later re-use


anjohnson:
To replicate: Type `iocshLoad /dev/null` into a softIoc IOC shell (which does nothing), then press the up-arrow to recall the previous command.

ralphlange:
Isn't that the correct behavior?

When loading another file, I would expect all commands from the loaded file to appear, but the load command itself?
Might have to be commented-out or marked in the history to not give a wrong impression.

(Assuming that the history file should contain the list of commands that have been executed. Including the load command and the loaded commands would execute them twice if you replayed the history file.)

anjohnson:
We disabled adding commands to the shell history that were read from a file in commit 057eb87101f201a76fa00af344734a7e024ddc9d, but the `iocshLoad` command itself wasn't read from a file (hence the "hand-typed" qualifier in my Subject line). Why would it not be saved in the history?

mdavidsaver:
> We disabled adding commands to the shell history that were read from a file in commit [057eb87](https://github.com/epics-base/epics-base/commit/057eb87101f201a76fa00af344734a7e024ddc9d) ...

This was my first thought as well, but not so.  I can see `add_history("iocshLoad /dev/null")` being called.

Instead the problem is that the history file is then re-read when the sub-shell evaluates, loading a whole second copy of the history afterwards.  History search will find it.

mdavidsaver:
Candidate fix https://github.com/epics-base/epics-base/pull/848

anjohnson:
Weird, deleting the `.iocsh_history` file seemed to clear it up, but the new file contents after a second test had some extra lines, which looked to be related to running the iocshLoad command. Your speedy analysis seems to make sense, thanks. I'll give #848 a try and report back.

mdavidsaver:
> Isn't that the correct behavior?

Appending a second copy of the history file is certainly not correct.

> When loading another file, I would expect all commands from the loaded file to appear, but the load command itself?

Otherwise, I think that the current behavior is as I expect.  An interactive command `iocshLoad other.cmd` should appear in history, but the contents of `other.cmd` should not.

ralphlange:
> Otherwise, I think that the current behavior is as I expect. An interactive command `iocshLoad other.cmd` should appear in history, but the contents of `other.cmd` should not.

Ok, so the other way round: load command appears, loaded file does not.
Fine with me, I was just against both the load command and the commands from the file appearing.

PR #848, opened 2026-03-31
Skip readline interactive setup when not interactive

mdavidsaver:
Fix incorrect re-initialization of readline for recursive shell (#847).  Only add the hooks for interactive iocsh for actual interactive sessions (no `FILE*`).


anjohnson (approved):
Works for me, solves the problem I was seeing.

Thanks!

mdavidsaver:
I have added a second commit which will `clear_history()` before `read_history()`.  I had not realized that `read_history()` appends to the global history array.  This will only have effect in a case where eg. some code calls `iocsh(NULL)` multiple times.

PR #849, opened 2026-04-02
Minor doc and strings

anjohnson:
Reformatted the ReleaseChecklist, other minor stuff

### DIFF
--- a/documentation/Makefile.headers
+++ b/documentation/Makefile.headers
@@ -71,7 +71,6 @@ database_HEADERS += chfPlugin
 database_HEADERS += dbBase
 database_HEADERS += dbChannel
 database_HEADERS += dbCommon
-database_HEADERS += dbDefs
 database_HEADERS += dbEvent
 database_HEADERS += dbExtractArray
 database_HEADERS += dbLink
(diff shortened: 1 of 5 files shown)


anjohnson:
Adjusted the iocInit error message, I was apparently thinking the IOC would be started even if asInit() fails, this makes it clear it doesn't.

I also disabled the failing Codacy rule.

ralphlange:
> I also disabled the failing Codacy rule.

Globally?
Does it fail in every instance? Otherwise, marking as false positive would be more appropriate.

anjohnson:
> Globally?

Yes. The [markdownlint rule that failed](https://github.com/DavidAnson/markdownlint/blob/main/doc/md040.md) says that every fenced markdown code-block (inside triple back-tick lines) should have a language specified. We have 70 such code-blocks under the documentation directory (85 with this PR) and only 36 of them have a language specified. Quite a few of those aren't actually code in any programming language. It looks like those could be marked as "text", but the language indicator is supposed to be optional, and different tools support different languages anyway.

Thus I don't think we should insist on this particular rule.

ralphlange:
Fine.

Personally, I like syntax highlighting, i.e. language indicators, so I would have graded it down to a warning, but removing also works.

anjohnson:
I like syntax highlighting too, but I'm not sure how to downgrade this from a felony to a misdemeanor. Codacy reported it thusly in the email:

image

Feel free to fix the global setting if you can make it appear less severe than that (a suggestion maybe?).

mdavidsaver (approved):
No objections.

PR #852, opened 2026-04-07
Potential fix for test 70 in regressTest

anjohnson:
Bug fix found by Claude Opus 4.6

Doesn't fully resolve the TODO on test 69, it wasn't asked to do that.

### DIFF
--- a/modules/database/src/ioc/db/dbCa.c
+++ b/modules/database/src/ioc/db/dbCa.c
@@ -830,6 +830,27 @@ static lset dbCa_lset = {
     scanForward, doLocked
 };
 
+/* Check if all deferred init work for a local CA link is complete.
+ * Must be called with pca->lock held.
+ *
+ * During connectionCallback, CA_INIT_READY may be deferred to both
+ * eventCallback (for native/string data) and getAttribEventCallback
+ * (for attributes). Each callback independently re-adds CA_INIT_READY,
+ * but only the first to be processed by the worker thread takes effect.
+ * This check ensures CA_INIT_READY is only signaled once ALL deferred
+ * work is complete.
+ */
+static int dbCaLinkInitComplete(caLink *pca, struct link *plink)
+{
+    if ((plink->value.pv_link.pvlMask & pvlOptInpNative) && !pca->gotInNative)
+        return 0;
+    if ((plink->value.pv_link.pvlMask & pvlOptInpString) && !pca->gotInString)
+        return 0;
+    if (pca->dbrType != DBR_STRING && !pca->gotAttributes)
+        return 0;
+    return 1;
+}
+
 static void connectionCallback(struct connection_handler_args arg)
 {
     caLink *pca;
@@ -996,7 +1017,8 @@ static void eventCallback(struct event_handler_args arg)
             link_action |= CA_DBPROCESS;
         }
     }
-    if (pca->flags & DBCA_CALLBACK_INIT_WAIT)
+    if ((pca->flags & DBCA_CALLBACK_INIT_WAIT) &&
+            dbCaLinkInitComplete(pca, plink))
         link_action |= CA_INIT_READY;
 done:
     if (link_action) addAction(pca, link_action);
@@ -1123,7 +1145,8 @@ static void getAttribEventCallback(struct event_handler_args arg)
     pca->alarmLimits[3] = pdbr->upper_alarm_limit;
     pca->precision = pdbr->precision;
     memcpy(pca->units, pdbr->units, MAX_UNITS_SIZE);
-    if (pca->flags & DBCA_CALLBACK_INIT_WAIT)
+    if ((pca->flags & DBCA_CALLBACK_INIT_WAIT) &&
+            dbCaLinkInitComplete(pca, plink))
         addAction(pca, CA_INIT_READY);
     epicsMutexUnlock(pca->lock);
     if (getAttributes) getAttributes(getAttributesPvt);


modules/database/test/std/rec/regressTest.c
```diff
@@ -196,6 +197,10 @@ void testLinkSevr(void)
     testOk1(dbIsLinkConnected(dbGetDevLink(testdbRecordPtr("si2"))));
     testOk1(dbIsLinkConnected(dbGetDevLink(testdbRecordPtr("li2"))));
 
+    /* Wait for CA links to receive initial monitor data */
+    testdbCaWaitForUpdateCount(dbGetDevLink(testdbRecordPtr("si2")), 1);
+    testdbCaWaitForUpdateCount(dbGetDevLink(testdbRecordPtr("li2")), 1);
```
mdavidsaver: This is a partial revert of https://github.com/epics-base/epics-base/pull/768.  If this wait is necessary, it is a symptom of a bug with that PR.

mdavidsaver (changes requested):

modules/database/test/std/rec/regressTest.c
anjohnson: Ack. This is how Claude described the race condition it found:

----

Race condition between CA link connection and initial data delivery.

CA links go through multiple stages:
1. Connection — `isConnected` flag set by `connectionCallback()`    
2. Monitor subscription — queued to worker thread after connection
3. Initial data delivery — `gotInNative`/`gotInString` flags set by `eventCallback()`

The test at line 196-197 checks `dbIsLinkConnected()`, which only verifies stage 1. But `dbCaGetLink()` (called when the record processes) requires stage 3 — if `gotInNative` is still `FALSE`, it returns -1 and the record's `VAL` is not updated, leaving it at the default value of 0.

The time window between stages 1 and 3 is non-deterministic. On macOS it typically resolves before the test proceeds; on Windows CI under load, it often doesn't.     

Test 69 (si2, stringin via CA) has the same issue and is already marked TODO "Not working" — but that appears to be a separate deeper problem with string conversion through CA links for SEVR fields, not just timing.

anjohnson: I asked it

> `startRegressTestIoc()` doesn't return until `testIocInitOk()` completes, doesn't that wait for the initial data delivery stage for all CA links?

The response was:

----

You're right that `iocInit()` is designed to wait. Here's the detailed mechanism and where it breaks down:

  The intended sync mechanism:
  1. Local CA links are created with `DBCA_CALLBACK_INIT_WAIT` flag, incrementing `initOutstanding`
  2. `dbCaRun()` (called from `iocRun()`) blocks until `initOutstanding` reaches 0
  3. `CA_INIT_READY` processing decrements `initOutstanding` and unblocks `dbCaRun()`

  The race is in `connectionCallback` (lines 862-914). For li2 (longin reading ai.SEVR, type `DBR_ENUM`):
  - Line 863: `CA_INIT_READY` is initially set
  - Line 898: `CA_INIT_READY` is cleared because `pvlOptInpNative` is set → deferred to `eventCallback`
  - Line 914: `CA_INIT_READY` is cleared again because `dbrType != DBR_STRING` → deferred to `getAttribEventCallback`

  Then both callbacks independently re-add `CA_INIT_READY`:
  - `eventCallback` line 999-1000: sets `gotInNative = TRUE`, adds `CA_INIT_READY`
  - `getAttribEventCallback` line 1126-1127: sets `gotAttributes = TRUE`, adds `CA_INIT_READY`

  But in the worker thread (line 1298-1299), the first `CA_INIT_READY` to arrive clears `DBCA_CALLBACK_INIT_WAIT`, and the second is a no-op.

  If `getAttribEventCallback` fires first (the `ca_get` for control limits/precision/units), then `CA_INIT_READY` is processed, `initOutstanding` decremented, `dbCaRun()` unblocks — but `gotInNative` is still `FALSE`. When li2 later processes and calls `dbCaGetLink()`, it sees `gotInNative == FALSE` and returns -1, leaving VAL at 0.

  So the `testdbCaWaitForUpdateCount` fix is correct as a test-level workaround — it waits specifically for `nUpdate >= 1` (set in `eventCallback` at line 941), which guarantees `gotInNative == TRUE`. But the deeper issue is that the `DBCA_CALLBACK_INIT_WAIT` mechanism allows `getAttribEventCallback` to complete the init before data arrives.

mdavidsaver: > The response was:

@anjohnson Have you verified the correctness and relevance of this output?

anjohnson:
Replacement fix, adjusting dbCa.c instead to wait for all deferred init conditions.

mdavidsaver:
Is it intended that both `pca->evidString && pca->evidNative` be subscribed simultaneously?  I see this does happen.  Seems like this would result in (almost) duplicate updates in an uncertain order.  Maybe  better if `CA_MONITOR_NATIVE` cancelled `evidString`, and `CA_MONITOR_STRING` cancelled `evidNative`.

mdavidsaver:
Ok.  Bottom line, the three conditions in `connectionCallback()` which clear `CA_INIT_READY` are not mutually exclusive.  As far as I can tell this evening, they could all match.

Since the state of `caLink` is mutated in some many places, I think the approach should be to change the one bit `DBCA_CALLBACK_INIT_WAIT` flag into 3 bits for the three callbacks which might be waited.  In `eventCallback()`, `arg.chid` could be used to distinguish between the two concurrent subscriptions (unless we can agree that this is also a bug).

mdavidsaver:
I do not think this change correctly distinguishes the two possible subscription updates.  I am hesitant to depend on existing `caLink` state as I am not certain how it could be changed by `lset` functions during initial processing.

See #856

PR #853, opened 2026-04-07
Configure Static IP with RTEMS libbsd

brendanchandler:
These changes allow epics to use RTEMS with libbsd to configure a static IP.  If no static IP configuration is in nvram, it falls back to trying DHCP.  

The changes also begin separating the BSP specific querying of network information from the actual initialization/setup of the network, with the hope that we can separate the two completely in the future soon.

### DIFF
--- a/modules/libcom/RTEMS/posix/rtems_config.c
+++ b/modules/libcom/RTEMS/posix/rtems_config.c
@@ -64,7 +64,6 @@ extern void *POSIX_Init(void *argument);
 #define CONFIGURE_APPLICATION_NEEDS_CONSOLE_DRIVER
 #define CONFIGURE_APPLICATION_NEEDS_STUB_DRIVER
 #define CONFIGURE_APPLICATION_NEEDS_ZERO_DRIVER
-#define CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER
 
 /* Note: The select() system call can only be used with the first FD_SETSIZE
  *       File Descriptors (newlib default is 64).  Beginning RTEMS 5.1, FDs are
--- a/modules/libcom/RTEMS/posix/rtems_config.c
+++ b/modules/libcom/RTEMS/posix/rtems_config.c
@@ -47,13 +47,7 @@ extern void *POSIX_Init(void *argument);
 
 #ifndef RTEMS_LEGACY_STACK
 #define CONFIGURE_USE_IMFS_AS_BASE_FILESYSTEM
-/*
- * Configure LibBSD.
- * Note: bsp.h must be included before rtems-bsd-config.h so that
- * BSP-specific macros (e.g., LIBBSP_POWERPC_MVME3100_BSP_H) are
- * defined when nexus-devices.h selects the appropriate drivers.
- */
-#include <bsp.h>
+
 //#define RTEMS_BSD_CONFIG_NET_PF_UNIX
 //#define RTEMS_BSD_CONFIG_NET_IF_BRIDGE
 //#define RTEMS_BSD_CONFIG_NET_IF_LAGG
--- a/modules/libcom/RTEMS/posix/rtems_config.c
+++ b/modules/libcom/RTEMS/posix/rtems_config.c
@@ -49,7 +49,11 @@ extern void *POSIX_Init(void *argument);
 #define CONFIGURE_USE_IMFS_AS_BASE_FILESYSTEM
 /*
  * Configure LibBSD.
+ * Note: bsp.h must be included before rtems-bsd-config.h so that
+ * BSP-specific macros (e.g., LIBBSP_POWERPC_MVME3100_BSP_H) are
+ * defined when nexus-devices.h selects the appropriate drivers.
  */
+#include <bsp.h>
 //#define RTEMS_BSD_CONFIG_NET_PF_UNIX
 //#define RTEMS_BSD_CONFIG_NET_IF_BRIDGE
 //#define RTEMS_BSD_CONFIG_NET_IF_LAGG
@@ -66,6 +70,7 @@ extern void *POSIX_Init(void *argument);
 #define CONFIGURE_APPLICATION_NEEDS_CONSOLE_DRIVER
 #define CONFIGURE_APPLICATION_NEEDS_STUB_DRIVER
 #define CONFIGURE_APPLICATION_NEEDS_ZERO_DRIVER
+#define CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER
 
 /* Note: The select() system call can only be used with the first FD_SETSIZE
  *       File Descriptors (newlib default is 64).  Beginning RTEMS 5.1, FDs are
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -990,8 +990,8 @@ POSIX_Init ( void *argument __attribute__((unused)))
     if (epicsRtemsInitPreSetBootConfigFromNVRAM(&rtems_bsdnet_config) != 0)
         delayedPanic("epicsRtemsInitPreSetBootConfigFromNVRAM");
     if (rtems_bsdnet_config.bootp == NULL) {
-        extern void setBootConfigFromNVRAM(void);
-        setBootConfigFromNVRAM();
+        extern int setBootConfigFromNVRAM(char *, size_t);
+        setBootConfigFromNVRAM(NULL, 0);
     }
     if (epicsRtemsInitPostSetBootConfigFromNVRAM(&rtems_bsdnet_config) != 0)
         delayedPanic("epicsRtemsInitPostSetBootConfigFromNVRAM");
@@ -1047,14 +1047,23 @@ POSIX_Init ( void *argument __attribute__((unused)))
     printf("\n***** ifconfig lo0 *****\n");
     rtems_bsd_ifconfig_lo0();
 
-    printf("\n***** add dhcpcd hook *****\n");
-    dhcpDone = epicsEventMustCreate(epicsEventEmpty);
-    rtems_dhcpcd_add_hook(&dhcpcd_hook);
+    /* Check whether global environment static network config exists.
+     * If so, use that, else, fall back to DHCP. */
+    extern int setBootConfigFromNVRAM(char *, size_t);
+    int status = setBootConfigFromNVRAM(rtemsInit_NTP_server_ip,
+                                             sizeof(rtemsInit_NTP_server_ip));
+    bool try_dhcp = status != 0;
 
-    printf("\n***** Start default network dhcpcd *****\n");
-    // if MY_BOOTP???
-    default_network_dhcpcd();
+    if (try_dhcp) {
+        printf("\n***** add dhcpcd hook *****\n");
+        dhcpDone = epicsEventMustCreate(epicsEventEmpty);
+        rtems_dhcpcd_add_hook(&dhcpcd_hook);
 
+        printf("\n***** Start default network dhcpcd *****\n");
+        // if MY_BOOTP???
+        default_network_dhcpcd();
+    }
+    
     /* this seems to be hard coded in the BSP -> Sebastian Huber ? */
     printf("\n--Info (hpj)-- bsd task prio IRQS: %d  -----\n", rtems_bsd_get_task_priority("IRQS"));
     printf("\n--Info (hpj)-- bsd task prio TIME: %d  -----\n", rtems_bsd_get_task_priority("TIME"));
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -34,6 +34,9 @@
 #include <sys/stat.h>
 #include <sys/socket.h>
 #include <sys/syslog.h>
+#include <net/route.h>
+#include <net/if_dl.h>
+#include <rtems/bsd/iface.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <assert.h>
@@ -65,6 +68,8 @@
 #include <rtems/bsd/modules.h>
 #include <rtems/dhcpcd.h>
 #include <machine/rtems-bsd-commands.h>
+#include <bsp.h>
+#include <bsp/nexus-devices.h>
 #endif
 
 #include <rtems/telnetd.h>
@@ -873,6 +878,50 @@ default_network_dhcpcd(void)
     sc = rtems_dhcpcd_start(NULL);
     assert(sc == RTEMS_SUCCESSFUL);
 }
+
+/*
+ * Block until the named interface reports link up via an RTM_IFINFO routing
+ * message, or until timeout_secs elapses.
+ *
+ * route_sock must already be open (opened before the interface was configured
+ * so that no RTM_IFINFO event can be missed).
+ *
+ * This function was modeled after the RTM_IFINFO handling in
+ * rtems-libbsd/dhcpcd/if-bsd.c (manage_link).  There's a
+ * simpler implementation in a test that uses a sleep loop in
+ * rtems-libbsd/testsuite/include/rtems/bsd/test/default-init.h, but
+ * we chose this more responsive event based implementation.
+ * 
+ * Returns 0 if link came up, -1 on timeout or error.
+ */
+static int
+wait_for_link_up(int route_sock, const char *ifname, int timeout_secs)
+{
+    printf("Waiting for link on %s (timeout %ds)...\n", ifname, timeout_secs);
+    
+    struct timeval tv = { .tv_sec = timeout_secs, .tv_usec = 0 };
+    setsockopt(route_sock, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv));
+
+    char buf[sizeof(struct if_msghdr) + sizeof(struct sockaddr_dl)];
+    ssize_t n;
+    while ((n = recv(route_sock, buf, sizeof(buf), 0)) > 0) {
+        struct rt_msghdr *rtm = (struct rt_msghdr *)(void *)buf;
+        if (rtm->rtm_type == RTM_IFINFO) {
+            struct if_msghdr *ifm = (struct if_msghdr *)(void *)buf;
+            char name[IFNAMSIZ];
+            if (if_indextoname(ifm->ifm_index, name) != NULL &&
+                strcmp(name, ifname) == 0 &&
+                ifm->ifm_data.ifi_link_state == LINK_STATE_UP) {
+                return 0;
+            }
+        }
+    }
+    
+    /* recv returned <= 0: SO_RCVTIMEO expired (EAGAIN) or socket error */
+    printf("Warning: link did not come up on %s within %ds\n",
+           ifname, timeout_secs);
+    return -1;
+}
 #endif // not RTEMS_LEGACY_STACK
 
 #if __RTEMS_MAJOR__>4
@@ -951,7 +1000,9 @@ POSIX_Init ( void *argument __attribute__((unused)))
     char timeBuff[100];
 
     initConsole ();
-
+    
+    printf("epics-base build date %s, %s\n", __DATE__, __TIME__);
+    
     /*
      * Use BSP-supplied time of day if available otherwise supply default time.
      * It is very likely that other time synchronization facilities in EPICS
@@ -1050,11 +1101,25 @@ POSIX_Init ( void *argument __attribute__((unused)))
     /* Check whether global environment static network config exists.
      * If so, use that, else, fall back to DHCP. */
     extern int setBootConfigFromNVRAM(char *, size_t);
+
+    /* Open route socket before configuring the interface so that no
+     * events are missed. */
+    int route_sock = socket(PF_ROUTE, SOCK_RAW, 0);
+    if (route_sock < 0)
+        printf("Warning: could not open PF_ROUTE socket: %s\n", strerror(errno));
+
+    printf("\n***** checking for static network configuration *****\n");
     int status = setBootConfigFromNVRAM(rtemsInit_NTP_server_ip,
                                              sizeof(rtemsInit_NTP_server_ip));
     bool try_dhcp = status != 0;
 
     if (try_dhcp) {
+        /* Route socket not needed — DHCP path has its own event mechanism */
+        if (route_sock >= 0) {
+            close(route_sock);
+            route_sock = -1;
+        }
+
         printf("\n***** add dhcpcd hook *****\n");
         dhcpDone = epicsEventMustCreate(epicsEventEmpty);
         rtems_dhcpcd_add_hook(&dhcpcd_hook);
@@ -1062,36 +1127,49 @@ POSIX_Init ( void *argument __attribute__((unused)))
         printf("\n***** Start default network dhcpcd *****\n");
         // if MY_BOOTP???
         default_network_dhcpcd();
+
+
+        /* this seems to be hard coded in the BSP -> Sebastian Huber ? */
+        printf("\n--Info (hpj)-- bsd task prio IRQS: %d  -----\n", rtems_bsd_get_task_priority("IRQS"));
+        printf("\n--Info (hpj)-- bsd task prio TIME: %d  -----\n", rtems_bsd_get_task_priority("TIME"));
+
+
+        // wait for dhcp done ... should be if SYNCDHCP is used
+        epicsEventWaitStatus stat;
+        printf("\n ---- Waiting for DHCP ...\n");
+        stat = epicsEventWaitWithTimeout(dhcpDone, 600);
+        if (stat == epicsEventOK)
+            epicsEventDestroy(dhcpDone);
+        else if (stat == epicsEventWaitTimeout)
+            printf("\n ---- DHCP timed out!\n");
+        else
+            printf("\n ---- dhcpDone Event Unknown state %d\n", stat);
+    } else {
+        /* Static IP: block until the physical link comes up before
+         * proceeding to NTP and NFS, analogous to the DHCP event wait. */
+        if (route_sock >= 0) {
+            char ifnamebuf[IF_NAMESIZE];
+            char *ifname = if_indextoname(1, ifnamebuf);
+            if (ifname != NULL)
+                wait_for_link_up(route_sock, ifname, 30);
+            close(route_sock);
+        }
+    }
+
+    { // Dump ifconfig and netstat info to stdout
+        const char* ifconfg_args[] = {
+            "ifconfig", NULL
+        };
+        const char* netstat_args[] = {
+            "netstat", "-rn", NULL
+        };
+
+        printf("-------------- IFCONFIG -----------------\n");
+        rtems_bsd_command_ifconfig(1, (char**) ifconfg_args);
+        printf("-------------- NETSTAT ------------------\n");
+        rtems_bsd_command_netstat(2, (char**) netstat_args);
     }
     
-    /* this seems to be hard coded in the BSP -> Sebastian Huber ? */
-    printf("\n--Info (hpj)-- bsd task prio IRQS: %d  -----\n", rtems_bsd_get_task_priority("IRQS"));
-    printf("\n--Info (hpj)-- bsd task prio TIME: %d  -----\n", rtems_bsd_get_task_priority("TIME"));
-
-
-    // wait for dhcp done ... should be if SYNCDHCP is used
-    epicsEventWaitStatus stat;
-    printf("\n ---- Waiting for DHCP ...\n");
-    stat = epicsEventWaitWithTimeout(dhcpDone, 600);
-    if (stat == epicsEventOK)
-        epicsEventDestroy(dhcpDone);
-    else if (stat == epicsEventWaitTimeout)
-        printf("\n ---- DHCP timed out!\n");
-    else
-        printf("\n ---- dhcpDone Event Unknown state %d\n", stat);
-
-    const char* ifconfg_args[] = {
-        "ifconfig", NULL
-    };
-    const char* netstat_args[] = {
-        "netstat", "-rn", NULL
-    };
-
-    printf("-------------- IFCONFIG -----------------\n");
-    rtems_bsd_command_ifconfig(1, (char**) ifconfg_args);
-    printf("-------------- NETSTAT ------------------\n");
-    rtems_bsd_command_netstat(2, (char**) netstat_args);
-
     /* until now there is no NTP support in libbsd -> Sebastian Huber ... */
     printf("\n***** Until now no NTP support in RTEMS 5 with rtems-libbsd *****\n");
     printf("\n***** Ask ntp server once... *****\n");
@@ -1168,6 +1246,15 @@ POSIX_Init ( void *argument __attribute__((unused)))
    // printf (" telnetd initialized with result %d\n", result);
 #endif
 
+#if 1
+// Start an rtems shell before main, for debugging RTEMS system issues
+    rtems_shell_init("SHLL", RTEMS_MINIMUM_STACK_SIZE * 4,
+                     100, "/dev/console",
+                     false, true,
+                     NULL);
+#endif
+
+    
     printf ("***** Preparing EPICS application *****\n");
     iocshRegisterRTEMS ();
     set_directory (argv[1]);
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -1122,9 +1122,7 @@ POSIX_Init ( void *argument __attribute__((unused)))
         rtems_dhcpcd_add_hook(&dhcpcd_hook);
 
         printf("\n***** Start default network dhcpcd *****\n");
-        // if MY_BOOTP???
         default_network_dhcpcd();
-
 
         /* this seems to be hard coded in the BSP -> Sebastian Huber ? */
         printf("\n--Info (hpj)-- bsd task prio IRQS: %d  -----\n", rtems_bsd_get_task_priority("IRQS"));
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -890,14 +890,14 @@ default_network_dhcpcd(void)
  * simpler implementation in a test that uses a sleep loop in
  * rtems-libbsd/testsuite/include/rtems/bsd/test/default-init.h, but
  * we chose this more responsive event based implementation.
- * 
+ *
  * Returns 0 if link came up, -1 on timeout or error.
  */
 static int
 wait_for_link_up(int route_sock, const char *ifname, int timeout_secs)
 {
     printf("Waiting for link on %s (timeout %ds)...\n", ifname, timeout_secs);
-    
+
     struct timeval tv = { .tv_sec = timeout_secs, .tv_usec = 0 };
     setsockopt(route_sock, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv));
 
@@ -915,7 +915,7 @@ wait_for_link_up(int route_sock, const char *ifname, int timeout_secs)
             }
         }
     }
-    
+
     /* recv returned <= 0: SO_RCVTIMEO expired (EAGAIN) or socket error */
     printf("Warning: link did not come up on %s within %ds\n",
            ifname, timeout_secs);
@@ -999,7 +999,7 @@ POSIX_Init ( void *argument __attribute__((unused)))
     char timeBuff[100];
 
     initConsole ();
-    
+
     /*
      * Use BSP-supplied time of day if available otherwise supply default time.
      * It is very likely that other time synchronization facilities in EPICS
@@ -1166,7 +1166,7 @@ POSIX_Init ( void *argument __attribute__((unused)))
         printf("-------------- NETSTAT ------------------\n");
         rtems_bsd_command_netstat(2, (char**) netstat_args);
     }
-    
+
     /* until now there is no NTP support in libbsd -> Sebastian Huber ... */
     printf("\n***** Until now no NTP support in RTEMS 5 with rtems-libbsd *****\n");
     printf("\n***** Ask ntp server once... *****\n");
@@ -1251,7 +1251,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
                      NULL);
 #endif
 
-    
     printf ("***** Preparing EPICS application *****\n");
     iocshRegisterRTEMS ();
     set_directory (argv[1]);
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -36,7 +36,6 @@
 #include <sys/syslog.h>
 #include <net/route.h>
 #include <net/if_dl.h>
-#include <rtems/bsd/iface.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <assert.h>
@@ -1001,8 +1000,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
 
     initConsole ();
     
-    printf("epics-base build date %s, %s\n", __DATE__, __TIME__);
-    
     /*
      * Use BSP-supplied time of day if available otherwise supply default time.
      * It is very likely that other time synchronization facilities in EPICS
@@ -1246,7 +1243,7 @@ POSIX_Init ( void *argument __attribute__((unused)))
    // printf (" telnetd initialized with result %d\n", result);
 #endif
 
-#if 1
+#if 0
 // Start an rtems shell before main, for debugging RTEMS system issues
     rtems_shell_init("SHLL", RTEMS_MINIMUM_STACK_SIZE * 4,
                      100, "/dev/console",
--- a/modules/libcom/RTEMS/setBootConfigFromNVRAM.c
+++ b/modules/libcom/RTEMS/setBootConfigFromNVRAM.c
@@ -102,23 +102,27 @@ splitNfsMountPath(char *nfsString)
 #endif /* HAVE_MOTLOAD || __mcf528x__ */
 
 struct boot_net_config {
-    char *ip_address;
-    char *netmask;
-    char *gateway;
-    char *server;        /* boot server; used as NTP/DNS fallback */
-    char *ntp_server;
-    char *hostname;
-    char *dns_server;
-    char *domainname;
-    char *bootp_boot_file;
+    char *ip_address;           /* static IP address of this device */
+    char *netmask;              /* subnet mask */
+    char *gateway;              /* default gateway */
+    char *server;               /* boot server; used as NTP/DNS fallback if not overridden */
+    char *ntp_server;           /* NTP server (falls back to server if NULL) */
+    char *hostname;             /* this device's hostname */
+    char *dns_server;           /* DNS resolver (falls back to server if NULL) */
+    char *domainname;           /* DNS domain name */
+    char *bootp_boot_file;      /* boot filename reported via BOOTP */
     uint32_t bootp_server_addr; /* server IPv4 address as stored in PPCBUG NVRAM */
 };
 
-#ifdef RTEMS_LEGACY_STACK
+/* Configure network from boot_net_config.
+ * RTEMS_LEGACY_STACK: populate rtems_bsdnet_config fields.
+ * libbsd: configure the first hardware interface (index 1) with a static IP.
+ * Returns 0 on success, -1 on failure. */
 static int
 applyNetConfig(const struct boot_net_config *cfg,
                char *ntp_server_ip, size_t ntp_server_ip_size)
 {
+#ifdef RTEMS_LEGACY_STACK
     (void)ntp_server_ip;
     (void)ntp_server_ip_size;
     rtems_bsdnet_bootp_server_name           = cfg->server;
@@ -134,16 +138,7 @@ applyNetConfig(const struct boot_net_config *cfg,
         rtems_bsdnet_bootp_boot_file_name = cfg->bootp_boot_file;
     if (cfg->bootp_server_addr)
         rtems_bsdnet_bootp_server_address.s_addr = cfg->bootp_server_addr;
-    return 0;
-}
 #else
-/* Configure the first hardware network interface with a static IP.
- * Assumes loopback is at index 0 and the first hardware device at index 1.
- * Returns 0 on success, -1 on failure. */
-static int
-applyNetConfig(const struct boot_net_config *cfg,
-               char *ntp_server_ip, size_t ntp_server_ip_size)
-{
     char ifnamebuf[IF_NAMESIZE];
     char *ifname = if_indextoname(1, ifnamebuf);
     if (ifname == NULL) {
@@ -156,7 +151,7 @@ applyNetConfig(const struct boot_net_config *cfg,
            cfg->gateway    ? cfg->gateway    : "NULL");
     if (cfg->ip_address && cfg->netmask) {
         int exit_code = rtems_bsd_ifconfig(ifname, cfg->ip_address,
-                                            cfg->netmask, cfg->gateway);
+                                           cfg->netmask, cfg->gateway);
         if (exit_code != EX_OK) {
             printf("rtems_bsd_ifconfig failed (exit code %d)\n", exit_code);
             return -1;
@@ -167,9 +162,9 @@ applyNetConfig(const struct boot_net_config *cfg,
     }
     if (ntp_server_ip != NULL && ntp_server_ip_size > 0 && cfg->ntp_server != NULL)
         snprintf(ntp_server_ip, ntp_server_ip_size, "%s", cfg->ntp_server);
+#endif
     return 0;
 }
-#endif
 
 #endif /* HAVE_MOTLOAD || HAVE_PPCBUG || __mcf528x__ */
 
@@ -275,14 +270,14 @@ setBootConfigFromNVRAM(char *ntp_server_ip, size_t ntp_server_ip_size)
     if ((cfg.ip_address = gev("mot-/dev/enet0-cipa", nvp)) == NULL) {
         cfg.ip_address = motScriptParm(mot_script_boot, 'c');
     }
-    
+
     if ((cfg.netmask = gev("mot-/dev/enet0-snma", nvp)) == NULL) {
         cfg.netmask = motScriptParm(mot_script_boot, 'm');
     }
     if ((cfg.gateway = gev("mot-/dev/enet0-gipa", nvp)) == NULL) {
         cfg.gateway = motScriptParm(mot_script_boot, 'g');
     }
-    
+
     if ((cfg.server = gev("mot-/dev/enet0-sipa", nvp)) == NULL) {
         cfg.server = motScriptParm(mot_script_boot, 's');
     }
--- a/modules/libcom/RTEMS/setBootConfigFromNVRAM.c
+++ b/modules/libcom/RTEMS/setBootConfigFromNVRAM.c
@@ -101,30 +101,44 @@ struct boot_net_config {
     char *ip_address;
     char *netmask;
     char *gateway;
-    char *server;      /* boot server; used as NTP/DNS fallback */
+    char *server;        /* boot server; used as NTP/DNS fallback */
     char *ntp_server;
+    char *hostname;
+    char *dns_server;
+    char *domainname;
+    char *bootp_boot_file;
+    uint32_t bootp_server_addr; /* server IPv4 address as stored in PPCBUG NVRAM */
 };
 
 #ifdef RTEMS_LEGACY_STACK
-/* Apply the six core bsdnet config assignments common to all BSPs.
- * Each BSP supplements with its own hostname/dns/boot-file overrides. */
-static void
-applyNetConfig(const struct boot_net_config *cfg)
+static int
+applyNetConfig(const struct boot_net_config *cfg,
+               char *ntp_server_ip, size_t ntp_server_ip_size)
 {
+    (void)ntp_server_ip;
+    (void)ntp_server_ip_size;
     rtems_bsdnet_bootp_server_name           = cfg->server;
-    rtems_bsdnet_config.name_server[0]       = cfg->server;
+    rtems_bsdnet_config.name_server[0]       = cfg->dns_server ? cfg->dns_server : cfg->server;
     rtems_bsdnet_config.ntp_server[0]        = cfg->ntp_server ? cfg->ntp_server : cfg->server;
     rtems_bsdnet_config.gateway              = cfg->gateway;
     rtems_bsdnet_config.ifconfig->ip_netmask = cfg->netmask;
     rtems_bsdnet_config.ifconfig->ip_address = cfg->ip_address;
-    rtems_bsdnet_config.hostname             = cfg->ip_address;
+    rtems_bsdnet_config.hostname             = cfg->hostname ? cfg->hostname : cfg->ip_address;
+    if (cfg->domainname)
+        rtems_bsdnet_config.domainname = cfg->domainname;
+    if (cfg->bootp_boot_file)
+        rtems_bsdnet_bootp_boot_file_name = cfg->bootp_boot_file;
+    if (cfg->bootp_server_addr)
+        rtems_bsdnet_bootp_server_address.s_addr = cfg->bootp_server_addr;
+    return 0;
 }
 #else
 /* Configure the first hardware network interface with a static IP.
  * Assumes loopback is at index 0 and the first hardware device at index 1.
  * Returns 0 on success, -1 on failure. */
 static int
-applyNetConfig(const struct boot_net_config *cfg)
+applyNetConfig(const struct boot_net_config *cfg,
+               char *ntp_server_ip, size_t ntp_server_ip_size)
 {
     char ifnamebuf[IF_NAMESIZE];
     char *ifname = if_indextoname(1, ifnamebuf);
@@ -147,6 +161,8 @@ applyNetConfig(const struct boot_net_config *cfg)
         printf("Skipping static IP address and netmask from NVRAM\n");
         return -1;
     }
+    if (ntp_server_ip != NULL && ntp_server_ip_size > 0 && cfg->ntp_server != NULL)
+        snprintf(ntp_server_ip, ntp_server_ip_size, "%s", cfg->ntp_server);
     return 0;
 }
 #endif
@@ -222,7 +238,6 @@ setBootConfigFromNVRAM(char *ntp_server_ip, size_t ntp_server_ip_size)
 {
     const char *mot_script_boot;
     volatile char *nvp;
-    char *cp;
     struct boot_net_config cfg = {0};
 
 # if defined(BSP_NVRAM_BASE_ADDR)
@@ -271,6 +286,12 @@ setBootConfigFromNVRAM(char *ntp_server_ip, size_t ntp_server_ip_size)
         char *ntp_gev = gev("epics-ntpserver", nvp);
         cfg.ntp_server = ntp_gev ? ntp_gev : cfg.server;
     }
+    cfg.hostname   = gev("rtems-client-name", nvp);
+    cfg.dns_server = gev("rtems-dns-server", nvp);
+    cfg.domainname = gev("rtems-dns-domainname", nvp);
+    cfg.bootp_boot_file  = gev("mot-/dev/enet0-file", nvp);
+    if (cfg.bootp_boot_file == NULL)
+        cfg.bootp_boot_file = motScriptParm(mot_script_boot, 'f');
 
     /*
      * Apply network configuration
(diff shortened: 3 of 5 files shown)


brendanchandler:
Windows test failure seems unrelated.  It appears the test abruptly ended for some reason?  Is this a known issue possibly?

brendanchandler:
All the codacity checks are related to unused fields for certain BSPs.  However, I personally prefer keeping the struct generic and allowing BSPs to optionally set fields they want to set.

That said, there are two ways to deal with these codacity issues.  I don't favor either one, and propose the method used in the pull request.  But here are some alternatives:
1. Add manifests around the fields in the struct based on which BSP we are using.  I think it complicates things unnecessarily to do this, and propagates more manifests through the code without a good reason.  The struct is only a few pointers, and is short lived, so memory increase is trivial.
2. Force the BSPs to set NULL or similar values for things that are unused.  I don't care for this because we already zero initialize the struct with the unused fields, so this would just be adding unnecessary code to satisfy codacity.

anjohnson:
Some of our tests are flaky on Windows and do fail at random on both the Appveyor and GitHub CI runners. I re-ran that GitHub Win-22 job and it passed as I expected. There's a discussion at #162 with some details of the frequent flapping tests; Claude developed some fixes for the one which failed here, I plan to turn it into a PR soon.

Apparently your branch has developed one or more merge conflicts with 7.0, please rebase it and drop/squash your commits when you get a chance.

mdavidsaver:
@brendanchandler Any thoughts or plans on how to unittest the NVRAM code(s)?

brendanchandler:
> @brendanchandler Any thoughts or plans on how to unittest the NVRAM code(s)?

I think there's a fair bit I could get tested.  Functions like splitRtemsBsdnetBootpCmdline, splitNfsMountPath, gev, motScriptParm, and so on should be fairly straight forward to get some unit tests for.  We could create a test that contains a memory layout that matches what a device would have, and point these functions at that test data address and make sure they parse things correctly.

I think the very BSP specific details I'd leave untested unless we had qemu support or similar.  I'm thinking of things like:

1. Getting a pointer to some hardware specific address (    nvp = (volatile char *)(BSP_NVRAM_BASE_ADDR+0x70f8);
)
2. opening and reading from I2C device (    if ((fd = open(BSP_I2C_VPD_EEPROM_DEV_NAME, 0)) < 0)   )

I find these types of things more difficult to test, and you're kinda testing interface code to access data from some other system, so you'd have to create some sort of mock or simulation which adds complexity and I am not sure how many problems such tests would actually catch.

So in short, I think we can get some tests for our own code parsing or doing operations on data, but I don't think I'd try to test interface code like opening I2C and reading from its file descriptor.

Hopefully that's similar to your line of thinking, but let me know if not.  I'll see about making another pull request to add those tests.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -873,6 +877,50 @@ default_network_dhcpcd(void)
...
+ * This function was modeled after the RTM_IFINFO handling in
+ * rtems-libbsd/dhcpcd/if-bsd.c (manage_link).  There's a
+ * simpler implementation in a test that uses a sleep loop in
+ * rtems-libbsd/testsuite/include/rtems/bsd/test/default-init.h, but
+ * we chose this more responsive event based implementation.
+ *
+ * Returns 0 if link came up, -1 on timeout or error.
+ */
+static int
+wait_for_link_up(int route_sock, const char *ifname, int timeout_secs)
```
kiwichris, line 897: RTEMS LibBSD provides an interface called `rtems_bsd_iface_get()` and `rtems_bsd_iface_link_state()`. Could they be used?

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -1047,41 +1095,77 @@ POSIX_Init ( void *argument __attribute__((unused)))
...
+            close(route_sock);
+            route_sock = -1;
+        }
 
-    printf("\n***** Start default network dhcpcd *****\n");
-    // if MY_BOOTP???
-    default_network_dhcpcd();
+        printf("\n***** add dhcpcd hook *****\n");
+        dhcpDone = epicsEventMustCreate(epicsEventEmpty);
+        rtems_dhcpcd_add_hook(&dhcpcd_hook);
```
kiwichris, line 1122: _Note:_ I am not sure about this call in RTEMS as an interface and so I cannot guarantee it will always be present. Our DHCP client is not maintained upstream and we may move to something else in LibBSD.

modules/libcom/RTEMS/setBootConfigFromNVRAM.c
```diff
@@ -172,52 +248,71 @@ setBootConfigFromNVRAM(void)
...
-    if ((rtems_bsdnet_config.ifconfig->ip_address = gev("mot-/dev/enet0-cipa", nvp)) == NULL)
-        rtems_bsdnet_config.ifconfig->ip_address = motScriptParm(mot_script_boot, 'c');
-    rtems_bsdnet_config.hostname = gev("rtems-client-name", nvp);
-    if (rtems_bsdnet_config.hostname == NULL)
-        rtems_bsdnet_config.hostname = rtems_bsdnet_config.ifconfig->ip_address;
-
-    if ((rtems_bsdnet_bootp_boot_file_name = gev("mot-/dev/enet0-file", nvp)) == NULL)
-        rtems_bsdnet_bootp_boot_file_name = motScriptParm(mot_script_boot, 'f');
+
+    /*
```
kiwichris, line 267: Do the changes in this block of code change the functionality in any way? I ask because I am currently tasked to move this code to RTEMS. You can view the code in my personal repo for [motload-gev-net-env.c](https://gitlab.rtems.org/chris/rtems/-/blob/chris/5573-gev-env-vars/bsps/powerpc/shared/start/motload-gev-net-env.c).

brendanchandler: It shouldn't modify the functionality in any way.  Basically, rather than directly set rtems_bsdnet_bootp_server_name and similar, we store the values in a struct and then later set rtems_bsdnet_bootp_server_name from the values stored in the struct.  You can see we still query "mot-/dev/enet0-sipa" in this function (line 281 in new diff), and then set rtems_bsdnet_bootp_server_name in the applyNetConfig() function above:

    if (cfg->bootp_boot_file)
        rtems_bsdnet_bootp_boot_file_name = cfg->bootp_boot_file;

The general idea here is to separate collecting all the params from applying them all to set up the network, and this "cfg" struct is the interface for separating the two.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: This is an existing call from before this pull request.  This pull request moves it under a "try_dhcp" block but otherwise doesn't change from previous behavior.  That said, it sounds like we might need to make an issue to look at changing this if we know a better or more supported interface to use.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I'm a little surprised that this socket interface wouldn't be meant for client code to use.  Why expose it via the socket interface then?

We can certainly change to use rtems_bsd_iface_link_state().  I don't have the code in front of me at the moment, but I'm guessing this is what the sleep loop in test/default_init.h does, which I referred to in the comment of this function.  I think this pull request's implementation is:
a. Slightly more complicated
b. event based rather than poll and sleep based

Do you prefer the poll and sleep method instead?  Its an easy code change, but I find the event based a nicer behavior.

kiwichris: > I'm a little surprised that this socket interface wouldn't be meant for client code to use. Why expose it via the socket interface then?

The socket interface is present in FreeBSD and we support that so you are most welcome to make use of it. I was only offering a version we maintain 😄  

> We can certainly change to use rtems_bsd_iface_link_state(). I don't have the code in front of me at the moment, but I'm guessing this is what the sleep loop in test/default_init.h does, which I referred to in the comment of this function. I think this pull request's implementation is: a. Slightly more complicated b. event based rather than poll and sleep based

Yes it seems seems so ...

> Do you prefer the poll and sleep method instead? Its an easy code change, but I find the event based a nicer behavior.

... and I had not picked up on the event based version. It is a nice approach. Maybe we should move this back into LibBSD?

modules/libcom/RTEMS/posix/rtems_init.c
kiwichris: Yes I am aware it has been in a while now. My understanding is the call was used to determine a valid IP address had been set on the interface. Is it used for any other purpose?

modules/libcom/RTEMS/setBootConfigFromNVRAM.c
kiwichris: Thanks. If we can maintain the behaviour and then work to push this down into the BSPs to use environment variables we can remove these global variables and possible references to things that were deep in the legacy stack. The catch is getting the support for the BSPs into RTEMS. My beatnik work is first and an example I hope results in support for other BSPs being added.

If we have an agreed direction how we get there is not as important.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: Yes, I believe that's the only purpose.  I'll double check tomorrow when I have the code in front of me.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: That sounds good.  I'll admit, in my mind I sorta blur the line between FreeBSD and LibBSD whereas you have more of a clear separation (from what I infer reading your comments).  I suppose you're saying if LibBSD provided a similar interface it would be maintaned by RTEMS (rather than upstream FreeBSD).  That certainly sounds fine to me.

kiwichris: This was written about LibBBSD a while ago now but it is still valid .. [FreeBSD and RTEMS, UNIX in
A REAL-TIME OPERATING SYSTEM](https://freebsdfoundation.org/wp-content/uploads/2016/08/FreeBSD-and-RTEMS-Unix-in-a-Real-Time-Operating-System.pdf).

anjohnson: @kiwichris Would moving "this back into LibBSD" change the API being called here at all? If it would, we'd need a way to know which API to call.

modules/libcom/RTEMS/posix/rtems_config.c
```diff
@@ -66,6 +64,7 @@ extern void *POSIX_Init(void *argument);
 #define CONFIGURE_APPLICATION_NEEDS_CONSOLE_DRIVER
 #define CONFIGURE_APPLICATION_NEEDS_STUB_DRIVER
 #define CONFIGURE_APPLICATION_NEEDS_ZERO_DRIVER
+#define CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER
```
anjohnson: Is this going to be set for all BSPs? It looks more like something that the BSP would know about, I'm not sure that all our boards have an I2C bus.

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -1159,6 +1243,14 @@ POSIX_Init ( void *argument __attribute__((unused)))
    // printf (" telnetd initialized with result %d\n", result);
 #endif
 
+#if 0
```
anjohnson, line 1244: Could/should this be controlled by looking for an environment variable instead?

I might even suggest a command-line switch, but I don't know if a command-line equivalent even exists.
If it does, supporting switches and the startup script path would make this more like the workstation IOCs. That idea is purely for discussion though…

anjohnson (commented):
2 questions, the I2C one is the most important.

modules/libcom/RTEMS/posix/rtems_init.c
kiwichris: It  maybe moved into `net-services` or it maybe provided in the networking stack with a standard interface. I do not yet know. The interface would be in the `rtems_*` namespace and specific a single purpose. Either implementation would allow the legacy and libbsd stacks to be supported.

I can take a look at the call `decl` if this is important for you or feel free to make one. I would have the call contain the internal workings, ie any socket open, close and `fd` handling.

modules/libcom/RTEMS/posix/rtems_config.c
kiwichris: Where does this define come from?

```
(rpy3.11) ruru rtems.git $ grep -r 'CONFIGURE_APPLICATION_NEEDS_.*_DRIVER' cpukit/include/ | grep I2C
(rpy3.11) ruru rtems.git $
```

modules/libcom/RTEMS/posix/rtems_init.c
kiwichris: You can access the RTEMS shell via the EPICS shell with the EPICS `rt` command. Maybe `rt` with no arguments could leave you at the RTEMS shell prompt and `exit` returns you to the EPICS shell?

This looks like it is a test feature for debugging this code. Would that be something you see users needing to do at an operational level? It could accidentally be enabled in production isolating the node?

anjohnson: I see the purpose of this to allow an IOC engineer to stop the initialization after RTEMS is up but before it runs the EPICS startup script. They could use RTEMS shell commands to look at the network configuration, what's been mounted (maybe they might want to change that mount temporarily) etc.

The only way to enable it would be to change the boot parameters, so they must have access to the device console to do that. Yes it could be done accidentally, but there are other ways to prevent the IOC from starting (set a bad IP address or gateway, change the path to the st.cmd file etc.) so I don't see this as any extra risk, more of a useful tool that might be helpful at 2am when trying to solve a problem with the machine.

On a VxWorks system we might currently clear the startup script parameter to reach a VxWorks prompt without loading anything. I'm not sure that doing the same thing on an RTEMS IOC would give you control at the same stage.

brendanchandler: Adding this as a boot parameter sounds useful.  I'll consider this for a future PR.

modules/libcom/RTEMS/posix/rtems_config.c
brendanchandler: @hongran helped me with this one.  Without it, there was some issue at boot, maybe the network interface on MVME3100 wasn't available... its a bit foggy for me now.

I agree it seems like something that might be better if epics-base didn't have to worry about--but at the same time, there is some precedent for similar options specified here.  I'm happy to remove this, but we'll need to find another place (in the BSP?) to address the need for i2c on mvme3100.  Ran might be able to help us with more explanation of what this option does and how it affects other boards without an I2C bus.

Regarding Chris's question:

It seems like the define is used by iodrivers.h

```
$ rg LIBI2C_DRIVER
src/rtems/cpukit/include/rtems/libi2c.h
142:#define RTEMS_LIBI2C_DRIVER_TABLE_ENTRY   \

src/rtems/cpukit/include/rtems/confdefs/iodrivers.h
63:  defined(CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER) || \
116:#ifdef CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER
165:  #ifdef CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER
166:    RTEMS_LIBI2C_DRIVER_TABLE_ENTRY,

src/rtems/testsuites/libtests/i2c01/init.c
803:#define CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER
```

I see a bunch of these options documented here, but this specific one is missing:  https://docs.rtems.org/docs/main/c-user/config/index.html.

hongran: mvme3100 needs this configuration. Vijay was working on this issue and he tried to fix the BSP so that this line in the application is not needed. I tried his last version but it didn't work. I think when he started working at ANL, he may prioritize this task.

joelsherrill: The code above in the grep output is **NOT** in RTEMS 5, 6, or main. That's why it is not in the C Users Guide.

The grep indicates that it is missing the code piece that would have included the header. The pattern in _iodrivers.h_ is to use the CONFIGURE_APPLICATION_NEEDS_xxx to conditionally include the driver header and then later conditionally include the driver table entry in the configured device driver table.

kiwichris: Thanks for the details. 

I am starting to wonder if some parts of `confdefs.h` work OK with EPICS and other parts cause problems? My concern is ending up with lots of conditional code in EPICS for BSP _related_ features in the `confdefs.h` and that means you need to add more when new boards and BSPs are added. Drivers and other things can be dynamically registered or maybe revisiting some driver set ups in BSPs to have drivers be available automatically could work?

brendanchandler: > The code above in the grep output is **NOT** in RTEMS 5, 6, or main. That's why it is not in the C Users Guide.
> 
> The grep indicates that it is missing the code piece that would have included the header. The pattern in _iodrivers.h_ is to use the CONFIGURE_APPLICATION_NEEDS_xxx to conditionally include the driver header and then later conditionally include the driver table entry in the configured device driver table.

Ahh, right.  This must be Vijay’s patches that Ran applied in our local fork of RTEMS.  I didn’t realize that.

It will be a priority to get those patches merged into upstream RTEMS, but for now its even more reason not to include this define in epics.  I will update this pull request to remove it tomorrow.

anjohnson: I'd like to clarify the fundamental issue, which I _think_ is that on the MVME3100 board the network driver requires that I2C driver to be initialized early enough for it to work. Have I understood that correctly, and if so how should that dependency be properly expressed within the RTEMS BSP sources so that EPICS doesn't have to put BSP-specific code into our initialization? As @kiwichris suggests we'd much rather not have that for _any_ BSP.

We can get Vijay to work on the implementation once he starts here next week so it's not really urgent, but given that there seems to be some confusion about it is there agreement on how to encode that dependency properly in the BSP?

kiwichris: Thanks @anjohnson. Is it needed to get the boot arguments to determine the network settings?

modules/libcom/RTEMS/posix/rtems_init.c
```diff
@@ -1047,41 +1095,77 @@ POSIX_Init ( void *argument __attribute__((unused)))
...
-    default_network_dhcpcd();
+        printf("\n***** add dhcpcd hook *****\n");
+        dhcpDone = epicsEventMustCreate(epicsEventEmpty);
+        rtems_dhcpcd_add_hook(&dhcpcd_hook);
 
-    /* this seems to be hard coded in the BSP -> Sebastian Huber ? */
-    printf("\n--Info (hpj)-- bsd task prio IRQS: %d  -----\n", rtems_bsd_get_task_priority("IRQS"));
-    printf("\n--Info (hpj)-- bsd task prio TIME: %d  -----\n", rtems_bsd_get_task_priority("TIME"));
+        printf("\n***** Start default network dhcpcd *****\n");
+        // if MY_BOOTP???
```
anjohnson: I'm aware this comment line was just moved from line 1055, but could it be resolved or removed if no longer appropriate?

`git blame` says it was added by @hjunkes in 2020.

anjohnson (changes requested):
This PR (and others tagged `RTEMS`) still has several unresolved conversations. It's not clear if any areas are still being worked on and hence whether the PR is ready to be reviewed & merged.

This may be cultural; I believe in RTEMS the person making the original comment resolves the thread after the implementer indicates they've completed the requested change, but EPICS doesn't have a written rule about that.

One thing still missing is an entry for the Release Notes. The file `documentation/new-notes/README.txt` explains what to do.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I think the original question here was resolves, so I will mark this resolved.

kiwichris: I have moved this code into the my RTEMS Init work.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: I removed the old comment.  At best it would probably be better tracked through a github issue if something doesn't work, rather than a comment in the code.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: @kiwichris, Do you have a link to a merge request or similar for tracking?  When this gets into RTEMS, we can look at removing it from epics-base.

modules/libcom/RTEMS/posix/rtems_init.c
brendanchandler: release note added too.

modules/libcom/RTEMS/posix/rtems_init.c
kiwichris: Sorry, I should have been clearer. The RTEMS Init work is for EPICS and currently a work-in-progress branch in my `epics-base` fork. I intend to have an updated version ready for the next RTEMS Support meeting.

When I reviewed this topic in more detail I remembered RTEMS already has support in LibBSD and my RTEMS Init work for EPCIS moves us to the `rc.conf` initialization and that waits for valid routes to be established using that support. As part of that work I brought [your code](https://github.com/kiwichris/epics-base/blob/chris/rtems-init/modules/libcom/RTEMS/epicsRtemsInit_net.cpp#L28) across to that branch to capture it. Your solution is better than the support I added to LibBSD however I do not have the time to update the RTEMS version as what is there is working.

modules/libcom/RTEMS/posix/rtems_config.c
thelunatic: The CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER is now added to upstream RTEMS, and this option should be available. As @anjohnson mentioned already, the i2c needs to be initialized much earlier in the mvme3100 for it to be accessible.

This configure option is not a BSP specific option, and it just enables the libi2c driver, following the same iodrivers pattern used by CONFIGURE_APPLICATION_NEEDS_CONSOLE_DRIVER and other drivers that are required by the application and needs to be initialized at early stages.

I am currently working on adding documentation of this option for upstream RTEMS for future reference.

anjohnson (approved):
This looks good to me now, thanks.

@hjunkes would you like to test this before we merge it, to make sure we didn't break your configuration? Last call!

hjunkes:
I am canoeing in Poland this weekFHI, Heinz JunkesOn 29. Jun 2026, at 19:11, Andrew Johnson ***@***.***> wrote:﻿
@anjohnson approved this pull request.

This looks good to me now, thanks.
@hjunkes would you like to test this before we merge it, to make sure we didn't break your configuration? Last call!

hjunkes:
As I am currently already working on RTEMS7 on the beatnik board (MVME6100) in parallel with the improvements to DHCP and NTP boot, I have used cherry-picking for this PR.

```
rtems@rtems-dev:~/EPICS/epics-base$ git log
commit 08a157e603a98268451912b7cd6856304e8ba19a (HEAD -> rtems-ntpd-7_0-with-5-and-6-v2)
Author: RTEMS developments for FEL <rtems@rtems-dev.fhi.mpg.de>
Date:   Mon Jul 6 11:33:49 2026 +0200

    reapply FHI NTP/mve local changes on top of PR-853

commit 4ba93c3e695c56d80a495efe6020cbfe0c4a72a2
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Wed Jun 24 14:30:53 2026 -0500

    Add release note for static IP config under libbsd RTEMS

commit 337d1e9fb35925ed8e871d981f9b20074b722692
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Wed Jun 24 10:26:14 2026 -0500

    Remove old comment about if MY_BOOTP

commit ebac22afa9b0de02ddeec3f409199f8eb66ecc9c
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Tue Jun 16 10:34:04 2026 -0500

    Remove CONFIGURE_APPLICATION_NEEDS_LIBI2C_DRIVER

    This option doesn't exist (yet?) in RTEMS and shouldn't have been
    included in this pull request.

commit 3261af9ca83f47ba85f4461994cb587a8f78f12f
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Wed Apr 8 10:51:10 2026 -0500

    Address some codacity issues in libcom/RTEMS/

commit 383981899ff1b277658f97885e0559aaad62a35a
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Tue Apr 7 09:17:19 2026 -0500

    Fix unused function warnings for functions unused on certain BSPs

commit b3891d4f874eab2534876e6a49191f23d38b1618
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Tue Apr 7 09:16:02 2026 -0500

    rtems_init: clean up debug code accidentally added in earlier commit

commit 28d390744f532e1e6d78b7ff3b1c859e79beab8e
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Mon Apr 6 17:23:46 2026 -0500

    Move the rest of networking setup into applyNetConfig

    This is a follow on to the previous commmit.  There are some
    networking options that weren't fully common across all the BSPs, so
    that config was left out.  Now its included and optionally used only
    when the BSP sets the field appropriately.

    This lets the setBootConfigFromNVRAM functions not have any rtems
    network stack #if's in them, and only focus on collecting the
    configuration from each BSP configuration as appropriate.

commit a6dbdb942d1cca6ad6fb97c492df4207b436dfe2
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Mon Apr 6 16:30:09 2026 -0500

    Separate common network setup from BSP specific implementations

    All the BSP setBootConfigFromNVRAM() functions ultimately configured
    the RTEMS networking the same way.  This commit pulls the common
    networking code out into a separate function and has each BSP specific
    code call this new function.

    There still remains some BSP specific networking setup in each
    function.  This change only moves out the common networking setup.

commit 622532136634c088c76fce37c6021199d17f3be2
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Tue Mar 24 12:54:14 2026 -0500

    Wait for link to come up when starting RTEMS/libbsd/static IP

    When configuring a static IP on RTEMS/libbsd, we need to intentionally
    wait for the link to come up before continuing to configure ntp, etc.

commit b4c52ba536ee22065814cb896cc38a0c05accebd
Author: Brendan Chandler <bchandler@anl.gov>
Date:   Tue Jan 6 11:48:03 2026 -0600

    Add RTEMS libbsd static network config for motload nvram

    setBootConfigFromNVRAM() now configures static networking via the
    libbsd rtems_bsd_ifconfig() API when building against RTEMS with
    libbsd.  It only uses static networking if it can find a configured IP
    and netmask from GEV or the mot script boot parameters, otherwise it
    falls back to DHCP.

    Also note, NTP was part of setBootConfigFromNVRAM() when using legacy
    RTEMS networking, but is handled outside this function for LibBSD.  To
    preserve the legacy behavior, setBootConfigFromNVRAM() still pulls out
    NTP server info, but now stores it in an output parameter for outside
    code to use.

commit 29f9ca63f29939173c4aee68b034beb0404c0701
Author: Heinz Junkes <junkes@fhi-berlin.mpg.de>
Date:   Fri Jun 19 16:32:17 2026 +0200

    Restructure NTP/DHCP init sequence on top of upstream ntpd support

    - Resolved merge conflicts from rebase onto kiwi/rtems-ntpd-7_0-with-5-and-6-v2
    ```
   After adding the following to configure/os/CONFIG.Common.RTEMS-beatnik
    ARCH_DEP_CFLAGS += -DHAVE_MOTLOAD
    I get the following on boot:
    ```
    vmeTsi148 IRQ manager: looking for registers on VME...
Trying to find CSR on VME...
vmeTsi148 - IRQ manager using VME CSR to flush FIFO
Registering /dev/console as minor 0 (==/dev/ttyS0)

*** FATAL ***
fatal source: 9 (RTEMS_FATAL_SOURCE_EXCEPTION)
exception vector 3 (0x3)
  next PC or address of fault = 0x003557c0
  saved MSR = 0x0200b032
  context = task, ISR nest level = 0
  thread dispatch disable level = 1
  R0  = 0x0035edf8 R1  = 0x00bca730 R2  = 0x00bd1ed0 R3  = 0x0001effc
  R4  = 0x00000000 R5  = 0x00000002 R6  = 0x487a9bd9 R7  = 0x4e800020
  R8  = 0x00003041 R9  = 0x00b92e60 R10 = 0x00007bff R11 = 0x00000008
  R12 = 0x00000001 R13 = 0x00b5a188 R14 = 0x00000000 R15 = 0x00000000
  R16 = 0x00000000 R17 = 0x00000000 R18 = 0x00000000 R19 = 0x00000000
  R20 = 0x00000000 R21 = 0x00000000 R22 = 0x00000000 R23 = 0x00000000
  R24 = 0x00bca918 R25 = 0x00000000 R26 = 0x00000006 R27 = 0x00000000
  R28 = 0x00bca8b8 R29 = 0x00a1c060 R30 = 0x00000002 R31 = 0x00000000
  CR  = 0x40000222
  CTR = 0x00000000
  XER = 0x00000000
  LR  = 0x0035edf8
  DAR = 0x00000110
  executing thread ID = 0x0b010001, name =
Stack Trace:
  IP: 0x003557c0, LR: 0x0035edf8
--^ 0x0035edf8--^ 0x002bcca8--^ 0x00359c70--^ 0x003acd40--^ 0x003b0754
--^ 0x0022e45c--^ 0x0022d438--^ 0x006bbf14--^ 0x006bc0a8--^ 0x006bbff8
RTEMS version: 7.0.0.4420185b4455ea7cbd56cb6c84a2f5effa857883
RTEMS tools: 15.2.0 20250808 (RTEMS 7, Build 2026.06.12, RSB 657823dc22d42a5d4ee81255540c2aab4aaf954d-modified, Newlib 7d4336cf)
executing thread ID: 0x0b010001
```

I need to look into this further. I just wanted to let you know that I’m on it now.

thelunatic:
> I need to look into this further. I just wanted to let you know that I’m on it now.

Hi @hjunkes, could you please paste the whole boot log somewhere (github gist maybe?) and post a link here?

hjunkes:
There was an incorrect ordering in rtems_init.c. That may, of course, have been caused by my cherry-picking.

```
diff --git a/modules/libcom/RTEMS/posix/rtems_init.c b/modules/libcom/RTEMS/posix/rtems_init.c
index 38aacfd86..753f190b2 100644
--- a/modules/libcom/RTEMS/posix/rtems_init.c
+++ b/modules/libcom/RTEMS/posix/rtems_init.c
@@ -1076,7 +1076,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
     char timeBuff[100];

     initConsole ();
-        setBootConfigFromNVRAM(NULL, 0);

     /*
      * Create a reasonable environment
@@ -1112,6 +1111,13 @@ POSIX_Init ( void *argument __attribute__((unused)))
     if(0) rtems_bsd_setlogpriority("debug");
     on_exit(default_network_on_exit, NULL);

+    printf("\n***** Initializing network (libbsd) *****\n");
+    sc = rtems_bsd_initialize();
+    assert(sc == RTEMS_SUCCESSFUL);
+
+    /* Let the callout timer allocate its resources */
+    rtems_task_wake_after(RTEMS_MILLISECONDS_TO_TICKS(500));
+
     /* Write leap seconds - rtems_ntpd_client_pool_config normally does this */
     rtems_ntpd_add_etc_services();   /* writes /etc/services */

@@ -1142,12 +1148,6 @@ POSIX_Init ( void *argument __attribute__((unused)))
     /* Tell osdNTP_Configure to use this file, skipping pool config */
     epicsEnvSet("EPICS_TS_NTP_CONF_FILE", "/etc/ntp.conf");

-    printf("\n***** Initializing network (libbsd) *****\n");
-    sc = rtems_bsd_initialize();
-    assert(sc == RTEMS_SUCCESSFUL);
-
-    /* Let the callout timer allocate its resources */
-    rtems_task_wake_after(RTEMS_MILLISECONDS_TO_TICKS(500));


     printf("\n***** ifconfig lo0 *****\n");
 ```

I have to set all the NVRAM GEV variables, otherwise it looks like this ;-)

```
...

***** Setting up file system *****
***** Initializing NFS *****
"(null)" is not a valid command pathname.
"%s" is not a valid command pathname.

*** FATAL ***
fatal source: 11 (RTEMS_FATAL_SOURCE_PANIC)
...

hjunkes:
Almost done ;-)
```
MVME6100> gevShow
mot-script-boot
tftpGet -a04000000 -c141.14.128.12 -s141.14.131.192 -m255.255.240.0 -g141.14.128.128 -frtems6/myExample.boot
waitProbe
netShut
go -a04000000

mot-/dev/enet0-cipa=141.14.128.12
mot-/dev/enet0-snma=255.255.240.0
mot-/dev/enet0-gipa=141.14.128.128
mot-/dev/enet0-sipa=141.14.131.192
epics-ntpserver=141.14.138.238
mot-/dev/enet0-file=/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample/st.cmd
epics-script=141.14.131.192:/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample:st.cmd
Total Number of GE Variables =8, Bytes Utilized =499, Bytes Free =3093
MVME6100> reset



Copyright(C)2008-2009,Emerson Network Power-Embedded Computing,Inc.
All Rights Reserved
Copyright Motorola Inc. 1999-2007, All Rights Reserved
MOTLoad RTOS Version 2.0,  PAL Version 2.3 RM01
Fri Jan 23 14:47:54 MST 2009

MPU-Type             =MPC74x7
MPU-Int Clock Speed  =1266MHz
MPU-Ext Clock Speed  =133MHz
MPU-Int Cache(L2) Enabled, 512KB, L2CR =C0000000
MPU-Ext Cache(L3) Enabled, 2MB, 211MHz, L3CR =DC026300

PCI bus instance 0   =64 bit, 133 Mhz, PCI-X
PCI bus instance 1   =64 bit, PCI

Reset/Boot Vector    =Flash1

Local Memory Found   =20000000 (&536870912)
User Download Buffer =006B7000:008B6FFF

MVME6100> tftpGet -a04000000 -c141.14.128.12 -s141.14.131.192 -m255.255.240.0 -g141.14.128.128 -frtems6/myExample.boot
Network Loading from: /dev/enet0
Loading File: rtems6/myExample.boot
Load Address: 04000000
Download Buffer Size = User Defined

Client IP Address      = 141.14.128.12
Server IP Address      = 141.14.131.192
Gateway IP Address     = 141.14.128.128
Subnet IP Address Mask = 255.255.240.0

Network File Load in Progress...

Bytes Received =&11873936, Bytes Loaded =&11873936
Bytes/Second   =&3957978, Elapsed Time =3 Second(s)
MVME6100> waitProbe
System I/O Probe Complete
MVME6100> netShut
/dev/enet0     -----------------------------------------
config addr is 0xf1000cf8
config data is 0xf1000cfc
Welcome to RTEMS rtems-7.0.0 (PowerPC/Generic (classic FPU)/beatnik)
CPU: MPC7457
Board Type: MVME6100-0163 (S/N E173D27)
Bus Clock Freq:   133333333 Hz
CPU Clock Freq:  1266666654 Hz
Memory:           536870912 bytes
-----------------------------------------
Now BSP_mem_size = 0x1fe00000
Configuration.work_space_size = 33837216
Page table setup finished; will activate it NOW...
Going to start PCI buses scanning and initialization
BAT0 raw(upper=0x00000000, lower=0x00000001)  range(0x00000000, 0x0001ffff) ps gmiw Read Only
BAT1 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT2 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT3 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT4 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT5 raw(upper=0x80003ffe, lower=0x8000002a)  range(0x80000000, 0x8fffffff) pS GmIw Read/Write
BAT6 raw(upper=0xe0001ffe, lower=0xe000002a)  range(0xe0000000, 0xefffffff) pS GmIw Read/Write
BAT7 raw(upper=0xf0001ffe, lower=0xe000002a)  range(0xf0000000, 0xffffffff) pS GmIw Read/Write
Number of PCI buses found is : 3
MSR 0x2003032
Exit from bspstart
unable to find the universe in pci config space
Tundra Tsi148 PCI-VME bridge detected at 0x81100000, IRQ 84
Tsi148 Outbound Ports:
Port  VME-Addr   Size       PCI-Adrs   Mode:
0:    0x20000000 0x0e000000 0x90000000 A32, SUP, D32, SCT
1:    0x00000000 0x00ff0000 0x9f000000 A24, SUP, D32, SCT
2:    0x00000000 0x00010000 0x9fff0000 A16, SUP, D32, SCT
7:    0x00000000 0x01000000 0x9e000000 CSR, SUP, D32, SCT
Tsi148 Inbound Ports:
Port  VME-Addr   Size       PCI-Adrs   Mode:
0:    0x90000000 0x1fe00000 0x00000000 A32, PGM, DAT, SUP, USR, MBLT, BLT
vmeTsi148 IRQ manager: looking for registers on VME...
Trying to find CSR on VME...
vmeTsi148 - IRQ manager using VME CSR to flush FIFO
Registering /dev/console as minor 0 (==/dev/ttyS0)

***** RTEMS Version: rtems-7.0.0 (PowerPC/Generic (classic FPU)/beatnik) *****

***** Initializing network (libbsd) *****
nexus0: <RTEMS Nexus device>
pcib0 pcibus 0
pci0: <PCI bus> on pcib0
pci0: <memory> at device 0.0 (no driver attached)
pci0: <memory> at device 0.1 (no driver attached)
pci0: <memory> at device 0.2 (no driver attached)
pci0: <memory> at device 0.3 (no driver attached)
pci0: <memory> at device 0.4 (no driver attached)
pci0: <memory> at device 0.5 (no driver attached)
pci0: <memory> at device 0.6 (no driver attached)
pci0: <memory> at device 0.7 (no driver attached)
pci0: <bridge> at device 5.0 (no driver attached)
pcib1: <PCI-PCI bridge> at device 6.0 on pci0
pci1: <PCI bus> on pcib1
cpu0: <legacy CPU>
mve0 on nexus0
miibus0: <MII bus> on mve0
ukphy0: <Generic IEEE 802.3u media interface> PHY 0 on miibus0
ukphy0:  none, 10baseT, 10baseT-FDX, 100baseTX, 100baseTX-FDX, 1000baseT, 1000baseT-master, 1000baseT-FDX, 1000baseT-FDX-master, auto
info: mve0: Ethernet address: ec:9e:cd:1a:12:3f

***** Prepare ntp.conf (server 141.14.138.238) *****
NTP: overwrote /etc/ntp.conf with server 141.14.138.238 iburst

***** ifconfig lo0 *****
info: lo0: link state changed to UP

***** checking for static network configuration *****
Configuring ifconfig with ip=141.14.128.12, netmask=255.255.240.0, gateway=141.14.128.128
info: mve0: link state changed to DOWN
Waiting for link on mve0 (timeout 30s)...
info: mve0: link state changed to UP
-------------- IFCONFIG -----------------
mve0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=80000<LINKSTATE>
        ether ec:9e:cd:1a:12:3f
        inet 141.14.128.12 netmask 0xfffff000 broadcast 141.14.143.255
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
        media: Ethernet autoselect (1000baseT <full-duplex>)
        status: active
        inet 127.0.0.1 netmask 0xffffff00
        inet6 ::1 prefixlen 128
        groups: lo
-------------- NETSTAT ------------------
Routing tables

Internet:
Destination        Gateway            Flags         Netif Expire
default            141.14.128.128     UGS            mve0
127.0.0.1          link#2             UH              lo0
141.14.128.0/20    link#1             U              mve0
141.14.128.12      link#1             UHS             lo0

Internet6:
Destination                       Gateway                       Flags         Netif Expire
::1                               link#2                        UHS             lo0
fe80::%lo0/64                     link#2                        U               lo0
fe80::1%lo0                       link#2                        UHS             lo0

***** Ask ntp server once... *****
Received 48 bytes
time from ntp : 07/06/26 15:30:51.000000000 UTC

***** One-shot NTP time set from 141.14.138.238 *****
Received 48 bytes
***** Clock set to 2026/07/06 15:30:51 UTC *****


------ ntp server address : 141.14.138.238 -------
 Initialising EPICS time provider...
ntpd: RTEMS service init

***** Setting up file system *****
***** Initializing NFS *****
nfsmMoouunntt:( "114411..1144..113311..119922:"/,V o"l/uVmoelsu/mEepsi/cEsp/iMcVsM/EM6V1M0E06/1m0y0E/xmaymEpxlaem/piloec/BioooctB/oiootc/miyoEcxmaymEpxlaem p-l>e "/,V o"l/uVmoelsu/mEepsi/cEsp/iMcVsM/EM6V1M0E06/1m0y0E/xmaymEpxlaem/piloec/BioooctB/oiootc/miyoEcxmaymEpxlaem polpet"i)o
ns:nfsv4,minorversion=1
nfs: mount: nfs -> 141.14.131.192:/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample (nfsv4,minorversion=1)
nfs: mount args: 8
  16 addr=10 02 08 01 8d 0e 83 c0 00 00 00 00 00 00 00 00
   0 nfsv4
  55 dirpath=/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample
   4 fstype=nfs
  16 fspath=/iocmyExample-0
  70 hostname=141.14.131.192:/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample
   0 rw
   2 minorversion=1
nfs: mount: (0) Success
 Will try to start telnetd with prio 0 ...
 telnetd initialized with result 0
***** Preparing EPICS application *****
chdir("/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample/")
***** Starting EPICS application *****
#< envPaths
epicsEnvSet("IOC","iocmyExample")
epicsEnvSet("TOP","/Volumes/Epics/MVME6100/myExample")
#epicsEnvSet("PVXS_LOG","*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.timer*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.iface*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.client*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.remote.log=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.udp*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.loop=INFO")
#epicsEnvSet("EVENT_NOKQUEUE", "1")
#epicsEnvSet("EVENT_NOEPOLL", "1")
#epicsEnvSet("EVENT_NOPOLL", "1")
epicsEnvSet("EPICS_PVAS_INTF_ADDR_LIST","141.14.128.12")
epicsEnvSet("EPICS_CAS_INTF_ADDR_LIST","141.14.128.12")
cd /Volumes/Epics/MVME6100/myExample
## Register all support components
dbLoadDatabase("dbd/myExample.dbd")
ERROR: Can't open file 'dbd/myExample.dbd'
myExample_registerRecordDeviceDriver(pdbbase)
Warning: IOC is booting with TOP = "/Volumes/Epics/MVME6100/myExample"
          but was built with TOP = "/home/rtems/MVME6100_7/EPICS/FHI/myExample"
registerRecordDeviceDriver failed aSub
registerRecordDeviceDriver failed aai
registerRecordDeviceDriver failed aao
registerRecordDeviceDriver failed ai
registerRecordDeviceDriver failed ao
registerRecordDeviceDriver failed bi
registerRecordDeviceDriver failed bo
registerRecordDeviceDriver failed calc
registerRecordDeviceDriver failed calcout
registerRecordDeviceDriver failed compress
registerRecordDeviceDriver failed dfanout
registerRecordDeviceDriver failed event
registerRecordDeviceDriver failed fanout
registerRecordDeviceDriver failed histogram
registerRecordDeviceDriver failed int64in
registerRecordDeviceDriver failed int64out
registerRecordDeviceDriver failed longin
registerRecordDeviceDriver failed longout
registerRecordDeviceDriver failed lsi
registerRecordDeviceDriver failed lso
registerRecordDeviceDriver failed mbbi
registerRecordDeviceDriver failed mbbiDirect
registerRecordDeviceDriver failed mbbo
registerRecordDeviceDriver failed mbboDirect
registerRecordDeviceDriver failed permissive
registerRecordDeviceDriver failed printf
registerRecordDeviceDriver failed sel
registerRecordDeviceDriver failed seq
registerRecordDeviceDriver failed state
registerRecordDeviceDriver failed stringin
registerRecordDeviceDriver failed stringout
registerRecordDeviceDriver failed sub
registerRecordDeviceDriver failed subArray
registerRecordDeviceDriver failed waveform
registerRecordDeviceDriver failed xxx
registryJLinkAdd failed calc
registryJLinkAdd failed const
registryJLinkAdd failed debug
registryJLinkAdd failed state
registryJLinkAdd failed trace
registryJLinkAdd failed pva
INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.
 enable event logging ....
 new event_base with config ....
 add new event 1000 sec ...
Enter loop worker for 0x17e1910 using poll
 enable event logging ....
 go into event base loop ...
 new event_base with config ....
 add new event 1000 sec ...
Enter loop worker for 0x18f3260 using poll
 go into event base loop ...
## Load record instances
dbLoadTemplate("db/user.substitutions")
dbLoadTemplate: error opening sub file db/user.substitutions: No such file or directory
dbLoadRecords("db/myExampleVersion.db", "user=rtems")
ERROR: Can't open file 'db/myExampleVersion.db'
ERROR: Failed to load 'db/myExampleVersion.db'
dbLoadRecords("db/dbSubExample.db", "user=rtems")
ERROR: Can't open file 'db/dbSubExample.db'
ERROR: Failed to load 'db/dbSubExample.db'
dbLoadRecords ("db/IOC-stats.db", "IOC=iocmyExample")
ERROR: Can't open file 'db/IOC-stats.db'
ERROR: Failed to load 'db/IOC-stats.db'
iocInit
Starting iocInit
checkDatabase: ERROR menuConvert not defined.
iocBuild: ERROR Aborting, bad database definition (DBD)!
## Start any sequence programs
#seq(sncExample, "user=rtems")
zoneset "CET-1CEST,M3.5.0/2,M10.5.0/3"
>
```
Have to go home, I’ll carry on working on it later... but it’s looking good so far....

brendanchandler:
> ***** Setting up file system *****
> ***** Initializing NFS *****
> "(null)" is not a valid command pathname.
> "%s" is not a valid command pathname.
> 
> *** FATAL ***
> fatal source: 11 (RTEMS_FATAL_SOURCE_PANIC)
> ...
> ```

I created Issue #903 because I think this error message could be done a little better.

hjunkes:
Its working with  these settings:

```
MVME6100> gevShow
mot-script-boot
tftpGet -a04000000 -c141.14.128.12 -s141.14.131.192 -m255.255.240.0 -g141.14.128.128 -frtems6/myExample.boot
waitProbe
netShut
go -a04000000

mot-/dev/enet0-cipa=141.14.128.12
mot-/dev/enet0-snma=255.255.240.0
mot-/dev/enet0-gipa=141.14.128.128
mot-/dev/enet0-sipa=141.14.131.192
epics-ntpserver=141.14.138.238
epics-script=141.14.131.192:/Volumes:Epics/MVME6100/myExample/iocBoot/iocmyExample/st.cmd
Total Number of GE Variables =7, Bytes Utilized =417, Bytes Free =3175

```

boot ....

```

MVME6100> reset



Copyright(C)2008-2009,Emerson Network Power-Embedded Computing,Inc.
All Rights Reserved
Copyright Motorola Inc. 1999-2007, All Rights Reserved
MOTLoad RTOS Version 2.0,  PAL Version 2.3 RM01
Fri Jan 23 14:47:54 MST 2009

MPU-Type             =MPC74x7
MPU-Int Clock Speed  =1266MHz
MPU-Ext Clock Speed  =133MHz
MPU-Int Cache(L2) Enabled, 512KB, L2CR =C0000000
MPU-Ext Cache(L3) Enabled, 2MB, 211MHz, L3CR =DC026300

PCI bus instance 0   =64 bit, 133 Mhz, PCI-X
PCI bus instance 1   =64 bit, PCI

Reset/Boot Vector    =Flash1

Local Memory Found   =20000000 (&536870912)
User Download Buffer =006B7000:008B6FFF

MVME6100> tftpGet -a04000000 -c141.14.128.12 -s141.14.131.192 -m255.255.240.0 -g141.14.128.128 -frtems6/myExample.boot
Network Loading from: /dev/enet0
Loading File: rtems6/myExample.boot
Load Address: 04000000
Download Buffer Size = User Defined

Client IP Address      = 141.14.128.12
Server IP Address      = 141.14.131.192
Gateway IP Address     = 141.14.128.128
Subnet IP Address Mask = 255.255.240.0

Network File Load in Progress...

Bytes Received =&11873936, Bytes Loaded =&11873936
Bytes/Second   =&3957978, Elapsed Time =3 Second(s)
MVME6100> waitProbe
System I/O Probe Complete
MVME6100> netShut
/dev/enet-----------------------------------------
config addr is 0xf1000cf8
config data is 0xf1000cfc
Welcome to RTEMS rtems-7.0.0 (PowerPC/Generic (classic FPU)/beatnik)
CPU: MPC7457
Board Type: MVME6100-0163 (S/N E173D27)
Bus Clock Freq:   133333333 Hz
CPU Clock Freq:  1266666654 Hz
Memory:           536870912 bytes
-----------------------------------------
Now BSP_mem_size = 0x1fe00000
Configuration.work_space_size = 33837216
Page table setup finished; will activate it NOW...
Going to start PCI buses scanning and initialization
BAT0 raw(upper=0x00000000, lower=0x00000001)  range(0x00000000, 0x0001ffff) ps gmiw Read Only
BAT1 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT2 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT3 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT4 raw(upper=0x00000000, lower=0x00000000)  range(0x00000000, 0x0001ffff) ps gmiw No Access
BAT5 raw(upper=0x80003ffe, lower=0x8000002a)  range(0x80000000, 0x8fffffff) pS GmIw Read/Write
BAT6 raw(upper=0xe0001ffe, lower=0xe000002a)  range(0xe0000000, 0xefffffff) pS GmIw Read/Write
BAT7 raw(upper=0xf0001ffe, lower=0xe000002a)  range(0xf0000000, 0xffffffff) pS GmIw Read/Write
Number of PCI buses found is : 3
MSR 0x2003032
Exit from bspstart
unable to find the universe in pci config space
Tundra Tsi148 PCI-VME bridge detected at 0x81100000, IRQ 84
Tsi148 Outbound Ports:
Port  VME-Addr   Size       PCI-Adrs   Mode:
0:    0x20000000 0x0e000000 0x90000000 A32, SUP, D32, SCT
1:    0x00000000 0x00ff0000 0x9f000000 A24, SUP, D32, SCT
2:    0x00000000 0x00010000 0x9fff0000 A16, SUP, D32, SCT
7:    0x00000000 0x01000000 0x9e000000 CSR, SUP, D32, SCT
Tsi148 Inbound Ports:
Port  VME-Addr   Size       PCI-Adrs   Mode:
0:    0x90000000 0x1fe00000 0x00000000 A32, PGM, DAT, SUP, USR, MBLT, BLT
vmeTsi148 IRQ manager: looking for registers on VME...
Trying to find CSR on VME...
vmeTsi148 - IRQ manager using VME CSR to flush FIFO
Registering /dev/console as minor 0 (==/dev/ttyS0)

***** RTEMS Version: rtems-7.0.0 (PowerPC/Generic (classic FPU)/beatnik) *****

***** Initializing network (libbsd) *****
nexus0: <RTEMS Nexus device>
pcib0 pcibus 0
pci0: <PCI bus> on pcib0
pci0: <memory> at device 0.0 (no driver attached)
pci0: <memory> at device 0.1 (no driver attached)
pci0: <memory> at device 0.2 (no driver attached)
pci0: <memory> at device 0.3 (no driver attached)
pci0: <memory> at device 0.4 (no driver attached)
pci0: <memory> at device 0.5 (no driver attached)
pci0: <memory> at device 0.6 (no driver attached)
pci0: <memory> at device 0.7 (no driver attached)
pci0: <bridge> at device 5.0 (no driver attached)
pcib1: <PCI-PCI bridge> at device 6.0 on pci0
pci1: <PCI bus> on pcib1
cpu0: <legacy CPU>
mve0 on nexus0
miibus0: <MII bus> on mve0
ukphy0: <Generic IEEE 802.3u media interface> PHY 0 on miibus0
ukphy0:  none, 10baseT, 10baseT-FDX, 100baseTX, 100baseTX-FDX, 1000baseT, 1000baseT-master, 1000baseT-FDX, 1000baseT-FDX-master, auto
info: mve0: Ethernet address: ec:9e:cd:1a:12:3f

***** Prepare ntp.conf (server 141.14.138.238) *****
NTP: overwrote /etc/ntp.conf with server 141.14.138.238 iburst

***** ifconfig lo0 *****
info: lo0: link state changed to UP

***** checking for static network configuration *****
Configuring ifconfig with ip=141.14.128.12, netmask=255.255.240.0, gateway=141.14.128.128
info: mve0: link state changed to DOWN
Waiting for link on mve0 (timeout 30s)...
info: mve0: link state changed to UP
-------------- IFCONFIG -----------------
mve0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=80000<LINKSTATE>
        ether ec:9e:cd:1a:12:3f
        inet 141.14.128.12 netmask 0xfffff000 broadcast 141.14.143.255
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
        media: Ethernet autoselect (1000baseT <full-duplex>)
        status: active
        inet 127.0.0.1 netmask 0xffffff00
        inet6 ::1 prefixlen 128
        groups: lo
-------------- NETSTAT ------------------
Routing tables

Internet:
Destination        Gateway            Flags         Netif Expire
default            141.14.128.128     UGS            mve0
127.0.0.1          link#2             UH              lo0
141.14.128.0/20    link#1             U              mve0
141.14.128.12      link#1             UHS             lo0

Internet6:
Destination                       Gateway                       Flags         Netif Expire
::1                               link#2                        UHS             lo0
fe80::%lo0/64                     link#2                        U               lo0
fe80::1%lo0                       link#2                        UHS             lo0

***** Ask ntp server once... *****
Timeout - no response
***** Can't get time from ntp ...

***** One-shot NTP time set from 141.14.138.238 *****
Received 48 bytes
***** Clock set to 2026/07/06 18:37:22 UTC *****


------ ntp server address : 141.14.138.238 -------
 Initialising EPICS time provider...
ntpd: RTEMS service init

***** Setting up file system *****
***** Initializing NFS *****
nfsmMoouunntt:( "114411..1144..113311..119922:"/,V o"l/uVmoelsu m-e>s "/,V o"l/uVmoelsu moepst"i)o
ns:nfsv4,minorversion=1
nfs: mount: nfs -> 141.14.131.192:/Volumes (nfsv4,minorversion=1)
nfs: mount args: 8
  16 addr=10 02 08 01 8d 0e 83 c0 00 00 00 00 00 00 00 00
   0 nfsv4
   9 dirpath=/Volumes
   4 fstype=nfs
  11 fspath=/Volumes-0
  24 hostname=141.14.131.192:/Volumes
   0 rw
   2 minorversion=1
nfs: mount: (0) Success
 Will try to start telnetd with prio 0 ...
 telnetd initialized with result 0
***** Preparing EPICS application *****
chdir("/Volumes/Epics/MVME6100/myExample/iocBoot/iocmyExample/")
***** Starting EPICS application *****
#< envPaths
epicsEnvSet("IOC","iocmyExample")
epicsEnvSet("TOP","/Volumes/Epics/MVME6100/myExample")
#epicsEnvSet("PVXS_LOG","*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.timer*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.iface*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.client*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.remote.log=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.udp*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs*=DEBUG")
#epicsEnvSet("PVXS_LOG", "pvxs.loop=INFO")
#epicsEnvSet("EVENT_NOKQUEUE", "1")
#epicsEnvSet("EVENT_NOEPOLL", "1")
#epicsEnvSet("EVENT_NOPOLL", "1")
epicsEnvSet("EPICS_PVAS_INTF_ADDR_LIST","141.14.128.12")
epicsEnvSet("EPICS_CAS_INTF_ADDR_LIST","141.14.128.12")
cd /Volumes/Epics/MVME6100/myExample
## Register all support components
dbLoadDatabase("dbd/myExample.dbd")
myExample_registerRecordDeviceDriver(pdbbase)
Warning: IOC is booting with TOP = "/Volumes/Epics/MVME6100/myExample"
          but was built with TOP = "/home/rtems/MVME6100_7/EPICS/FHI/myExample"
INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.
 enable event logging ....
 new event_base with config ....
 add new event 1000 sec ...
Enter loop worker for 0x1605650 using poll
 enable event logging ....
 go into event base loop ...
 new event_base with config ....
 add new event 1000 sec ...
Enter loop worker for 0x1606568 using poll
 go into event base loop ...
## Load record instances
dbLoadTemplate("db/user.substitutions")
dbLoadRecords("db/myExampleVersion.db", "user=rtems")
dbLoadRecords("db/dbSubExample.db", "user=rtems")
dbLoadRecords ("db/IOC-stats.db", "IOC=iocmyExample")
iocInit
Starting iocInit
## EPICS R7.0.10.1-DEV
## Rev. R7.0.10-89-g805cf4c94df0bc1a1622-dirty
## Rev. Date Git: 2026-07-06 17:04:51 +0200
 enable event logging ....
 new event_base with config ....
 add new event 1000 sec ...
Enter loop worker for 0x212ef40 using poll
 go into event base loop ...
## Start any sequence programs
#seq(sncExample, "user=rtems")
zoneset "CET-1CEST,M3.5.0/2,M10.5.0/3"
> pwd
/Volumes/Epics/MVME6100/myExample
> ls
Command 'ls' not registered.
> dbl
rtems:aSubExample
rtems:line:b
rtems:aiExample
rtems:aiExample1
rtems:ai1
rtems:aiExample2
rtems:ai2
rtems:aiExample3
rtems:ai3
iocmyExample:M_CaClientCnt
iocmyExample:M_CaConnectionCnt
iocmyExample:M_RecordCnt
iocmyExample:M_MemoryUsed
iocmyExample:M_MemoryFree
iocmyExample:M_MemoryTotal
iocmyExample:M_SysCpuCnt
iocmyExample:M_SysCpu
iocmyExample:M_SysCpu:Avg1
iocmyExample:M_SysCpu:Avg2
iocmyExample:M_SysCpu:Avg5
iocmyExample:M_Cpu
iocmyExample:M_Cpu:Avg1
iocmyExample:M_Cpu:Avg2
iocmyExample:M_Cpu:Avg5
iocmyExample:M_FdMax
iocmyExample:M_FdCnt
iocmyExample:M_IfErrorsInp
iocmyExample:M_IfErrorsOut
rtems:circle:step
rtems:circle:period
iocmyExample:S_FdScanPeriod
iocmyExample:S_CpuScanPeriod
iocmyExample:S_MemoryScanPeriod
rtems:circle:angle
rtems:line:a
rtems:circle:x
rtems:circle:y
rtems:calcExample
rtems:calcExample1
rtems:calc1
rtems:calcExample2
rtems:calc2
rtems:calcExample3
rtems:calc3
iocmyExample:M_Memory
iocmyExample:M_Fd
rtems:circle:tick
rtems:compressExample
iocmyExample:M_SysCpu:History
iocmyExample:M_SysCpu:Avg1Buf60
iocmyExample:M_SysCpu:Avg2Buf30
iocmyExample:M_SysCpu:Avg5Buf12
iocmyExample:M_Cpu:History
iocmyExample:M_Cpu:Avg1Buf60
iocmyExample:M_Cpu:Avg2Buf30
iocmyExample:M_Cpu:Avg5Buf12
iocmyExample:M_Memory:History
iocmyExample:M_Fd:History
iocmyExample:C_TimeBase10-360
rtems:myExample:version
iocmyExample:C_TimeBase:init
iocmyExample:M_TimeStamp
iocmyExample:M_BootTime
iocmyExample:M_Uptime
iocmyExample:M_BspVersion
iocmyExample:M_KernelVersion
iocmyExample:M_EpicsVersion
iocmyExample:M_Hostname
rtems:subExample
iocmyExample:S_Reboot
iocmyExample:M_SysCpu:Avg1Buf1
iocmyExample:M_SysCpu:Avg2Buf1
iocmyExample:M_SysCpu:Avg5Buf1
iocmyExample:M_Cpu:Avg1Buf1
iocmyExample:M_Cpu:Avg2Buf1
iocmyExample:M_Cpu:Avg5Buf1
rtems:xxxExample
>
```
Will check for first ntp timeout and print garbage ...

Issue #855, opened 2026-04-10
dbCa duplicate subscriptions in CA_MONITOR_STRING case


mdavidsaver:
When the `CA_MONITOR_STRING` special case is triggered (`dbGetLink()` of DBF_ENUM as DBR_STRING), a second subscription is started.  However, the first "native" subscription is not stopped.  Both subscriptions remain active and deliver events.  I think the relative ordering of the two subscriptions is unspecified, but probably stable.

For CP/CPP links, since at least 717d69e1fcf3cb8dc4533ba53bdae838a396d274 this does not result in duplicate processing (due to the `CA_DBPROCESS` action bit).  However, there is a race between dbLocal and dbCa threads over whether the native or string update is used.

What to do about this?  Cancel the "native" subscription when creating the "string" subscription?

mdavidsaver:
This [patch](https://github.com/user-attachments/files/26641461/dbCa.diff.txt) adds some debug prints.  With [this database](https://github.com/user-attachments/files/26641464/dbca-string.db.txt) loaded, I see the following.  Notice that duplicate `string: event` lines.

```
$ ./bin/linux-x86_64/softIoc -m P= -d dbca-string.db 
Starting iocInit
...
integer: native
string: native
iocRun: All initialization complete
integer: event
string: event
dbCaLink: dbProcess of 'integer'
integer: process
dbCaLink: dbProcess of 'string'
string: process
string: string
dbCaLink: dbProcess of 'string'
string: event
string: process
dbCaLink: dbProcess of 'integer'
integer: event
string: event
string: event
integer: process
string: process
...
```

mdavidsaver:
A pathological example to illustrate one aspect of the weird handling of the "string" subscription.  The "enumstr:first" record ends up with VAL="" and AMSG="field INP".  This seems unavoidable without client side ENUM handling, which would best be accomplished with a second subscription with DBE_PROPERTY.  Of course, with the history of how DBE_PROPERTY was introduced, this is not practical as a default behavior.

```
record(bi, "enumstr:bi") {
    field(ZNAM, "zero")
    field(ONAM, "one")
    field(VAL , "1")
}
record(stringin, "enumstr:string") {
    field(INP , "enumstr:bi CPP MSS")
    field(FLNK, "enumstr:first")
    field(TPRO, "1")
}
record(stringin, "enumstr:first") {
    field(INP , "enumstr:string MSS")
    field(FLNK, "enumstr:gate")
}
[...]

PR #856, opened 2026-04-12
dbCa: iocInit wait for all conditions

mdavidsaver:
Address the issue mentioned by https://github.com/epics-base/epics-base/pull/852#issue-4219442078 .  See #768 for history.

```
# Testing with regressLinkSevr.db
...
not ok 69 - dbGetField("si2", DBR_STRING) -> "INVALID" == "" # TODO Not working
ok 70 - dbGetField("li2", DBR_LONG) -> 3 == 3
```

It is necessary to wait for up to three concurrent actions after connection.  Native and string subscriptions, as well as GET of metadata.  Decide in `connectCallback()` which will be waited for.

One complication is that `struct event_handler_args` does not identify the `evid`.  So need to introduce two entry point functions to distinguish the two subscriptions.  (see #855 for this tangent)

Instead of one `caLink::flags` bit, switch to four.  bit 0 `_WAITING` is set while the `initOutstanding` is incremented for this link.  Three `_WAIT_*` bits (1 -> 3) track the 3 concurrent conditions which may be outstanding to declare readiness.  From `dbInitLink()`, all 4 bits are set for local links.  `testInitReady()` holds the state logic for clearing bits 1-3, then issuing `CA_INIT_READY` when the last of these bits are cleared.  Bit 0 is then cleared by the dbCa worker when it decrements `initOutstanding`.

```c
#define DBCA_CALLBACK_INIT_WAITING     (0x1)
#define DBCA_CALLBACK_INIT_WAIT_NATIVE (0x2)
#define DBCA_CALLBACK_INIT_WAIT_STRING (0x4)
#define DBCA_CALLBACK_INIT_WAIT_ATTRIB (0x8)
```

### DIFF
--- a/modules/database/src/ioc/db/dbCa.c
+++ b/modules/database/src/ioc/db/dbCa.c
@@ -65,7 +65,7 @@ extern int dbServiceIsolate;
 static ELLLIST workList = ELLLIST_INIT;    /* Work list for dbCaTask */
 static epicsMutexId workListLock; /*Mutual exclusions semaphores for workList*/
 static epicsEventId workListEvent; /*wakeup event for dbCaTask*/
-static size_t initOutstanding; // one count for every link with DBCA_CALLBACK_INIT_WAIT
+static size_t initOutstanding; // one count for every link with a DBCA_CALLBACK_INIT_WAIT bit
 static int removesOutstanding = 0;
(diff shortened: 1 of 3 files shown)


anjohnson:
Core Group approved.

Issue #860, opened 2026-04-17
Update dllPath.bat handling?


mdavidsaver:
I was [surprised](https://github.com/epics-base/pvxs/pull/167#issuecomment-4264235817) to discover `dllPath.bat`.  Apparently this has escaped my notice because it is only included when `makeBaseApp` is asked to expand for a windows target, which I never do when starting a new project.

Since it is now fairly common for one `iocBoot/*/st.cmd` to work on multiple targets, would it be reasonable to either always include `dllPath.bat` in `Makefile@Common`?  Or better, have an `ifeq` or enable something like `TARGETS_WIN32 = dllPath.bat`?

```
$ git grep TARGETS modules/database/src/template/top/iocBoot/ioc/
modules/database/src/template/top/iocBoot/ioc/Makefile@Common:TARGETS = envPaths
modules/database/src/template/top/iocBoot/ioc/Makefile@cygwin:TARGETS = envPaths relPaths.sh
modules/database/src/template/top/iocBoot/ioc/Makefile@vxWorks:TARGETS = cdCommands
modules/database/src/template/top/iocBoot/ioc/Makefile@win32:TARGETS = envPaths dllPath.bat
modules/database/src/template/top/iocBoot/ioc/Makefile@windows:TARGETS = envPaths dllPath.bat
```

@FreddieAkeroyd fyi.

FreddieAkeroyd:
I added it to `Makefile@Common` in our local installations but having it automatically pick up target arch via a `TARGETS_WIN32`  sounds a good idea.  One issue we have had with `dllPath.bat` is that its contents are  `EPICS_HOST_ARCH` specific so if you do have multiple windows builds (e.g. windows-x64 and windows-x64-debug) in the same tree it doesn't work as it has expanded the host arch first time. Options could be:
- write `dllPath.bat` with explicit `bin\%EPICS_HOST_ARCH%`  references rather than expanded ones
- generate target specific `dllPath-$(ARCH).bat` files, then have `dllPath.bat` call `dllPath-%EPICS_HOST_ARCH%.bat`. For backward compatibility we could have a default `EPICS_HOST_ARCH` still specified in `dllPath.bat` if it hasn't been set before it is called.

mdavidsaver:
> * write `dllPath.bat` with explicit `bin\%EPICS_HOST_ARCH%`  references rather than expanded ones

[...]

PR #861, opened 2026-04-17
Windows: emit .bat wrappers with DLL builds

mdavidsaver:
For windows targets, when linking dynamically, emit one

- `$(INSTALL_LOCATION)/bin/$(T_A)/dllPath.bat`

And for each install exe, also emit a bat file wrapper to prepend PATH

- `$(INSTALL_LOCATION)/bin/$(T_A)/%.bat`

cf. #860

### DIFF
--- a/configure/RULES_BUILD
+++ b/configure/RULES_BUILD
@@ -371,6 +371,15 @@ $(MODNAME): %$(MODEXT): %$(EXE)
 	@$(MKDIR) $(dir $@)
 	$(CONVERTRELEASE) -T $(TOP) $@
 
+ifeq (WIN32,$(OS_CLASS))
+install: $(INSTALL_BIN)/dllPath.bat
+$(INSTALL_BIN)/dllPath.bat: \
+    $(wildcard $(TOP)/configure/RELEASE*) \
+    $(wildcard $(TOP)/configure/CONFIG_SITE*)
+	$(CONVERTRELEASE) -a $(T_A) -T $(TOP) -t $(IOCS_APPL_TOP) dllPath.bat
+	$(INSTALL) -d dllPath.bat $(INSTALL_BIN)
+endif
+
 #---------------------------------------------------------------
 # Automated testing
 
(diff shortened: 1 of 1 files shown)


configure/RULES_BUILD
```diff
@@ -480,6 +490,12 @@ $(INSTALL_BIN)/%: ../os/$(OS_CLASS)/%
 $(INSTALL_BIN)/%: %
 	$(ECHO) "Installing created executable $@"
 	@$(INSTALL_PRODUCT) -d -m $(BIN_PERMISSIONS) $< $(INSTALL_BIN)
+ifeq (WIN32,$(OS_CLASS))
+ifeq (NO,$(STATIC_BUILD))
```
FreddieAkeroyd, line 493: Also i think you may need to build a `dllPath` on a static epics build. We have some external vendor libraries that are dll only and their dlls are in the `bin` of a support module even on a static build.

configure/RULES_BUILD
```diff
@@ -480,6 +490,12 @@ $(INSTALL_BIN)/%: ../os/$(OS_CLASS)/%
...
+	echo "CALL %~dp0/dllPath.bat" > $(basename $(INSTALL_BIN)/$*).bat
+	echo "%~dp0/$* %*" >> $(basename $(INSTALL_BIN)/$*).bat
```
FreddieAkeroyd: `%~dp0` includea a final `\` so `%~dp0$*` is probably fine

FreddieAkeroyd: probably good to add echo setlocal so that PATH is reset after IOC exits
Maybe Wine is different, but i think the " get passed through on windows - I will confirm that

mdavidsaver:
@FreddieAkeroyd Requested changed made.

mdavidsaver:
Potential Makefile issues:

- `bin/*/dllPath.bat` is created even when no `$(PROD)` defined
- Will use of the `$(INSTALL_BIN)/%: %` rule create `.bat` wrappers for files which are not `.exe`?  Would this matter?

mdavidsaver:
I am not able to find a microsoft.com reference for this weird looking `%~dp0` incantation.  The best I can find is [ss64.com](https://ss64.com/nt/syntax-args.html).

Assuming the WINE implementation is consistent, then with `foo.bat` containing:

```cmd
@echo off
echo %%0 %0
echo %%~dp0 %~dp0
echo %%~f0 %~f0
echo %%~d0 %~d0
echo %%~p0 %~p0
echo %%~n0 %~n0
echo %%~x0 %~x0
```

Running `wine foo.bat` emits:

```
%0 "Z:\tmp\junk\foo.bat"
%~dp0 Z:\tmp\junk\
%~f0 Z:\tmp\junk\foo.bat
%~d0 Z:
%~p0 \tmp\junk\
%~n0 foo
%~x0 .bat
```

FreddieAkeroyd:
I found https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/call if that is any help

FreddieAkeroyd:
And I get the same results on windows 11, except my `%0` does not have quotes unless i execute in it quotes e.g. as  `"a.bat"`  did you add the quotes or does wine always add quotes?  Also note if i run `a.bat` my `%0` is `a.bat` - if i run `c:\Users\faa59\a.bat` my `%0` is `c:\Users\faa59\a.bat` - so does wine always make things run if they were an absolute path as it looks like you invoked wine with a relative path?

mdavidsaver:
> ... did you add the quotes or does wine always add quotes?

It seems that `wine ...` always adds quotes to `%0`.  Maybe not surprising as the quotes are handled by `bash` before `wine` executes.  If instead I run `wineconsole` and the enter `foo` or `"foo"`, then any quoting is passed through.  In either case `%~0` strips quotes, as specified (something to remember).

PR #862, opened 2026-04-17
as/ca: keep HAG and CA hostname state current by DNS TTL

derekste:
## Summary

This PR makes EPICS Base treat hostname-based policy and search destinations as runtime state instead of one-time startup resolution.

It covers two areas:

1. **Access Security HAG hostnames** now refresh lazily by DNS TTL/retry.
2. **CA client `EPICS_CA_ADDR_LIST` / `EPICS_CA_NAME_SERVERS` hostnames** now stay current for the life of a CA client context.

The goal is to avoid stale behavior in long-lived clients after DNS changes, without requiring reconnect, reload, or restart.

## Why

Before this branch:

- HAG hostnames were resolved once at ACF load time.
- Existing `ASCLIENTPVT` clients could keep stale cached access rights even after HAG DNS state changed.
- CA client hostname destinations were resolved once during setup and then fixed for the lifetime of the context.

So DNS changes could update external reality while EPICS Base continued using stale policy or stale search routing.

### Access Security / HAG

- Refresh HAG hostnames lazily by DNS TTL or retry interval.
- Preserve existing address selection by resolving through `aToIPAddr()`.
- Use DNS A-record TTLs on supported POSIX resolver platforms when the selected IPv4 address is present in the reply.
- Fall back to `300s` after successful lookups without a usable TTL and `60s` for unresolved retry.
- Track the next HAG expiry/retry time so ordinary checks stay cheap when no refresh is due.
- Add `asCheckClientAccess(ASCLIENTPVT, asAccessRights)` and route `asCheckGet()` / `asCheckPut()` through it as source-compatible macros.
- Treat HAG DNS expiry as an access-security state transition: refresh HAG state, recompute active clients, and fire existing `asClientCOAR` callbacks when permissions change.
- Split recomputation paths and guard against recursive refresh while recomputing clients or running callbacks.
- Preserve current behavior for `asCheckClientIP=0`, numeric HAG entries, no-source HAG entries, and existing IPv4-only HAG matching.

### CA client search / name-server destinations

- Add a private stateful resolver/cache helper for CA hostname destinations.
- Keep source token, resolved address, resolved/unresolved state, and expiry/retry time instead of discarding the hostname after setup.
- Refresh `EPICS_CA_ADDR_LIST` hostname entries before UDP search sends.
- Refresh `EPICS_CA_NAME_SERVERS` hostname entries before TCP name-server use.
- Disconnect and reconnect a CA name-server circuit if the hostname resolves to a different address.
- Keep unresolved hostname entries present but fail closed until retry succeeds.
- Leave numeric and auto-discovered broadcast destinations static.
- Preserve initial duplicate suppression for resolved destinations.

### Static analysis follow-up

- Use bounded local string handling for environment/token parsing.
- Replace analyzer-flagged zero-initialization patterns with analyzer-friendly initialization.
- Compile DNS TTL parsing only on supported platforms.

## PVXS Follow-up

This PR is scoped to EPICS Base.

PVXS already benefits from the HAG-side access propagation through the existing AS client/check path. A similar follow-up PR will address **stateful PVXS client `addressList` / `nameServers` DNS behavior**.

## Testing

- `make -C epics-base/modules/libcom test`
- `make -C epics-base/modules/libcom/test runtests TESTS=aslibtest`
- `epicsSockResolveTest`
- `make -C epics-base/modules/ca/src/client`
- `make -C epics-base/modules/ca/src/client runtests TESTSCRIPTS_HOST=caStatefulAddrTest.t`
- `git -C epics-base diff --check`

### DIFF
--- a/modules/libcom/src/Makefile
+++ b/modules/libcom/src/Makefile
@@ -61,6 +61,10 @@ API_HEADER += dbCoreAPI.h
 LIBRARY=Com
 
 Com_SYS_LIBS_WIN32 = ws2_32 advapi32 user32 dbghelp
+Com_SYS_LIBS_Darwin += resolv
+Com_SYS_LIBS_Linux += resolv
+Com_SYS_LIBS_freebsd += resolv
+Com_SYS_LIBS_solaris += resolv
 
 Com_RCS = Com.rc
 
--- a/modules/ca/src/client/Makefile
+++ b/modules/ca/src/client/Makefile
@@ -127,7 +127,8 @@ EXPAND_VARS = INSTALL_BIN=$(FINAL_LOCATION)/bin/$(T_A)
 
 SRC_DIRS += $(CURDIR)/test
 TESTPROD_HOST += caStatefulAddrTest
-caStatefulAddrTest_SRCS = caStatefulAddrTest.cpp
+caStatefulAddrTest_SRCS = caStatefulAddrTest.cpp caStatefulAddr.cpp
+caStatefulAddrTest_LIBS = Com
 TESTSCRIPTS_HOST += caStatefulAddrTest.t
 
 PROD_HOST += ca_test
--- a/modules/ca/src/client/Makefile
+++ b/modules/ca/src/client/Makefile
@@ -48,6 +48,7 @@ LIBSRCS += cacWriteNotify.cpp
 LIBSRCS += cacStateNotify.cpp
 LIBSRCS += access.cpp
 LIBSRCS += iocinf.cpp
+LIBSRCS += caStatefulAddr.cpp
 LIBSRCS += convert.cpp
 LIBSRCS += test_event.cpp
 LIBSRCS += repeater.cpp
@@ -94,6 +95,10 @@ ca_RCS = ca.rc
 ca_LIBS = Com
 
 ca_SYS_LIBS_WIN32 = ws2_32 advapi32 user32
+ca_SYS_LIBS_Darwin += resolv
+ca_SYS_LIBS_Linux += resolv
+ca_SYS_LIBS_freebsd += resolv
+ca_SYS_LIBS_solaris += resolv
 
 #   libs needed for PROD and TESTPRODUCT
 PROD_LIBS = ca Com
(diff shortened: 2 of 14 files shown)


derekste:
💚

derekste:
Putting the current PR back to draft. I'd like to finish the existing HAG DNS TTL PR by making refreshed HAG DNS state propagate to already-created ASCLIENTPVT objects. The current PR refreshes HAG hash entries during asComputePvt(), but normal permission checks read cached access/trapMask state, so long-lived CA/PVXS clients can remain stale. The fix should treat HAG DNS expiry as an access-security state transition: refresh HAG entries, recompute all active clients, and fire existing change-of-access callbacks.

derekste:
The earlier PR made HAG DNS entries refreshable, but normal authorization checks still read per-client cached access rights. Long-lived CA/PVXS clients could therefore keep stale permissions until reconnect, reload, or some unrelated AS recompute. This update makes HAG DNS expiry an access-security state transition: ordinary asCheckGet() / asCheckPut() calls now trigger a due refresh, rebuild HAG hash entries, recompute active clients, and fire the existing change-of-access callbacks when permissions actually change. PVXS should benefit through its existing AS client/check path without PVXS-specific changes.

derekste:
crossposting this here for reference https://github.com/epics-base/pvxs/pull/172

modules/libcom/src/Makefile
```diff
@@ -61,6 +61,10 @@ API_HEADER += dbCoreAPI.h
 LIBRARY=Com
 
 Com_SYS_LIBS_WIN32 = ws2_32 advapi32 user32 dbghelp
+Com_SYS_LIBS_Darwin += resolv
+Com_SYS_LIBS_Linux += resolv
+Com_SYS_LIBS_freebsd += resolv
+Com_SYS_LIBS_solaris += resolv
```
mdavidsaver, line 67: Adding a dependency to libCom is likely to cause problems with dependent code which is linking statically.

What extra information does this BSD API provide?  Just TTL?

derekste: Yes, the BSD resolver API is only there for TTL inspection. HAG hostname resolution still goes through the existing aToIPAddr() path, so the selected IPv4 address and matching behavior stay exactly as they were before; the res_query() / ns_* logic is only a follow-up to ask “how long should this chosen result stay valid?”

So the extra information is specifically: the TTL of the selected A-record result, if that selected address is present in the DNS answer. If that information is not available, the code falls back to the fixed refresh/retry intervals already in the PR (300s success fallback, 60s unresolved retry).

I added -lresolv at the libCom level because the implementation lives in modules/libcom/src/as/asLibRoutines.c. If that is too broad for static-link consumers, I’m happy to revisit it. The simplest fallback would be to keep the same stateful-refresh behavior but drop TTL parsing on those builds and use the fixed intervals only, rather than forcing a wider resolver-library dependency.

derekste:
Thank you for the feedback and discussion on this earlier version.  I have
opened #945 as a focused successor.

The replacement isolates the Access Security HAG refresh API, removes the
CA-client destination changes and the `libresolv` dependency, preserves the
installed Access Security structure layouts, and keeps DNS out of ordinary
access checks.  CA-client hostname refresh remains deferred to a separate
asynchronous design.

I am closing this PR so there is one review target.  I will keep the branch
available for historical context until the replacement merges.

Issue #836, opened 2026-03-13
Slow search_reply_udp in RTEMS IOCs


hongran:
When I tested my IOC build with RTEMS7 (the main branch) and libbsd (branch 7-freebsd-14) deployed on mvme3100, I found that it took significant longer time to load all the PVs on a CSSPhoebus screen. After some investigation, I found that it is slowed down by the search_reply_udp function, which is called in the CAS-UDP thread. Further investigation showed that it is the function call osiSufficentSpaceInPool that slows down the search_reply_udp. It takes ~6ms to execute, and this occurs for almost every message sent. 

We are using the POSIX API in RTEMS, and in the epics-base code, the RTEMS-POSIX version of osiSufficentSpaceInPool calls the malloc_info(Heap_Information_block info) function, and this function runs slow. According to some documentation, malloc_info may check the entire heap so it takes long time. 

I think the purpose of the osiSufficentSpaceInPool is to check the free space and determine whether there is enough space. Is there a better way to check it that is faster than malloc_info?

Thanks!

@kiwichris

kiwichris:
Hi @hongran , thanks for the issue and including me.

I am not surprised the call is taking time. The intention of the call is a diagnostic and not in a place where performance is needed. I suspect the more blocks and allocations in the heap the time the call takes.

I have assigned the issue to myself.

kiwichris:
I cannot assign it. If someone could I would appreciate it

anjohnson:
Hi @kiwichris My understanding is that the `osiSufficentSpaceInPool()` routine is being used to check whether the remaining free memory in this IOC should be sufficient to accept another CA client connection — it doesn't really need to know how much RAM is left, just answer a go/no-go question. I assume it's being called by the CA search handler code to stop it replying to name searches from clients when the IOC is in a low-memory state, and that it is also checked before the server will accept a new TCP connection (which requires space for network buffers and 2 additional threads).

If there is a faster way to answer that question in RTEMS that you can point to we can work on coding it. If not we can copy the VxWorks solution from modules/libcom/src/osi/os/vxWorks/osdPoolStatus.c, which runs a background thread at 1Hz to keep track of the largest free memory block that's available.

joelsherrill:
Checking how much free space is available before an allocation is an anti-pattern. The malloc() function returns NULL if the allocation fails. If another thread runs between the check and allocation, memory could be allocated by another thread.

[...]

PR #866, opened 2026-05-05
Slow search_reply_udp in RTEMS IOCs (#836)

hongran:
malloc_info( &info )  takes a long time to get the memory statistics. This is called each time before establishing CA link, which slows down the clients when they try to connect many PVs.
Use malloc_free_space() instead of malloc_info( &info ) in osiSufficentSpaceInPool

### DIFF
--- a/modules/libcom/src/osi/os/RTEMS-posix/osdPoolStatus.c
+++ b/modules/libcom/src/osi/os/RTEMS-posix/osdPoolStatus.c
@@ -33,11 +33,7 @@ LIBCOM_API int epicsStdCall osiSufficentSpaceInPool ( size_t contiguousBlockSize
  */
 LIBCOM_API int epicsStdCall osiSufficentSpaceInPool ( size_t contiguousBlockSize )
 {
-    unsigned long n;
-    Heap_Information_block info;
-
-    malloc_info( &info );
-    n = info.Stats.size - (unsigned long)(info.Stats.lifetime_allocated - info.Stats.lifetime_freed);
+    size_t n = malloc_free_space();
     return (n > (50000 + contiguousBlockSize));
 }
 #endif


joelsherrill:
It is ironic that EPICS has gone back to using _malloc_info()_ to determine free space. This was reported around RTEMS 4.6.3 and _malloc_free_space()_ was added in 4.6.5 to avoid the cost of walking all the used blocks. Back then, the EPICS system in question had ~1500 used blocks and 6 free blocks if I remember correctly. [1]

In general, there is no value in calling this before _malloc()_ or _new_. First, those return NULL if the allocation failed. And yes, _new_ can throw an exception but this can be avoided. Second, if the thread is switched out between _malloc_free_space()_ and _malloc()_/_new)_, there may have been allocations which changed the amount of free space before the thread gets to try the allocation. 

[1] The 4.6.4 release was never made public. We received a robustness report done by a space agency after it was cut but before it was published. This was done to have time to read the report and address all the issues reported there. Funny enough, the report was long but the majority were reporting a fatal error because they would configure more than would git on a 1MB RAM target board. The BSP correctly reported not enough memory to initialize that configuration of RTEMS.

kiwichris:
> It is ironic that EPICS has gone back to using _malloc_info()_ to determine free space. 

It was changed when moving RTEMS Init over to POSIX from the classic API and it is the wrong call. It happens.

> This was reported around RTEMS 4.6.3 and _malloc_free_space()_ was added in 4.6.5 to avoid the cost of walking all the used blocks. Back then, the EPICS system in question had ~1500 used blocks and 6 free blocks if I remember correctly. [1]

And we had no mobile phones, no social media and people going around the moon was a dream :smile:

EPICS has grown and the environments it has to operate in are more complex. For example NFSv4 means LibBSD and that alone adds a lot of blocks to the heap and we are doing that.

> In general, there is no value in calling this before _malloc()_ or _new_. First, those return NULL if the allocation failed. And yes, _new_ can throw an exception but this can be avoided.

My understanding is this approach is a softer landing to avoid hitting the hard limit of no free space in the heap. FreeBSD says your disks are 100% used but it holds a bit back so the system does not wedge itself.

> Second, if the thread is switched out between _malloc_free_space()_ and _malloc()_/_new)_, there may have been allocations which changed the amount of free space before the thread gets to try the allocation.

It is common to implement _low water marks_ in systems and to trigger an alarm if the system moves below a set threshold. A working IOCstats helps monitor an active system. I do not expect this code is wanting to operate close to the edge of no free memory in the heap. 

> [1] The 4.6.4 release was never made public. We received a robustness report done by a space agency after it was cut but before it was published. This was done to have time to read the report and address all the issues reported there. Funny enough, the report was long but the majority were reporting a fatal error because they would configure more than would git on a 1MB RAM target board. The BSP correctly reported not enough memory to initialize that configuration of RTEMS.

I am OK with this change. It is simple and returns the system to previous level of performance. I will however document what I said in today's RTEMS Support meeting that the call may change or go away if we implement a new heap allocator and getting the free space in a timely manner is not possible.

anjohnson (approved):
Good enough for now, we can revisit this later if necessary.

Issue #868, opened 2026-05-08
Possibly incorrect condition when adding events to event queue


simon-ess:
See https://github.com/epics-base/epics-base/blob/ac0fc2effff43c9db116ae68c91f3a9767c991ec/modules/database/src/ioc/db/dbEvent.c#L813
```c
    rngSpace = ringSpace ( ev_que );
    if ( pevent->npend>0u &&
        (ev_que->evUser->flowCtrlMode || rngSpace<=EVENTSPERQUE) ) {
```
`rngSpace` will give the total amount of free space in the queue (which is of size 144 = 4 * 36), but we then compare it to `EVENTSPERQUE` which is 36. This ought not to be the correct test for how full the queue is.

anjohnson:
Claude agrees, but suggested:
> The check is a conservative approximation, assuming 36 monitors. It's "correct" in the sense that it errs on the side of caution (starts flow-controlling earlier than strictly necessary in the common case), but it doesn't actually match the comment's stated intent of checking for "one more of each monitor attached."

It also found this 1-line change that fixes the expression:
```patch
diff --git a/modules/database/src/ioc/db/dbEvent.c b/modules/database/src/ioc/db/dbEvent.c
index 3e5506a2d..e213b03b0 100644
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -810,7 +810,7 @@ static void db_queue_event_log (evSubscrip *pevent, db_field_log *pLog)
      */
     rngSpace = ringSpace ( ev_que );
     if ( pevent->npend>0u &&
-        (ev_que->evUser->flowCtrlMode || rngSpace<=EVENTSPERQUE) ) {
+        (ev_que->evUser->flowCtrlMode || rngSpace <= ev_que->quota / EVENTENTRIES) ) {
         /*
          * replace last event if no space is left
          */
```
It ran the database unit tests with that change which all passed, but on questioning admitted that we don't currently have any test programs which would exercise that code.

I will create a PR with the test code that Claude came up with for dbEvent, and the above fix.

ralphlange:
I agree that the Claude-suggested expression is matching the intention expressed in the comment.

But the design of that segmented queue is pretty weird, and I would think that the original expression is correct and the comment was not worded correctly.

E.g., consider the following situation:
In a new queue block, only one monitor is assigned.
We're in a hiccup situation, so updates for this monitor are filling up the queue.
[...]

Issue #867, opened 2026-05-08
db_queue_event_log seems to drop the most _recent_ event and keep the last one


simon-ess:
See https://github.com/epics-base/epics-base/blob/ac0fc2effff43c9db116ae68c91f3a9767c991ec/modules/database/src/ioc/db/dbEvent.c#L797:
```c
    if (pevent->npend > 0u
            && !dbfl_has_copy(*pevent->pLastLog)
            && !dbfl_has_copy(pLog)) {
        db_delete_field_log(pLog);
        UNLOCKEVQUE (ev_que);
        return;
    }
```

Note that this occurs when we are dealing with reference values in the event queues; if the last one and the current one are both references, we drop the _current_ one instead of the older one, which should potentially result in some out-of-sync states. This probably should be dropping the last one instead, but this ought to be investigated.

ralphlange:
Isn't that the case when (for arrays and such) the field log has a pointer to the record?

If true, you might be formally right, but two field logs pointing to the same piece of data are identical - dropping one or the other wouldn't make a difference.

mdavidsaver:
> but two field logs pointing to the same piece of data are identical

The alarm/time meta-data will not be identical.  So the current code would throw away the newer/latest alarm state.

ralphlange:
Correct.

So - apart from the rare case that only the alarm state changes - this means for an assumed series of field logs in the queue created at times A, B, C:

| Current behavior | comment | Correct behavior | comment |
|-----|-----|-----|-----|
| A | wrong data (points to C data) | A | wrong data (points to C data)  |
| B | wrong data (points to C data) | B | wrong data (points to C data)  |
| C | correct | C | correct  |

After a new field log D arrives:

| Current behavior | comment | Correct behavior | comment |
|-----|-----|-----|-----|
| A | wrong data (points to D data) | A | wrong data (points to D data)  |
| B | wrong data (points to D data) | B | wrong data (points to D data)  |
| C | wrong data (points to D data);<br />update D is dropped | D | correct;<br />update C is dropped  |

You're right.
If the queue gets flushed to the network in that overflow state, we used to send only wrong updates to the client, while the correct  behavior eventually sends a single correct update.
Overall, this behavior of sending updates known to be wrong is awful. Not having a queue would be more transparent

[...]

PR #869, opened 2026-05-12
Add dbEvent tests and fixes

anjohnson:
Adds tests for #867 and #868 with fixes, then more API tests since dbEvent.c didn't have any before now.

### DIFF
--- a/modules/database/src/ioc/db/dbEvent.c
+++ b/modules/database/src/ioc/db/dbEvent.c
@@ -789,12 +789,14 @@ static void db_queue_event_log (evSubscrip *pevent, db_field_log *pLog)
 
     /* if we have an event on the queue and both the last
      * event on the queue and the current event reference
-     * a record field, simply ignore duplicate events.
+     * a record field, replace the queued event so that
+     * the current alarm and timestamp metadata is kept.
      */
     if (pevent->npend > 0u
             && !dbfl_has_copy(*pevent->pLastLog)
             && !dbfl_has_copy(pLog)) {
-        db_delete_field_log(pLog);
+        db_delete_field_log(*pevent->pLastLog);
+        *pevent->pLastLog = pLog;
         UNLOCKEVQUE (ev_que);
         return;
     }
@@ -810,7 +812,8 @@ static void db_queue_event_log (evSubscrip *pevent, db_field_log *pLog)
      */
     rngSpace = ringSpace ( ev_que );
     if ( pevent->npend>0u &&
-        (ev_que->evUser->flowCtrlMode || rngSpace<=EVENTSPERQUE) ) {
+        (ev_que->evUser->flowCtrlMode
+            || rngSpace <= ev_que->quota / EVENTENTRIES) ) {
         /*
          * replace last event if no space is left
          */
--- a/modules/database/test/ioc/db/Makefile
+++ b/modules/database/test/ioc/db/Makefile
@@ -43,6 +43,12 @@ testHarness_SRCS += dbTestIoc_registerRecordDeviceDriver.cpp
 
 PROD_LIBS = dbTestIoc dbCore ca Com
 
+TESTPROD_HOST += dbEventTest
+dbEventTest_SRCS += dbEventTest.c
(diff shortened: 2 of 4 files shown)


modules/database/src/ioc/db/dbEvent.c
```diff
@@ -810,7 +812,8 @@ static void db_queue_event_log (evSubscrip *pevent, db_field_log *pLog)
      */
     rngSpace = ringSpace ( ev_que );
     if ( pevent->npend>0u &&
-        (ev_que->evUser->flowCtrlMode || rngSpace<=EVENTSPERQUE) ) {
+        (ev_que->evUser->flowCtrlMode
+            || rngSpace <= ev_que->quota / EVENTENTRIES) ) {
```
mdavidsaver, line 816: I do not believe this condition is correct.  I think the optimal case would be to reserve one entry for every `evSubscrip` which is currently unassigned or empty.  The proposed condition checks for the former, but not the later.  Doing so would likely need to add an `event_que` level count of non-empty `evSubscrip`.  Something like a "reserve" counter, initialized to `EVENTSPERQUE`, decremented each time an `evSubscrip` becomes not empty, and incremented each time one becomes empty or is cancelled.

mdavidsaver (changes requested):
@anjohnson have you reviewed this new `dbEventTest.c`?

anjohnson:
@mdavidsaver I have reviewed the tests in the first 2 commits and they look okay to me, although I'm not very experienced with the dbEvent code. I accept that Claude's analysis and fix for the issue #868 could be wrong (per your comment above and @ralphlange's comment in the issue) and I'm happy to remove or change that as necessary. The test and fix for #867 might still be worthwhile.

In the absence of any other tests for dbEvent.c I'm hoping we might still review and use versions of some of the other tests that it created in the 4th commit here, but I haven't looked at them as closely.

We should move (some of?) the `EVENT*` macros from dbEvent.c into an `EPICS_PRIVATE_API` area in dbEvent.h to avoid having to duplicate them in dbEventTest.c

modules/database/src/ioc/db/dbEvent.c
ralphlange: Yep. That's what I called the "reasonable simplification" in my comment over in [#868](https://github.com/epics-base/epics-base/issues/868#issuecomment-4440669437).

mdavidsaver: Ok.  I'm fine with leaving this test as is.  Maybe the preceding comment would be:

> Replace last log if this subscription already has an event queued, and flow control blocking or insufficent space for one more event from all possible subscriptions.

mdavidsaver:
> ... I haven't looked at them as closely.

The first thing I look for in an attempt at a deterministic test of dbEvent are the points where the worker thread is forced to stop to allow the queue to (over)fill.  I do not find any.  There is one place where an `epicsEvent` is created and signaled, but there is no associated wait.  So dead code, except...

The calls to `cleanCbContext()` come before the calls to `db_close_events()` which stops the worker thread.  This probably leaves open a possible race between `epicsEventDestroy()` and `epicsEventSignal()`.  Still, I guess this is not too much of an issue since many of the test sequences end with some variation on "this space intentionally left blank" in lieu of cleanup code.  No obvious reason for this.  I know that dbEvent code can be shutdown safely.  cf. the `testMonitor*` code from `dbUnitTest.h`.

`testSingleSubscription()` ends with the comment:

```
    /*
     * Cleanup is intentionally skipped here.
     * The subscriptions have pending events that would be deferred to
     * the event task for cleanup, but no event task is running.
     * The event context is leaked; the test process exits immediately.
     */
```

There is in fact still an event task running.  Also, `testSingleSubscription()` is the first sequence called in `dbEventTest()`, so the test process assuredly does not exit immediately after `testSingleSubscription()`.

`syncEventTask()` is an almost verbatim copy of `testMonitorSync()` from `dbUnitTest.h`, including the "dummylabor" name.  Nothing wrong with this, just a reminder of how LLMs actually work.

I could go on, but I've run out of steam for the day.  There may be some pieces of this `dbEventTest.c` which could be salvaged, but I think not many.

PR #870, opened 2026-05-13
epicsStrPrintEscaped() operate on unterminated string

mdavidsaver:
Until 3a2d225682d79f79c84852dbcb98d54cb250d62f (#310) `epicsStrPrintEscaped()` would operate on an unterminated string.  The added call to `strlen(s)` requires that the input string be null terminated, although `len <= strlen(s)` would work.  
f488765631998c9d0b27a14c13aa412ec81a5fe9 explicitly tests this case.

So far this bug is theoretical in epics-base as all usage pass `len==strlen(s)`.

Which raises a question.  Is this bug worth fixing with this PR?  Or do we instead document that the input string must be null terminated?

### DIFF
--- a/modules/libcom/src/misc/epicsString.c
+++ b/modules/libcom/src/misc/epicsString.c
@@ -234,7 +234,7 @@ int epicsStrPrintEscaped(FILE *fp, const char *s, size_t len)
    if (fp == NULL)
        return -1;
 
-   if (s == NULL || strlen(s) == 0 || len == 0)
+   if (s == NULL || len == 0)
        return 0; // No work to do
 
    while (len--) {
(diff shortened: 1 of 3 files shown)


anjohnson:
I think the original point of taking the `len` argument was to allow the string to contain zero bytes, which ought to be printed as an escaped zero `\0` or similar. This would be useful for something like AsynTrace, where you want to see those zeros; and yes, the asyn module [does exactly that](https://github.com/search?q=repo%3Aepics-modules%2Fasyn%20epicsStrPrintEscaped&type=code).

Please remove the `strlen()`.

mdavidsaver:
> asyn module [does exactly that](https://github.com/search?q=repo%3Aepics-modules%2Fasyn%20epicsStrPrintEscaped&type=code).

So this bug is not theoretical...  @MarkRivers fyi

Since the fix seems simple enough, I'm going to go ahead and merge this PR.

PR #871, opened 2026-05-13
asDump* quote UAG and HAG entries

mdavidsaver:
UAG entries at least may contain characters outside the set `a-zA-Z0-9_\-+:.\[\]<>;` (https://github.com/slac-epics/pvxs-cms/issues/8#issuecomment-4435999344), and since at least 9f01c475421ccfcad25607773e2d9b72a002bce3 can contain escape sequences.  I think that quoted string were always accepted, so unconditionally escape in `asDump*` functions.

### DIFF
--- a/modules/libcom/src/as/asLibRoutines.c
+++ b/modules/libcom/src/as/asLibRoutines.c
@@ -20,6 +20,7 @@
 #include "epicsStdio.h"
 #include "dbDefs.h"
 #include "epicsThread.h"
+#include "epicsString.h"
 #include "cantProceed.h"
 #include "epicsMutex.h"
 #include "errlog.h"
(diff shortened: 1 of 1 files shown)


anjohnson (approved):
I agree that quoting the names will solve the issue. When humans might read the result I try to only quote strings that actually need it when I print them, but that's a tiny refinement and much easier to do in Perl than C.

Where does Codacy get the idea that these UAG and HAG names aren't NULL-terminated?

mdavidsaver:
> Where does Codacy get the idea that these UAG and HAG names aren't NULL-terminated?

No idea.  I consider this a false positive, and will ignore.

ralphlange:
This is a false positive, correct.

It's a very weak pattern in flawfinder that jumps on any `strlen()` unless you write a zero in the string immediately before.

(I marked it as false positive in Codacy - which is the appropriate thing to do unless we want to disable that check completely.)

> Where does Codacy get the idea that these UAG and HAG names aren't NULL-terminated?

It doesn't claim that, actually.
It says that `strlen()` is unsafe on non NULL-terminated strings, which is true, and flags it as the stupid pattern is not smart enough to check if the strings in this case are terminated or not.

I thankfully take that as a reminder to double-check, which I do and then flag it in Codacy as false positive. Done. Overall, I still find that a reasonable help. (But that's personal preference, or noise-tolerance, of course.)

mdavidsaver:
With one approval, and one non-rejection, I am going to merge this almost trivial change.

PR #873, opened 2026-05-13
Add a job to build with INSTALL_LOCATION set in CONFIG_SITE

simon-ess:
We have noticed that the behaviour of INSTALL_LOCATION depends on where it is set; it works differently if it is set as an environment or command-line variable than if set in CONFIG_SITE(.local), despite this being explicitly what is recommended in that file.

This may not be the best approach. One could also update cue.py. This would be necessary as at the moment cue.py only sets environment/commandline variables, so would have to be modified to allow writing to a file.

I have opted for a very not DRY-friendly approach instead of that for the time being since while _in theory_ updating cue.py is nicer, it's not clear that any other feature other than INSTALL_LOCATION really cares about where it is defined.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -237,6 +237,45 @@ jobs:
       if: ${{ always() }}
       run: python .ci/cue.py -T 5M test-results
 
+  install_location:
+    name: "Ub-24 gcc install location"
+    runs-on: ubuntu-24.04
+    env:
+      CMP: gcc
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Automatic core dumper analysis
+        uses: mdavidsaver/ci-core-dumper@master
+      - name: "apt-get install"
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install qemu-system-x86 g++-mingw-w64-x86-64 gdb
+        if: runner.os == 'Linux'
+      - name: Prepare and compile dependencies
+        run: python .ci/cue.py prepare
+      - name: Update CONFIG_SITE.local to set INSTALL_LOCATION
+        run: |
+          cat <<EOF > configure/CONFIG_SITE.local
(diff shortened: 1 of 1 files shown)


anjohnson:
Core: May need updates to GHA actions & node changes. We didn't see this had been added, like it.

anjohnson:
Note to core: Today's GHA rerun of [the new check](https://github.com/epics-base/epics-base/actions/runs/25786079816/job/75739706821?pr=873) passed, so actions and node changes don't seem to be needed at the moment.

Should this job also try to run some programs from the install location? It should be possible to run the generated `modules/database/test/std/rec/O.$EPICS_HOST_ARCH/netget.t` test in `perl` to check the `softIocPVA`, `caget` and `pvget` programs, which it runs from the install location.

.github/workflows/ci-scripts-build.yml
```diff
@@ -237,6 +237,45 @@ jobs:
...
+          sudo apt-get update
+          sudo apt-get -y install qemu-system-x86 g++-mingw-w64-x86-64 gdb
+        if: runner.os == 'Linux'
+      - name: Prepare and compile dependencies
+        run: python .ci/cue.py prepare
+      - name: Update CONFIG_SITE.local to set INSTALL_LOCATION
+        run: |
+          cat <<EOF > configure/CONFIG_SITE.local
+          INSTALL_LOCATION=/tmp/epics
+          EOF
```
mdavidsaver, line 263: It looks like this is the only additional step.  Maybe add it conditionally to the `native:` matrix?

```yaml
      - name: Update CONFIG_SITE.local to set INSTALL_LOCATION
        if: ${{ matrix.install_location }}
...
```

mdavidsaver (changes requested):
> ... so actions and node changes don't seem to be needed at the moment.

Not yet, but it would be good if this PR was rebased and updated to follow 69b91e8074d1f43b0804b41634b823ac8e341736 which quiets some deprecation warnings.

PR #875, opened 2026-05-13
dbAccess.c: Improve logging of illegal field_type

tboegi:
Debugging a problem and looking at the message

recGblDbaddrError: dbPut: Request type is 0 Illegal Database Request Type...
does not help very much: request type 0 is a legal value.

Improve the message to show the field type as well: recGblDbaddrError: dbPut: Request type is 0 field_type=14...

Do the same for both dbPut and dbGet.

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -956,7 +956,9 @@ long dbGet(DBADDR *paddr, short dbrType,
     if (INVALID_DB_REQ(dbrType) || field_type > DBF_DEVICE) {
         char message[80];
 
-        sprintf(message, "dbGet: Request type is %d\n", dbrType);
+        sprintf(message, "dbGet: dbrType = %d, field_type = %.12s (%d).",
+                dbrType, dbGetFieldTypeString(field_type), field_type);
+
         recGblDbaddrError(S_db_badDbrtype, paddr, message);
         status = S_db_badDbrtype;
         goto done;
@@ -1338,7 +1340,9 @@ long dbPut(DBADDR *paddr, short dbrType,
     } else if (INVALID_DB_REQ(dbrType) || field_type > DBF_DEVICE) {
         char message[80];
 
-        sprintf(message, "dbPut: Request type is %d", dbrType);
+        sprintf(message, "dbPut: dbrType = %d, field_type = %.12s (%d).",
+                dbrType, dbGetFieldTypeString(field_type), field_type);
+
         recGblDbaddrError(S_db_badDbrtype, paddr, message);
         return S_db_badDbrtype;
     }


anjohnson:
There is a function `const char * dbGetFieldTypeString(int dbfType)` declared in dbStaticLib.h that converts a field type index into a string. It would be nicer to report the string than the number (14 is "DBF_INLINK" BTW).

A request type of 0 is not legal for the request types used inside the IOC; see dbAccessDefs.h for those bit definitions, and don't confuse them with the similar enum that is used by CA clients. I'm not sure that we have a way to stringify the IOC request type values.

tboegi:
@anjohnson Thanks for the review.
I don't think that trying to translate possible illegal numeric
values into a string (which one ?) gives us much.
 I have seen this message once within 12 years, and digging up
 what is what is not that hard.
 Please let us know if this PR is counted as 
 "improvement, will be merged" or
 "improvement, but will not be merged as is"

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -956,7 +956,7 @@ long dbGet(DBADDR *paddr, short dbrType,
     if (INVALID_DB_REQ(dbrType) || field_type > DBF_DEVICE) {
         char message[80];
 
-        sprintf(message, "dbGet: Request type is %d\n", dbrType);
+        sprintf(message, "dbGet: Request type is %d field_type=%d", dbrType, field_type);
```
anjohnson: I would edit the `sprintf()` calls to something like this:
```
sprintf(message, "dbGet: dbrType = %d, target field is %.12s (%d)",
    dbrType, dbGetFieldTypeString(field_type), field_type);
```

mdavidsaver:
> I don't think that trying to translate possible illegal numeric
> values into a string (which one ?) gives us much.

`dbGetFieldTypeString()` should only be applied to `field_type` (DBF), not `dbrType` (DBR).  The later has different meanings after `DBR_ENUM`.

> "improvement, will be merged" or

I see an improvement either way.

anjohnson (commented):
I was wrong about the request type value in my comments last night, getting that confused with the `long options` bitmap; 0 is of course `DBR_STRING` as I would expect from your database. I probably shouldn't be responding to issues/PRs after 9pm…

I'm all for making the IOC's error messages better and easier to understand. Adding the string translation and adjusting the message strings as shown inline would make this definitely worth merging.

modules/database/src/ioc/db/dbAccess.c
tboegi: Lets try a new version, which gives
recGblDbaddrError: dbPut: dbrType = 0, field_type = DBF_INLINK (14). Illegal Database Request Type PV: IOC:m1.DOL

mdavidsaver (approved):
Looks ok to me.

PR #877, opened 2026-05-15
Fix #859 in template CONFIG file

### LINKED ISSUE #859: Support modules that set INSTALL_LOCATION in CONFIG_SITE(.local) do not correctly load custom CFG rules
See https://github.com/epics-base/pvxs/issues/165, https://github.com/epics-base/epics-base/issues/724.

In short, if we override `INSTALL_LOCATION` locally within a module (as is suggested in our templated `CONFIG_SITE` files, see e.g. https://github.com/ChannelFinder/recsync/blob/master/client/configure/CONFIG_SITE#L25, https://github.com/epics-base/epics-base/blob/7.0/src/template/base/top/configure/CONFIG_SITE#L28) then the custom CFG files are not loaded correctly.

This arises in PVXS due to the custom CFG files implementing a strict check that CONFIG_PVXS_MODULE has been loaded; if you do not override INSTALL_LOCATION (or of you do it from the command line) then it loads correctly. If you do not, it does not.

This can be tested by adding two files to a configure directory:
```make
# CONFIG_OOPS
OOPS=BAR
```
and
```make
# RULES_OOPS
ifeq ($(strip $(OOPS)),)
$(error OOOPSIE)
endif
```
and add these to the `CFG` variable, then:
* If you override INSTALL_LOCATION you will get an error
* If you do not you do not

In particular, if you add `$(info ...)` lines to see when these are loaded, only the RULES file is loaded, and even that is only done on a _second_ pass through.

Note that this is visible in https://github.com/epics-base/pvxs/issues/165 due to the cross-compilation; on the first pass through for the host architecture you install the CFG files; on the second pass the RULES file is loaded and the error happens.

anjohnson:
Fixes #859, build breakage when INSTALL_LOCATION is set.

Instructions on updating existing modules in Notes entries.

Also prints warning messages from the modules/Makefile if $(INSTALL_LOCATION) is empty.

### DIFF
--- a/configure/RULES_MODULES
+++ b/configure/RULES_MODULES
@@ -25,10 +25,19 @@
 # Add checked-out submodules to DIRS, unless INSTALL_LOCATION is empty
 LIVE_SUBMODULES = $(subst /Makefile,, \
     $(wildcard $(addsuffix /Makefile, $(SUBMODULES))))
-live = $(if $(wildcard $(INSTALL_CONFIG)/RULES_TOP),LIVE,DEAD)
+live := $(if $(wildcard $(INSTALL_CONFIG)/RULES_TOP),LIVE,DEAD)
 DIRS += $($(live)_SUBMODULES)
 
 include $(CONFIG)/RULES_DIRS
+
+ifeq ($(live),DEAD)
+ifneq ($(filter $(ACTIONS),$(MAKECMDGOALS)),)
+  $(info *** WARNING: The install directory $(INSTALL_LOCATION))
+  $(info ***   is absent or empty. The "make $(MAKECMDGOALS)" rule)
+  $(info ***   cannot recurse into these external modules as a result:)
+  $(info ***     $(LIVE_SUBMODULES))
+endif
+endif
 
 RELEASE_LOCAL := RELEASE.$(EPICS_HOST_ARCH).local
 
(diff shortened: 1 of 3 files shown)


mdavidsaver (approved):
Looks ok to me.

simon-ess:
Same here.

ralphlange:
Fine with me.
Codacy seems to have a point, though: trailing whitespace in `documentation/new-notes/
fix-859.md`, line 12.

configure/RULES_MODULES
```diff
@@ -25,11 +25,22 @@
...
-live = $(if $(wildcard $(INSTALL_CONFIG)/RULES_TOP),LIVE,DEAD)
+live := $(if $(wildcard $(INSTALL_CONFIG)/RULES_TOP),LIVE,DEAD)
 DIRS += $($(live)_SUBMODULES)
 
 include $(CONFIG)/RULES_DIRS
 
+ifeq ($(live),DEAD)
+ifneq ($(filter $(ACTIONS),$(MAKECMDGOALS)),)
+    z :=
+    i := $z   #
```
simon-ess: Strictly speaking you don't need the `i` variable. You can just do
```make
z :=
$(info $z       this is indented)
```

anjohnson:
Group: Fix the trailing white-space, look at Simon's suggestion.

PR #882, opened 2026-05-27
fix build on Windows using cmd shell

### LINKED ISSUE #881: Modules build broken on Windows
Two subtle changes to `configure/RULES_MODULE` broke the generated file `modules/RELEASE.$(EPICS_HOST_ARCH).local` on Windows.

In the rule for `$(RELEASE_LOCAL)`, commit d44ddf44 added quotes and a terminating space (before the output redirect) to the string written to the file. On Windows, unlike on Linux, the `echo` command prints quotes and terminating spaces verbatim, so that the generated file content looks something like: `"EPICS_BASE = C:/Users/zimoch/epics-base" ` (note the terminating space after the closing quote).

Later in submodules like pvData, this is parsed in unexpected ways and leads to the error:
```
configure/CONFIG:17: /configure/CONFIG: No such file or directory
gmake: *** No rule to make target '/configure/CONFIG'. Stop.
```

The quotes cause `$(EPICS_BASE)` to be empty in the module. But even without the quotes, the terminating space causes the `include` command to have 2 arguments, a directory (which leads to a "Permission denied" error when trying to include it) and a non-existing file.

dirk-zimoch:
Fix usage of quotes incompatible with the `echo` command as implemented by the Windows `cmd` shell.
Fixes #881 
and creation of broken dependency files which probably never worked as intended on Windows when using the `cmd` shell.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -175,10 +175,11 @@ jobs:
 
           # Windows builds
 
-          - name: "Win-22 MSC-22"
+          - name: "Win-22 MSC-22 (cmd shell)"
             os: windows-2022
             cmp: vs2022
             configuration: default
+            extra: "SHELL=cmd"
 
           - name: "Win-22 MSC-22, static"
             os: windows-2022
--- a/configure/RULES.Db
+++ b/configure/RULES.Db
@@ -223,60 +223,60 @@ realclean: clean
 %Record.h$(DEP): $(COMMON_DIR)/%Record.dbd $(DBDTORECTYPEH_dep)
 	@$(RM) $@
 	@$(DBTORECORDTYPEH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 %Record.h$(DEP): %Record.dbd $(DBDTORECTYPEH_dep)
 	@$(RM) $@
 	@$(DBTORECORDTYPEH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 %Record.h$(DEP): ../%Record.dbd $(DBDTORECTYPEH_dep)
 	@$(RM) $@
 	@$(DBTORECORDTYPEH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 menu%.h$(DEP): $(COMMON_DIR)/menu%.dbd $(DBDTOMENUH_dep)
 	@$(RM) $@
 	@$(DBTOMENUH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 menu%.h$(DEP): menu%.dbd $(DBDTOMENUH_dep)
 	@$(RM) $@
 	@$(DBTOMENUH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 menu%.h$(DEP): ../menu%.dbd $(DBDTOMENUH_dep)
 	@$(RM) $@
 	@$(DBTOMENUH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 %.dbd$(DEP): %.dbd.pod
 	@$(RM) $@
-	@echo "$(COMMONDEP_TARGET): ../Makefile" > $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile > $@.stdout
 	@$(MV) $@.stdout $@
 
 %.dbd$(DEP): %Include.dbd
 	@$(RM) $@
 	@$(DBEXPAND) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
 	@$(MV) $@.stdout $@
 
 %.dbd$(DEP): ../%Include.dbd
 	@$(RM) $@
 	@$(DBEXPAND) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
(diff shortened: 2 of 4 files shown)


anjohnson:
Core: Accept the change, please add any other similar issues before we merge this.

anjohnson:
Q: Why didn't any of our CI builds fail this?

dirk-zimoch:
Commit d44ddf4 had not changed any other quotes or spaces, but I checked anyway for other undesired side effects of Windows `cmd` printing quotes and spaces (including additional leading spaces starting from the second space after `echo`, and trailing spaces for example between last word and output redirection `>`) verbatim:

 * `gmake help` prints the quotes to the screen. Ugly but harmless.
 * Many debug output lines do the same, e.g `"Installing ..."`. Harmless.
 * `Rules.Db` generates `*.d` files containing `"file: ../Makefile" ` (with quotes). That probably never worked. (It means that a file whose path starts with a quote depends on a file whose name ends with a quote, none of those exist.) Modifying `../Makefile` does not cause the file to be remade. Probably nobody ever noticed since it had been added in 2008. The CI tests do not cover that. Without the quotes it works.

Of course if instead of `cmd` the shell used is `bash`, everything is fine.

`Rules.Db` is fixable. Simply remove the quotes.

Fixing debug and help output is a bit more tricky. Required escaping differs between `cmd` and other shells:
 * Literal quotes: `"` vs `\"`
 * Special characters`^>` vs `\>`
 * Empty lines: `echo:`(no space!) vs `echo` without arguments
 * Indentation (leading space): `    text` (just spaces) vs `"    text"` (quoted spaces)

Finding out which shell is called by `gmake` is not trivial. Examining the `SHELL` variable does not help much (contains `sh.exe` on Windows when using `cmd`). It can only be determined reliably by observing the behavior of the `echo` command like in `ifeq ($(shell echo "test"), "test")`.

dirk-zimoch:
Fixing help text and debug output would become really ugly. I will leave it as it is because the problem is purely cosmetic.

dirk-zimoch:
> Q: Why didn't any of our CI builds fail this?

According to the `gmake` documentation if it finds `sh` in the `PATH` on Windows, it uses it. Else it uses the „ridiculously limited“ Windows command shell. On github, bash is installed in the Windows runners, which probably puts `sh` in the `PATH`. On my computer I had no `bash`.

https://www.gnu.org/software/make/manual/make.html#Choosing-the-Shell

mdavidsaver:
> According to the `gmake` documentation if it finds `sh` in the `PATH` on Windows, it uses it.

Could we override this for one of the CI jobs?  `make SHELL=cmd`?

configure/RULES.Db
```diff
@@ -223,60 +223,60 @@ realclean: clean
 %Record.h$(DEP): $(COMMON_DIR)/%Record.dbd $(DBDTORECTYPEH_dep)
 	@$(RM) $@
 	@$(DBTORECORDTYPEH) -D $(DBDFLAGS) -o $(COMMONDEP_TARGET) $< > $@.stdout
-	@echo "$(COMMONDEP_TARGET): ../Makefile" >> $@.stdout
+	@echo $(COMMONDEP_TARGET): ../Makefile >> $@.stdout
```
anjohnson, line 226: Shouldn't you also remove the spaces before the `>>` operators since the output is going to a file? Maybe it doesn't matter in this case because the generated `.d` file is parsed by `make` which will ignore them, but I would consider doing this for all `@echo` redirections to a file, setting a pattern that later maintainers will hopefully see and follow.

These are the kinds of local practices that probably ought to be documented somewhere, but it feels a bit petty have what could be a long list of them all in the CONTRIBUTORS.md file.

simon-ess:
@mdavidsaver - that seems a great idea. I'll create an issue to remind us to do that.

configure/RULES.Db
dirk-zimoch: The space is indeed copied into the .d file but make has no problem with that. Trailing (or leading) spaces are only a problem if the value is concatenated with another string. Maybe we should better sanitize all strings read from a file with `$(strip)` before concatenating them or use `$(join)` or `$(addsuffix)` to concatenate the strings.

dirk-zimoch:
> Could we override this for one of the CI jobs? `make SHELL=cmd`?

Fails (without this PR) as expected:
https://github.com/dirk-zimoch/epics-base/actions/runs/26625721239/job/78461851484

dirk-zimoch:
I could fix the ugly formatting of `gmake help` by replacing `echo` with `$(info)`.

configure/RULES_TOP
```diff
@@ -77,42 +77,43 @@ rm-failure-file:
...
+	$(info $()     distclean        - Does realclean cvsclean realuninstall and deletes any)
+	$(info $()                        generated modules/RELEASE.<host>.local files)
 endif
-	@echo "     help           - Prints this list of valid make targets "
-	@echo "Object targets are supported by the O.<arch> level Makefile .e.g"
-	@echo "     xxxRecord.o"
+	$(info $()     help             - Prints this list of valid make targets )
+	$(info $()Object targets are supported by the O.<arch> level Makefile .e.g)
+	$(info $()     xxxRecord.o)
+	@:
```
dirk-zimoch, line 116: `@:` gives make "something" to do in order to avoid message `make: Nothing to be done for 'help'.`
`:` does nothing (successfully) in `cmd` as well as in posix shells.

`$()` expands to nothing because you cannot define a variable with empty name but you can use it and it allows leading spaces.

anjohnson: An alternative approach: I discovered recently that both GNU Coreutils and Bash provide the command `printf` which allows the use of string widths and justification in the format string argument. For Windows builds that don't use bash, providing a `printf.pl` script in Base should be pretty trivial.

I've started using that for `make help` output in some local projects because I don't have to manually realign the text.

Example excerpts, but I'm not suggesting we follow exactly what I use here, and I doubt that Windows has `tput`:

```makefile
# Variables for help text
b := $(shell tput bold)
r := $(shell tput sgr0)
P = printf "    $b%-25s$r - %s\n"

help:
        @echo "Usage: make [target]"
        @echo "The $(NAME) IOCs run on: $(HOSTS)."
        @echo "Targets for general use:"
        @$P help "Print this list of make targets"
        @$P hosts "List all hosts used"
        @$P iocs "List all IOCs managed"
        @$P units "List all unit files installed"
```

The output looks like this (the target names are in **bold** if the terminal supports that):

```
voltctl% make help
Usage: make [target]
The xbpm IOCs run on: c2ioc07.
Targets for general use:
    help                      - Print this list of make targets
    hosts                     - List all hosts used
    iocs                      - List all IOCs managed
    units                     - List all unit files installed
```

configure/RULES.Db
anjohnson: Janet Anderson used to use `$(strip …)` even in places where it wasn't necessary (or maybe GNUmake changed to handle that automatically in more places). I found that annoying since it reduces code readability, so I have removed many of those calls over the years.

simon-ess (approved):

configure/RULES.Db
dirk-zimoch: Also using `$(strip ...)` is actually fragile. It needs to be used correctly in all other modules in order to fix a problem that should not have been created by base in the first place.

configure/RULES_TOP
dirk-zimoch: Not yet another dependency plus workaround for Windows please.

.github/workflows/ci-scripts-build.yml
```diff
@@ -175,10 +175,11 @@ jobs:
 
           # Windows builds
 
-          - name: "Win-22 MSC-22"
+          - name: "Win-22 MSC-22 (cmd shell)"
             os: windows-2022
             cmp: vs2022
             configuration: default
+            extra: "SHELL=cmd"
```
mdavidsaver, line 182: Hopefully this is a silly question.  How can we tell if `cmd` is actually used by `make`?

At least something to check occasionally?  Maybe leave the quote on one line of `make help`?

simon-ess: Good question, but I would say that the documentation is pretty clear about this: https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

dirk-zimoch: If the messages `Installing ...` are quoted like `"Installing ..."` in the log, then it is `cmd`.
One could add `make PRINT.SHELL` at the beginning of the build process, but that does not really help, because if it is the default `'sh.exe'`, still `cmd` may be used if there is no `sh.exe` in the `PATH`.
Observing if `echo` prints quotes is the best way to find out which shell is really used.

mdavidsaver:
```
  make[2]: Entering directory 'D:/a/epics-base/epics-base/configure/O.windows-x64'
  "Installing CFG file ../../cfg/TOOLCHAIN.windows-x64.windows-x64"
```

```
   make[2]: Entering directory 'D:/a/epics-base/epics-base/configure/O.windows-x64-debug'
  Installing CFG file ../../cfg/TOOLCHAIN.windows-x64-debug.windows-x64-debug
```

This is the kind of test I had in mind.

mdavidsaver (approved):
Looks ok to me.

PR #885, opened 2026-05-29
Shorten CONFIG_ENV for pthread priority scheduling

jwlodek:
The current name is read by iocStats as a PV suffix, and is long enough that even reasonably short PV prefixes overshoot the 60 character limit


ralphlange:
Are you (still) aware of https://github.com/epics-modules/iocStats/issues/75 ?

I am in favor of fixing an issue in one place, if possible.
@anjohnson, @simon-ess - what do you think?

jwlodek:
I've made a PR to iocStats that truncates the variable name in the DESC field to avoid the 40 character limit. The issue is that the long variable name plus a prefix approaches the 60 character record name limit. I don't know if truncating the name in iocStats for the record name is a good idea - what is a reasonable record length given the prefix length is not known at build time? A truncated name also may lose the context/usefulness of reporting the value of the env var.

ralphlange:
Ok, so it is a different issue, sorry.

Generally, I don't like the idea of changing Base because of an issue in a Device Support module. The problem is using an unabridged environment variable name (very high length limit that is not even fixed) in narrow places (40 or 60 char length limit).

This is caused by the limits from Base, obviously.
But the proper fix for that would be making the string fields in the EPICS database variable-length.

mdavidsaver:
> Generally, I don't like the idea of changing Base because of an issue in a Device Support module.

Agreed.

Personally, I think it would be preferable if iocStats would switch to a [manual variable list, and abbreviate some names](https://github.com/mdavidsaver/linStat/blob/bf2ba2b1859a98475228f759e82fcfdb99e1d742/statApp/Db/linStatProc.substitutions#L98-L132).

If we were to accept a retroactive name change in this case, then we should also adopt an upper limit on any future CONFIG variable names.  I would not mind this part as I think `EPICS_ALLOW_POSIX_THREAD_PRIORITY_SCHEDULING` is excessively long.

ralphlange:
> I think `EPICS_ALLOW_POSIX_THREAD_PRIORITY_SCHEDULING` is excessively long.

I agree. I was simply using the variable name in the EPICS build system, adding the usual prefix. Too bad that the [review of the PR in November](https://github.com/epics-base/epics-base/pull/737) did not turn up anything.

As a consequence, for CODAC Core System I will have to add a paragraph to the migration guide, create and test a migration script for applications (including managing this in our ticketing system) and create new versions of the user guides of the two applications that are using the feature, then push all documents through the review process. Total of 6-8 hours between three people over two months.
Note: We are not using this feature on IOCs and we are not using the relevant part of iocStats. For the parts of iocStats that we use, all record names have been changed anyway.

simon-ess:
As the one who implemented the iocstats change to fetch vars from EPICS base, I simply didn't want to maintain a duplicate list of variables. However, I didn't really reckon with how long these variable names can be...

My preference would of course be to limit config variable lengths, but if that isn't feasible I can revert my changes to iocstats. It would probably also be good as much as possible if linstats and iocstats didn't diverge toooooo much.

jwlodek:
I don't have a strong opinion on where the fix should be applied, but at the moment we cannot use the most recent base with the most recent IOC stats. We'll likely downgrade to iocStats 3.** in the interim, but I suspect we won't be the only ones to run into this. It would be good to decide on a solution and apply the fix + have a patch release when possible.

mdavidsaver:
> As the one who implemented the iocstats change to fetch vars from EPICS base, I simply didn't want to maintain a duplicate list of variables. However, I didn't really reckon with how long these variable names can be...

I understand the impulse for brevity, but we don't add variables that often :)


> It would probably also be good as much as possible if linstats and iocstats didn't diverge toooooo much.

This is my goal as well.

anjohnson:
Core Group: After long discussion we conclude that this problem should be fixed in iocStats, possibly by ignoring names that are too long to be represented in the DESC field, or by shortening that string. Thus we are closing this PR. Thanks for making us discuss it and coming to a proper solution in the correct place.

PR #887, opened 2026-06-02
biRecord.c/boRecord.c: Fix gcc -Wsizeof-pointer-memaccess

AbdallaDalleh:
On RL8 and 9, `gcc` prints the following warning:
```
../rec/biRecord.c:191:42: warning: argument to ‘sizeof’ in ‘strncpy’ 
call is the same expression as the source; did you mean to use the 
size of the destination? [-Wsizeof-pointer-memaccess]
  191 |         strncpy(pstring,prec->znam,sizeof(prec->znam));
      |                                          ^
```
No way to tell `pstring` size since it is provided from support modules. `snprintf` is used instead to copy `prec->znam` to `pstring`.

The warning was reproduced and tested outside of EPICS base with this simple snippet
```c
#include <stdio.h>
#include <string.h>

#define MAX_STRING_SIZE 40

typedef struct
{
    char znam[26];
} biRecord;

long get_enum_str(biRecord *prec, char *pstring)
{
    // Same strncpy generates -Wsizeof-pointer-memaccess
    // strncpy(pstring,prec->znam,sizeof(prec->znam));
    snprintf(pstring, MAX_STRING_SIZE, "%s", prec->znam);
    pstring[sizeof(prec->znam)] = 0;
    return(0);
}

int main()
{
    biRecord record;
    char buffer[2];
    snprintf(record.znam, sizeof(record.znam), "HELLO WORLD");

    get_enum_str(&record, buffer);

    printf("Buffer: %s\n", buffer);

    return 0;
}
```

### DIFF
--- a/modules/database/src/std/rec/biRecord.c
+++ b/modules/database/src/std/rec/biRecord.c
@@ -35,6 +35,7 @@
 #include "recSup.h"
 #include "recGbl.h"
 #include "special.h"
+#include "epicsTypes.h"
 
 #define GEN_SIZE_OFFSET
 #include "biRecord.h"
@@ -188,10 +189,10 @@ static long get_enum_str(const DBADDR *paddr, char *pstring)
     if(index!=biRecordVAL) {
         strcpy(pstring,"Illegal_Value");
     } else if(*pfield==0) {
-        strncpy(pstring,prec->znam,sizeof(prec->znam));
+        snprintf(pstring, MAX_STRING_SIZE, "%s", prec->znam);
         pstring[sizeof(prec->znam)] = 0;
     } else if(*pfield==1) {
-        strncpy(pstring,prec->onam,sizeof(prec->onam));
+        snprintf(pstring, MAX_STRING_SIZE, "%s", prec->onam);
         pstring[sizeof(prec->onam)] = 0;
     } else {
         strcpy(pstring,"Illegal_Value");
@@ -205,9 +206,9 @@ static long get_enum_strs(const DBADDR *paddr,struct dbr_enumStrs *pes)
 
     pes->no_str = 2;
     memset(pes->strs,'\0',sizeof(pes->strs));
-    strncpy(pes->strs[0],prec->znam,sizeof(prec->znam));
+    snprintf(pes->strs[0], MAX_STRING_SIZE, "%s", prec->znam);
     if(*prec->znam!=0) pes->no_str=1;
-    strncpy(pes->strs[1],prec->onam,sizeof(prec->onam));
+    snprintf(pes->strs[1], MAX_STRING_SIZE, "%s", prec->onam);
     if(*prec->onam!=0) pes->no_str=2;
     return(0);
 }
--- a/modules/database/src/std/rec/biRecord.c
+++ b/modules/database/src/std/rec/biRecord.c
@@ -35,7 +35,6 @@
 #include "recSup.h"
 #include "recGbl.h"
 #include "special.h"
-#include "epicsTypes.h"
 
 #define GEN_SIZE_OFFSET
 #include "biRecord.h"
@@ -186,16 +185,16 @@ static long get_enum_str(const DBADDR *paddr, char *pstring)
 
 
     index = dbGetFieldIndex(paddr);
-    if(index!=biRecordVAL) {
-        strcpy(pstring,"Illegal_Value");
-    } else if(*pfield==0) {
-        snprintf(pstring, MAX_STRING_SIZE, "%s", prec->znam);
+    if(index != biRecordVAL) {
+        strcpy(pstring, "Illegal_Value");
+    } else if(*pfield == 0) {
+        memcpy(pstring, prec->znam, sizeof(prec->znam));
         pstring[sizeof(prec->znam)] = 0;
-    } else if(*pfield==1) {
-        snprintf(pstring, MAX_STRING_SIZE, "%s", prec->onam);
+    } else if(*pfield == 1) {
+        memcpy(pstring, prec->onam, sizeof(prec->onam));
         pstring[sizeof(prec->onam)] = 0;
     } else {
-        strcpy(pstring,"Illegal_Value");
+        strcpy(pstring, "Illegal_Value");
     }
     return(0);
 }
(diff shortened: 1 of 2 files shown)


modules/database/src/std/rec/biRecord.c
```diff
@@ -35,6 +35,7 @@
 #include "recSup.h"
 #include "recGbl.h"
 #include "special.h"
+#include "epicsTypes.h"
```
anjohnson: This addition isn't necessary, the `MAX_STRING_SIZE` macro must be available before the `struct dbr_enumStrs` type can be defined in the dbAccessDefs.h header (as included by dbAccess.h).

modules/database/src/std/rec/biRecord.c
```diff
@@ -205,9 +206,9 @@ static long get_enum_strs(const DBADDR *paddr,struct dbr_enumStrs *pes)
 
     pes->no_str = 2;
     memset(pes->strs,'\0',sizeof(pes->strs));
-    strncpy(pes->strs[0],prec->znam,sizeof(prec->znam));
+    snprintf(pes->strs[0], MAX_STRING_SIZE, "%s", prec->znam);
     if(*prec->znam!=0) pes->no_str=1;
-    strncpy(pes->strs[1],prec->onam,sizeof(prec->onam));
+    snprintf(pes->strs[1], MAX_STRING_SIZE, "%s", prec->onam);
```
anjohnson: The `snprintf()` function is significantly slower than `strncpy()` or `memcpy()`. The `struct dbr_enumStrs` does guarantee that the `pes->strs[]` elements are large enough to hold the `znam` and `onam` strings so there's no problem with the target buffer length here, but the copying should be done more efficiently. Using `memcpy()` would be functionally correct and avoid the warning messages:
```C
      memcpy(pes->strs[0], prec->znam, sizeof(prec->znam));
      if (*prec->znam != 0) pes->no_str = 1;
      memcpy(pes->strs[1], prec->onam, sizeof(prec->onam));
```

modules/database/src/std/rec/biRecord.c
```diff
@@ -188,10 +189,10 @@ static long get_enum_str(const DBADDR *paddr, char *pstring)
     if(index!=biRecordVAL) {
         strcpy(pstring,"Illegal_Value");
     } else if(*pfield==0) {
-        strncpy(pstring,prec->znam,sizeof(prec->znam));
+        snprintf(pstring, MAX_STRING_SIZE, "%s", prec->znam);
         pstring[sizeof(prec->znam)] = 0;
     } else if(*pfield==1) {
-        strncpy(pstring,prec->onam,sizeof(prec->onam));
+        snprintf(pstring, MAX_STRING_SIZE, "%s", prec->onam);
         pstring[sizeof(prec->onam)] = 0;
```
anjohnson, line 195: See my comments about the `get_enum_strs()` routine; please use `memcpy()` here too.

anjohnson (changes requested):
My comments here are for the bi record, but the same code appears in the boRecord.c file. You could also look at the mbbi and mbbo records which make similar calls to `strncpy()` in their `get_enum_str()` and `get_enum_strs()` routines.

Thanks!

modules/database/src/std/rec/biRecord.c
mdavidsaver: Agreed.  Use of `snprintf()` is not appropriate.

@AbdallaDalleh Please look at the changes to `modules/database/src/ioc/db/dbAccess.c` in 8c9dd71f6e790dbe5d5963b0b569d84125a18447 which resolved a similar warning on the caller side of this API.  Add some `STATIC_ASSERT()` so that a build will fail obviously if ZNAM or ONAM is ever resized without due consideration.

anjohnson:
Another reason to not use `snprintf()` is that apparently the ancient MS Visual Studio 2013 compiler didn't support it, causing of most of the failures reported by the @AppVeyorBot.

modules/database/src/std/rec/biRecord.c
AbdallaDalleh: This is true for `struct dbr_enumStrs` in `get_enum_strs`, but for `get_enum_str`, if `pstring` is passed with small buffer this would cause serious problems, at the least overwriting local variables on the stack. For example:
```c
char znam[26];
snprintf(znam, sizeof(znam), "HELLO WORLD");
char buffer[2];
memcpy(buffer, znam, sizeof(znam));
```
The issue is that `sizeof(buffer) = 2` and `strlen(buffer) = 11`. You can extend it and have multiple buffers, for example take this `main` function for the snippet I provided:
```c
int main()
{
    biRecord record;
    char buffer[2];
    char buffer2[2];
    snprintf(record.znam, sizeof(record.znam), "HELLO WORLD");

    get_enum_str(&record, buffer);
    get_enum_str(&record, buffer2);

    printf("Buffer: %s\n", buffer);
    printf("Buffer: %s\n", buffer2);

    return 0;
}
```
It prints:
```
buffer: O WORLD
buffer: LLO WORLD
```

I updated the PR with `memcpy`, but I think this would require us to assume that the caller respects the API contract, i.e., pass a string with size `MAX_ENUM_STRING_SIZE`, what do you think?

mdavidsaver: > I think this would require us to assume that the caller respects the API contract, i.e., pass a string with size `MAX_ENUM_STRING_SIZE`, what do you think?

Correct.  The current definition of `dbGet()` has a complex contract, with many possibilities for programmer error.  fyi. see `dbPutGetTest.c` in Base, as well as usages in RSRV and QSRV.

anjohnson: Agreed; changing `strncpy()` into `memcpy()` in `get_enum_strs()` doesn't fix that problem, but it doesn't make it any worse. I asked Claude to check yesterday and it confirmed that all the current uses inside Base itself do provide an appropriately-sized buffer when fetching an enum as a string.

A proper fix isn't possible without requiring many out-of-tree modules to be updated at the same time, and that's not something we do very often here.

anjohnson (approved):
This looks good to me now, thanks!

PR #888, opened 2026-06-02
Support Windows on ARM64

### LINKED ISSUE #883: Support for windows-arm64
Windows can run on arm64 too. I think, EPICS should support that.

Some minor fixes needed:
 * EpicsHostArch.pl: strawberry Perl runs in x64 emulator mode and mis-identifies Windows on arm64 as "windows-x64"
 * epicsAtomicsCD.h: `epicsAtomicMemoryBarrier` needs implementation for windows-arm64
 * calcPerform: disable buggy optimization for arm64 as already for windows-x64
 * configure/os/\*windows-arm64\* files needed
 * github runner for windows-arm64 (requires update of .ci submodule)

dirk-zimoch:
Resolves #883.

Waiting for https://github.com/epics-base/ci-scripts/issues/114 to be resolved.

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -210,6 +210,21 @@ jobs:
             cmp: gcc
             configuration: default
 
+          - name: "Win11 (aarch64)"
+            os: windows-11-arm
+            cmp: vs2022
+            configuration: default
+
+          - name: "Win11 (aarch64), debug"
+            os: windows-11-arm
+            cmp: vs2022
+            configuration: debug
+
+          - name: "Win11 (aarch64), static"
+            os: windows-11-arm
+            cmp: vs2022
+            configuration: static
+
     steps:
     - uses: actions/checkout@v6
       with:
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,5 +24,5 @@
     branch = master
 [submodule ".ci"]
     path = .ci
-    url = https://github.com/epics-base/ci-scripts
-    branch = master
+	url = https://github.com/ralphlange/ci-scripts.git
+	branch = windows-11-arm-support-10659373095315842272
--- a/modules/libcom/src/calc/calcPerform.c
+++ b/modules/libcom/src/calc/calcPerform.c
@@ -35,7 +35,7 @@ static int cond_search(const char **ppinst, int match);
 #endif
 
 /* Turn off global optimization for 64-bit MSVC builds */
-#if defined(_WIN32) && defined(_M_X64) && !defined(_MINGW)
+#if defined(_WIN64) && !defined(_MINGW)
 #  pragma optimize("g", off)
 #endif
 
@@ -422,7 +422,7 @@ LIBCOM_API long
     return 0;
 }
 
-#if defined(_WIN32) && defined(_M_X64) && !defined(_MINGW)
+#if defined(_WIN64) && !defined(_MINGW)
 #  pragma optimize("", on)
 #endif
 
--- a/modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
+++ b/modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
@@ -17,8 +17,6 @@
 #ifndef epicsAtomicCD_h
 #define epicsAtomicCD_h
 
-#include "epicsAssert.h"
-
 #ifndef _MSC_VER
 #   error this header file is only for use with with the Microsoft Compiler
 #endif
@@ -27,33 +25,19 @@
 
 #define EPICS_ATOMIC_CMPLR_NAME "MSVC-INTRINSIC"
 
-#include <intrin.h>
+// Avoid including old winsock.h automatically with windows.h
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
 
-#if defined ( _M_IX86 )
-#   pragma warning( push )
-#   pragma warning( disable : 4793 )
-    EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
-    {
-        long fence;
-        __asm { xchg fence, eax }
-    }
-#   pragma warning( pop )
-#elif defined ( _M_X64 )
+EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
+{
+    MemoryBarrier();
+}
+
+#if defined ( _WIN64 )
 #   define MS_ATOMIC_64
-#   pragma intrinsic ( __faststorefence )
-    EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
-    {
-        __faststorefence ();
-    }
-#elif defined ( _M_IA64 )
-#   define MS_ATOMIC_64
-#   pragma intrinsic ( __mf )
-    EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
-    {
-        __mf ();
-    }
-#else
-#   error unexpected target architecture, msvc version of epicsAtomicCD.h
 #endif
 
 /*
--- a/src/tools/EpicsHostArch.pl
+++ b/src/tools/EpicsHostArch.pl
@@ -102,6 +102,39 @@ HELP_MESSAGE() if !getopts('hg:') || $opt_h;
 #
 # Some examples from the Debian project:
 # - https://wiki.debian.org/Multiarch/Tuples
+
+# On windows arm64, Perl may actually run in x64 emulator mode.
+# Find out "real" architecture using Win32::API
+sub realWinArch {
+    # Do we have Win32::API? (fails for older Windows)
+    my $have_api = eval {
+        require Win32::API;
+        Win32::API->import();
+        1;
+    };
+    return 'x64' unless $have_api;
+
+    # Call IsWow64Process2 if we have it
+    my $isWow64Process2 = Win32::API->new(
+        'kernel32', 'IsWow64Process2', 'NPP', 'I'
+    ) or return 'x64';
+    my ($proc, $native) = (pack('v', 0), pack('v', 0));
+    $isWow64Process2->Call(-1, $proc, $native)
+        or return 'x64';
+    my $arch = unpack('v', $native);
+
+    # Architecture code mapping
+    my %ARCH = (
+        0x014c => "x86",
+        0x8664 => "x64",
+        0xaa64 => "aarch64",
+        0x01c0 => "arm",
+    );
+
+    # Return known architecture or hex code
+    return $ARCH{$arch} // sprintf("0x%04x", $arch);
+}
+
 sub toEpicsArch {
     my $arch = shift;
     for ($arch) {
@@ -110,7 +143,7 @@ sub toEpicsArch {
         return 'linux-arm'      if m/^arm-linux/;
         return 'linux-aarch64'  if m/^aarch64-linux/;
         return 'linux-ppc64'    if m/^powerpc64-linux/;
-        return 'windows-x64'    if m/^MSWin32-x64/;
+        return 'windows-'.realWinArch() if m/^MSWin32-x64/;
         return 'win32-x86'      if m/^MSWin32-x86/;
         return "cygwin-x86_64"  if m/^x86_64-cygwin/;
         return "cygwin-x86"     if m/^i[3-6]86-cygwin/;
--- a/src/tools/makeTestfile.pl
+++ b/src/tools/makeTestfile.pl
@@ -30,6 +30,8 @@
 
 use 5.10.1;   # This script uses the defined-or operator //
 use strict;
+use Config;
+use POSIX;
 
 use File::Basename;
 my $tool = basename($0);
@@ -47,8 +49,13 @@ if ($TA =~ /^win32-x86/ && $HA !~ /^win/) {
     $wine32 = "/usr/bin/wine" if ! -x $wine32;
     $error = $exec = "$wine32 $exe";
 }
-elsif ($TA =~ /^windows-x64/ && $HA !~ /^win/) {
-    # Use WINE to run windows-x64 executables on non-windows hosts.
+elsif ($TA =~ /^windows-x64/ && $HA !~ /^win/ && $Config{'archname'} =~ /x86_64/) {
+    # Use WINE to run windows-x64 executables on non-windows x64 hosts.
+    $error = $exec = "wine64 $exe";
+}
+elsif ($TA =~ /^windows-aarch64/ && $HA !~ /^win/ && $Config{'archname'} =~ /arm64|aarch64/ && sysconf(_SC_PAGESIZE) == 4096) {
+    # Use WINE to run windows-aarch64 executables on non-windows aarch64 hosts.
+    # Requires that the host uses 4k memory pages as expeted by Windows.
     $error = $exec = "wine64 $exe";
 }
 elsif ($TA =~ /^RTEMS-pc[36]86-qemu$/) {
--- a/startup/windows.bat
+++ b/startup/windows.bat
@@ -15,6 +15,7 @@ rem Sets EPICS_HOST_ARCH and the environment for Microsoft Visual Studio.
 rem Optionally, resets PATH, adds Strawberry Perl to PATH, and adds the
 rem EPICS Base install host architecture bin directory to PATH.
 rem
+rem Uses first argument or %PROCESSOR_ARCHITECTURE% to select architecture.
 
 rem ----------------------------------------------------------------------
 rem Site serviceable parts (These definitions may be modified)
@@ -54,11 +55,8 @@ rem Strawberry Perl will be added to PATH.
 set _strawberry_perl_home=C:\Strawberry
 
 rem The location of Microsoft Visual Studio (pathname).
-set _visual_studio_home=C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
-
-rem The EPICS host architecture specification for EPICS_HOST_ARCH
-rem (<os>-<arch>[-<toolset>] as defined in configure/CONFIG_SITE).
-set _epics_host_arch=windows-x64
+rem If not set, vxwhere.exe will be called to find latest.
+set _visual_studio_home=
 
 rem The install location of EPICS Base (pathname).  If nonempty and
 rem _auto_path_append is yes, it will be used to add the host architecture
@@ -82,16 +80,37 @@ set "PATH=%PATH%;%_strawberry_perl_home%\perl\site\bin"
 set "PATH=%PATH%;%_strawberry_perl_home%\perl\bin"
 :after_add_strawberry_perl
 
+rem Find latest Microsoft Visual Studio automatically if not specified
+if "%_visual_studio_home%" == "" (
+    for /f "usebackq tokens=*" %%i in (
+        `"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property InstallationPath`
+    ) do ( set "_visual_studio_home=%%i" )
+)
+
 rem Set the environment for Microsoft Visual Studio
-call "%_visual_studio_home%\VC\Auxiliary\Build\vcvarsall.bat" x64
+rem Default to native architecture if no argument given
+rem Allow "aarch64" as alias for "arm64"
+if "%~1" == "aarch64" (
+    set _arch=arm64
+) else if "%~1" == "" (
+    set _arch=%PROCESSOR_ARCHITECTURE%
+) else (
+    set _arch=%~1
+)
+call "%_visual_studio_home%\VC\Auxiliary\Build\vcvarsall.bat" %_arch%
 
 rem Set the EPICS host architecture specification
-set "EPICS_HOST_ARCH=%_epics_host_arch%"
+if "%VSCMD_ARG_TGT_ARCH%" == "arm64" (
+    rem EPICS uses aarch64 where Visual Studio uses arm64
+    set EPICS_HOST_ARCH=windows-aarch64
+) else (
+    set EPICS_HOST_ARCH=windows-%VSCMD_ARG_TGT_ARCH%
+)
 
 rem Add the EPICS Base host architecture bin directory to PATH
 if "%_auto_path_append%" == "yes" (
   if not "%_epics_base%" == "" (
-    set "PATH=%PATH%;%_epics_base%\bin\%_epics_host_arch%"
+    set "PATH=%PATH%;%_epics_base%\bin\%EPICS_HOST_ARCH%"
   )
 )
 
@@ -102,5 +121,4 @@ set _auto_path_append=
 set _path_new=
 set _strawberry_perl_home=
 set _visual_studio_home=
-set _epics_host_arch=
 set _epics_base=
--- a/configure/CONFIG_SITE
+++ b/configure/CONFIG_SITE
@@ -41,6 +41,8 @@
 #	windows-x64            (MS Visual C++ compiler used for host builds)
 #	windows-x64-mingw      (MinGW compiler used for host builds)
 #	windows-x64-static     (MS Visual C++ compiler used for host builds)
+#	windows-aarch64        (MS Visual C++ compiler used for host builds)
+#	windows-aarch64-static (MS Visual C++ compiler used for host builds)
 
 #  Debugging builds:
 #	linux-arm-debug        (GNU compiler used for host builds)
@@ -49,6 +51,7 @@
 #	solaris-sparc-debug    (sun compiler no optimization,-g for debugging info)
 #	win32-x86-debug        (MS Visual C++ compiler with debug option for host builds)
 #	windows-x64-debug      (MS Visual C++ compiler with debug option for host builds)
+#	windows-aarch64-debug  (MS Visual C++ compiler with debug option for host builds)
 
 
 #  Do not set EPICS_HOST_ARCH in this file. 
--- a/configure/os/CONFIG.windows-aarch64-debug.Common
+++ b/configure/os/CONFIG.windows-aarch64-debug.Common
@@ -1,0 +1,9 @@
+# CONFIG.windows-aarch64-debug.Common
+#
+# Definitions for windows-aarch64-debug host arch
+# Override these definitions in CONFIG_SITE.windows-aarch64-debug.Common
+#-------------------------------------------------------
+
+#Include definitions common to windows-aarch64 hosts
+include $(CONFIG)/os/CONFIG.windows-aarch64.Common
+
--- a/configure/os/CONFIG.windows-aarch64-debug.windows-aarch64-debug
+++ b/configure/os/CONFIG.windows-aarch64-debug.windows-aarch64-debug
@@ -1,0 +1,14 @@
+# CONFIG.windows-aarch64-debug.windows-aarch64-debug
+#
+# Definitions for windows-aarch64-debug host - windows-aarch64-debug target build
+# Override these definitions in CONFIG_SITE.windows-aarch64-debug.windows-aarch64-debug
+#-------------------------------------------------------
+
+#Include definitions common to windows-aarch64 builds
+include $(CONFIG)/os/CONFIG.windows-aarch64.windows-aarch64
+
+# Override CONFIG.CrossCommon settings:
+BUILD_CLASS = HOST
+
+# Override CONFIG_SITE settings:
+HOST_OPT=NO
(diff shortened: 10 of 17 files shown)


.github/workflows/ci-scripts-build.yml
```diff
@@ -175,10 +175,11 @@ jobs:
 
           # Windows builds
 
-          - name: "Win-22 MSC-22"
+          - name: "Win-22 MSC-22 (cmd shell)"
             os: windows-2022
             cmp: vs2022
             configuration: default
+            extra: "SHELL=cmd"
```
FreddieAkeroyd, line 182: Was this related to `echo` in Makefiles or a different issue? I've something similar in out epics base and `echo` issues when `sh.exe` is in your PATH seems to be the comment i have there.

dirk-zimoch: This is from #881/#882.

modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
```diff
@@ -52,6 +52,13 @@
     {
         __mf ();
     }
+#elif defined ( _M_ARM64 )
+#   define MS_ATOMIC_64
+#   pragma intrinsic ( __dmb )
+    EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
+    {
+        __dmb (_ARM64_BARRIER_ISH);
```
mdavidsaver: Is there a reference for this?  I have never been so clear on all of the differences between `dmb` vs. `dsb` with their many sub-options.  Perhaps MSVC has a higher level intrinsic?

dirk-zimoch: AI says so. But also https://learn.microsoft.com/en-us/cpp/intrinsics/arm-intrinsics?view=msvc-170#BarrierRestrictions, just less clear.

dirk-zimoch: MSVC has the `MemoryBarrier` macro which is deprecated and not implemented for ARM64.
Microsoft suggests to use [std::atomic_thread_fence](https://en.cppreference.com/cpp/atomic/atomic_thread_fence) instead.

dirk-zimoch: See also https://medium.com/@wadixtech/embedded-systems-difference-between-dsb-dmb-and-isb-instructions-4b2fc1545132
It says which one to use (`dmb` vs `dsb` vs  `isb`) depends on the use case. Who would have thought. The only use case I found in base is in posix/osdThread.c to make sure `pthreadInfo->tid` is updated. I think `dmb`, the least invasive one, would be sufficient here, but Windows does not even use it.

dirk-zimoch: `dmb` is definitively sufficient for drivers that need to write hardware registers in programming order.

mdavidsaver: > MSVC has the `MemoryBarrier` macro which is deprecated and not implemented for ARM64.
> Microsoft suggests to use [std::atomic_thread_fence](https://en.cppreference.com/cpp/atomic/atomic_thread_fence) instead.

Some combination of these seem like a better approach (`MemoryBarrier` when < c++11).

> I think `dmb`,

I agree.  However, what about the modifier ISH vs SY?  It is this level of ARM details which I would rather avoid encoding in Base.

Quoting [ARM](https://developer.arm.com/documentation/100941/0101/Memory-attributes):

> The division between inner and outer is IMPLEMENTATION DEFINED, but typically the inner attributes are used by caches in the processor.  ...

So ISH seems like the correct choice to synchronize between CPUs, but not device DMA.  However, it would not be an ARM document without some IMPLEMENTATION DEFINED ambiguity.  Thus my preference to defer to the compiler/OS.

mdavidsaver: > `dmb` is definitively sufficient for drivers that need to write hardware registers in programming order.

Our documentation in `epicsAtomic.h` is not specific, but to my mind these barriers apply only to CPU synchronization.  For I/O synchronization see `epicsMMIO.h`, which currently has a no-op implementation except for the RTOS targets...  So some potential work here.

dirk-zimoch: winnt.h is the Windows header that defines `MemoryBarrier` -- in theory. It contains:
`// TODO: WOWXX - Unblock ARM. Make all alignment checks DWORD for now.`
And including it fails with
`#error: "No Target Architecture"`
If it worked, it would do:
`#define MemoryBarrier() __dmb(_ARM_BARRIER_SY)`

anjohnson: My AI suggested the same changes for this file.

modules/libcom/src/calc/calcPerform.c
```diff
@@ -35,7 +35,7 @@ static int cond_search(const char **ppinst, int match);
 #endif
 
 /* Turn off global optimization for 64-bit MSVC builds */
-#if defined(_WIN32) && defined(_M_X64) && !defined(_MINGW)
+#if defined(_WIN32) && ( defined(_M_X64) || defined(_M_ARM64) ) && !defined(_MINGW)
```
anjohnson: Claude suggested that the code-gen bug we're avoiding here might not be needed on ARM64, so it's worth checking whether it works without adding this.

anjohnson (changes requested):
Please also add these new target-archs in the following files:
- `configure/CONFIG_SITE`
- `startup/windows.bat` (or create a parallel equivalent)
- `documentation/README.md` (about windows.bat)
- `documentation/new-notes/PR-888.md`
- `src/tools/makeTestfile.pl` – is there any support in WINE for this?

Is there an updated `valgrind.h` as we provide at `modules/libcom/src/valgrind/valgrind.h` which would add this?

Do you know what the GNU architecture tuplet would be given for this? I'm looking at the `-g` option accepted by `EpicsHostArch.pl`. I'd like to see a comment from @minijackson about this since he added that option (that may not support Windows at all though).

The external PVA submodules have their own `.github/workflows/ci-scripts-build.yml` job configurations and could test this new build target (not sure if we're doing Windows builds on them yet though).

pvData has a `byteBuffer.h` file that does alignment checks which may need extending.

Did you test both the `-static` and `-debug` targets? Should we include those in the GHA builds, since there's no support for this CPU on Appveyor yet?

modules/libcom/src/calc/calcPerform.c
dirk-zimoch: Claude is wrong. It failed. That‘s why I extended the `#if` condition.

anjohnson: You already checked, that's fine. Thanks!

dirk-zimoch: BTW, I tried several alternatives like more casting, explicit assignment to intermediate integer variables,`volatile`. Nothing helped. The logic operations fail to produce correct results with optimization on consistently on x64 as well as on arm64.

dirk-zimoch:
Wine for arm64 seems still under development (in particular for gaming on arm64). I have not tried it yet.

dirk-zimoch:
Adding `-debug` and `-static` builds should be straight forward. I have not tried yet.

dirk-zimoch:
I trust Michael more with checking pvData for compatibility than myself. It builds, that‘s all I can say.

dirk-zimoch:
According to https://valgrind.org/, Windows is not on the list of supported platforms, neither on x64 nor arm64 CPUs.
The [valgrind4win](https://sourceforge.net/p/valgrind4win/code/HEAD/tree/) project does not seem to support arm. It states its status as „Pre-Alpha“. As its last commit is from 2013, I do not expect anything soon.

Looking into our valgrind.h, it seems to support only mingw and mingw64. Both are not the scope of this PR. This PR is about a MSVC build.

modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
dirk-zimoch: AI suggested ISH domain to me instead of SY domain. It probably only makes a difference when synchronizing with DMA. SY seems „safer“ to me but probably wastes time delaying processing until all DMAs are done. As Michael already noticed: ARM documentation is hard to understand „correctly“ and mentions implementation dependency.
One would need to know all possible use cases of `epicsAtomicMemoryBarrier` first before deciding on the „correct“ domain. I could not find ANY use case for Windows in base!
And all „implementations“ (processors? computers?)

dirk-zimoch: https://github.com/ARM-software/abi-aa/blob/main/atomicsabi64/atomicsabi64.rst#synchronization-fences seems to suggest ISH domain.

dirk-zimoch:
> Do you know what the GNU architecture tuplet would be given for this?

[`aarch64-w64-mingw32`](https://gcc.gnu.org/gcc-15/changes.html#aarch64)? (I am confused by the usage of the numbers 64 and 32 here.)

ralphlange:
Gemini says:

The standard GNU architecture triplet for Windows 11 on ARM64 is:

### **`aarch64-w64-mingw32`**

Although "Windows 11" is the specific OS version, the GNU build system (and tools like GCC and Binutils) uses a unified triplet for all modern 64-bit Windows versions on the ARM64 architecture.

### **Breakdown of the Triplet:**
*   **`aarch64`**: Represents the 64-bit ARM architecture (the instruction set).
*   **`w64`**: Refers to the MinGW-w64 project, which provides the headers and runtime for 64-bit Windows.
*   **`mingw32`**: This is a legacy convention. Even though you are targeting a 64-bit system, the operating system identifier remains `mingw32` for compatibility with the original MinGW build scripts.

---

### **Other Variations & Context**
Depending on the toolchain you are using (e.g., LLVM/Clang or Rust), you may encounter slightly different versions of this triplet:

1.  **LLVM/Clang (MSVC ABI)**:
    If you are building for Windows 11 ARM64 but want compatibility with Microsoft's Visual C++ (MSVC) runtime/libraries, the triplet is:
    *   **`aarch64-pc-windows-msvc`**
2.  **LLVM/Clang (MinGW ABI)**:
    If you are using Clang but targeting the MinGW environment (common in MSYS2 or cross-compiling from Linux):
    *   **`aarch64-w64-windows-gnu`** (Note: Clang often accepts `windows-gnu` as an alias for `mingw32`).
3.  **vcpkg/Microsoft Naming**:
    Microsoft’s package manager and build tools often simplify this to:
    *   **`arm64-windows`**

### **How to verify on your system:**
If you have a GNU-based compiler (like GCC) installed on your Windows ARM64 device, you can confirm the exact triplet the compiler was built for by running:
```bash
gcc -dumpmachine
```
In most MSYS2 or cross-compilation environments (like `aarch64-w64-mingw32-gcc`), this will return **`aarch64-w64-mingw32`**.

modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
dirk-zimoch: I tried to `#include <Windows.h>` instead of `winnt.h`, as suggested by https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-memorybarrier. Then, I can simply use `MemoryBarrier()` for all Windows CPU architectures and leave the implementation details to Microsoft.

However, compiling pvAccess now dies from allergic shock from too much Windows, it seems:
```
cl -EHsc -GR          -DNOMINMAX     -nologo -FC -D__STDC__=0 -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE    -Ox -GL -Oy-   -W3 -w44355 -w44344 -w44251 -D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING        -MD -DEPICS_BUILD_DLL -DEPICS_CALL_DLL -TP   -I. -I../O.Common -I. -I. -I.. -I../../src/pva -I../../src/utils -I../../src/client -I../../src/factory -I../../src/remote -I../../src/remoteClient -I../../src/server -I../../src/rpcService -I../../src/rpcClient -I../../src/pipelineService -I../../src/mb -IC:/Users/epics/epics-base/include/compiler/msvc -IC:/Users/epics/epics-base/include/os/WIN32 -IC:/Users/epics/epics-base/include -IC:/Users/epics/epics-base/include/compiler/msvc -IC:/Users/epics/epics-base/include/os/WIN32 -IC:/Users/epics/epics-base/include        -c ../../src/client/pvAccess.cpp
pvAccess.cpp
C:\Users\epics\epics-base\include\pv\pvData.h(521): warning C4250: 'epics::pvData::PVString': inherits 'epics::pvData::PVScalarValue<std::string>::epics::pvData::PVScalarValue<std::string>::deserialize' via dominance
C:\Users\epics\epics-base\include\pv\pvData.h(440): note: see declaration of 'epics::pvData::PVScalarValue<std::string>::deserialize'
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\shared\ws2def.h(103): warning C4005: 'AF_IPX': macro redefinition
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\um\winsock.h(457): note: see previous definition of 'AF_IPX'
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\shared\ws2def.h(147): warning C4005: 'AF_MAX': macro redefinition
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\um\winsock.h(476): note: see previous definition of 'AF_MAX'
[...]
```

dirk-zimoch: `#define WIN32_LEAN_AND_MEAN` solves that.
https://learn.microsoft.com/en-us/windows/win32/winsock/creating-a-basic-winsock-application

modules/libcom/src/calc/calcPerform.c
```diff
@@ -35,7 +35,7 @@ static int cond_search(const char **ppinst, int match);
 #endif
 
 /* Turn off global optimization for 64-bit MSVC builds */
-#if defined(_WIN32) && defined(_M_X64) && !defined(_MINGW)
+#if defined(_WIN64) && !defined(_MINGW)
 #  pragma optimize("g", off)
```
dirk-zimoch, line 39: `_WIN64` is defined for both, `_M_X64` and `_M_ARM64` targets (and only on Windows)
All 64 bit Windows builds seems to have the same optimizer bug with bit operations on converted floating point numbers.

modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
```diff
@@ -27,33 +25,19 @@
...
-    EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
-    {
-        __faststorefence ();
-    }
-#elif defined ( _M_IA64 )
+// Avoid including old winsock.h automatically with windows.h
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
```
dirk-zimoch, line 32: Need `WIN32_LEAN_AND_MEAN` to avoid pulling in old winsock.h which would later collide with code including winsock2.h.

modules/libcom/src/osi/compiler/msvc/epicsAtomicCD.h
```diff
@@ -27,33 +25,19 @@
...
+
+EPICS_ATOMIC_INLINE void epicsAtomicMemoryBarrier (void)
+{
+    MemoryBarrier();
```
dirk-zimoch, line 36: Leave it to Microsoft to find the correct implementation of a memory barrier for the used CPU architecture.

src/tools/EpicsHostArch.pl
```diff
@@ -102,6 +102,51 @@ sub HELP_MESSAGE {
...
+        1;
+    };
+    return $fallback->() unless $have_api;
+
+    # Call IsWow64Process2 if we have it
+    my $isWow64Process2 = Win32::API->new(
+        'kernel32', 'IsWow64Process2', 'NPP', 'I'
+    ) or return $fallback->();
+    my ($proc, $native) = (pack('v', 0), pack('v', 0));
+    $isWow64Process2->Call(-1, $proc, $native)
```
dirk-zimoch, line 122: This tells us the real CPU architecture, even if Perl is running in Windows's built-in emulator because Strawberry Perl is only available as a x64 program, even on arm64 Windows.

src/tools/makeTestfile.pl
```diff
@@ -47,8 +49,13 @@
...
 }
-elsif ($TA =~ /^windows-x64/ && $HA !~ /^win/) {
-    # Use WINE to run windows-x64 executables on non-windows hosts.
+elsif ($TA =~ /^windows-x64/ && $HA !~ /^win/ && $Config{'archname'} =~ /x86_64/) {
+    # Use WINE to run windows-x64 executables on non-windows x64 hosts.
+    $error = $exec = "wine64 $exe";
+}
+elsif ($TA =~ /^windows-arm64/ && $HA !~ /^win/ && $Config{'archname'} =~ /arm64|aarch64/ && sysconf(_SC_PAGESIZE) == 4096) {
+    # Use WINE to run windows-arm64 executables on non-windows arm64 hosts.
+    # Requires 4k memory pages.
```
dirk-zimoch: Check if host CPU (and page size) matches Windows architecture. Modern arm64 systems might use 16k pages which is incompatible with Windows and beyond what wine can do alone.
One could try to add emulators for running windows-x64 on linux-aarch64 or vice versa (beyond the scope of this PR).

startup/windows.bat
```diff
@@ -83,15 +79,17 @@ set "PATH=%PATH%;%_strawberry_perl_home%\perl\bin"
 :after_add_strawberry_perl
 
 rem Set the environment for Microsoft Visual Studio
-call "%_visual_studio_home%\VC\Auxiliary\Build\vcvarsall.bat" x64
+call "%_visual_studio_home%\VC\Auxiliary\Build\vcvarsall.bat" %PROCESSOR_ARCHITECTURE%
```
dirk-zimoch: cmd.exe already sets `PROCESSOR_ARCHITECTURE` to `x64` or `ARM64` (upper case)
`vcvarsall.bat` then sets `VSCMD_ARG_TGT_ARCH` to `x64` or `arm64` (lower case).

In principle, the user can ask to use `x64` on `arm64` here too because windows can use its built-in emulator.

.gitmodules
```diff
@@ -24,5 +24,5 @@
     branch = master
 [submodule ".ci"]
     path = .ci
-    url = https://github.com/epics-base/ci-scripts
-    branch = master
+	url = https://github.com/ralphlange/ci-scripts.git
+	branch = windows-11-arm-support-10659373095315842272
```
dirk-zimoch, line 28: Waiting for Ralph's fixes to be merged...

ralphlange: Yep, now, with your integration working, I can test the tests for ci-scripts.
Classical hen/egg situation.

dirk-zimoch: I will keep this PR a "draft" until then. For the final PR, I need to change this back.

src/tools/EpicsHostArch.pl
anjohnson: It might not be long before Strawberry Perl is available native for ARM64, would that break this code? Apparently Linaro has a [native Perl package for WoA](https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28697821212/Perl), it just doesn't have all the add-ons and MinGW GCC that Strawberry provides.

See also:

- https://github.com/StrawberryPerl/Perl-Dist-Strawberry/issues/218
- https://github.com/StrawberryPerl/spbuild/issues/20

dirk-zimoch: No. It would not break the code. That is the whole point of this code: to find out the real native host architecture, independent of any emulators.
This code asks Windows for the native architecture. Now, Perl running on arm64 in an emulator thinks it runs on x64 but this function reveals that the native architecture is indeed arm64. If in the future Perl is natively available for arm64, this function will still return the same native architecture: arm64.
Of course, on an x64 machine, this function returns x64. Even if an arm64 version of Perl would run on an x64 machine in an emulator, it would still return the native architecture x64.

dirk-zimoch: If the system has no WoW support, i.e. older Windows (before 10, I think) without an emulator or arm support, it falls back to `x64`.

dirk-zimoch:
What is your opinion about the architecture name? **arm64** (like Microsoft calls it) or **aarch64** (like GNU calls it)?

dirk-zimoch:
> ```shell
> gcc -dumpmachine
> ```
>   
> In most MSYS2 or cross-compilation environments (like `aarch64-w64-mingw32-gcc`), this will return `aarch64-w64-mingw32`.

I tested that on my arm64 windows. I got `x86_64-w64-mingw32`. The gcc is part of Strawberry, so it is running in x64 emulator mode (as well as the Perl interpreter).

src/tools/EpicsHostArch.pl
```diff
@@ -110,7 +143,9 @@ sub toEpicsArch {
         return 'linux-arm'      if m/^arm-linux/;
         return 'linux-aarch64'  if m/^aarch64-linux/;
         return 'linux-ppc64'    if m/^powerpc64-linux/;
-        return 'windows-x64'    if m/^MSWin32-x64/;
+        return 'windows-'.realWinArch() if m/^MSWin32-x64/;
+        return 'windows-arm64'  if m/^aarch64-w64/;
+        return 'windows-x64'    if m/^x86_64-w64/;
```
dirk-zimoch: Support for `-g` (gnu architecture) flag:
`perl EpicsHostArch.pl -g aarch64-w64-mingw32` and similar returns `windows-arm64` and
`perl EpicsHostArch.pl -g x86_64-w64-mingw32` and similar returns `windows-x64`.

mdavidsaver:
> What is your opinion about the architecture name? **arm64** (like Microsoft calls it) or **aarch64** (like GNU calls it)?

aarch64.  Consistent with other platforms.  Also, I find myself mis-reading arm64 as amd64.

src/tools/EpicsHostArch.pl
anjohnson: The `EpicsHostArch.pl` script is used if the user hasn't explicitly set `EPICS_HOST_ARCH` in the environment; it's also called from some of the scripts in the `startup` directory (if anyone actually uses them). IMHO when running on an emulated host-arch, the script's answer could reasonably be either the emulated or the native architecture — I can imagine users of one wanting to build for the other CPU.

A user can also set `EPICS_HOST_ARCH` to request building the `-static` or `-debug` alternative target-arch's. Does setting `EPICS_HOST_ARCH=windows-x64` on a `windows-aarch64` system produce Intel binaries, and _vice versa_? That would make supporting both easy, so if that works we should document it. Supporting cross-builds between the two could also be interesting, but maybe as a separate PR by someone who wants that.

src/tools/EpicsHostArch.pl
dirk-zimoch: Is this actually correct? If the idea of the `-g` flag was to convert the output of `gcc -dumpmachine` to a matching EPICS architecture, then it should probably return some `...-mingw` string on Windows. In that case, it would be out of scope for this PR which targets MSVC builds and I will remove it. @minijackson?
BTW: On my Linux host, `perl EpicsHostArch.pl -g $(gcc -dumpmachine)` fails and prints `Architecture 'x86_64-redhat-linux' not recognized`. I wonder what is the intended use of `-g` then?
Fixing the pattern match from `/^x86_64-linux/` to something like `/^x86_64(-.*)?-linux/` is out of scope for this PR as well.

src/tools/EpicsHostArch.pl
dirk-zimoch: >  IMHO when running on an emulated host-arch, the script's answer could reasonably be either the emulated or the native architecture 

I think if the native host system is Windows on arm64 and the user has not set `EPICS_HOST_ARCH`, the build should be for the native arm64 architecture. The user should not have to worry about Strawberry Perl running in an emulation. That should not force EPICS into using the emulation as well if it can be built natively instead.

Unfortunately actually getting a cross build is not as easy as setting `EPICS_HOST_ARCH` accordingly because at the moment, the built Windows target depends on how vcvarsall.bat has been called for example by startup\windows.bat and set up the environment and not on compiler arguments. I can look into that.

src/tools/EpicsHostArch.pl
minijackson: The original intent was to make it easier to integrate the EPICS build system in other build system.

Originally, I just wanted to find out what the standard EPICS-way of naming architectures was, for cross-compilation, and I just made minimal modifications to `EpicsHostArch.pl` to output the EPICS naming without resolving the one from the host machine.

The issue for `perl EpicsHostArch.pl -g $(gcc -dumpmachine)` is that `EpicsHostArch.pl` was made for Perl's `$Config{archname}` which is different from `gcc -dumpmachine`. It worked for me when I made the PR, because I could ask Nix either for the `x86_64-linux` form (which works) or the `x86_64-unknown-linux-gnu` form (which doesn't). All are valid for GCC, but other tools (LLVM, Rust) might use slightly different conventions, which I think we could handle as they come.

I still think this could be useful for Windows, maybe someone one day will try to package EPICS using Conan or something, and they'll want to know which `configure/` files to edit.

If someone's curious, autoconf comes with a `config.sub` bash script which does the normalization of target triplets.

dirk-zimoch: I will remove the two lines. They have not much to do with this PR.

dirk-zimoch:
Architecture name changed to `windows-aarch64`. Now @ralphlange has to update the .ci-scripts module to reflect that change.

startup/windows.bat
```diff
@@ -82,16 +80,37 @@ set "PATH=%PATH%;%_strawberry_perl_home%\perl\site\bin"
 set "PATH=%PATH%;%_strawberry_perl_home%\perl\bin"
 :after_add_strawberry_perl
 
+rem Find latest Microsoft Visual Studio automatically if not specified
+if "%_visual_studio_home%" == "" (
+    for /f "usebackq tokens=*" %%i in (
+        `"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property InstallationPath`
+    ) do ( set "_visual_studio_home=%%i" )
+)
+
```
dirk-zimoch, line 89: Visual Studio installs `vswhere.exe` (always at the same location) which helps to find where Visual Studio is installed. This avoids hard-coding the Visual Studio path here and having to change it after each upgrade. The user can still explicitly set `_visual_studio_home` above if required.

startup/windows.bat
```diff
@@ -82,16 +80,37 @@ set "PATH=%PATH%;%_strawberry_perl_home%\perl\site\bin"
...
+rem Default to native architecture if no argument given
+rem Allow "aarch64" as alias for "arm64"
+if "%~1" == "aarch64" (
+    set _arch=arm64
+) else if "%~1" == "" (
+    set _arch=%PROCESSOR_ARCHITECTURE%
+) else (
+    set _arch=%~1
+)
+call "%_visual_studio_home%\VC\Auxiliary\Build\vcvarsall.bat" %_arch%
```
dirk-zimoch, line 100: Allow the user to pass `x64` or `arm64` or `aarch64` on the command line. `aach64` is treated as an alias for `arm64`, which is the name Visual Studio uses.

startup/windows.bat
```diff
@@ -82,16 +80,37 @@ set "PATH=%PATH%;%_strawberry_perl_home%\perl\site\bin"
...
 
 rem Set the EPICS host architecture specification
-set "EPICS_HOST_ARCH=%_epics_host_arch%"
+if "%VSCMD_ARG_TGT_ARCH%" == "arm64" (
+    rem EPICS uses aarch64 where Visual Studio uses arm64
+    set EPICS_HOST_ARCH=windows-aarch64
+) else (
+    set EPICS_HOST_ARCH=windows-%VSCMD_ARG_TGT_ARCH%
+)
```
dirk-zimoch, line 108: Convert the name `arm64` as used by Visual Studio to `aarch64` for use in `EPICS_HOST_ARCH`.

dirk-zimoch:
Rebased.

src/tools/EpicsHostArch.pl
dirk-zimoch: I have now modified `startup/windows.bat` so that the user can pass the processor architecture (one of those supported by Visual Studio as installed on the host, or `aarch64` as an alias for `arm64`) as an argument. For example on an ARM64 windows, `startup/windows.bat x64` will set up the Visual Studio Environment for x64 compilation and set `EPICS_HOST_ARCH=windows-x64` while without arguments `startup/windows.bat` will set up Visual Studio for the native architecture, here arm64, and set `EPICS_HOST_ARCH=windows_aarch64`.

Issue #879, opened 2026-05-26
makeRPath: Perl port from PR #589 regressed `$ORIGIN` rpath generation — analysis, fix, and tests


jeonghanlee:
# makeRPath: Perl port from PR #589 regressed `$ORIGIN` rpath generation — analysis, fix, and tests

## Summary

PR #589 ("Remove Python build dependency when `LINKER_USE_RPATH=ORIGIN`", commit `f4c474e`) converted `src/tools/makeRPath.py` to `makeRPath.pl`, and was later reverted in full by `a3d8531` ("Revert PR #589"). The revert commit does not record a reason.

On inspection the Perl port carried three behavioral regressions against the Python original. The most significant one ties the emitted `$ORIGIN` rpath to the *current working directory*, which works against the goal of `LINKER_USE_RPATH=ORIGIN` (a relocatable build tree) and may well have led to the revert.

The Python-to-Perl direction is a natural fit for `src/tools`. `makeRPath` is the only Python script among the Perl build tools there, so porting it both restores consistency with the rest of the directory and removes the Python dependency from the normal (non-doc) build path — the stated goal of PR #589. The unversioned `python` command is fragile (depending on the distribution it may be absent or resolve to Python 2); this can be steered with `PYTHON` in CONFIG_SITE, but narrowing the build's interpreter dependencies toward Perl alone is the more sustainable path. The port adds no new Perl dependency: it uses only core modules and the defined-or operator `//`, which is within EPICS's stated minimum (Perl 5.10.1) and already used in `src/tools` by `makeTestfile.pl` (which likewise declares `use 5.10.1` for it). The blocker in #589 was the regression, not the language choice.

So this issue includes a **corrected port and a comparison test driver inline for review**, not just a bug report. If the approach looks right I'll open it as a PR.

## Reproduction

Given a relocatable tree:

```
build/lib
build/module/lib
other/lib        # outside the root
```

and the call

```
makeRPath -F build/lib -R build build/lib build/module/lib other/lib
```

**Python (`makeRPath.py`, current):**

```
-Wl,-rpath-link,/.../build/lib -Wl,-rpath,$ORIGIN/. \
-Wl,-rpath-link,/.../build/module/lib -Wl,-rpath,$ORIGIN/../module/lib \
-Wl,-rpath,/.../other/lib
```

**Perl (`makeRPath.pl` from PR #589):**

```
-Wl,-rpath-link,/.../build/lib \
[...]

PR #890, opened 2026-06-04
configure: fail hard if RPATH ORIGIN requested, but makeRPath errors out

mdavidsaver:
I am given to understand that a contributing factor to #879 was the fact that the Makefile `$(shell ` function calls fail soft.  If there is no `python` executable in `$PATH`, then combining `PYTHON=python` and `LINKER_USE_RPATH=ORIGIN` results in "warning" messages and apparent success, but with incorrect results.

There are currently 34 places in Base where `$(shell` is used.  All of these will fail "soft".  I'm not going to try to "solve" this whole problem.  eg. An absence of `perl` results in a failure which is messy enough to error out.  However, as `python` vs. `python3` confusion is likely to be with us until RHEL7 derivatives age out (some time around the heat death of the universe) I think it is worth making this optional configuration error explicitly.

The PR makes use of the `$(shell ...)` construct storing a numeric exit code in `$(.SHELLSTATUS)`.  The `$(filter-out)` maps zero to empty string, so that non-zero values select `$(error ...)`.

```sh
$ git grep '$(shell'|wc -l
34
```


simon-ess:
I should note that as I understand, this won't work on mac OS (gnu make 3.81...), but it also won't _hurt_.

mdavidsaver:
> I should note that as I understand, this won't work on mac OS (gnu make 3.81...), but it also won't _hurt_.

Due to Make version?

anjohnson:
> Due to Make version?

Yup, and GPLv3:

```
Kato$ /usr/bin/make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
Kato$ alias make
alias make='gmake'
Kato$ which gmake
/usr/local/bin/gmake
Kato$ gmake --version
GNU Make 4.4.1
Built for x86_64-apple-darwin22.3.0
Copyright (C) 1988-2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

My 4.4.1 is from the Homebrew package.

simon-ess:
@mdavidsaver correct. This variable was release as a part of make 4.2: https://lists.gnu.org/archive/html/info-gnu/2016-05/msg00013.html

simon-ess:
That said, I want to be clear that I'm 100% fine with this change. It silently does nothing for mac os and helps in other cases, so that's a win in my books.

simon-ess (approved):

anjohnson:
Core: Approved subject to American => English translation.

PR #894, opened 2026-06-17
Document Built-in Device Support

anjohnson:
The documentation for the soft device support included in Base has always lacking. The major record types mention some of their soft device layers, but many are missing. Some of the newer supports have no documentation at all (outside of a mention in the Release Notes at the time, if at all).

I tasked Claude with finding where documentation is missing, and writing an initial version of the appropriate text as a starting point. That was the initial commit in this series (which didn't mention JSON links at all for example), and I plan to continue editing this both by hand and with Claude's help. Other edits or suggestions would be welcome.


mdavidsaver:
Should this PR be marked as a draft?

modules/database/src/std/rec/biRecord.dbd.pod
anjohnson, line 124: Blank line needed above the `=head3` here

PR #897, opened 2026-06-22
macOS: Drop the XCode -flat_namespace flag when possible

### LINKED ISSUE #895: Usage of `flat_namespace` is not recommended on Apple platforms
It seems the builds for epics-base use `-flat_namespace` but there is no good reason to use that anymore since the move to Intel by Apple.

For reference: https://github.com/Homebrew/homebrew-core/pull/114543#issuecomment-1330719444

anjohnson:
Exactly when this flag was no longer required is unclear, see some discussion in the linked issue.

Fixes #895.

### DIFF
--- a/configure/os/CONFIG.darwinCommon.darwinCommon
+++ b/configure/os/CONFIG.darwinCommon.darwinCommon
@@ -60,7 +60,9 @@ GNU = NO
 #
 # Darwin shared libraries
 #
-SHRLIB_LDFLAGS = -dynamiclib -flat_namespace \
+FLAT_NAMESPACE = $(shell \
+    test $(or $(CLANG_MAJOR),99) -lt 15 && echo -flat_namespace)
+SHRLIB_LDFLAGS = -dynamiclib $(FLAT_NAMESPACE) \
     -install_name $(abspath $(INSTALL_LIB))/$@ \
     $(addprefix -compatibility_version , $(SHRLIB_VERSION)) \
     $(addprefix -current_version , $(SHRLIB_VERSION))
--- a/documentation/new-notes/PR-897.md
+++ b/documentation/new-notes/PR-897.md
@@ -1,0 +1,9 @@
+### macOS: Drop the `-flat_namespace` linker flag
+
+It's not clear when this flag was no longer required on macOS, we
+tried dropping it
+[in 2019](https://github.com/epics-base/epics-base/commit/cbb13bf6b170c6ba6660d594bdd9ca44a5bdaea5)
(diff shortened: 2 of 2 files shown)


configure/os/CONFIG.darwinCommon.darwinCommon
```diff
@@ -60,7 +60,9 @@ GNU = NO
 #
 # Darwin shared libraries
 #
-SHRLIB_LDFLAGS = -dynamiclib -flat_namespace \
+FLAT_NAMESPACE = $(shell \
+    test $(or $(CLANG_MAJOR),99) -lt 15 && echo -flat_namespace)
+SHRLIB_LDFLAGS = -dynamiclib $(FLAT_NAMESPACE) \
```
simon-ess, line 65: Too bad we can't guarantee a sufficiently new version of gnu make, which includes the `intcmp` function: https://www.gnu.org/software/make/manual/html_node/Conditional-Functions.html

(Release in make 4.4 as of 2022... https://lists.gnu.org/archive/html/info-gnu/2022-10/msg00008.html)

anjohnson: Ack; unfortunately Apple is violently allergic to GPLv3:
```
Cox$ /usr/bin/make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

simon-ess: It's not just apple; I think something released in 2022 is not going to be standard in many cases. Our Ubuntu22 VMs don't have this version of make either...

It's really too bad because it's quite a nice feature!

simon-ess: (Now if we want to be _really_ fancy  we can can do:
```make
MAKE_PARTS=$(subst ., ,$(MAKE_VERSION))
MAKE_MAJOR=$(firstword $(MAKE_PARTS))
MAKE_MINOR=$(word 2,$(MAKE_PARTS))

MAKE_NEW_ENOUGH=$(shell test $(MAKE_MAJOR) -gt 3 && test $(MAKE_MINOR) -gt 3 && echo YES)

ifeq ($(MAKE_NEW_ENOUGH),YES)
FLAT_NAMESPACE=$(intcmp $(CLANG_MAJOR),15,-flat_namespace)
else
FLAT_NAMESPACE = $(shell \
    test $(or $(CLANG_MAJOR),99) -lt 15 && echo -flat_namespace)
endif
```
But let's be clear, this is an extremely silly idea. 😁

PR #899, opened 2026-06-24
database: Replace epicsEventMustSignal -> epicsEventMustTrigger in comments

JJL772:
Very minor comment update


mdavidsaver:
`epicsEventTrigger()` has been in since 3.15.0.1 (2011).  Long enough for me to forget why this new name was added.  So, sure.  Let's update the examples.

anjohnson:
History: In my self-appointed role as the EPICS Naming Czar, I decided that since it wasn't immediately obvious when reading code whether the `Signal()` routines announced events or waited for them (think traffic signals) I would rename them to `Trigger()` and create aliases for backwards-compatibility. The period between 3.14.x and 3.15.x _was_ a bit like the wild west…

PR #900, opened 2026-06-25
Make EPICS_ABORT_ON_ASSERT affect cantProceed()

JJL772:
When I implemented `EPICS_ABORT_ON_ASSERT` (in #565) I completely forgot about the `xxxMustBlah` APIs that EPICS has. Unfortunately, as a result, that PR introduced some inconsistency in the behavior of those functions. All `xxxMustBlah` APIs call `cantProceed` except for `epicsThreadMustCreate` and `epicsMutexMustLock` which use `assert()`, and are thus affected by `EPICS_ABORT_ON_ASSERT`.

This PR does two things:
- Switches `epicsMutexMustLock` and `epicsThreadMustCreate` to `cantProceed`
- Makes `EPICS_ABORT_ON_ASSERT` change the behavior of `cantProceed`

When `EPICS_ABORT_ON_ASSERT=YES`, it is probably desirable to kill the IOC on failure of `xxxMustBlah` instead of suspending the calling thread. I'm not sure if this is worth a new configuration option or not.

### DIFF
--- a/modules/libcom/src/misc/cantProceed.c
+++ b/modules/libcom/src/misc/cantProceed.c
@@ -15,6 +15,7 @@
 #include <stdarg.h>
 
 #include "errlog.h"
+#include "envDefs.h"
 #include "cantProceed.h"
 #include "epicsThread.h"
 #include "epicsStackTrace.h"
@@ -53,20 +54,30 @@ LIBCOM_API void * mallocMustSucceed(size_t size, const char *msg)
 
 LIBCOM_API void cantProceed(const char *msg, ...)
 {
+    int shouldAbort = 0;
     va_list pvar;
     va_start(pvar, msg);
     if (msg)
         errlogVprintf(msg, pvar);
     va_end(pvar);
 
-    errlogPrintf(ANSI_RED("CRITICAL ERROR") " Thread %s (%p) can't proceed, suspending.\n",
-            epicsThreadGetNameSelf(), epicsThreadGetIdSelf());
+    envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &shouldAbort);
+
+    errlogPrintf(ANSI_RED("CRITICAL ERROR") " Thread %s (%p) can't proceed, %s.\n",
+            epicsThreadGetNameSelf(), epicsThreadGetIdSelf(), shouldAbort ? "aborting" : "suspending");
 
     epicsStackTrace();
 
     errlogFlush();
 
     epicsThreadSleep(1.0);
-    while (1)
-        epicsThreadSuspendSelf();
+
+    if (shouldAbort) {
+        errlogPrintf("cantProceed() Calling abort()\n");
+        errlogFlush();
+        abort();
+    } else {
+        while (1)
+            epicsThreadSuspendSelf();
+    }
 }
--- a/modules/libcom/src/misc/cantProceed.h
+++ b/modules/libcom/src/misc/cantProceed.h
@@ -34,12 +34,14 @@
 extern "C" {
 #endif
 
-/** \brief Suspend this thread, the caller cannot continue or return.
+/** \brief Suspend this thread or call abort(), the caller cannot continue
+ *         or return.
  *
  * The effect of calling this is to print the error message followed by
- * the name of the thread that is being suspended. A stack trace will
+ * the name of the thread that can't proceed. A stack trace will
  * also be shown if supported by the OS, and the thread is suspended
- * inside an infinite loop.
+ * inside an infinite loop, or the process aborted, depending on
+ * EPICS_ABORT_ON_ASSERT.
  * \param errorMessage A printf-style error message describing the error.
  * \param ... Any parameters required for the error message.
  */
--- a/modules/libcom/src/osi/epicsMutex.h
+++ b/modules/libcom/src/osi/epicsMutex.h
@@ -213,12 +213,12 @@ LIBCOM_API epicsMutexLockStatus epicsStdCall epicsMutexLock(
  * This routine does not return if the identifier is invalid.
  * \param ID The mutex identifier.
  **/
-#define epicsMutexMustLock(ID) {                            \
-    epicsMutexLockStatus status = epicsMutexLock(ID);       \
-    if(status != epicsMutexLockOK) {                        \
-        cantProceed("epicsMutexMustLock() failed at %s:%d", \
-            __FILE__, __LINE__);                            \
-    }                                                       \
+#define epicsMutexMustLock(ID) {                                \
+    epicsMutexLockStatus status = epicsMutexLock(ID);           \
+    if(status != epicsMutexLockOK) {                            \
+        cantProceed("epicsMutexMustLock() failed at %s:%d\n",   \
+            __FILE__, __LINE__);                                \
+    }                                                           \
 }
 
 /**\brief Similar to epicsMutexLock() except that the call returns immediately,
--- a/modules/libcom/src/osi/epicsMutex.h
+++ b/modules/libcom/src/osi/epicsMutex.h
@@ -41,6 +41,7 @@
 #ifndef epicsMutexh
 #define epicsMutexh
 
+#include "cantProceed.h"
 #include "epicsAssert.h"
 
 #include "libComAPI.h"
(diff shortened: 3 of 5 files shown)


mdavidsaver:
> * Switches `epicsMutexMustLock` and `epicsThreadMustCreate` to `cantProceed`

Why make this change?

This would be a behavior change as `assert()` can be disabled by`-DNDEBUG`.

Otherwise, this seems reasonable.

modules/libcom/src/misc/cantProceed.c
```diff
@@ -67,6 +69,11 @@ LIBCOM_API void cantProceed(const char *msg, ...)
     errlogFlush();
 
     epicsThreadSleep(1.0);
-    while (1)
-        epicsThreadSuspendSelf();
+
+    if (envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &doAbort) == 0 && doAbort) {
+        abort();
```
mdavidsaver: For consistence with `epicsAssert()` add:

```c
        errlogPrintf("cantProceed() Calling abort()\n");
        errlogFlush();
```

JJL772:
> This would be a behavior change as assert() can be disabled by -DNDEBUG.

Is this the desired behavior though? It seems odd to me that `epicsMutexMustLock`/`epicsThreadMustCreate` would specifically change behavior based on `NDEBUG`, but the rest don't.

mdavidsaver:
I can not dispute the "oddness", and would personally not mind this change `cantProceed()`.  I am only flagging this for @anjohnson who has in the past held Opinions on topics like this.

modules/libcom/src/misc/cantProceed.c
anjohnson: This error message probably shouldn't say "suspending" any more if it's about to call `abort()`. Maybe move the code that sets `doAbort` above here and change the message to `(doAbort ? "aborting process" : "suspending task")` or similar?

modules/libcom/src/osi/epicsMutex.h
```diff
@@ -212,9 +213,12 @@ LIBCOM_API epicsMutexLockStatus epicsStdCall epicsMutexLock(
...
  **/
-#define epicsMutexMustLock(ID) {                        \
-    epicsMutexLockStatus status = epicsMutexLock(ID);   \
-    assert(status == epicsMutexLockOK);                 \
+#define epicsMutexMustLock(ID) {                            \
+    epicsMutexLockStatus status = epicsMutexLock(ID);       \
+    if(status != epicsMutexLockOK) {                        \
+        cantProceed("epicsMutexMustLock() failed at %s:%d", \
+            __FILE__, __LINE__);                            \
+    }                                                       \
```
anjohnson: The old code here was calling `assert()` which becomes `epicsAssert()` unless `NDEBUG` is defined. Defining `NDEBUG` actually breaks the promise of the `epicsMutexMustLock()` to not return if there's an error, so this change fixes a bug there. The choice between `abort()` and suspending the task is controlled by the `EPICS_ABORT_ON_ASSERT` variable, which is correct.

modules/libcom/src/osi/epicsThread.cpp
```diff
@@ -345,7 +345,8 @@ extern "C" {
     {
         epicsThreadId id = epicsThreadCreate (
             name, priority, stackSize, funptr, parm );
-        assert ( id );
+        if ( !id )
+            cantProceed("epicsThreadMustCreate: epicsThreadCreate failed\n");
```
anjohnson, line 349: As above, the original `assert()` call would be elided it `NDEBUG` is defined at build time, so I think this is also a correct fix.

anjohnson (changes requested):
Minor change requested in the error message from `cantProceed()`, but otherwise this looks good to me.

anjohnson:
I forgot to mention in my review just now that both RTEMS-4 CI builds of this PR are failing with the same compiler errors, I think this is the right command-line:
```
  make[4]: Entering directory '/home/runner/work/epics-base/epics-base/modules/libcom/src/O.RTEMS-pc386-qemu'
  In file included from ../misc/cantProceed.h:28,
  /opt/rtems/4.9/bin/i386-rtems4.9-g++ -B/opt/rtems/4.9/i386-rtems4.9/pc386/lib/ -specs bsp_specs -qrtems    -mtune=i386               -DUNIX      -O2 -g -g   -Wall         -D__LINUX_ERRNO_EXTENSIONS__ -DRTEMS_LEGACY_STACK   -I. -I../O.Common -I. -I../osi/compiler/gcc -I../osi/compiler/default -I. -I../osi/os/RTEMS-score -I../osi/os/RTEMS -I../osi/os/posix -I../osi/os/default -I.. -I../as -I../bucketLib -I../calc -I../cvtFast -I../cppStd -I../cxxTemplates -I../dbmf -I../ellLib -I../env -I../error -I../fdmgr -I../flex -I../freeList -I../gpHash -I../iocsh -I../log -I../macLib -I../misc -I../osi -I../pool -I../ring -I../taskwd -I../timer -I../yacc -I../yacc -I../yajl -I../../../../include/compiler/gcc -I../../../../include/os/RTEMS -I../../../../include         -c ../osi/os/RTEMS-score/osdTime.cpp
                   from ../osi/epicsMutex.h:44,
                   from ../osi/epicsThread.h:365,
                   from ../timer/epicsTimer.h:21,
                   from ../osi/generalTimeSup.h:15,
                   from ../osi/os/RTEMS-score/osdTime.cpp:29:
  /opt/rtems/4.9/lib/gcc/i386-rtems4.9/4.3.2/../../../../i386-rtems4.9/include/stdlib.h:80:25: error: macro "free" requires 2 arguments, but only 1 given
  /opt/rtems/4.9/lib/gcc/i386-rtems4.9/4.3.2/../../../../i386-rtems4.9/include/stdlib.h:87:35: error: macro "malloc" requires 3 arguments, but only 1 given
  In file included from ../misc/cantProceed.h:28,
                   from ../osi/epicsMutex.h:44,
                   from ../osi/epicsThread.h:365,
                   from ../timer/epicsTimer.h:21,
                   from ../osi/generalTimeSup.h:15,
                   from ../osi/os/RTEMS-score/osdTime.cpp:29:
  /opt/rtems/4.9/lib/gcc/i386-rtems4.9/4.3.2/../../../../i386-rtems4.9/include/stdlib.h:80: error: variable or field 'free' declared void
  ../osi/os/RTEMS-score/osdTime.cpp:172: warning: 'done' defined but not used
  make[4]: *** [../../../../configure/RULES_BUILD:268: osdTime.o] Error 1
```

I have no idea what's causing those failures, but they do need to be fixed.

mdavidsaver:
> I have no idea what's causing those failures, but they do need to be fixed.

The trigger must be the changes in `epicsMutex.h`.

JJL772:
`rtems_bsdnet_internal` does `#define free ...` and `#define malloc ...` which breaks `stdlib.h` in that case.

modules/libcom/src/misc/cantProceed.c
```diff
@@ -53,20 +54,30 @@ LIBCOM_API void * mallocMustSucceed(size_t size, const char *msg)
...
+    int doAbort = 0;
     va_list pvar;
     va_start(pvar, msg);
     if (msg)
         errlogVprintf(msg, pvar);
     va_end(pvar);
 
-    errlogPrintf(ANSI_RED("CRITICAL ERROR") " Thread %s (%p) can't proceed, suspending.\n",
-            epicsThreadGetNameSelf(), epicsThreadGetIdSelf());
+    envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &doAbort);
```
mdavidsaver: Probably just pedantry.  If we are going to introduce duplication between `canProceed.c` and `osdAssert.c`, then I think the two conditions should be identical.

```c
    if (envGetBoolConfigParam(&EPICS_ABORT_ON_ASSERT, &shouldAbort) == 0 && shouldAbort) {
```

mdavidsaver (changes requested):
It looks like that weird header ordering issue has been sorted out, so I think I have just one more request (famous !last words).

modules/libcom/src/misc/cantProceed.h
```diff
@@ -34,12 +34,13 @@
...
 
-/** \brief Suspend this thread, the caller cannot continue or return.
+/** \brief Suspend this thread or call abort(), the caller cannot continue
+ *         or return.
  *
  * The effect of calling this is to print the error message followed by
  * the name of the thread that is being suspended. A stack trace will
  * also be shown if supported by the OS, and the thread is suspended
- * inside an infinite loop.
+ * inside an infinite loop, or aborted, depending on EPICS_ABORT_ON_ASSERT.
```
anjohnson: A couple of nits on the wording, since this is used to generate the online documentation.
- s/that is being suspended./that can't proceed./
- s/or aborted/or the process is aborted/ — clarify that abort means the whole process, we can't just stop that thread.

modules/libcom/src/misc/cantProceed.c
JJL772: I would like to avoid duplicating the entire `errlogPrintf` call. `envGetBoolConfigParam` does not modify `&shouldAbort` on failure, so the return value check would be redundant (clearly I missed that part of the doc string on my previous PR :))

nonetheless I have renamed `doAbort` -> `shouldAbort`

anjohnson (approved):
LGTM.

modules/libcom/src/misc/cantProceed.c
mdavidsaver: > clearly I missed that part of the doc string on my previous PR :))

Will you remember this understanding 5 years from now?  I am not arguing for one or the other logic, just that they be consistent.  If you like the approach of this PR, then change osdAssert.c to match.

Issue #901, opened 2026-06-30
7.0: too many arguments to function ‘pdrvet->report’


matwey:
Hello,

I am running gcc 15.3.0 and trying to build epics-base from current `7.0` branch (the latest release also has this issue).
I see the following inconsistent behavior.

When I do (to trigger the issue is enough to just specify empty `USR_CFLAGS` and `USR_CXXFLAGS`)
```bash
make USR_CFLAGS='' USR_CXXFLAGS=''
```
I see the following
```
/usr/bin/gcc  -D_GNU_SOURCE -D_DEFAULT_SOURCE        -DUSE_TYPED_RSET -DUSE_TYPED_DSET  -D_X86_64_ -DUNIX  -Dlinux   -DBUILDING_dbCore_API   -O3 -g   -Wall -Werror-implicit-function-declaration      -mtune=generic     -m64  -fPIC -I. -I../O.Common -I. -I. -I.. -I../as -I../bpt -I../db -I../dbStatic -I../dbtemplate -I../misc -I../registry -I../rsrv -I../../../../../include/compiler/gcc -I../../../../../include/os/Linux -I../../../../../include         -c ../db/dbTest.c
../db/dbTest.c: In function ‘dbior’:
../db/dbTest.c:744:13: error: too many arguments to function ‘pdrvet->report’; expected 0, have 1
  744 |             pdrvet->report(interest_level);
      |             ^~~~~~         ~~~~~~~~~~~~~~
In file included from ../dbStatic/dbBase.h:26,
                 from ../db/dbAccessDefs.h:20,
                 from ../db/dbTest.c:27:
../dbStatic/drvSup.h:48:15: note: declared here
   48 |     DRVSUPFUN report;   /* print report */
      |               ^~~~~~
```
When I do
```bash
make
```
then the build is successful. Comparing build commands for `dbTest.c` I see that `-DUSE_TYPED_DRVET` is missed in the former case.

matwey:
From what I can see, the compiler is right:
https://github.com/epics-base/epics-base/blob/cf85a1a5eb86928323d4c62d4d3c553f362a7d04/modules/database/src/ioc/dbStatic/drvSup.h#L26
`struct typed_drvet` and `struct drvet` are inconsistent with each other. Field `report` clearly has different type.

anjohnson:
Hi Matwey,

[...]

PR #902, opened 2026-06-30
Fix drvet registrations

anjohnson:
We missed a change related to `typed_drvet` in the generated `RecordDeviceDriver.cpp` file, which fails to compile when the Makefile building the IOC defines the `USE_TYPED_DRVET` macro in `USR_CXXFLAGS` or `USR_CPPFLAGS`.

The fix is in the Perl script, the other code here adds a driver to the exampleApp template as a regression test. This also removes an anachronism so we always use `USR_CPPFLAGS` to define the `USE_TYPED_…` variables in Base (see issue #901).


modules/database/src/template/top/exampleApp/src/devXxxSoft.c
```diff
@@ -48,3 +49,22 @@ xxxdset devXxxSoft = {
...
 };
 epicsExportAddress(dset, devXxxSoft);
+
+
+static long xxxReport(int level) {
+    printf("xxxDriver: Report called\n");
+    return 0;
+}
+
+static long xxxInitialise(void) {
```
mdavidsaver: Personally, I would not demonstrate `drvet::init()`.  It is called too early to do much.

modules/database/src/template/top/exampleApp/src/devXxxSoft.c
```diff
@@ -48,3 +49,22 @@ xxxdset devXxxSoft = {
     read_xxx,
 };
 epicsExportAddress(dset, devXxxSoft);
+
+
+static long xxxReport(int level) {
+    printf("xxxDriver: Report called\n");
```
mdavidsaver: Perhaps some mention of `dbior()` as a hint as to why this callback is useful?

mdavidsaver (approved):
A couple of thoughts, but not objections.

modules/database/src/template/top/exampleApp/src/devXxxSoft.c
anjohnson: Good idea, done.

modules/database/src/template/top/exampleApp/src/devXxxSoft.c
anjohnson: There are only 2 methods in the drvet, it looks lost without this!
I made your point in an added comment.

mdavidsaver: "Let me call your attention to that which you should ignore."  Far from the only distraction in the example template.  I won't pursue this point further than to suggest:

```c
drvet xxxDriver = {
    2,
    xxxReport,
    NULL /* legacy */
};
```

anjohnson: Removed.
I also made some other internal changes in the latest force-push including making variable names more like what we use in the base record and device code, which I now realize I should have put in a separate commit, sorry about that.

modules/database/src/template/top/exampleApp/src/devXxxSoft.c
```diff
@@ -1,50 +1,70 @@
...
+/* Driver Routines */
+
+/* The report method is called by the iocsh command 'dbior'.
+ * Use level to show more or different information about
+ * the state of the driver software, connections etc.
+ */
+static long xxxReport(int level) {
+    printf("xxxDriver: Report called, level %d\n", level);
+    return 0;
+}
```
mdavidsaver, line 59: Also, should include `epicsStdio.h` so that stdout/err redirections are done.  Maybe use `epicsStdoutPrintf()` explicitly?  This omission is one of the more common mistakes I see (and make) with functions called through iocsh.

anjohnson: Ack, done.

PR #907, opened 2026-07-10
Add protection to CAC library from network messages

anjohnson:
5 issues that could allow messages from a malicious CA server to affect client applications.

### DIFF
--- a/modules/ca/src/client/cac.cpp
+++ b/modules/ca/src/client/cac.cpp
@@ -1107,6 +1107,9 @@ bool cac::exceptionRespAction ( callbackManager & cbMutexIn, tcpiiu & iiu,
     if ( hdr.m_postsize < bytesSoFar ) {
         return false;
     }
+    // Ensure the context string is NULL-terminated
+    static_cast < char * > ( pMsgBdy ) [hdr.m_postsize - 1] = '\0';
+
     caHdrLargeArray req;
     req.m_cmmd = AlignedWireRef < const epicsUInt16 > ( pReq->m_cmmd );
     req.m_postsize = AlignedWireRef < const epicsUInt16 > ( pReq->m_postsize );
--- a/modules/ca/src/client/cac.cpp
+++ b/modules/ca/src/client/cac.cpp
@@ -895,6 +895,13 @@ bool cac::readNotifyRespAction ( callbackManager &, tcpiiu & iiu,
             // this does *not* assign a new resource id
             this->ioTable.add ( *pmiu );
         }
+        if ( INVALID_DB_REQ ( hdr.m_dataType ) ||
+             hdr.m_postsize < dbr_size_n ( hdr.m_dataType, hdr.m_count ) ) {
+            pmiu->exception ( guard, *this, ECA_BADTYPE,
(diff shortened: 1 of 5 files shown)


modules/ca/src/client/cac.cpp
```diff
@@ -895,6 +895,13 @@ bool cac::readNotifyRespAction ( callbackManager &, tcpiiu & iiu,
             // this does *not* assign a new resource id
             this->ioTable.add ( *pmiu );
         }
+        if ( INVALID_DB_REQ ( hdr.m_dataType ) ||
+             hdr.m_postsize < dbr_size_n ( hdr.m_dataType, hdr.m_count ) ) {
```
anjohnson, line 899: Core:

Q: Has `hdr.m_postsize` been validated against the buffer length before the action routines are called?
Similarly are there bounds on `hdr.m_count`? Can there be?

modules/ca/src/client/getCopy.cpp
```diff
@@ -59,7 +59,7 @@ void getCopy::completion (
     epicsGuard < epicsMutex > & guard, unsigned typeIn,
     arrayElementCount countIn, const void *pDataIn )
 {
-    if ( this->type == typeIn ) {
+    if ( this->type == typeIn && countIn <= this->count ) {
         unsigned size = dbr_size_n ( typeIn, countIn );
```
anjohnson, line 63: Core: Possible overflows in `dbr_size_n()`, have checks been done before here?
Find the code that zero-fills the remainder of the buffer.

modules/ca/src/client/syncGroupReadNotify.cpp
```diff
@@ -88,7 +90,8 @@ void syncGroupReadNotify::completion (
         return;
     }
 
-    if ( this->pValue ) {
+    if ( this->pValue && type == this->requestedType &&
+         count <= this->requestedCount ) {
```
anjohnson, line 94: Core: See previous commit questions.

anjohnson:
Core: Suggestion for bounds checking, check the channel's element count (ensure the smaller counts can't overflow an uint32), use that count as the array limit for incoming message checks (if not already done).

modules/ca/src/client/cac.cpp
anjohnson: Yes, `hdr.m_postsize` has been checked before these 3 routines are called. tcpiiu.cpp:1208 grows the buffer if necessary; tcpiiu.cpp:1252 bounds-checks it and the code then copies the message into `pCurData` before `executeResponse()` is called.

However the `dbr_size_n()` calculation can wrap when `hdr.m_count` comes from a large array header, where it's a uint32. In the small array header it's a uint16 where the values are safe. The changes here don't currently fix that.

I asked about making `dbr_size_n()` an inline function, which [Claude liked](https://gist.github.com/anjohnson/afe91e41b99f40906b32c63903ca44ab) and tested – there's a git patch file in that gist. The "Fix 2" that it mentions are the changes in this commit, 6c449aa.

anjohnson:
Superceded by #931

Issue #667, opened 2025-06-09
asDbLib only once


mdavidsaver:
It is currently not possible to restart a isolated test IOC which includes ACF rules.  The stall point is slightly different if no ACF is loaded the first time around, but the effect is the same.

There is currently no unit test covering `asDbLib.c`.

PR #912, opened 2026-07-15
Assorted libcom/database memory-safety hardening

jacquelinegarrahan:
Reworked per review: the CA/RSRV header-validation changes have been removed
from this PR (they are being handled in #934 for RSRV and #931 for libca),
and what remains is a sequence of one-commit-per-change, independent
memory-safety fixes in libcom/database. Apologies again for the earlier
bundling and for cutting across the private disclosure on the CA parts.

All six commits build cleanly (`make`, no new warnings). The reworked
`dbLoadTemplate` accumulator was smoke-tested with `softIoc` (records expand
correctly, including quoted values containing `,` and `=`).

### Commits
1. **asLib: bound numeric parse-error message formatting** — over-range numeric
   literal in a `.acf` formatted unbounded `yytext` into `message[40]`; use
   `epicsSnprintf`.
2. **asShutdown: clear asActive to keep the asActive/pasbase invariant** — plus
   a comment documenting the relationship at their definitions (may partly
   address #667).
3. **calc: bound the postfix expression translation stack** — nested parens
   could overflow the fixed `stack[80]`; guard each push with
   `CALC_ERR_OVERFLOW`.
4. **dbStaticLib: size CALC validation buffer from the field capacity** — the
   `RPCL_LEN`/`MAX_POSTFIX_SIZE` invariant note is still open; happy to switch
   to `MAX_POSTFIX_SIZE` (fixing the second use in `dbVerify()` too) once you
   and @anjohnson settle on the preferred form.
5. **dbLoadTemplate: bound the substitution accumulator** — the four append
   groups now go through a bounded `sub_append(name, value, quoteme)` that
   tracks the write position (no `strlen(sub_collect)`) and returns non-zero on
   overflow so the yacc actions `YYABORT` instead of loading a truncated
   substitution.
6. **iocLogServer: bound the getDirectory() filename copy** — `strcpy` of an up
   to 511-char name into `temp[256]`; use bounded `epicsSnprintf`. (Noted this
   is low priority / effectively deprecated.)

### DIFF
--- a/modules/database/src/ioc/as/asDbLib.c
+++ b/modules/database/src/ioc/as/asDbLib.c
@@ -157,6 +157,10 @@ int asInit(void)
 int asShutdown(void) {
     volatile ASBASE *pbase = pasbase;
     pasbase = NULL;
+    /* restore the invariant that asActive implies pasbase != NULL; leaving
+       asActive TRUE here would let later AS calls dereference the freed/NULL
+       base or consult stale cached access rights */
+    asActive = FALSE;
     firstTime = TRUE;
     if(pbase)
         asFreeAll((ASBASE*)pbase);
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -2591,7 +2591,10 @@ long dbPutString(DBENTRY *pdbentry,const char *pstring)
         ((char *)pfield)[pflddes->size-1] = 0;
 
         if((pflddes->special == SPC_CALC) && !stringHasMacro) {
-            char  rpcl[RPCL_LEN];
+            /* size the postfix output from the field's actual capacity;
+               a fixed RPCL_LEN sized for an 80-char infix overflows for the
+               larger CALC-type fields (e.g. size 160) */
+            char *rpcl = dbmfMalloc(INFIX_TO_POSTFIX_SIZE(pflddes->size));
             short err;
 
             if (postfix(pstring,rpcl,&err)) {
@@ -2599,6 +2602,7 @@ long dbPutString(DBENTRY *pdbentry,const char *pstring)
                 errlogPrintf("%s in CALC expression '%s'\n",
                               calcErrorStr(err), pstring);
             }
+            dbmfFree(rpcl);
         }
         break;
 
@@ -2746,10 +2750,14 @@ char * dbVerify(DBENTRY *pdbentry, const char *pstring)
         }
 
         if (pflddes->special == SPC_CALC) {
-            char  rpcl[RPCL_LEN];
+            /* size the postfix output from the field's actual capacity;
+               a fixed RPCL_LEN sized for an 80-char infix overflows for the
+               larger CALC-type fields (e.g. size 160) */
+            char *rpcl = dbmfMalloc(INFIX_TO_POSTFIX_SIZE(pflddes->size));
             short err;
 
             status = postfix(pstring, rpcl, &err);
+            dbmfFree(rpcl);
             if (status)  {
                 sprintf(message,"%s in CALC expression '%s'",
                         calcErrorStr(err), pstring);
--- a/modules/database/src/ioc/dbtemplate/dbLoadTemplate.y
+++ b/modules/database/src/ioc/dbtemplate/dbLoadTemplate.y
@@ -31,9 +31,40 @@ static int yyerror(char* str);
 
 static char *sub_collect = NULL;
 static char *sub_locals;
+static char *sub_end;             /* write position (the trailing NUL) */
 static char **vars = NULL;
 static char *db_file_name = NULL;
 static int var_count, sub_count;
+static size_t sub_collect_size;   /* capacity of the sub_collect buffer */
+
+/* Append ",name=value" (or ",name=\"value\"" when quoteme) at sub_end.
+ * The names and values come from the substitution file and the command-line
+ * macro string and are not otherwise length limited, so the append is bounded
+ * against the fixed sub_collect allocation.  Returns non-zero (having logged)
+ * when the buffer is full so the caller can YYABORT rather than silently
+ * loading a truncated substitution. */
+static int sub_append(const char *name, const char *value, int quoteme)
+{
+    size_t need = 1 + strlen(name) + 1 + strlen(value) + (quoteme ? 2 : 0);
+
+    if (sub_end + need + 1 > sub_collect + sub_collect_size) {
+        fprintf(stderr, "dbLoadTemplate: substitution too long near line %d; "
+            "increase dbTemplateMaxVars\n", line_num);
+        return 1;
+    }
+    *sub_end++ = ',';
+    strcpy(sub_end, name);
+    sub_end += strlen(name);
+    *sub_end++ = '=';
+    if (quoteme)
+        *sub_end++ = '"';
+    strcpy(sub_end, value);
+    sub_end += strlen(value);
+    if (quoteme)
+        *sub_end++ = '"';
+    *sub_end = '\0';
+    return 0;
+}
 
 /* We allocate MAX_VAR_FACTOR chars in the sub_collect string for each
  * "variable=value," segment, and will accept at most dbTemplateMaxVars
@@ -92,7 +123,7 @@ global_definitions: GLOBAL O_BRACE C_BRACE
     #ifdef ERROR_STUFF
         fprintf(stderr, "global_definitions: %s\n", sub_collect+1);
     #endif
-        sub_locals += strlen(sub_locals);
+        sub_locals = sub_end;
     }
     ;
 
@@ -191,7 +222,8 @@ pattern_definition: global_definitions
         fprintf(stderr, "    dbLoadRecords(%s)\n", sub_collect+1);
     #endif
         if(msiLoadRecords(db_file_name, sub_collect+1)) YYABORT;
-        *sub_locals = '\0';
+        sub_end = sub_locals;
+        *sub_end = '\0';
         sub_count = 0;
     }
     | WORD O_BRACE pattern_values C_BRACE
@@ -207,7 +239,8 @@ pattern_definition: global_definitions
     #endif
         if(msiLoadRecords(db_file_name, sub_collect+1)) YYABORT;
         dbmfFree($1);
-        *sub_locals = '\0';
+        sub_end = sub_locals;
+        *sub_end = '\0';
         sub_count = 0;
     }
     ;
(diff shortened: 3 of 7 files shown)


pr.md
lucmaga, line 1: I guess this file was added by mistake.

mdavidsaver:
This PR includes a grab bag of unrelated issues.  I will make suggestions for decomposing it as I triage.

The header validation issues with the CA code had previously been reported privately, and were being worked on as part of a responsible disclosure process.  So much for that idea.  Since that cat is now out of the bag, I have opened https://github.com/epics-base/epics-base/pull/934.  The libca issues overlap with what is being worked on with https://github.com/epics-base/epics-base/pull/931 .

jacquelinegarrahan:
@mdavidsaver happy to close and break into sep prs

modules/database/src/ioc/as/asDbLib.c
```diff
@@ -157,6 +157,10 @@ int asInit(void)
 int asShutdown(void) {
     volatile ASBASE *pbase = pasbase;
     pasbase = NULL;
+    /* restore the invariant that asActive implies pasbase != NULL; leaving
+       asActive TRUE here would let later AS calls dereference the freed/NULL
+       base or consult stale cached access rights */
+    asActive = FALSE;
```
mdavidsaver, line 163: A reasonable change, which might be a partial solution to #667.

I think a comment describing the relationship between these two global variables would be a nice addition above their declarations/definitions in `asLibRoutines.c`.

mdavidsaver (commented):
One piece for a small PR.

jacquelinegarrahan:
> This PR includes a grab bag of unrelated issues. I will make suggestions for decomposing it as I triage.
> 
> The header validation issues with the CA code had previously been reported privately, and were being worked on as part of a responsible disclosure process. So much for that idea. Since that cat is now out of the bag, I have opened #934. The libca issues overlap with what is being worked on with #931 .

Big apology here — I didn't clock the header implications and ought to have moved into private channel. Happy to close this and re-open only the unrelated local fixes as separate small PRs if they're still useful.

modules/ca/src/client/convert.cpp
```diff
@@ -583,6 +583,11 @@ arrayElementCount   num         /* number of values     */
     pDest->status           = dbr_ntohs(pSrc->status);
     pDest->severity         = dbr_ntohs(pSrc->severity);
     pDest->no_str           = dbr_ntohs(pSrc->no_str);
+    /* the strs array has only MAX_ENUM_STATES entries; a hostile server can
+       supply a larger no_str, so clamp it to prevent out-of-bounds access by
+       consumers that iterate over strs[0..no_str-1] */
+    if ( pDest->no_str > MAX_ENUM_STATES )
+        pDest->no_str = MAX_ENUM_STATES;
```
mdavidsaver: Bounding `no_str` is reasonable.  Also check for negative values as `dbr_short_t no_str` is signed.

modules/ca/src/tools/tool_lib.c
```diff
@@ -311,6 +311,7 @@ char *val2str (const void *v, unsigned type, int index)
 
 #define PRN_DBR_X_ENUM(T)                               \
     n = ((struct T *)value)->no_str;                    \
+    if (n < 0 || n > MAX_ENUM_STATES) n = MAX_ENUM_STATES; \
```
mdavidsaver: With libca bounding `no_str`, is this check necessary?

mdavidsaver (commented):
`no_str` bounding in libca seems a separate piece.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -2591,14 +2591,18 @@ long dbPutString(DBENTRY *pdbentry,const char *pstring)
         ((char *)pfield)[pflddes->size-1] = 0;
 
         if((pflddes->special == SPC_CALC) && !stringHasMacro) {
-            char  rpcl[RPCL_LEN];
+            /* size the postfix output from the field's actual capacity;
+               a fixed RPCL_LEN sized for an 80-char infix overflows for the
+               larger CALC-type fields (e.g. size 160) */
+            char *rpcl = dbmfMalloc(INFIX_TO_POSTFIX_SIZE(pflddes->size));
```
mdavidsaver, line 2597: There was an unstated invariant between `RPCL_LEN` and `MAX_POSTFIX_SIZE` which has not been kept up to date since at least #655.

@anjohnson How to resolve?  Just replace `RPCL_LEN` with `MAX_POSTFIX_SIZE`?  Switch to `dbmfMalloc()` (with a `NULL` check)?  Something else?

There is another use of `RPCL_LEN` in `dbVerify()`.

modules/database/src/ioc/dbtemplate/dbLoadTemplate.y
```diff
@@ -34,6 +34,28 @@ static char *sub_locals;
 static char **vars = NULL;
 static char *db_file_name = NULL;
 static int var_count, sub_count;
+static size_t sub_collect_size;   /* capacity of the sub_collect buffer */
+static int sub_overflow;          /* set once an append has been rejected */
+
+/* Bounded append into sub_collect (via the moving sub_locals pointer).
```
mdavidsaver: Tracking only capacity seems a bit unusual.  Then again, the buffering scheme here is a bit unusual.

```
|,global=value,global=value,inst=value,inst=value\0..........|
 |                         |
 |                         \ - sub_local
 \ - sub_collect point to leading comma
```

modules/database/src/ioc/dbtemplate/dbLoadTemplate.y
```diff
@@ -34,6 +34,28 @@ static char *sub_locals;
...
+ * and the command-line macro string and are not otherwise length limited, so
+ * an unchecked strcat could overflow the fixed sub_collect allocation. */
+static void sub_append(const char *s)
+{
+    size_t used, len;
+    if (!sub_collect || sub_overflow)
+        return;
+    used = strlen(sub_collect);
+    len = strlen(s);
+    if (used + len + 1 > sub_collect_size) {
```
mdavidsaver: How about avoiding `strlen(sub_collect)` with `s + len + 1 < sub_collect + sub_collect_size`?

Also, instead of adding that special `sub_overflow` flag, yacc actions should do something like:

```c
if(sub_append(",")) YYABORT;
```

modules/database/src/ioc/dbtemplate/dbLoadTemplate.y
```diff
@@ -223,11 +245,11 @@ pattern_value: QUOTE
...
-            strcat(sub_locals, ",");
-            strcat(sub_locals, vars[sub_count]);
-            strcat(sub_locals, "=\"");
-            strcat(sub_locals, $1);
-            strcat(sub_locals, "\"");
+            sub_append(",");
+            sub_append(vars[sub_count]);
+            sub_append("=\"");
+            sub_append($1);
+            sub_append("\"");
```
mdavidsaver: I think an improvement would be to handle the four groups of `strcat()` with four calls to a helper function like:

```c
int sub_append(const char *value, int quoteme);
```

mdavidsaver (commented):
The `dbLoadTemplate.y` change can be separated.

modules/libcom/src/as/asLib_lex.l
```diff
@@ -54,7 +55,8 @@ INP{link} {
         char *end;
         if (epicsParseDouble((char *)yytext, &yylval.Float64, &end) ) {
             char message[40];
-            sprintf(message, "Error parsing Float64: %s", (char *)yytext);
+            epicsSnprintf(message, sizeof(message),
```
mdavidsaver, line 58: Use of `epicsSnprintf()` in asLib_lex.l seems a straightforward improvement.

mdavidsaver:
> @mdavidsaver happy to close and break into sep prs

Or multiple commits in sequence.  There are already a lot of PRs to keep track up.

modules/libcom/src/log/iocLogServer.c
```diff
@@ -989,11 +989,12 @@ static int getDirectory(void)
...
             char *slash = strrchr(ioc_log_file_name, '/');
-            char temp[256];
+            char temp[sizeof(ioc_log_file_name)];
 
             if (slash != NULL) name = slash + 1;
-            strcpy(temp,name);
-            sprintf(ioc_log_file_name,"%s/%s",dir,temp);
+            epicsSnprintf(temp, sizeof(temp), "%s", name);
+            epicsSnprintf(ioc_log_file_name, sizeof(ioc_log_file_name),
+                "%s/%s", dir, temp);
```
mdavidsaver, line 997: No objection.  `iocLocServer` is (to my mind at least) deprecated.  We generally recommend use of some (any) other logger receiver.  So I can't see a whole lot of reason to spend time on it.

jacquelinegarrahan:
Reorganized into commits!

modules/database/src/ioc/dbStatic/dbStaticLib.c
anjohnson: Our current approach to this is evidently fragile, and unnecessary.

`RPCL_LEN` is a macro defined earlier in this file, and IMHO it shouldn't exist at all now. I like the approach used here of calculating the output buffer size from the size of the field holding the infix expression. In both cases here the output from `postfix()` gets discarded anyway, so we don't need it to match any field lengths. We could even use the length of the new string's value instead of its capacity to calculate the buffer size, but that's not worth bothering about; this looks fine to me. The first line of the comment here is good, the other two aren't necessary though.

Please delete the `RPCL_LEN` macro, to ensure it can't be used again, and repeat the comment-shortening in `dbVerify()` below.

Thanks!

modules/ca/src/client/convert.cpp
anjohnson: Implemented in my cac-hardening branch.

modules/libcom/src/log/iocLogServer.c
anjohnson: I've disabled the build of the log-server in my ioc-hardening branch instead of applying this.

modules/libcom/src/as/asLib_lex.l
anjohnson: This commit included in my ioc-hardening branch.

modules/database/src/ioc/dbStatic/dbStaticLib.c
anjohnson: This commit is included in my ioc-hardening branch.

modules/database/src/ioc/as/asDbLib.c
anjohnson: This commit is included in my ioc-hardening branch.

PR #913, opened 2026-07-15
histogramRecord: fix heap OOB write when SGNL approaches ULIM

jacquelinegarrahan:
The bin-search loop in add_count() assumes nelm*wdth == ulim-llim, but
wdth is a rounded quotient so nelm*wdth can be strictly less than
ulim-llim.  A SGNL value just below ULIM then leaves the loop with
i == nelm+1, and pdest = bptr + nelm writes one epicsUInt32 past the
nelm-element buffer.  This is reachable over Channel Access by putting a
value near ULIM to the SGNL field (special SPC_MOD).  Clamp the index to
nelm.

### DIFF
--- a/modules/database/src/std/rec/histogramRecord.c
+++ b/modules/database/src/std/rec/histogramRecord.c
@@ -338,7 +338,8 @@ static long add_count(histogramRecord *prec)
         return 0;
 
     temp = prec->sgnl - prec->llim;
-    for (i = 1; i <= prec->nelm; i++){
+    /* stop at nelm-1 so a rounded wdth can't push i past the last bin */
+    for (i = 1; i < prec->nelm; i++){
         if (temp <= (double) i * prec->wdth)
             break;
     }


modules/database/src/std/rec/histogramRecord.c
```diff
@@ -342,6 +342,11 @@ static long add_count(histogramRecord *prec)
         if (temp <= (double) i * prec->wdth)
             break;
     }
+    /* wdth is a rounded quotient of (ulim-llim)/nelm, so nelm*wdth can be
+       slightly less than (ulim-llim); a sgnl just below ulim then leaves the
+       loop with i == nelm+1, which would index bptr one element past its end */
+    if (i > prec->nelm)
+        i = prec->nelm;
```
anjohnson: My Claude says "Hi!" to your Claude, and agrees that there is a problem with accessing one element beyond the end of the allocated storage.

However a simpler fix is to change the end condition in the `for ()` loop immediately above to be `i < prec->nelm` (dropping the `=` case) since the result is the same. That needs a different comment, but please try to keep it to 1 line to match the other comments in this source file.

anjohnson (changes requested):
Alternative and cleaner fix suggested.

jacquelinegarrahan:
Updated to suggested, thanks!

anjohnson (approved):
Thanks!

ralphlange (approved):
LGTM

PR #914, opened 2026-07-15
printfRecord: fix one-byte VAL overflow in %ls conversion

jacquelinegarrahan:
In the long-string (%ls) branch, n was vspace+1 and dbGetLink(DBR_CHAR)
fills up to n characters without reserving a terminator, so pval[n] = 0
writes at pval[vspace+1], which is one byte past the sizv-byte VAL buffer
(pval + vspace + 1 == prec->val + sizv).  Read at most vspace characters
(and precision, not precision+1) so the terminator lands in bounds.

### DIFF
--- a/modules/database/src/std/rec/printfRecord.c
+++ b/modules/database/src/std/rec/printfRecord.c
@@ -250,6 +250,8 @@ static void doPrintf(printfRecord *prec)
                             int padding;
 
                             /* Terminate string and measure its length */
+                            if (n > vspace)
+                                n = vspace;
                             pval[n] = 0;
                             added = strlen(pval);
                             padding = width - added;


modules/database/src/std/rec/printfRecord.c
```diff
@@ -228,18 +228,26 @@ static void doPrintf(printfRecord *prec)
 
                 case 's':
                     if (flags & F_LONG) {
-                        long n = vspace + 1;
+                        /* reserve one byte for the terminator written at
+                         * pval[n] below; pval has only vspace+1 usable bytes
+                         * (pval + vspace + 1 == prec->val + sizv), so reading
+                         * vspace+1 chars would push that NUL one past the end */
+                        long n = vspace;
```
anjohnson: So here you change `n` from holding the total buffer size including the null terminator, to holding the amount of space left for non-NULL characters. Then to compensate for that you then have to adjust its value for the 3 places lower down where the previous definition was already being relied on correctly.

There is a simpler fix just before the actual site of the bug, which only requires adding 2 lines:
```C
if (n > vspace)
    n = vspace;
```

pr.md
anjohnson, line 1: Please don't add verbose AI-written lists of bug descriptions to our source tree, they don't belong here.

A brief description of the specific problem the PR fixes is appropriate to include in the commit log message (which you did), and a slightly longer explanation in the GitHub PR description is fine if it's warranted. I suggest publishing this file as/in a Gist and including a link to it in the descriptions of all of the PRs that it covers if you want to make it available to us.

anjohnson (changes requested):
Thanks for the bug report. I accept that the problem analysis is correct, but I don't like the specific fix here, see my comment below.

jacquelinegarrahan:
Updated to suggested, thanks!

anjohnson (approved):
Thank-you!

ralphlange (approved):
LGTM

PR #915, opened 2026-07-15
epicsRingBytes: fix reversed operands in ResetHighWaterMark

jacquelinegarrahan:
The used-bytes count is nextPut - nextGet everywhere else in this file
(epicsRingBytesPut, epicsRingBytesUsedBytes), but
epicsRingBytesResetHighWaterMark computed nextGet - nextPut.  For a ring
holding u bytes it therefore stored size-u; since Put only raises the
mark, the high-water mark stayed permanently inflated and wrong.

### DIFF
--- a/modules/libcom/src/ring/epicsRingBytes.c
+++ b/modules/libcom/src/ring/epicsRingBytes.c
@@ -244,7 +244,7 @@ LIBCOM_API void epicsStdCall epicsRingBytesResetHighWaterMark(epicsRingBytesId i
     ringPvt *pring = (ringPvt *)id;
     int used;
     if (pring->lock) epicsSpinLock(pring->lock);
-    used = pring->nextGet - pring->nextPut;
+    used = pring->nextPut - pring->nextGet;
     if (used < 0) used += pring->size;
     pring->highWaterMark = used;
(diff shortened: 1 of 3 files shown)


modules/libcom/src/ring/epicsRingBytes.c
```diff
@@ -244,7 +244,7 @@ LIBCOM_API void epicsStdCall epicsRingBytesResetHighWaterMark(epicsRingBytesId i
     ringPvt *pring = (ringPvt *)id;
     int used;
     if (pring->lock) epicsSpinLock(pring->lock);
-    used = pring->nextGet - pring->nextPut;
+    used = pring->nextPut - pring->nextGet;
```
anjohnson, line 247: My Claude agrees with your Claude that this is a bug, and approves of this fix.

I asked mine why our unit test program didn't discover this. Apparently we're missing any test cases that would show the bug, as currently it only resets the HWM when the ring is empty. I have some extra tests to catch regressions in both ring implementations which I will append to this PR.

anjohnson (approved):
Thanks for this PR, it also points out where we have some missing tests.

ralphlange (approved):
LGTM

PR #916, opened 2026-07-15
epicsTime: fix normalization of pre-epoch epicsTimeAddSeconds results

jacquelinegarrahan:
When the accumulated nanosecond total went negative, secPastEpoch was
truncated toward zero while nsec used |nsec| % nSecPerSec, producing an
inconsistent (seconds, fraction) pair off by up to ~1 s.  Use a floored
division with a borrow so the fraction stays in [0, nSecPerSec).

### DIFF
--- a/modules/libcom/src/osi/epicsTime.cpp
+++ b/modules/libcom/src/osi/epicsTime.cpp
@@ -437,8 +437,17 @@ void epicsStdCall epicsTimeAddSeconds (epicsTimeStamp *pDest, double seconds)
     nsec *= nSecPerSec;
     nsec += epicsInt64(pDest->nsec);
     nsec += epicsInt64(seconds*1e9 + (seconds>=0.0 ? 0.5 : -0.5));
-    pDest->secPastEpoch = nsec/nSecPerSec;
-    pDest->nsec         = (nsec>=0 ? nsec : -nsec)%nSecPerSec;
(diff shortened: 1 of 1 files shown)


modules/libcom/src/osi/epicsTime.cpp
```diff
@@ -437,8 +437,17 @@ void epicsStdCall epicsTimeAddSeconds (epicsTimeStamp *pDest, double seconds)
     nsec *= nSecPerSec;
     nsec += epicsInt64(pDest->nsec);
     nsec += epicsInt64(seconds*1e9 + (seconds>=0.0 ? 0.5 : -0.5));
-    pDest->secPastEpoch = nsec/nSecPerSec;
-    pDest->nsec         = (nsec>=0 ? nsec : -nsec)%nSecPerSec;
+    /* normalise with a floored division so the fractional part stays in
+       [0, nSecPerSec); truncating toward zero and taking |nsec|%nSecPerSec
+       gives an inconsistent (seconds, nsec) pair when the total is negative */
```
anjohnson, line 442: Your PR's title refers to "pre-epoch results". An `epicsTimeStamp` structure consists of a pair of unsigned 32-bit integers which together represent a date/time between 1990-01-01T00:00Z and some moment in the year 2106. Dates before that epoch cannot be represented, so we aren't worried about incorrect normalization of any results in that range.

PR #917, opened 2026-07-15
errlog: fix out-of-bounds read when resizing the print buffer

jacquelinegarrahan:
errlogBufResize copied the print buffer from &pvt.print->base (the
address of the char* field inside the buffer_t struct) instead of
pvt.print->base (the buffer it points to), reading bufSize bytes from an
~8-byte struct field.  Drop the stray & so it reads the actual buffer,
matching the adjacent log-buffer copy.


simon-ess:
Seems correct to me. Mea culpa here as well!

simon-ess (approved):

PR #918, opened 2026-07-15
iocsh: fix inverted parse test in 'on error wait <delay>'

jacquelinegarrahan:
epicsParseDouble returns 0 on success and non-zero on failure, but the
branches were swapped: a valid delay printed "Invalid 'on error wait'
delay" and the usage text, while an invalid delay was silently accepted
with timeout forced to 5.0.  Report the error on parse failure and accept
the value on success.

### DIFF
--- a/modules/libcom/src/iocsh/iocsh.cpp
+++ b/modules/libcom/src/iocsh/iocsh.cpp
@@ -1559,13 +1559,14 @@ static void onCallFunc(const iocshArgBuf *args)
                 USAGE();
             } else if (epicsParseDouble(args->aval.av[3],
                           &context->scope->timeout, NULL)) {
-                context->scope->timeout = 5.0;
-            } else {
+                /* epicsParseDouble returns non-zero on failure */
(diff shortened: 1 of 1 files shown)


anjohnson:
@mdavidsaver I'm not completely sure what your original code was trying to do here. This change might be correct, but it's not obvious.

simon-ess (approved):
We've seen this sort of conditial oopsy in a few other places; I think that this fix is correct; currently if the parse succeeds we would hit the else branch and print usage. If the parse fails, it sets it to 5 but doesn't print a message. This would correct that.

This makes me wonder how often this is used given that no one has ever reported that
```
on error wait 1

iocInit
```
produces
```
on error wait 1
Usage: on error [continue | break | halt | wait <delay>]
Invalid 'on error wait' delay '1'.
```
as output...

mdavidsaver (approved):
It looks like my original code handles parse errors correctly, but prints the error in the wrong case.

PR #919, opened 2026-07-15
dbConvert: fix UInt64->String conversion truncating to 32 bits

jacquelinegarrahan:
putUInt64String used cvtUlongToString, which is cvtUInt32ToString, so the
64-bit source was narrowed to its low 32 bits (e.g. 0x1_0000_0007 printed
as "7").  Use cvtUInt64ToString, matching getUInt64String and every other
width-specific putXxxString.

### DIFF
--- a/modules/database/src/ioc/db/dbConvert.c
+++ b/modules/database/src/ioc/db/dbConvert.c
@@ -1528,12 +1528,12 @@ static long putUInt64String(dbAddr *paddr,
 
 
     if (nRequest==1 && offset==0) {
-        cvtUlongToString(*psrc, pdst);
+        cvtUInt64ToString(*psrc, pdst);
         return 0;
     }
     pdst += (size*offset);
     while (nRequest--) {
-        cvtUlongToString(*psrc, pdst);
+        cvtUInt64ToString(*psrc, pdst);
         psrc++;
         if (++offset == no_elements)
             pdst = (char *) paddr->pfield;


anjohnson (approved):
_mea culpa_

A comprehensive test of dbConvert.c is apparently called for, but may not actually be necessary now!

mdavidsaver (approved):
Agreed.

mdavidsaver:
I suppose this class of issue is hidden because we do not enable warnings for narrowing casts because of the high number of (false?) positives.

As a note, I also see warnings about deference before NULL check in `getFloatString()`, `getDoubleString()`, `putStringDevice()`, `putFloatString()`, and `putDoubleString()`.  I suspect there are not practically an issue as I do not think that there will be calls with null `dbAddr*`.

anjohnson:
The `if (paddr)` checks in those 4 converters between string and float/double are useless; passing a NULL for `paddr` omits either the source or destination value. I assume removing those checks will stop the warnings you're seeing.

I'm not sure why `putStringDevice()` might be generating a warning, that's must be a different cause.

I've added deleting those checks to my "minor-fixes" branch and will merge that sometime.

mdavidsaver:
> I assume removing those checks will stop the warnings you're seeing.

It should.

> I'm not sure why `putStringDevice()` might be generating a warning, that's must be a different cause.

Right, that case is a check of `dbFldDes *pdbFldDes`, which should also never be NULL.

anjohnson:
> that case is a check of `dbFldDes *pdbFldDes`

Thanks, added to my minor-fixes branch.

PR #920, opened 2026-07-15
dbStaticLib: fix dbGetStringNum reading UINT64 through a 32-bit pointer

jacquelinegarrahan:
The DBF_UINT64 case dereferenced the 8-byte field as *(epicsUInt32 *),
reading only its low 32 bits, so large UINT64 values printed wrong
(e.g. 0x1_0000_0000 as "0") and save/restore round-tripping broke.  Cast
to epicsUInt64*, matching the adjacent DBF_INT64 case.


anjohnson (approved):
Another _mea culpa_ from my int64 additions.

`modules/database/test/ioc/db/dbPutGetTest.c` might be the right place to test this, but that can come later.

ralphlange (approved):
LGTM - a pretty obvious one.

PR #922, opened 2026-07-15
dbJLink: fix assignment-instead-of-compare in dbJLinkMapAll

jacquelinegarrahan:
The guard used recname[0] = '\0' (assignment) instead of == , which wrote
a NUL into the caller's string, truncating it to "", and evaluated to 0
so recname was left non-NULL pointing at "".  Every record then failed
the name comparison and nothing was mapped.  Use ==, matching dbjlr.


anjohnson (approved):
Another oops from me.

PR #923, opened 2026-07-16
convert dbr_size_n from macro to inline function

dirk-zimoch:
The macro returned `unsigned`, the function now returns `size_t`.

The new function now asserts valid ranges of its arguments.

### DIFF
--- a/modules/ca/src/client/db_access.h
+++ b/modules/ca/src/client/db_access.h
@@ -63,8 +63,8 @@ typedef epicsOldString dbr_class_name_t;
 #define DBF_DOUBLE      6
 #define DBF_NO_ACCESS   7
 #define LAST_TYPE       DBF_DOUBLE
-#define VALID_DB_FIELD(x)       ((x >= 0) && (x <= LAST_TYPE))
-#define INVALID_DB_FIELD(x)     ((x < 0) || (x > LAST_TYPE))
+#define VALID_DB_FIELD(x)       ((unsigned)(x) <= LAST_TYPE)
+#define INVALID_DB_FIELD(x)     !VALID_DB_FIELD(x)
 
 /* data request buffer types */
 #define DBR_STRING      DBF_STRING
@@ -112,8 +112,8 @@ typedef epicsOldString dbr_class_name_t;
 #define DBR_STSACK_STRING DBR_PUT_ACKS + 1
 #define DBR_CLASS_NAME DBR_STSACK_STRING + 1
 #define LAST_BUFFER_TYPE        DBR_CLASS_NAME
-#define VALID_DB_REQ(x) ((x >= 0) && (x <= LAST_BUFFER_TYPE))
-#define INVALID_DB_REQ(x)       ((x < 0) || (x > LAST_BUFFER_TYPE))
+#define VALID_DB_REQ(x)     ((unsigned)(x) <= LAST_BUFFER_TYPE)
+#define INVALID_DB_REQ(x)   !VALID_DB_REQ(x)
 
 /*
  * The enumeration "epicsType" is an index to this array
@@ -668,8 +668,8 @@ union db_access_val{
 #define db_state_dim            MAX_ENUM_STATES
 #define db_state_text_dim       MAX_ENUM_STRING_SIZE
 
-#define dbf_type_is_valid(type)   ((type) >= 0 && (type) <= LAST_TYPE)
-#define dbr_type_is_valid(type)   ((type) >= 0 && (type) <= LAST_BUFFER_TYPE)
+#define dbf_type_is_valid(type)   VALID_DB_FIELD(type)
+#define dbr_type_is_valid(type)   VALID_DB_REQ(type)
 #define dbr_type_is_plain(type)   \
                 ((type) >= DBR_STRING && (type) <= DBR_DOUBLE)
 #define dbr_type_is_STS(type)   \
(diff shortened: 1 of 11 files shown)


modules/ca/src/client/db_access.h
```diff
@@ -63,8 +64,8 @@ typedef epicsOldString dbr_class_name_t;
 #define DBF_DOUBLE      6
 #define DBF_NO_ACCESS   7
 #define LAST_TYPE       DBF_DOUBLE
-#define VALID_DB_FIELD(x)       ((x >= 0) && (x <= LAST_TYPE))
-#define INVALID_DB_FIELD(x)     ((x < 0) || (x > LAST_TYPE))
+#define VALID_DB_FIELD(x)       ((unsigned)(x) <= LAST_TYPE)
+#define INVALID_DB_FIELD(x)     !VALID_DB_FIELD(x)
```
dirk-zimoch, line 69: Simplification to save one compare and avoid code duplication.

modules/ca/src/client/db_access.h
```diff
@@ -112,8 +113,8 @@ typedef epicsOldString dbr_class_name_t;
 #define DBR_STSACK_STRING DBR_PUT_ACKS + 1
 #define DBR_CLASS_NAME DBR_STSACK_STRING + 1
 #define LAST_BUFFER_TYPE        DBR_CLASS_NAME
-#define VALID_DB_REQ(x) ((x >= 0) && (x <= LAST_BUFFER_TYPE))
-#define INVALID_DB_REQ(x)       ((x < 0) || (x > LAST_BUFFER_TYPE))
+#define VALID_DB_REQ(x)     ((unsigned)(x) <= LAST_BUFFER_TYPE)
+#define INVALID_DB_REQ(x)   !VALID_DB_REQ(x)
```
dirk-zimoch, line 117: Simplification to save one compare and avoid code duplication.

modules/ca/src/client/db_access.h
```diff
@@ -668,8 +674,8 @@ union db_access_val{
 #define db_state_dim            MAX_ENUM_STATES
 #define db_state_text_dim       MAX_ENUM_STRING_SIZE
 
-#define dbf_type_is_valid(type)   ((type) >= 0 && (type) <= LAST_TYPE)
-#define dbr_type_is_valid(type)   ((type) >= 0 && (type) <= LAST_BUFFER_TYPE)
+#define dbf_type_is_valid(type)   ((unsigned)(type) <= LAST_TYPE)
+#define dbr_type_is_valid(type)   VALID_DB_REQ(type)
```
dirk-zimoch, line 678: Avoid code duplication.

modules/ca/src/client/db_access.h
```diff
@@ -681,26 +687,19 @@ union db_access_val{
...
-                 (type)%(LAST_TYPE+1) == DBR_CHAR)
+                (dbr_type_is_valid(type) && (type)%(LAST_TYPE+1) == DBR_CHAR)
 #define dbr_type_is_LONG(type)   \
-                ((type) >= 0 && (type) <= LAST_BUFFER_TYPE && \
-                 (type)%(LAST_TYPE+1) == DBR_LONG)
+                (dbr_type_is_valid(type) && (type)%(LAST_TYPE+1) == DBR_LONG)
 #define dbr_type_is_DOUBLE(type)   \
-                ((type) >= 0 && (type) <= LAST_BUFFER_TYPE && \
-                 (type)%(LAST_TYPE+1) == DBR_DOUBLE)
+                (dbr_type_is_valid(type) && (type)%(LAST_TYPE+1) == DBR_DOUBLE)
```
dirk-zimoch, line 702: Avoid code duplication.

modules/ca/src/client/db_access.h
```diff
@@ -553,6 +537,28 @@ LIBCA_API extern const unsigned short dbr_size[];
...
+ * \param[in] count The element count.
+ * \returns The size in bytes of the specified type
+ * with the specified number of elements.
+ */
+
+static EPICS_ALWAYS_INLINE size_t dbr_size_n(unsigned dbr_type, long count)
+{
+    assert(VALID_DB_REQ(dbr_type));
+    assert(count < 0 || (((size_t)~0UL)-dbr_size[dbr_type])/dbr_value_size[dbr_type] > (unsigned long)count);
+    return count < 0 ? dbr_size[dbr_type] : dbr_size[dbr_type]+((size_t)(count)-1)*dbr_value_size[dbr_type];
```
dirk-zimoch, line 559: Assert that the calculation does not overflow `size_t`.

modules/ca/src/client/db_access.h
```diff
@@ -36,20 +37,20 @@ extern "C" {
...
+typedef epicsInt32      dbr_long_t;
+typedef epicsUInt32     dbr_ulong_t;
+typedef epicsFloat32    dbr_float_t;
+typedef epicsFloat64    dbr_double_t;
 typedef epicsUInt16     dbr_put_ackt_t;
 typedef epicsUInt16     dbr_put_acks_t;
-typedef epicsOldString dbr_stsack_string_t;
-typedef epicsOldString dbr_class_name_t;
+typedef epicsOldString  dbr_stsack_string_t;
+typedef epicsOldString  dbr_class_name_t;
```
dirk-zimoch, line 53: Alignment

modules/ca/src/client/db_access.h
```diff
@@ -107,13 +108,13 @@ typedef epicsOldString dbr_class_name_t;
...
-#define DBR_STSACK_STRING DBR_PUT_ACKS + 1
-#define DBR_CLASS_NAME DBR_STSACK_STRING + 1
-#define LAST_BUFFER_TYPE        DBR_CLASS_NAME
-#define VALID_DB_REQ(x) ((x >= 0) && (x <= LAST_BUFFER_TYPE))
-#define INVALID_DB_REQ(x)       ((x < 0) || (x > LAST_BUFFER_TYPE))
+#define DBR_PUT_ACKT        DBR_CTRL_DOUBLE + 1
+#define DBR_PUT_ACKS        DBR_PUT_ACKT + 1
+#define DBR_STSACK_STRING   DBR_PUT_ACKS + 1
+#define DBR_CLASS_NAME      DBR_STSACK_STRING + 1
+#define LAST_BUFFER_TYPE    DBR_CLASS_NAME
```
dirk-zimoch, line 115: Alignment

anjohnson:
@dirk-zimoch Did you see [this gist](https://gist.github.com/anjohnson/afe91e41b99f40906b32c63903ca44ab), which I linked to in a discussion on #907? I haven't compared the implementations at all and your version is likely to be better, just wanted you to see that.

dirk-zimoch:
No, I had not seen that. I briefly considered returning 0 in the error case, but considered that unsafe because often the result goes straight into `malloc` unchecked.

modules/ca/src/client/db_access.h
anjohnson: I'm not happy with adding assertions here.

The existing code doesn't expect them. `dbr_size_n()` is used 8 times in the modules/database/src/rsrv/camessage.c file, so a client might be able to cause the IOC to abort (or a server task to be suspended) just by passing in values that trigger the assertions. Some places there might check those arg's before calling this, but I'd like to minimize how many times we do check them.

The version that Claude wrote explicitly returns 0 for a bad data type or a negative element count. That version is much easier to understand than yours, which looks like line-noise in places.
```C
static EPICS_ALWAYS_INLINE
epicsUInt64 dbr_size_n(unsigned type, epicsInt64 count)
{
    if (type > (unsigned) LAST_BUFFER_TYPE)
        return 0;
    if (count < 0)
        return dbr_size[type];
    return (epicsUInt64) dbr_size[type]
        + (epicsUInt64) (count - 1) * dbr_value_size[type];
}
```
Claude also claimed that its version was a direct drop-in replacement with no other code changes necessary, although code using it could be simplified given the checks it does.

anjohnson (changes requested):
There are a lot of code changes here, many more than I would want to include in a bug-fix patch. I haven't reviewed most of them, but I do want to raise one objection.

modules/ca/src/client/db_access.h
dirk-zimoch: I was not sure how to handle this.
Before, passing an invalid value for `TYPE` simply accessed the `dbr_size[]` and `dbr_value_size[]` arrays out of range, which might or might not crash the IOC. If it did, it was hard to see why, if not it would return some nonsense value which might crash the IOC later or lead to other funny effects.
The same for overflowing the result.
I took the approach of "fail early and loud". The caller should **never** pass invalid arguments. Before, it would crash or do nonsense, now it would at least print a human readable reason for crashing.

The problem with returning a "magic value" like 0 is that it only works if the caller checks the results. No caller I found ever did that. But then, they could as well have checked the arguments before calling `dbr_size_n`. In the end, the responsibility would lie with the caller but nothing checks that the caller actually does what is required.

In case the caller checks all arguments beforehand, and the compiler can reason that the asserts will never fail, I expect the optimizer to remove the asserts. (I will check that theory.) In that case the asserts would come with no costs.

It is not correct that Claude's approach does not require code changes on the caller side. Compilers that check for possible data loss (e.g. Windows with our standard settings) will warn when assigning the `epicsUInt64` result to a 32 bit `unsigned (int)`. It would just move the overflow from the macro to the assignment.

dirk-zimoch: Confirmed: Changing for example dbContext.cpp to properly check the arguments **before** calling dbr_size_n() completely eliminates the assert checks from the compiled code.

mdavidsaver: I also don't like the assertions, nor changing the function contract by returning zero.  We can't code away a poor API design.

How about a two part approach?

1. Replace the `dbr_size_n()` macro with an out-of-line function, marked as deprecated, which asserts when given invalid inputs.
2. Add a new "safe" function designed to communicate an error when given invalid inputs.
3. Change internal call sites to use the "safe" version.

PR #924, opened 2026-07-16
seqRecord: fix upper display limit of the DLYn fields

physwkim:
`get_graphic_double()` assigns `lower_disp_limit` twice and never writes `upper_disp_limit`:

```c
case 0: /* DLYn */
    pgd->lower_disp_limit = 0.0;
    pgd->lower_disp_limit = 10.0;
    return 0;
```

so a `DBR_GR_DOUBLE` read of any DLYn field reports the inverted range lower=10, upper=0. The second assignment was evidently meant for the upper limit; the 0 in the upper slot is the zeroed buffer `dbAccess.c` hands to the RSET, not a chosen value.

Both assignments date from 27e80be2e5 which introduced the DLYn limits, so every release since 3.14.11 reports the inverted range over both CA (`DBR_GR`/`DBR_CTRL`) and PVA (`display.limitLow`/`limitHigh`).

Found by differential testing of all record.FIELD channels against another implementation.

The PR also extends `seqTest` with a `DBR_GR_DOUBLE` read of `DLY0`; the new cases fail with lower=10, upper=0 before the fix and pass after (68/68).

### DIFF
--- a/modules/database/src/std/rec/seqRecord.c
+++ b/modules/database/src/std/rec/seqRecord.c
@@ -327,7 +327,7 @@ static long get_graphic_double(DBADDR *paddr, struct dbr_grDouble *pgd)
         switch (fieldOffset & 3) {
         case 0: /* DLYn */
             pgd->lower_disp_limit = 0.0;
-            pgd->lower_disp_limit = 10.0;
+            pgd->upper_disp_limit = 10.0;
             return 0;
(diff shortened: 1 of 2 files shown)


dirk-zimoch (approved):
Thanks for finding this. I had obviously made a mistake when introducing display limits. Your change is what I had intended to do.

PR #925, opened 2026-07-16
calcPerform: fix the undefined integer narrowing in MODULO and NINT

physwkim:
Two related defects in `calcPerform()`.

**MODULO by −1 can crash the whole IOC.** `%` is a signed integer remainder on values narrowed from `double`. The zero divisor is guarded; `INT_MIN % -1` is not — undefined in C, and on x86 the `idiv` behind `%` traps the quotient overflow as `#DE`/SIGFPE. It is reachable from ordinary data, because the dividend is narrowed by a raw `(epicsInt32)` cast that maps any out-of-range double or NaN onto `0x80000000` = `INT_MIN`. Commit 1 guards the whole `divisor == -1` column: every integer is an exact multiple of −1, so the remainder is 0 for any dividend, keeping `%` the true mathematical remainder and leaving NaN to mean only the zero divisor.

**NINT and MODULO skip the `d2i` guard their siblings use.** Every bitwise/shift operator narrows through the `d2i()`/`d2ui()` macros written precisely to make out-of-range double→int conversions well-defined (the macro comment says so); NINT and MODULO were left on the raw cast. Commit 2 hoists `d2i`/`d2ui` to file scope and routes both MODULO operands and the NINT result through `d2i`.

Commit 2 **changes results for out-of-range operands**: `NINT(3e9)` → `-1294967296` (matching what `3e9 & 0xffffffff` already returns), `3e9 % 7` → `0`. In-range arithmetic is unchanged. This brings NINT/MODULO to parity with the sibling operators; making `d2i` itself total (it currently defines only the `[2^31, 2^32)` window) would change those siblings too and is left as a separate decision.

`epicsCalcTest` grows 687 → 698; the new MODULO cases take the test binary down with SIGFPE before commit 1 and pass after. Full `libcom` suite (4549 tests) passes.

The synApps `calc` sCalc/aCalc engines carry the same MODULO SIGFPE; a separate fix is submitted there (epics-modules/calc#38).

### DIFF
--- a/modules/libcom/src/calc/calcPerform.c
+++ b/modules/libcom/src/calc/calcPerform.c
@@ -161,10 +161,17 @@ LIBCOM_API long
 
         case MODULO:
             itop = (epicsInt32) *ptop--;
-            if (itop)
-                *ptop = (epicsInt32) *ptop % itop;
-            else
+            if (itop == 0)
                 *ptop = epicsNAN;
+            else if (itop == -1)
+                /* Every integer is an exact multiple of -1, so the remainder
+                 * is always zero.  C's % operator is undefined for this
+                 * divisor when the dividend is the most negative integer,
+                 * since the quotient it is defined in terms of overflows.
+                 */
+                *ptop = 0;
+            else
+                *ptop = (epicsInt32) *ptop % itop;
             break;
 
         case POWER:
--- a/modules/libcom/src/calc/calcPerform.c
+++ b/modules/libcom/src/calc/calcPerform.c
@@ -34,6 +34,20 @@ static int cond_search(const char **ppinst, int match);
 #define PI 3.14159265358979323
 #endif
 
+/* Be VERY careful converting double to int in case bit 31 is set!
+ * Out-of-range errors give very different results on different systems.
+ * Convert negative doubles to signed and positive doubles to unsigned
+ * first to avoid overflows if bit 32 is set.
+ * The result is always signed, values with bit 31 set are negative
+ * to avoid problems when writing the value to signed integer fields
+ * like longout.VAL or ao.RVAL. However unsigned fields may give
+ * problems on some architectures. (Fewer than giving problems with
+ * signed integer. Maybe the conversion functions should handle
+ * overflows better.)
+ */
+#define d2i(x) ((x)<0?(epicsInt32)(x):(epicsInt32)(epicsUInt32)(x))
+#define d2ui(x) ((x)<0?(epicsUInt32)(epicsInt32)(x):(epicsUInt32)(x))
+
 /* Turn off global optimization for 64-bit MSVC builds */
 #if defined(_WIN32) && defined(_M_X64) && !defined(_MINGW)
 #  pragma optimize("g", off)
@@ -160,7 +174,8 @@ LIBCOM_API long
             break;
 
         case MODULO:
-            itop = (epicsInt32) *ptop--;
+            top = *ptop--;
+            itop = d2i(top);
             if (itop == 0)
                 *ptop = epicsNAN;
             else if (itop == -1)
@@ -171,7 +186,7 @@ LIBCOM_API long
                  */
                 *ptop = 0;
             else
-                *ptop = (epicsInt32) *ptop % itop;
+                *ptop = d2i(*ptop) % itop;
             break;
 
         case POWER:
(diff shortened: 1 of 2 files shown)


simon-ess (commented):
Perhaps this is architecture-dependent, but these tests do not crash on my system before the fix. This is on the latest EPICS base on darwin-aarch64, for what it is worth.

anjohnson:
@physwkim Thanks for including the extra test cases here. What EPICS target architecture were you seeing failures on?

I wonder if the new tests fail on a 32-bit ARM CPU, I don't think our CI builds have one of those.

@dirk-zimoch Could you take a look at this PR please? I have a vague idea that you might have an appropriate target. I can see this is fixing some other places where `calcPerform()` does value conversions, so it looks good to me, but I'd like to be able to replicate the problems before we approve.

physwkim:
@anjohnson @simon-ess I saw the failures on `linux-x86_64`, and just re-ran there to be precise about it.

The MODULO crash is specifically an **x86 `idiv` trap**, not a portable failure. `INT_MIN % -1` is undefined in C because the *quotient* `INT_MIN / -1 == +2^31` overflows `INT_MAX`, and x86's `idiv` raises `#DE` for that quotient overflow → `SIGFPE`. Re-measured just now:

- **Without the `-1` guard**, `epicsCalcTest` runs up through `ok 272 - -2147483647 % -1` (dividend `INT_MAX`, quotient in range, fine) and then dies with `Floating point exception (core dumped)`, exit 136 / signal 8, exactly at the next case `-2147483648 % -1` (dividend `INT_MIN`).
- **With the fix**, 698/698 pass, exit 0.

So to your ARM question: I'd expect these **not** to fail on 32-bit ARM, for two separate reasons:

1. **The crash won't reproduce.** ARM integer division doesn't trap on overflow — `SDIV` defines `INT_MIN / -1 == INT_MIN` with no exception, so `INT_MIN % -1` computes `0` (which is the right answer) without a `SIGFPE`. The `% -1` tests would pass on ARM even *without* commit 1. The crash is what makes this an x86-only IOC-killer.

2. **The narrowing tests are architecture-independent after the fix.** `d2i(x)` converts a positive out-of-range double in two well-defined steps — `double → epicsUInt32` (in range, defined) then `epicsUInt32 → epicsInt32` (two's-complement wrap, the same on every real target) — so `NINT(3e9) == -1294967296` and `3e9 % 7 == 0` hold on ARM as well as x86. Note this is the fixed path; the *old* raw `(epicsInt32)` cast is exactly where x86 and ARM diverge (x86 yields `0x80000000`, ARM saturates), which is the portability wart the PR removes.

Caveat: I don't have a 32-bit ARM target here, so the ARM half is reasoned from the hardware division semantics and `d2i`'s definition, not measured — @simon-ess / @dirk-zimoch, if one of you can run `epicsCalcTest` on real ARM (armv7 in particular) that would confirm both halves: the pre-fix binary should *not* SIGFPE there, and the post-fix binary should pass 698/698.

dirk-zimoch:
> Caveat: I don't have a 32-bit ARM target here, so the ARM half is reasoned from the hardware division semantics and `d2i`'s definition, not measured — @simon-ess / @dirk-zimoch, if one of you can run `epicsCalcTest` on real ARM (armv7 in particular) that would confirm both halves: the pre-fix binary should _not_ SIGFPE there, and the post-fix binary should pass 698/698.

I am on vacation since today for 3 weeks and have no access to 32 bit ARM (or PPC) systems during this time.

However, I have tested `A%-1` with different values for `A` and can confirm the crashes on linux-x86_64 for big values of A (positive and negative). Crashing one architecture is justification enough for this fix, regardless of other architectures being more forgiving or not.

simon-ess:
> However, I have tested `A%-1` with different values for `A` and can confirm the crashes on linux-x86_64 for big values of A (positive and negative). Crashing one architecture is justification enough for this fix, regardless of other architectures being more forgiving or not.

That's quite sensible. I feel it's at least worth ensuring that someone checks this given the sudden increase in LLM-generated PRs with at least a few instances of hallucinated issues.

dirk-zimoch:
Modulo -1 has of course no real use case, that’s why nobody noticed. But neither has modulo 0 and we catch that. It is an attack surface to crash the IOC if anyone can manipulate the modulo denominator or even the CALC field itself.

PR #926, opened 2026-07-16
selRecord: fix the precision reported for the A-L and LA-LL fields

physwkim:
`get_precision()` compares `paddr->pfield` against `&pvalue`/`&plvalue` — the addresses of two local pointer variables — so the test has never matched, and every A-L / LA-LL precision request has fallen through to `recGblGetPrec()` since the loop was added in 1992 (`e87551df24`). Those fields report `recGblGetPrec`'s clamped 15 for a record PREC of 17, while VAL reports 17. (A `(void *)` cast on the comparison hid the type error from the compiler until it was removed in 2025; the logic was already dead.)

Fix: compare against `pvalue`/`plvalue` — the field pointers themselves. New `selTest` covers VAL, A, L, LA, LL and HIHI; the A-L/LA-LL cases fail before the change (report 15 for PREC 17) and pass after.

### DIFF
--- a/modules/database/src/std/rec/selRecord.c
+++ b/modules/database/src/std/rec/selRecord.c
@@ -156,8 +156,8 @@ static long get_precision(const DBADDR *paddr, long *precision)
     pvalue = &prec->a;
     plvalue = &prec->la;
     for(i=0; i<SEL_MAX; i++, pvalue++, plvalue++) {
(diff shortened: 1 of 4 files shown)


dirk-zimoch (approved):
Thanks for spotting this mistake.
Approved.

PR #927, opened 2026-07-16
Disable MSVC warning C4250 "inherit via dominance"

dirk-zimoch:
This warning is utterly pointless and an annoyance when compiling pvData or pvAccess because it is reported by every compilation unit that includes for example pvData.h.

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4250

### DIFF
--- a/configure/os/CONFIG.win32-x86.win32-x86
+++ b/configure/os/CONFIG.win32-x86.win32-x86
@@ -104,7 +104,9 @@ CODE_CPPFLAGS += -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE
 # -w44355 "'this' used in the base initializer list"
 # -w44344 "behavior change: use of explicit template arguments results in ..."
 # -w44251 "class needs to have dll-interface to be used by clients of ..."
-WARN_CXXFLAGS_YES = -W3 -w44355 -w44344 -w44251
+# -w44250 "class1 inherits class2::member via dominance"
+
+WARN_CXXFLAGS_YES = -W3 -w44355 -w44344 -w44251 -w44250
 WARN_CXXFLAGS_NO  = -W1
 
 # Silence tr1 namespace deprecation warnings


mdavidsaver:
I have no objection to disabling this MSVC specific warning globally.

That said, the situation being warned about is a symptom of the "dreaded diamond" inheritance pattern.  Which is not seen as good c++ practice these days.  I think it is worth at least a small effort to limit scope, so that new occurrences in code unrelated to pvDataCPP triggers this warning.  Is there some way it could be limited to pvDataCPP and dependent code?  A `#pragma` in `pvData.h`?  I am not suggesting any complex push/pop, just a simple:

```c
#ifdef _MSC_VER
#  pragma warning (disable:4250)
#endif
```

dirk-zimoch:
Yes, that works and it was my first approach. I had wrapped that around the class where the diamond of doom comes together (PVString), with an additional `#pragma warning (push)`… `#pragma warning (pop)` because it is a header and I wanted to avoid the warning disabling to „spill out“.
But then I found the next instance in pvAccess and thought maybe fix it globally if it is a common pattern in pvXXX.

dirk-zimoch:
BTW: There are multiple similar warning disabling pragmas for windows in the code, most of them checking for `defined (_WIN32) && !defined (cygwin)` or similar. I find the check for `_MSC_VER` more elegant.

dirk-zimoch:
https://github.com/epics-base/pvDataCPP/pull/108

mdavidsaver:
> I wanted to avoid the warning disabling to „spill out“.

What I am suggesting is to have it "spill" only where pvData.h is included.  This will limit the spill to the pvDataCPP ecosystem.   An improvement on globally disabling, without needing a lot of extra effort on your part.

ralphlange:
> BTW: There are multiple similar warning disabling pragmas for windows in the code, most of them checking for `defined (_WIN32) && !defined (cygwin)` or similar. I find the check for `_MSC_VER` more elegant.

It's a different check.
Windows platform can also be handled using Gnu (MinGW) or - in the past - Cygwin.
The first checks for the platform OS, yours checks for the compiler being used.

ralphlange:
> What I am suggesting is to have it "spill" only where pvData.h is included. This will limit the spill to the pvDataCPP ecosystem. An improvement on globally disabling, without needing a lot of extra effort on your part.

+1 on that.
I see globally disabling as a last-resort measure.

anjohnson:
Core: This PR would be dropped in Base and only disable the warning globally in pvData.h

anjohnson:
Punt to pvData as described above.

PR #921, opened 2026-07-15
dbStaticLib: fix stack overflows in dbReportDeviceConfig

jacquelinegarrahan:
dbGetString() returns up to messagesize-1 (275) characters, but
dbReportDeviceConfig copied DTYP into dtypValue[50] and built
"cvt(EGUL,EGUF)" in cvtValue[40] with strcpy/strcat.  A long DTYP choice
or high-precision EGUL/EGUF (e.g. 1.7976931348623157e+308) overflowed
these stack buffers.  Size them from messagesize so the copies fit.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #930, opened 2026-07-17
Fix possible stack overflows in dbReportDeviceConfig

anjohnson:
Writes to both dtypValue and cvtValue strings could overflow. Instead of just expanding the string buffers, change the output. Long DTYP strings are truncated to 30 char's by this, and the cvt(<EGUL>,<EGUF>) range added for ai/ao LINR conversions is limited to avoid overflow.

This version also adds a header line (in BOLD) to the report.

Replaces #921

### DIFF
--- a/modules/database/src/ioc/dbStatic/dbStaticLib.c
+++ b/modules/database/src/ioc/dbStatic/dbStaticLib.c
@@ -22,6 +22,7 @@
 #include "dbDefs.h"
 #include "dbmf.h"
 #include "ellLib.h"
+#include "epicsMath.h"
 #include "epicsPrint.h"
 #include "epicsStdio.h"
 #include "epicsStdlib.h"
@@ -3591,11 +3592,12 @@ void  dbReportDeviceConfig(dbBase *pdbbase, FILE *report)
         status = dbFirstRecord(pdbentry);
         while (!status) {
             int ilink;
+            int titlesPrinted = 0;
 
             for (ilink=0; ilink<nlinks; ilink++) {
                 char linkValue[messagesize];
-                char dtypValue[50];
-                char cvtValue[40];
+                char dtypValue[31];
+                char cvtValue[40] = "";
                 struct link *plink;
                 int linkType;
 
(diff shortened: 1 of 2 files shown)


modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -3629,28 +3631,32 @@ void  dbReportDeviceConfig(dbBase *pdbbase, FILE *report)
                 if (status)
                     break;  /* Next record type */
 
-                strcpy(dtypValue, dbGetString(pdbentry));
-                status = dbFindField(pdbentry, "LINR");
-                if (status) {
-                    cvtValue[0] = 0;
+                {
+                    const char *dtyp = dbGetString(pdbentry);
+                    size_t srclen = strlen(dtyp);
```
lucmaga, line 3637: strnlen seams appropriate, and you can remove the following line.

modules/database/src/ioc/dbStatic/dbStaticLib.c
```diff
@@ -3629,28 +3631,32 @@ void  dbReportDeviceConfig(dbBase *pdbbase, FILE *report)
...
+                    size_t copylen = srclen > 30 ? 30 : srclen;
+
+                    memcpy(dtypValue, dtyp, copylen);
+                    dtypValue[copylen] = '\0';
+                    if (srclen > 30)                    /* value truncated */
+                        memcpy(&dtypValue[27], "...", 3);  /* copylen == 30 */
```
lucmaga: Maybe, instead we should log a warning?

anjohnson: Thanks for reviewing this PR! The code is printing a report requested by the user, and one aim of my changes was to format the report nicely. The ellipsis `...` is included when a legal but unusually long DTYP choice string that is too wide for the report column is seen; that _is_ the warning.

modules/database/src/ioc/dbStatic/dbStaticLib.c
anjohnson: Unfortunately both the `srclen` and `copylen` values are needed by the code below to insert the ellipsis `...` if the `dtyp` string is too wide for its report column.

However I asked Claude to look for a simpler approach that also moves the elision to the middle of the string instead of the end, and it suggested just one call to `epicsSnprintf()`, which I will use instead.

modules/database/src/ioc/dbStatic/dbStaticLib.c
lucmaga: IMHO warning statements are more visible and impactful than just the ellipses. But I see that this it is a pattern EPICS is already using other places.

anjohnson: Giving a device support a choice name longer than 30 characters isn't illegal, although anything too long will be truncated when reading the field choices. If we're going to add warnings about the name length they should appear when loading the DBD file which defines that device support, not when the user runs this iocsh command (which probably almost never happens anyway, it's pretty obscure).

PR #931, opened 2026-07-17
More CA client protection

anjohnson:
This is an extension of my cac-protection branch PR #907 with the added `inline dbr_size_n()` function and action method checks using it within the CA Client code.

The original PR also included several other CA client hardening changes.

### DIFF
--- a/modules/ca/src/client/cac.cpp
+++ b/modules/ca/src/client/cac.cpp
@@ -1109,7 +1109,7 @@ bool cac::exceptionRespAction ( callbackManager & cbMutexIn, tcpiiu & iiu,
 
     // execute the exception message
     pExcepProtoStubTCP pStub;
-    if ( hdr.m_cmmd >= NELEMENTS ( cac::tcpExcepJumpTableCAC ) ) {
+    if ( req.m_cmmd >= NELEMENTS ( cac::tcpExcepJumpTableCAC ) ) {
         pStub = &cac::defaultExcep;
     }
     else {
--- a/modules/ca/src/client/cac.cpp
+++ b/modules/ca/src/client/cac.cpp
@@ -896,6 +896,13 @@ bool cac::readNotifyRespAction ( callbackManager &, tcpiiu & iiu,
             // this does *not* assign a new resource id
             this->ioTable.add ( *pmiu );
         }
+        if ( INVALID_DB_REQ ( hdr.m_dataType ) ||
+             hdr.m_postsize < dbr_size_n ( hdr.m_dataType, hdr.m_count ) ) {
+            pmiu->exception ( guard, *this, ECA_BADTYPE,
+                "server response inconsistent with payload size",
+                hdr.m_dataType, hdr.m_count );
+            return true;
+        }
         if ( caStatus == ECA_NORMAL ) {
             /*
              * convert the data buffer from net
@@ -960,6 +967,13 @@ bool cac::eventRespAction ( callbackManager &, tcpiiu &iiu,
     //
     baseNMIU * pmiu = this->ioTable.lookup ( hdr.m_available );
     if ( pmiu ) {
+        if ( INVALID_DB_REQ ( hdr.m_dataType ) ||
+             hdr.m_postsize < dbr_size_n ( hdr.m_dataType, hdr.m_count ) ) {
+            pmiu->exception ( guard, *this, ECA_BADTYPE,
+                "server response inconsistent with payload size",
+                hdr.m_dataType, hdr.m_count );
+            return true;
+        }
         /*
          * convert the data buffer from net format to host format
          */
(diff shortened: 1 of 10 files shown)


modules/ca/src/client/cac.cpp
```diff
@@ -895,6 +911,9 @@ bool cac::readNotifyRespAction ( callbackManager &, tcpiiu & iiu,
             // this does *not* assign a new resource id
             this->ioTable.add ( *pmiu );
         }
+        if ( this->badResponsePayload ( guard, *pmiu, hdr ) ) {
+            return true;
+        }
```
mdavidsaver, line 917: Is `return true;` the desired behavior here?

When `readNotifyRespAction()` is executed (though `cac::executeResponse()`) the result bool is named `msgOk`.  A `false` breaks from the loop, and closes the TCP connection.  A `true` will continue to process messages.

modules/ca/src/client/cac.cpp
```diff
@@ -1086,6 +1111,9 @@ bool cac::exceptionRespAction ( callbackManager & cbMutexIn, tcpiiu & iiu,
     if ( hdr.m_postsize < bytesSoFar ) {
         return false;
     }
+    // Ensure the context string is NULL-terminated
+    static_cast < char * > ( pMsgBdy ) [hdr.m_postsize - 1] = '\0';
```
mdavidsaver, line 1116: To answer my own question from an earlier conversation.  `m_postsize` is check against the payload size (`curDataBytes`) in `tcpiiu::processIncoming` prior to calling `executeResponse()`.

It also looks like the case which resulted in an overflow of `m_postsize` in RSRV is avoided in libca.

modules/ca/src/client/db_access.h
```diff
@@ -552,6 +536,38 @@ LIBCA_API extern const unsigned short dbr_size[];
...
+ * \param[in] count The element count.
+ * \returns The size in bytes of the specified type
+ * with the specified number of elements.
+ */
+#ifndef db_accessHFORdb_accessC
+static EPICS_ALWAYS_INLINE
+epicsUInt64 dbr_size_n(unsigned type, epicsInt64 count)
+{
+    if (type > (unsigned) LAST_BUFFER_TYPE)
+        return 0;
```
mdavidsaver, line 581: Repeating discussion from https://github.com/epics-base/epics-base/pull/923#discussion_r3593461577 .  I am concerned about changing the contract of `dbr_size_n()` to begin returning zero on error.

Changing to return a wider `epicsUInt64` seems reasonable though.

modules/ca/src/client/udpiiu.cpp
```diff
@@ -811,15 +811,17 @@ bool udpiiu::exceptionRespAction (
...
-    else{
-        errlogPrintf (
-            ERL_ERROR " condition \"%s\" detected by %s at %s\n",
+    int len = msg.m_postsize - sizeof( caHdr );
+    if ( len > 0 ) {
+        const char *ctx = reinterpret_cast<const char *>( &reqMsg + 1 );
+        len = epicsStrnLen ( ctx, len );
+        errlogPrintf ( ERL_ERROR
+            " condition \"%s\" detected by %s with context \"%.*s\" at %s\n",
+            ca_message ( msg.m_available ), name, len, ctx, date );
```
mdavidsaver, line 823: I think I have convinced myself the this `len > 0` case is essentially ok.  For a UDP message, `msg.m_postsize` is 16-bits.  So no over/under flow.  I think the `msg` -> `reqMsg` -> `ctx` flow adjusts the remaining buffer size accordingly.

Putting on my paranoia hat, printing unescaped `ctx` as `char*` might be a vector for malfeasance (unprintables, ANSI escapes, ...).  Also, maybe bound `len` to 64 bytes to avoid potentially printing 65500 bytes of garbage to the console?

modules/ca/src/client/db_access.h
anjohnson: Core: Use a different name for this internal function.
Application code might be using the macro, caLib should never pass it a bad dbrType value, so the macro can stay and should never cause problems.

modules/ca/src/client/cac.cpp
anjohnson: Core: That's a choice between keeping connections open in the face of slightly flakey servers (robustness, or other (bad) server implementations), vs protection from malicious servers. Breaking TCP connections immediately makes it hard to debug issues.

modules/ca/src/client/udpiiu.cpp
anjohnson: Core: Length limit seems reasonable, not sure if it's worth escaping too.

simon-ess: As discussed in the meeting, this is passing through `errlogPrintf`  as well which would truncate the message regardless.

modules/ca/src/client/db_access.h
anjohnson: The new inline function `dbr_size_chk()` now exists alongside the original `dbr_size_n()` macro.

mdavidsaver: Should the `count` argument be `epicsUInt32` to make clear that the result can not overflow?

PR #934, opened 2026-07-19
RSRV message validation

mdavidsaver:
Extend validation of CA messages by RSRV to reduce the scope for remote maleficence.  Principally avoiding integer overflows, and cross-checking between interrelated `msgsize`, `dtype`, and `count` header fields.

This change set was developed in response to a confidential security report by Shubham Raj shubham@causalsecurity.com.  It is being released early after a subsequent unconfidential report of an overlapping set of issues.  Thank you to @anjohnson @dirk-zimoch @simon-ess for review during the embargo period.

Anyone attempting to backport these fixes to releases before 2017 would also need: e430d097b727de920aeb22af0e282678036d8a0d .


ralphlange (approved):
Looks all good and appropriate to me.

anjohnson (approved):
LGTM, merging.

PR #944, opened 2026-08-03
Fix for #943, ca_put(DBR_STRING,chid,"1")

### LINKED ISSUE #943: Regression: commit 4128a7c0 breaks CA writes ?
A possible regression that came up with Gateway testing...

Testing with a softIoc hosting a single ao record `record(ao, "ao") {}`.

**Pinning EPICS Base to 28d54063 everything seems to work fine.**
Client side:
```
langer@FR-IWL-LANGER:~/work/EPICS/7.0/base$ ./bin/linux-x86_64/caput -c ao 1.23
CA Repeater: Attached and initialized
Old : ao                             0
New : ao                             1.23
```

IOC side:
```
langer@FR-IWL-LANGER:~/work/EPICS/7.0/base$ ./bin/linux-x86_64/softIoc -d ~/test.db
Starting iocInit
## EPICS R7.0.10.1-DEV
## Rev. R7.0.10-140-g28d540638538ddc5fb08
## Rev. Date Git: 2026-07-23 18:17:02 -0500
```

**Pinning EPICS Base to 4128a7c0 has all CA writes fail.**
Client side:
```
langer@FR-IWL-LANGER:~/work/EPICS/7.0/base$ ./bin/linux-x86_64/caput -c ao 1.23
CA Repeater: Attached and initialized
Old : ao                             0
ERROR occurred writing data: Virtual circuit disconnect
    Warning: "Virtual circuit disconnect"
    Context: "172.30.227.221:5064"
    Source File: ../cac.cpp line 1241
    Current Time: Sun Aug 02 2026 16:20:32.083876993
```

IOC side:
```
langer@FR-IWL-LANGER:~/work/EPICS/7.0/base$ ./bin/linux-x86_64/softIoc -d ~/test.db
Starting iocInit
## EPICS R7.0.10.1-DEV
## Rev. R7.0.10-141-g4128a7c078a313c98608
## Rev. Date Git: 2026-07-23 18:17:02 -0500
epics> CAS: forcing disconnect from 172.30.227.221:45454
```

Note that the CLI `caput` writes everything as a string - not sure if that matters.

anjohnson:
The CA client library truncates short scalar strings, RSRV must accept the result.

Fixes #943

### DIFF
--- a/modules/database/src/ioc/rsrv/camessage.c
+++ b/modules/database/src/ioc/rsrv/camessage.c
@@ -759,9 +759,18 @@ static int write_action ( caHdrLargeArray *mp,
         mp->m_count = dbChannelFinalElements(pciu->dbch);
     }
 
-    size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
-        return RSRV_ERROR;
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1) {
+        /* Accept short scalar string payloads that are
+         * null-terminated within the message received. */
+        if (epicsStrnLen(pPayload, mp->m_postsize) >= mp->m_postsize) {
+            return RSRV_ERROR;
+        }
+    }
+    else {
+        size = dbr_size_n (mp->m_dataType, mp->m_count);
+        if (size > mp->m_postsize) {
+            return RSRV_ERROR;
+        }
     }
 
     if(!rsrvCheckPut(pciu)){
@@ -1685,7 +1694,14 @@ static int write_notify_action ( caHdrLargeArray *mp, void *pPayload,
     }
 
     size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1) {
+        /* Accept short scalar string payloads that are
+         * null-terminated within the message received. */
+        if (epicsStrnLen(pPayload, mp->m_postsize) >= mp->m_postsize) {
+            return RSRV_ERROR;
+        }
+    }
+    else if (size > mp->m_postsize) {
         return RSRV_ERROR;
     }
 
--- a/modules/database/src/ioc/rsrv/camessage.c
+++ b/modules/database/src/ioc/rsrv/camessage.c
@@ -743,7 +743,6 @@ static int write_action ( caHdrLargeArray *mp,
     int                     status;
     long                    dbStatus;
     void                    *asWritePvt;
-    unsigned                size;
 
     pciu = MPTOPCIU(mp, client);
     if(!pciu){
@@ -766,11 +765,8 @@ static int write_action ( caHdrLargeArray *mp,
             return RSRV_ERROR;
         }
     }
-    else {
-        size = dbr_size_n (mp->m_dataType, mp->m_count);
-        if (size > mp->m_postsize) {
-            return RSRV_ERROR;
-        }
+    else if (dbr_size_n (mp->m_dataType, mp->m_count) > mp->m_postsize) {
+        return RSRV_ERROR;
     }
 
     if(!rsrvCheckPut(pciu)){
--- a/modules/database/src/ioc/rsrv/camessage.c
+++ b/modules/database/src/ioc/rsrv/camessage.c
@@ -1775,13 +1775,19 @@ static int write_notify_action ( caHdrLargeArray *mp, void *pPayload,
     pciu->pPutNotify->msg = *mp;
     pciu->pPutNotify->nRequest = mp->m_count;
 
-    status = caNetConvert (
-        mp->m_dataType, pPayload, pciu->pPutNotify->pbuffer,
-        FALSE /* net -> host format */, mp->m_count );
-    if ( status != ECA_NORMAL ) {
-        log_header ("invalid data type", client, mp, pPayload, 0);
-        putNotifyErrorReply ( client, mp, status );
-        return RSRV_ERROR;
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1) {
+        /* Scalar string, copy up to the NUL terminator and 0-pad */
+        strncpy(pciu->pPutNotify->pbuffer, pPayload, size);
+    }
+    else {
+        status = caNetConvert (
+            mp->m_dataType, pPayload, pciu->pPutNotify->pbuffer,
+            FALSE /* net -> host format */, mp->m_count );
+        if ( status != ECA_NORMAL ) {
+            log_header ("invalid data type", client, mp, pPayload, 0);
+            putNotifyErrorReply ( client, mp, status );
+            return RSRV_ERROR;
+        }
     }
 
     pciu->pPutNotify->dbrType = mp->m_dataType;


ralphlange:
LGTM.

mdavidsaver:
> LGTM.

Not so hasty.  Let's not compound the regression.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -759,9 +759,18 @@ static int write_action ( caHdrLargeArray *mp,
         mp->m_count = dbChannelFinalElements(pciu->dbch);
     }
 
-    size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
-        return RSRV_ERROR;
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1) {
```
mdavidsaver, line 761: What about the `m_count==0` case?

Remember that `dbr_size_n(X,1) == dbr_size_n(X,0)`.  So as written, this extended condition will probably still reject attempts to write zero elements.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -759,9 +759,18 @@ static int write_action ( caHdrLargeArray *mp,
...
+        /* Accept short scalar string payloads that are
+         * null-terminated within the message received. */
+        if (epicsStrnLen(pPayload, mp->m_postsize) >= mp->m_postsize) {
+            return RSRV_ERROR;
+        }
+    }
+    else {
+        size = dbr_size_n (mp->m_dataType, mp->m_count);
+        if (size > mp->m_postsize) {
+            return RSRV_ERROR;
```
mdavidsaver: I added the `size` local here for consistency with `write_notify_action()`.  Having it set only in one arm of this conditional seems a recipe for future confusion.  Either maintain this consistency, or just eliminate `size`.

mdavidsaver (commented):
Allowing DBR_STRING with less than 40 bytes as long as a null is present seems the only practical solution.

I was on the fence about whether to allow exactly 40 non-null bytes.  However, I find enough places where `MAX_STRING_SIZE - 1` is already enforced.

I think this PR needs to be paired with an audit for places where `memcpy()` assumes `MAX_STRING_SIZE`.  At least up to and including `dbPut()` and friends.  eg.

```
modules/ca/src/client/convert.cpp:    memcpy ( pDest, pSrc, num*MAX_STRING_SIZE );
modules/ca/src/client/convert.cpp:    memcpy ( pDest->value, pSrc->value, (MAX_STRING_SIZE * num) );
modules/ca/src/client/convert.cpp:        memcpy(pDest->value, pSrc->value, (MAX_STRING_SIZE * num));
modules/ca/src/client/convert.cpp:    memcpy(pDest->value, pSrc->value, (MAX_STRING_SIZE * num));
```

modules/database/src/ioc/rsrv/camessage.c
anjohnson: A `ca_array_put()` with COUNT=0 doesn't fail if that's what you mean, because it doesn't trigger the check for scalar string puts in the CA client code that shortens the message. In [`comQueSend::insertRequestWithPayLoad()`](https://github.com/epics-base/epics-base/blob/bc1b965f90cf3ee14de630a49a7f83ffcb3fab3b/modules/ca/src/client/comQueSend.cpp#L331):
```C
    if ( nElem == 1 ) {
        if ( dataType == DBR_STRING  ) {
            const char * pStr = static_cast < const char * >  ( pPayload );
            size = strlen ( pStr ) + 1u;
            if ( size > MAX_STRING_SIZE ) {
                throw cacChannel::outOfBounds();
            }
            payloadSize = CA_MESSAGE_ALIGN ( size );
            this->insertRequestHeader ( request, payloadSize,
                static_cast <ca_uint16_t> ( dataType ),
                nElem, cid, requestDependent, v49Ok );
            this->pushString ( pStr, size );
        }
```

modules/database/src/ioc/rsrv/camessage.c
anjohnson: I will eliminate it here. The `size` in the `write_notify_action()` routine below  is used again later so it needs to stay.

anjohnson:
> I think this PR needs to be paired with an audit for places where `memcpy()` assumes `MAX_STRING_SIZE`. At least up to and including `dbPut()` and friends. eg.
> 
> ```
> modules/ca/src/client/convert.cpp:    memcpy ( pDest, pSrc, num*MAX_STRING_SIZE );
> modules/ca/src/client/convert.cpp:    memcpy ( pDest->value, pSrc->value, (MAX_STRING_SIZE * num));
> modules/ca/src/client/convert.cpp:    memcpy(pDest->value, pSrc->value, (MAX_STRING_SIZE * num));
> modules/ca/src/client/convert.cpp:    memcpy(pDest->value, pSrc->value, (MAX_STRING_SIZE * num));
> ```

I'm happy to have Claude do that, could you be a bit more specific about what you're asking for?

I'm starting it on this question:
> Please audit the code-base for places where calls to `memcpy()` assume the destination string buffer
  is `MAX_STRING_SIZE` bytes (or `DB_UNITS_SIZE`, `MAX_UNITS_SIZE`, `MAX_ENUM_STRING_SIZE`,
  `DB_AMSG_SIZE` etc.) without being able to check that. Are there other locations which call those APIs
  with smaller buffers that could overflow the destinations? Don't develop fixes, just tabulate the
  locations with potential problems and prepare to write a report including your results.

anjohnson:
The summary for that question was:

The pattern is **systemic but sound in-tree.** A whole family of core APIs write
a fixed, compile-time size into a caller-supplied pointer and cannot verify the
destination — the size is a *convention*, not a checked argument (the RSET
callback signatures don't even carry a length). However, **every in-tree caller
of these APIs supplies a buffer at least as large as the convention**, so none of
the `memcpy(…, MAX_STRING_SIZE)` / units / enum-string paths actually overflow.

The full report is in [this gist](https://gist.github.com/anjohnson/8b7441b4536523324e9decf39066d413).

ralphlange:
> > LGTM.
> 
> Not so hasty. Let's not compound the regression.

Yes, you're right.
Basically all 7.0 branch tests of my downstream modules are currently failing, which is a pain to ignore or work around. Explanation, not justification of haste.

ralphlange:
I do remember discussions where - as far as I recall - we found that the fixed-size string does not require or have a \0 termination if all 40 characters are used.
Stupid but true.

modules/database/src/ioc/rsrv/camessage.c
anjohnson: Core: Needs more testing of the COUNT=0 case. However:

`dbr_size_n(type, 0)` for a `dbr_string_t` scalar type returns 0, which is correct, there are no other fields in that type, and the check below is using `>` so `0 > 0` is false and we don't trigger the error. RSRV doesn't accept containers in put operations anyway and this is in `write_action()`, so this logic is correct for scalar arrays.

mdavidsaver: > Remember that `dbr_size_n(X,1) == dbr_size_n(X,0)`.

Well, I'm wrong.  I was mis-reading the `(COUNT)<0?` as `<=`.  Instead the `(COUNT)-1` is casting the count to signed 32.  So a count of zero results in effectively `dbr_size[TYPE] - dbr_value_size[TYPE]`.  For the plain DBR types this will be zero.  In RSRV, a count overflow is now bounds checked, so no problem there.

Next question, what about attempts to PUT with meta-data?  eg. `DBR_TIME_DOUBLE`.  `INVALID_DB_REQ()` allows this.  Should PUT be restricted to `dataType <= DBR_DOUBLE`?

ralphlange: I haven't found where yet, but I'm pretty sure that writing the extended DBR_ types is blocked somewhere.

ralphlange: Metadata is not rejected. It's silently dropped in `dbChannel_put()` at
https://github.com/epics-base/epics-base/blob/6b866ddc67135ba47c3d3bf26b9af8891a98ca6d/modules/database/src/ioc/db/db_access.c#L845-L847 and friends.

ralphlange: So, at our place, `dbr_size_n(X,0)` would return the size of the metadata in the X structure (without value).

What is the payload size of a metadata DBR_* when count is 0? Should be the same size as for count == 1, shouldn't it?! (As we don't ship partial containers.)
If that is true,  `dbr_size_n(X,0)` is always smaller than the payload and the operation will not be rejected.

anjohnson: > Should be the same size as for count == 1, shouldn't it?

No, whether you're reading or writing a DBR type with count=0 is transferred with no value data.

For reading, including monitors: Since protocol 4.13 the client can _request_ a count of 0 to mean "all available elements", although older clients don't. RSRV in `read_reply()` sends 0-count arrays with no value data, using `payload_size = dbr_size(TYPE,0)`. The CA client-side converters also handle num=0 correctly, and pass the completion callback a zero count.

For writing: As you found, metadata is delivered but ignored in `dbChannel_put()`. I'm pretty sure the catools can't do that though. A zero-length array gets delivered as normal with count=0, but when it gets to `dbPut()` that sets a `LINK_ALARM/INVALID_ALARM` on the record and triggers processing normally (if the field is marked `PP(TRUE)` and DISP isn't set). I just confirmed that with my caExample program on IOCs with and without this PR applied.

anjohnson:
I just pasted Claude's plan for Null-terminating all 40-char string writes to [this gist](https://gist.github.com/anjohnson/9411d9bf2539997815cf1e974fc12a18) if anyone wants to look at it. I would like to separate out any action about that from this PR though; are we agreed yet that this correctly fixes the #943 regression?

modules/database/src/ioc/rsrv/camessage.c
ralphlange: > > Should be the same size as for count == 1, shouldn't it?
> 
> No, whether you're reading or writing a DBR type with count=0 is transferred with no value data.

The value is part of the DBR structure, so even with no value the size of the structure is the same. (Probably has NULL as value.)
Are we really doing two sizes of the same DBR structure on the wire for count==0 vs. count==1? I wouldn't think so.

mdavidsaver: > So, at our place, `dbr_size_n(X,0)` would return the size of the metadata in the X structure (without value).

Yes.  I am working on a unit-test to demonstrate this.

```c
    testEq(dbr_size_n(DBR_INT, 2), 4);
    testEq(dbr_size_n(DBR_INT, 1), 2);
    testEq(dbr_size_n(DBR_INT, 0), 0);
    testEq(dbr_size_n(DBR_STS_INT, 2), 4+2*2);
    testEq(dbr_size_n(DBR_STS_INT, 1), 4+2);
    testEq(dbr_size_n(DBR_STS_INT, 0), 4);
```

> What is the payload size of a metadata DBR_* when count is 0?

The size of only the meta-data.

> Should be the same size as for count == 1, shouldn't it?! (As we don't ship partial containers.)

No.  Since PUT with meta-data is not actually honored, this partial-ness is not so relevant.

ralphlange: Ok, then.
For count == 0, both `dbr_size_n(X,0)` and payload have the same size, the write is not rejected and silently dropped.

mdavidsaver:
> Allowing DBR_STRING with less than 40 bytes as long as a null is present seems the only practical solution.

Due to https://github.com/epics-base/epics-base/issues/943#issuecomment-5206963960 I'm having second thoughts.  Since only scalar string is involved, I'm going to explore an alternative of adding a 40 byte "scratch" buffer to `struct client` of use in this case.

Issue #863, opened 2026-04-27
Proposal: make DNS/interface-derived CA and access-security state refreshable at runtime


derekste:
## Summary

I'd like to get architectural feedback on a broader direction I've started pursuing in EPICS Base: treating DNS- and interface-derived client/server state as **runtime state** instead of one-time startup state.

The motivating problem is that long-lived clients and servers can hold stale behavior after DNS changes or interface changes, even though the underlying system state has changed. In practice this shows up as stale access-security decisions, stale CA search destinations, and stale beacon/search routing until reconnect, reload, or restart.

I already have one in-flight PR in this area:

- [epics-base/epics-base#862](https://github.com/epics-base/epics-base/pull/862)

That PR currently covers:

1. **Access Security HAG hostname refresh by DNS TTL/retry**
   - HAG hostnames are refreshed lazily instead of being resolved only once at ACF load time.
   - Existing `ASCLIENTPVT` clients are recomputed when due HAG DNS refresh changes access state, so cached access rights do not remain stale.

2. **CA client hostname destination refresh**
   - `EPICS_CA_ADDR_LIST` and `EPICS_CA_NAME_SERVERS` hostname entries are kept stateful for the lifetime of the CA client context.
   - Hostname-backed search/name-server destinations refresh lazily instead of being resolved once at startup and then fixed forever.

This issue is meant to ask whether the broader direction makes sense before I keep sending follow-up PRs in this area.

## Overall Goal

The overall goal is to:

- preserve existing resolver/address-selection behavior where possible
- avoid adding background worker threads unless they are clearly needed
- refresh runtime-derived state lazily through existing timers/check paths
- make long-lived clients/servers converge on new DNS/interface state without requiring restart
- add enough diagnostics that this behavior is inspectable and operable

### 1. Refresh CA client auto-broadcast destinations when interfaces change

Current CA client search destinations include auto-discovered broadcast/interface state that is still effectively startup-fixed. A laptop/VPN/container interface change can leave the client using stale broadcast destinations.

### 2. Make CA server beacon destination lists dynamic

Server beacon destination/interface state is still built once and assumed fixed.

### 3. Improve diagnostics for dynamic/stateful address behavior

Once runtime refresh exists, it should be inspectable:

- source hostname/token -> current resolved address
- next expiry/retry time
- last resolve failure
- static vs refreshable state

### 4. Possibly centralize DNS-with-TTL policy inside EPICS Base

After the HAG and CA work, there is now repeated logic around:

[...]

PR #945, opened 2026-08-03
as: add explicit hostname HAG refresh

derekste:
## Summary

Add an explicit Access Security API for refreshing hostname-based Host Access
Groups (HAGs) without reloading the full ACF:

```c
LIBCOM_API long epicsStdCall asRefreshHag(unsigned *changed);
```

Applications call this periodically from a control-plane thread.  Base limits
successful lookups to once every 300 seconds and failed lookups to once every
60 seconds.  Ordinary `asCheckGet()` and `asCheckPut()` calls remain simple
cached-rights checks and never perform DNS work.

## Why this belongs in Base

With `asCheckClientIP=1`, Access Security resolves HAG hostnames when an ACF is
loaded, but the resolved address is otherwise fixed for the lifetime of that
policy.  A long-running server therefore keeps stale host membership after DNS
changes unless it reloads the entire ACF or restarts.

Applications such as `redis-pvxs-ioc` can decide when a refresh should run,
but the existing public API does not expose enough state to carry it out
safely.  `asLib` owns the original HAG hostname, its lookup hash, the Access
Security lock, active `ASCLIENTPVT` objects, cached rights, and
change-of-access callbacks.  Keeping the state transition in `asLib` lets a
server request a refresh without depending on private layouts or treating a
full ACF reload as a DNS refresh operation.

## Behavior

- Keep numeric HAG entries static and preserve existing IPv4 matching.
- Retain hostname source text in private sidecar state without changing the
  installed `ASBASE` or `HAGNAME` layouts.
- Resolve due names outside the Access Security lock, then apply the complete
  result only if the same policy generation is still active.
- Serialize refresh callers and discard stale results when an ACF reload wins
  during DNS resolution.
- Rebuild the HAG lookup state when an entry changes, but recompute clients only
  when the normalized effective HAG/address set changes.
- Invoke existing change-of-access callbacks for affected connected clients.
- Treat resolution failure as fail-closed by removing the stale mapping and
  retrying after 60 seconds.
- Report `*changed = 1` only for an effective mapping change.  DNS failures are
  policy outcomes and are not returned as API failures.

The implementation continues to use `aToIPAddr()`.  It deliberately does not
inspect DNS record TTLs or add a `libresolv` dependency to `libCom`.

## Scope

This is intended as a focused successor to #862 and is limited to Access
Security HAG refresh.  The earlier PR also included CA-client search
destination refresh.  CA-client DNS behavior is left for a separate design
that can address asynchronous resolution and event-loop requirements on its
own.

## Testing

- `aslibtest`: 105/105 on macOS and Linux, shared and
  `SHARED_LIBRARIES=NO`
- Full Linux `modules/libcom/test` suite: 52 test programs and 4605 tests
- Full shared Base builds on macOS and Linux; `SHARED_LIBRARIES=NO` Base build
  on macOS and static `libCom` build on Linux
- `redis-pvxs-ioc`: all 6 consumer test executables passed after rebuilding
  Base and PVXS against this candidate on Linux
- `git diff --check`
- Confirmed with `ldd` that `libCom` has no resolver-library dependency

The deterministic tests cover refresh/retry timing, unchanged and duplicate
effective mappings, DNS loss and restoration, persistent-client callbacks,
numeric and hostname-string modes, concurrent refresh callers, and an ACF
reload completing while DNS resolution is pending.

### DIFF
--- a/documentation/new-notes/PR-945.md
+++ b/documentation/new-notes/PR-945.md
@@ -1,0 +1,17 @@
+### Refresh hostname-based Host Access Groups without reloading the ACF
+
+Access Security now provides `asRefreshHag()` for applications that need
+hostname-based Host Access Groups (HAGs) to follow DNS changes at runtime.
+Applications call the function periodically from a control-plane thread.
+Successful resolutions are refreshed after 300 seconds, and failed resolutions
+are retried after 60 seconds.
+
+When the effective HAG address set changes, Access Security recomputes existing
+clients and invokes their registered change-of-access callbacks.  Resolution
+failures remove stale addresses, so affected rules fail closed until a later
+retry succeeds.  Numeric HAG entries remain static.
+
+DNS resolution is performed only by `asRefreshHag()` and during ACF loading;
+ordinary access checks do not perform DNS work.  The refresh implementation
+uses the existing portable IPv4 resolver and adds no resolver-library
+dependency to `libCom`.
--- a/modules/libcom/src/as/asLib.h
+++ b/modules/libcom/src/as/asLib.h
@@ -25,7 +25,7 @@ extern "C" {
 struct dbChannel;
 
 /* 0 - Use (unverified) client provided host name string.
- * 1 - Use actual client IP address.  HAG() are resolved to IPs at ACF load time.
+ * 1 - Use actual client IP address.  HAG() host names are resolved to IPs.
  */
 LIBCOM_API extern int asCheckClientIP;
 
@@ -103,6 +103,21 @@ LIBCOM_API long epicsStdCall asComputeAllAsg(void);
 LIBCOM_API long epicsStdCall asComputeAsg(ASG *pasg);
 */
 LIBCOM_API long epicsStdCall asCompute(ASCLIENTPVT asClientPvt);
+/**
+ * Refresh due hostname HAG entries and update existing clients.
+ *
+ * Call this periodically from a control-plane thread.  Successful hostname
+ * resolutions are refreshed after 300 seconds and failures after 60 seconds.
+ * DNS is never performed by asCheckGet() or asCheckPut().
+ *
+ * If changed is not NULL, it is set to one only when the effective HAG to
+ * IPv4 mapping changes.  Existing change-of-access callbacks are run for
+ * clients whose rights change.  DNS failures remove stale mappings and are
+ * retried later; they are not reported as API errors.
+ *
+ * Returns S_asLib_asNotActive when Access Security is not active.
+ */
+LIBCOM_API long epicsStdCall asRefreshHag(unsigned *changed);
 LIBCOM_API int epicsStdCall asDump(
     void (*memcallback)(ASMEMBERPVT,FILE *),
     void (*clientcallback)(ASCLIENTPVT,FILE *),int verbose);
--- a/modules/libcom/src/as/asLibPvt.h
+++ b/modules/libcom/src/as/asLibPvt.h
@@ -1,0 +1,27 @@
+/* asLibPvt.h - Private Access Security test interfaces */
+
+#ifndef INC_asLibPvt_H
+#define INC_asLibPvt_H
+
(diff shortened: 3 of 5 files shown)


derekste:
@mdavidsaver : you raised the libCom static-linking concern on #862. This successor removes the resolver dependency and narrows the change to an explicit Access Security API.  
Could you give a design-level go/no-go on whether refresh belongs in asLib and whether asRefreshHag() is an acceptable public API? I’d prefer to settle that before asking anyone to spend time on a line-by-line review. The GitHub Actions runs also appear to be awaiting maintainer approval.  
I realize this is a feature/API change and the contribution guide prefers prior discussion; I’m happy to move the design discussion to an issue or Matrix if that would be better.

mdavidsaver:
> whether asRefreshHag() is an acceptable public API?

imo. It is not.

> Applications call this periodically from a control-plane thread ...

This reads a bit like an "excercise for the reader".  In EPICS base, the process database is the 99+% use case.  Any serious proposal will need to fully integrate with PDB "application".

Also, AI code generators have a bad habit of emitting a lot more code than is necessary, with the result being a bit "bolted on".  Please review the structure of the existing code and try to make this change a better fit, and smaller.  eg. `asHagParseNumeric()`  Please use existing helper functions to parse IP addresses.

derekste:
Thanks for the prompt feedback, @mdavidsaver. I agree that requiring each application to arrange periodic calls leaves the normal PDB/IOC integration incomplete—effectively an exercise for every downstream application.

I do think the underlying issue is worth addressing: hostname-based HAG membership can otherwise remain stale for the lifetime of a long-running IOC after DNS changes. From a security standpoint, I'd like to find a Base-native way to prevent that rather than requiring a full ACF reload or restart.

My understanding of your suggested direction is:

- remove the public `asRefreshHag()` API and make refresh automatic for the standard PDB/IOC case;
- tie the refresh lifecycle to the existing Access Security/PDB initialization and shutdown;
- simplify the implementation and reuse the existing address parsing and formatting helpers, eliminating code such as `asHagParseNumeric()`.

Is that broadly the shape you have in mind? In particular, would a private refresh worker owned by the PDB integration and tied to `asInit()`/`asShutdown()` be appropriate, or is there an existing scheduling mechanism you would prefer?

anjohnson:
Core Group: Why doesn't reloading the access security file (which already causes the DNS lookup to be redone) work for you? We won't accept all the large amount of extra code here that is duplicative of existing routines for a small benefit. We do agree with the goal, but not this implementation. Can this be distilled down to some additions to the asLib library?

anjohnson:
Following-up from our group meeting earlier today, I believe the Core group does understand the fundamental issue that you're trying to fix and want to accept changes to solve it. However it will take a more step-by-step approach to get there, explaining the details of your requirements and discussing possible approaches to implementing them in advance. Asking us to accept 500 new lines of near-network code isn't conducive to enthusiasm in volunteer reviewers.

Turning this PR into an issue that describes the problems and steps you've taken to investigate them would probably be a good idea.

derekste:
Thanks, @anjohnson and @mdavidsaver. I've been traveling and am catching up.

Reloading the ACF does repeat hostname resolution, but my concern is using that full reinitialization as routine DNS maintenance. The [IOC Access Security documentation](https://docs.epics-controls.org/en/latest/appdevguide/AccessSecurity.html#ioc-access-security-initialization) explicitly states:

> Restarting access security after ioc initialization is an expensive operation and should not be used as a regular procedure.

DNS mappings can change independently of any ACF changes. In a complex environment, the policy may remain unchanged while the addresses associated with its hostnames change dynamically. Periodically re-resolving those names—for example, every five minutes—and updating affected clients' rights seems like an appropriate operation to support without rebuilding the entire access-security configuration.

I accept the concerns about implementation size, duplicated helpers, and normal IOC integration. I'd like to carry this requirement into #863 and agree on the smallest change that supports it. The particular public API and scheduling mechanism remain open for discussion.

PR #949, opened 2026-08-06
More RSRV checks

mdavidsaver:
~Two~ Some RSRV related changes.

I found myself getting confused about where in `write_action()` and especially `write_notify_action()` the `pPayload` buffer was being accessed.  To clarify this I have first extended 33f4d15ff340d4af428d2b111e9d84bbaadf5087 to remove the now unused `pPayload` argument from `log_header()` and friends.  This adds a lot of churn.

A fix for #943, an alternative to #944.  Comes in two parts.

For `write_action()`.  Adds a 40 bytes "scratch" buffer to `struct client` for use in the case of a truncated PUT to a scalar string.  So downstream code will always see a buffer backed by the promised number of bytes.  Does enforce a null terminating byte, although I think this should not be necessary.

For `write_notify_action()`.  Signification rearranging to separate wait/cancel of previous PUT from handling of next PUT.  These two pieces of code are almost entirely unrelated, and having the cancel code right in the middle made it more difficult to follow the logic and usage of `pPayload`.  This adds a fair amount of churn, but I think the end result is easier to follow.

Finally, add checks on `nRequest` to `dbPut()` and `dbPutField()` code paths.  Error if negative counts get that far.  Also, handle zero count in `dbPutFieldLink()` with DBR_CHAR, and count not one in the DBR_STRING case..

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -1087,11 +1087,16 @@ static long dbPutFieldLink(DBADDR *paddr,
     switch (dbrType) {
     case DBR_CHAR:
     case DBR_UCHAR:
-        if (pstring[nRequest - 1] != '\0')
+        if (nRequest == 0) {
+            pstring = pbuffer = ""; // treat zero bytes the same as one null byte
+        } else if (pstring[nRequest - 1] != '\0') {
             return S_db_badDbrtype;
+        }
         break;
 
     case DBR_STRING:
+        if (nRequest != 1)
+            return S_db_onlyOne;
         break;
 
     default:
@@ -1253,6 +1258,8 @@ long dbPutField(DBADDR *paddr, short dbrType,
 
     if (special == SPC_ATTRIBUTE)
         return S_db_noMod;
+    if (nRequest < 0)
+        return S_db_badChoice;
 
     /*check for putField disabled*/
     if (precord->disp && paddr->pfield != &precord->disp)
@@ -1332,6 +1339,8 @@ long dbPut(DBADDR *paddr, short dbrType,
 
     if (special == SPC_ATTRIBUTE)
         return S_db_noMod;
+    if (nRequest < 0)
+        return S_db_badChoice;
 
     if (dbrType == DBR_PUT_ACKT && field_type <= DBF_DEVICE) {
         return putAckt(paddr, pbuffer, 1, 1, 0);
--- a/modules/database/src/ioc/rsrv/camessage.c
+++ b/modules/database/src/ioc/rsrv/camessage.c
@@ -743,7 +743,6 @@ static int write_action ( caHdrLargeArray *mp,
     int                     status;
     long                    dbStatus;
     void                    *asWritePvt;
-    unsigned                size;
 
     pciu = MPTOPCIU(mp, client);
     if(!pciu){
@@ -759,8 +758,19 @@ static int write_action ( caHdrLargeArray *mp,
         mp->m_count = dbChannelFinalElements(pciu->dbch);
     }
 
-    size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1 && mp->m_postsize < MAX_STRING_SIZE
+        && (client->sssb || !!(client->sssb = malloc(MAX_STRING_SIZE)))) {
+        // client sent DBR_STRING with truncated body.  This is allowed when count==1.
+        // we copy into a temporary buffer.
+        // copy into [0, postsize)
+        memcpy(client->sssb, pPayload, mp->m_postsize);
+        // zero remaining [postsize, max_string_size)
+        memset(client->sssb + mp->m_postsize, 0, MAX_STRING_SIZE - mp->m_postsize);
+        // use instead of socket buffer
+        pPayload = client->sssb;
+
+    } else if (dbr_size_n (mp->m_dataType, mp->m_count) > mp->m_postsize) {
+        log_header ( "truncated body", client, mp, 0);
         return RSRV_ERROR;
     }
 
--- a/modules/database/src/ioc/rsrv/camessage.c
+++ b/modules/database/src/ioc/rsrv/camessage.c
@@ -57,8 +57,8 @@
 
 static EVENTFUNC read_reply;
 
-#define logBadId(CLIENT, MP, PPL)\
-logBadIdWithFileAndLineno(CLIENT, MP, PPL, __FILE__, __LINE__)
+#define logBadId(CLIENT, MP)\
+logBadIdWithFileAndLineno(CLIENT, MP, __FILE__, __LINE__)
 
 /*
  * for tracking db put notifies
@@ -283,7 +283,6 @@ static void log_header (
     const char              *pContext,
     struct client           *client,
     const caHdrLargeArray   *mp,
-    const void              *pPayLoad,
     unsigned                mnum
 )
 {
@@ -312,12 +311,11 @@ static void log_header (
 static void logBadIdWithFileAndLineno(
 struct client       *client,
 caHdrLargeArray     *mp,
-const void          *pPayload,
 char                *pFileName,
 unsigned            lineno
 )
 {
-    log_header ( "bad resource ID", client, mp, pPayload, 0 );
+    log_header ( "bad resource ID", client, mp, 0 );
     SEND_LOCK ( client );
     send_err ( mp, ECA_INTERNAL, client, "Bad Resource ID at %s.%d",
         pFileName, lineno );
@@ -330,9 +328,10 @@ unsigned            lineno
 static int bad_udp_cmd_action ( caHdrLargeArray *mp,
                        void *pPayload, struct client *pClient )
 {
+    (void)pPayload;
     if (CASDEBUG > 0)
         log_header ("invalid (damaged?) request code from UDP",
-                    pClient, mp, pPayload, 0);
+                    pClient, mp, 0);
     return RSRV_ERROR;
 }
 
@@ -343,7 +342,8 @@ static int bad_tcp_cmd_action ( caHdrLargeArray *mp, void *pPayload,
                            struct client *client )
 {
     const char *pCtx = "invalid (damaged?) request code from TCP";
-    log_header ( pCtx, client, mp, pPayload, 0 );
+    (void)pPayload;
+    log_header ( pCtx, client, mp, 0 );
 
     /*
      *  by default, clients don't recover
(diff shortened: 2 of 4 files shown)


mdavidsaver:
Codecy is really on top of things.  One complaint that `struct member 'client::sssb' is never used.`.  And another complaint about a `memset()` of that member.

Also, GHA is having a [bad day](https://www.githubstatus.com/).  So spurious CI failures should be expected.

mdavidsaver:
fyi. `sizeof(client)==280` before this change (amd64).  If adding 41 bytes per-client is a significant concern, I can switch to allocate when needed.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -1087,11 +1087,16 @@ static long dbPutFieldLink(DBADDR *paddr,
     switch (dbrType) {
     case DBR_CHAR:
     case DBR_UCHAR:
-        if (pstring[nRequest - 1] != '\0')
+        if (nRequest == 0) {
+            pstring = pbuffer = ""; // treat zero bytes the same as one null byte
```
anjohnson, line 1091: The assignment to `pbuffer` here does nothing, only `pstring` is used below.

anjohnson (commented):
I approve of these changes to dbAccess.c.

I don't like the additional sssb[] buffer and all the extra complications that you added to camessage.c. Please look at my latest commit to #944 which is significantly simpler, and I believe now correct. As usual Codacy doesn't like the added `strncpy()` but in this case it provides exactly the needed functionality (see the notes in the log message).

mdavidsaver:
> which is significantly simpler,

A matter of opinion.  In terms of overall code complexity, I don't see much difference.

>  and I believe now correct

I do not like that a truncated `pbuffer` is allowed to propagate in `write_action()`.  Adding a special case in RSRV rather than address the fundamentals in `caNetConvert()` (and elsewhere) seems to me illustrative.  It is just too difficult to prove that code elsewhere does not blindly assume DBR_STRING is backed by 40 bytes.  Both now, and in future.  The "future" part is significant for me.  It would mean future developers/reviewers having to recognize that a change to code handling DBR_STRING, not in camessage.c, is called on this one special code path where the provided buffer may have less than 40 bytes.

Thus my preference to locally ensure that doing so will be safe.  With an alternative being a global change to all DBR_STRING handling to use null aware manipulations.

> I don't like the additional sssb[] buffer

In what way(s)?  The extra per-client memory usage?  Execution time?  The name?  Something else?

modules/database/src/ioc/db/dbAccess.c
mdavidsaver: Correct.  I opt to keep these two variables in sync. to avoid confusing some future developer.

mdavidsaver:
> I approve of these changes to dbAccess.c.

@anjohnson Your approval seems a bit reluctant.  So I will wait on merging until you have a chance to reply.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -1683,9 +1686,13 @@ static int write_notify_action ( caHdrLargeArray *mp, void *pPayload,
         putNotifyErrorReply (client, mp, ECA_NOWTACCESS);
         return RSRV_OK;
     }
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1 && mp->m_postsize < MAX_STRING_SIZE) {
+        memcpy(client->sssb, pPayload, mp->m_postsize);
+        memset(client->sssb+mp->m_postsize, 0, sizeof(client->sssb)-mp->m_postsize);
+        pPayload = client->sssb;
+        size = MAX_STRING_SIZE;
```
anjohnson: Buffer not required for putNotify, which has its own.

modules/database/src/ioc/rsrv/server.h
```diff
@@ -102,6 +102,7 @@ typedef struct client {
   unsigned              recvBytesToDrain;
   unsigned              priority;
   char                  disconnect; /* disconnect detected */
+  char                  sssb[MAX_STRING_SIZE+1]; // scalar string scratch buffer
```
anjohnson: Replace with char*, alloc when needed.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -759,8 +758,12 @@ static int write_action ( caHdrLargeArray *mp,
         mp->m_count = dbChannelFinalElements(pciu->dbch);
     }
 
-    size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1 && mp->m_postsize < MAX_STRING_SIZE) {
+        memcpy(client->sssb, pPayload, mp->m_postsize);
+        memset(client->sssb+mp->m_postsize, 0, sizeof(client->sssb)-mp->m_postsize);
```
anjohnson: Use strncpy(), and look at my epicsStrnLen() code for the check.

anjohnson (commented):
Thanks!

mdavidsaver:
Updated with significant changes.  I have edited the PR description to reflect the current state.

modules/database/src/ioc/rsrv/camessage.c
mdavidsaver: I do not think that `strncpy()` can be a one-and-done here.  `pPayload` may not be NULL terminated, and is known to be smaller than the destination buffer. So I would have to pass postsize to `strncpy()`, then still have a `memset()` of the remaining buffer.  I don't like the idea of leaving the trailing part of the buffer with old values.

I have added comments to make my intent clear.

modules/database/src/ioc/rsrv/server.h
mdavidsaver: The `char sssb[]` becomes `char* sssb`, and is only used from `write_action()`.

modules/database/src/ioc/rsrv/camessage.c
mdavidsaver: I have rearranged `write_notify_action()` to copy `pPayload` directly to the put notify buffer.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -1664,6 +1664,60 @@ void rsrvFreePutNotify ( client *pClient,
      }
 }
 
+/* Only one concurrent write_notify (aka. put with callback) is allowed for each client.
+ * If a second write_notify is attempted before the first has completed, then wait for
+ * some time for the first to complete.  Then cancel the first.
+ */
+static const double write_notify_timeout = 60.0;
```
anjohnson, line 1671: The addition of the write-notify timeout as described in this comment should be publicly discussed with the community before being merged, I believe it risks breaking many beamline databases.

In any case since it changes the [documented behavior](https://github.com/epics-base/epics-base/blob/67f7ee55bdc089a536b10a69f1a1ceefb4bef5b9/modules/ca/src/client/CAref.html#L2911) of the `ca_put_callback()` and `ca_array_put_callback()` routines, the change _must_ be described in the Release Notes in detail, and an update made to the CAref.html documentation.

The synApps [busy record](https://epics-modules.github.io/busy/busyRecord.html) and [swait record](https://epics-modules.github.io/calc/swaitRecord.html) types were designed to be used in long-running experiments making use of put-notify callbacks to signal acquisition or experiment completion. These experiments can run for many hours at some beamlines, with callbacks only being triggered at the end of long operations. 

If I've understood this proposal, an experimenter triggering some processing by mistake while a long-running experiment is under way could cause a second put-callback to be run and causing the first callback to happen early with an error, wasting experimental beam-time.

Adding a timeout might be acceptable if the _second_ write-notify operation is the one that gets cancelled instead of the first, but the synApps maintainers should be consulted and this option suggested to them. I suggest moving this change into an independent PR to allow for separate discussion on that topic without holding up the other changes.

mdavidsaver: > The addition of the write-notify timeout as described in this comment ...

There is no behavior change.  I am calling out and documenting a magic number which was already there.

mdavidsaver: The effective difference

```diff
+        epicsEventStatus status = epicsEventWaitWithTimeout(client->blockSem, write_notify_timeout);
-        status = epicsEventWaitWithTimeout(client->blockSem,60.0);
```

I guess that I am not the only one who has been looking past this line for many years without really noticing that magic `60.0` :)

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -743,24 +743,34 @@ static int write_action ( caHdrLargeArray *mp,
...
         return RSRV_ERROR;
     }
     if (mp->m_count > dbChannelFinalElements(pciu->dbch)) {
         mp->m_count = dbChannelFinalElements(pciu->dbch);
     }
 
-    size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
+    if (mp->m_dataType == DBR_STRING && mp->m_count == 1 && mp->m_postsize < MAX_STRING_SIZE
+        && (client->sssb || !!(client->sssb = malloc(MAX_STRING_SIZE)))) {
```
brendanchandler, line 762: Do we need to free this client->sssb?  It seems like destroy_client() should clean this up.

Optionally, initializing it to NULL in create_client is explicit and we do it for several other fields in client (pUserName, pHostName), though I think technically not needed due to the calloc creation of client.

mdavidsaver: > Do we need to free this client->sssb?

Yes.  Now added.  I remember having done so before, but apparently this did not make it up.

modules/database/src/ioc/rsrv/camessage.c
```diff
@@ -1684,94 +1748,59 @@ static int write_notify_action ( caHdrLargeArray *mp, void *pPayload,
         return RSRV_OK;
     }
 
-    size = dbr_size_n (mp->m_dataType, mp->m_count);
-    if (size > mp->m_postsize) {
-        return RSRV_ERROR;
-    }
-
     if ( pciu->pPutNotify ) {
+        write_notify_cancel(pciu);
```
anjohnson, line 1752: `write_notify_cancel()` is now being called before the payload size calculation  and the call to `rsrvExpandPutNotify()` (R1768-1770; checked at R1777 & R1784), whereas previously the check happened before both (L1687-8). A malformed write-notify packet can thus expand the PN-buffer and cancel & trigger the callback of an active put-callback operation.

The early buffer expansion may be more of a problem than the latter, were you aware of either?

modules/database/src/ioc/rsrv/camessage.c
anjohnson: Ack, sorry!

anjohnson (commented):
Okay I'm convinced. One final question from me below though about the payload size check(s) in `write_notify_action()`.

modules/database/src/ioc/rsrv/camessage.c
mdavidsaver: Both possibilities are gated by the `rsrvCheckPut()` test.  And `size` is based on a ~size~count which has already been clipped to `dbChannelFinalElements()`.

PR #956, opened 2026-09-03
No more silent rules

mdavidsaver:
It is super confusing to see error messages coming from hidden commands, leading to mis-attribution.  So remove all silent (`@`) makefile rules.

eg. seeing an `include` error coming after `podRemove.pl`, which is actually printed from a hidden `dbdExpand.pl -D`!


configure/RULES.Db
```diff
@@ -65,7 +65,7 @@ DBDCAT_SOURCE = $(firstword $(wildcard $(file) $(foreach dir,  \
 
 DBDCAT_COMMAND = $(if $(DBDCAT_SOURCES),\
     $(CAT) $(DBDCAT_SOURCES) > $(notdir $@),\
-    @echo "No input files for $(notdir $@)")
+    echo "No input files for $(notdir $@)")
```
simon-ess: I think that the `echo` statements should still be `@`ed. I think the output
```
cat foo.dbd bar.dbd > out.dbd
echo "No input files for out.dbd"
No input files for out.dbd
```
seems a bit silly.

(and of course this applies elsewhere)

mdavidsaver: Presumably this exception would be limited to `echo` to stdout/err?  As opposed to say `echo xxx > $@`?

simon-ess: That seems sensible, yes.

ralphlange: Absolutely.
Just to avoid the echo command and its output in two successive lines of the same output stream.

mdavidsaver: Updated.  I think I found all of the places with stdout prints.

configure/RULES_BUILD
```diff
@@ -342,33 +342,33 @@ $(INSTALL_LIB):
...
 
 $(TESTMUNCHNAME): %$(MUNCH_SUFFIX): $(MUNCH_DEPENDS) %$(EXE)
-	@$(RM) $@
+	$(RM) $@
 	$(MUNCH_CMD)
 
 #---------------------------------------------------------------
 # GeSys modules for RTEMS
 $(MODNAME): %$(MODEXT): %$(EXE)
 	@echo "Building module $@"
```
simon-ess, line 364: This _could_ be an `$(ECHO)` 🤷

documentation/Makefile
```diff
@@ -112,29 +112,29 @@ else
...
+	$(RM) $@
 	$(PERL) $(MAKENOTES) -o $@ -V $(EPICS_SHORT_VERSION) $(OLD_NOTES)
   endif
 
   $(HEADER_MD_FILES): %_h.md: ../HEADER_h.md
 	$(EXPAND_TOOL) -t $(INSTALL_LOCATION) -DHEADER=$* $< $@
 
   $(API_RST_FILES): %-api.rst: ../%-API.rst
-	@$(RM) $@
 	@$(ECHO) Creating $@
```
simon-ess, line 129: This one can be removed, since `ECHO` is define as
```
ECHO = @$(if $(make-s),$(NOP),echo)
```

src/template/ext/top/configure/RULES_IDL
```diff
@@ -24,11 +24,11 @@ buildInstall : $(INSTALL_IDLS)
 
 $(INSTALL_IDL)/%: %
 	@echo "Installing idl program $@"
-	@$(INSTALL_IDLFILE) -d -m 644 $< $(INSTALL_IDL)
+	$(INSTALL_IDLFILE) -d -m 644 $< $(INSTALL_IDL)
 
 $(INSTALL_IDL)/%: ../%
 	@echo "Installing idl program $@"
```
simon-ess, line 30: `echo` -> `$(ECHO)`

anjohnson:
Core: ANJ to review before his vacation.

src/template/ext/top/configure/RULES_IDL
mdavidsaver: I would like to keep the scope of this expansive PR narrow to aid in review.  Just about removing `@`.

PR #957, opened 2026-09-03
Add support to module version registry

lucmaga:
This change adds support to a registry of the linked EPICS modules and versions.  The registry works as a simple linked list that holds the information of each module that is printed out using the epicsModulesVersion command.

Each module register it self thought the <module>_registerModule.cpp file generated by the generate_module_version.py python script.  The code uses a similar trick as the registerDeviceDriver facility to call a initialization code that populates the registry.

For static libraries the anchor symbols are added to the ioc so the linker include the initialization code.  The anchor code is generated via a the generate_anchor_symbols.py python script.

This is what it looks like for softIocPVA, "Com" is using the result of genVersionHeader.pl:
image

Notice this changes libCom, in example, to add Com_VERSION and this is the version added to the registry.
So to enable it the modules need to add a similar, <library>_VERSION, variable to work.


modules/ca/src/client/Makefile
```diff
@@ -88,6 +88,7 @@ API_HEADER = libCaAPI.h
 ca_API = libCa
 
 LIBRARY=ca
+ca_VERSION=7.0.10
```
simon-ess, line 91: These should probably be fetched from various configuration files. Moreover, these are actually the wrong versions, see `CONFIG_CA_VERSION` (and the same for database and libcom)

modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -488,6 +489,13 @@ static void installLastResortEventProviderCallFunc(const iocshArgBuf *args)
     iocshSetError(installLastResortEventProvider());
 }
 
+static const iocshFuncDef epicsModulesVersionFuncDef = {"epicsModulesVersion", 0, NULL,
```
simon-ess: Nitpick, but I think this would be better named as e.g. `epicsShowModuleVersions`

modules/libcom/src/iocsh/Makefile
```diff
@@ -11,12 +11,14 @@ SRC_DIRS += $(LIBCOM)/iocsh
 INC += iocsh.h
 INC += initHooks.h
 INC += registry.h
+INC += registerModules.h
```
simon-ess, line 14: Should this have an ifdef around it as is done in the registry makefile?

src/tools/generate_anchor_symbols.py
```diff
@@ -0,0 +1,37 @@
...
+    prog="generate_weak_symbols",
+    description="generate the weak symbols for epics module registration",
+)
+
+parser.add_argument(
+    "-p",
+    "--prod",
+    required=True,
+)
+parser.add_argument("libs", nargs="+")
```
simon-ess, line 16: Could add help text to these 🔧

src/tools/generate_anchor_symbols.py
```diff
@@ -0,0 +1,37 @@
...
+
+args = parser.parse_args()
+filename = Path(f"{args.prod}_weakAnchor.cpp")
```
simon-ess, line 19: Might want to validate the arguments to ensure e.g. there are no spaces or any other invalid characters.

configure/RULES_BUILD
```diff
@@ -204,6 +204,31 @@ ifneq ($(DELAY_INSTALL_LIBS),YES)
...
+# for each PROD. The anchor is necessary for static builds otherwise the
+# linker will garbage collect the symbols.
+ifeq ($(REGISTER_LINKED_MODULES),YES)
+define MODULE_REGISTER_TEMPLATE
+$$(if $$(eq $$(strip $$($(strip $(1))_VERSION)),),$$(error "$(strip $(1))_VERSION is not defined."),)
+LIBRARY_OBJS += $$(strip $(1))_registerModule$(OBJ)
+$(1)_registerModule.cpp: $(REGISTERMODULE_py)
+	$(REGISTERMODULE) -n $$(strip $(1)) -v $$(strip $$($$(strip $(1))_VERSION))
+.PRECIOUS: $(1)_registerModule.cpp
+endef
```
simon-ess, line 218: You definitely don't need the `$(strip $(1))`s here; you're stripping on the `$(foreach )` outside.

I'm not sure you actually need the `$(strip )` in the foreach either; as I understand make's splitting the following should be equivalent:
```make
$(foreach foo,$(FOO),$(eval $(call TEMPLATE,$(foo))))
$(foreach foo,$(FOO),$(eval $(call TEMPLATE,$(strip $(foo)))))
```

configure/RULES_BUILD
```diff
@@ -204,6 +204,31 @@ ifneq ($(DELAY_INSTALL_LIBS),YES)
 $(TESTPRODNAME) $(PRODNAME): | $(INSTALL_LIBS) $(INSTALL_DLLSTUB_LIBS)
 endif
 
+# Module register facility
+# Creates the registerModule.cpp targets for each LIBRARY and a _weakAnchor
+# for each PROD. The anchor is necessary for static builds otherwise the
+# linker will garbage collect the symbols.
+ifeq ($(REGISTER_LINKED_MODULES),YES)
+define MODULE_REGISTER_TEMPLATE
+$$(if $$(eq $$(strip $$($(strip $(1))_VERSION)),),$$(error "$(strip $(1))_VERSION is not defined."),)
```
simon-ess: Should this be a warning or an error? I would be inclined to say that this should be a warning and to only proceed if the version is defined. This way you would have EPICS base define that you _want_ module versions represented but missing versions from dependencies wouldn't stop the relevant builds.

You could also say something like
```make
define MODULE_REGISTER_TEMPLATE
LIBRARY_OBJS += $$(strip $(1))_registerModule$(OBJ)
$(1)_registerModule.cpp: $(REGISTERMODULE_py)
	$(REGISTERMODULE) -n $(1) -v $$(or $$(strip $$($(1)_VERSION)),$$(warning Version missing!!!!!!)dev)
endef
```
i.e. if the version is missing use "dev" but warn at the same time.

modules/libcom/src/iocsh/libComRegister.c
lucmaga: I thought about it, but I found it too long. Maybe its better anyway.

modules/libcom/src/iocsh/registerModules.h
```diff
@@ -0,0 +1,8 @@
+#ifndef INC_registerModules_H
+#define INC_registerModules_H
+
+
+void registerPrintModules();
+int registerModule(char *name, char *version);
```
mdavidsaver: Public functions need to be decorated with an `*_API` macro.  In this case `LIBCOM_API` from `libComAPI.h`.

Good practice would have the argument types being `const char *` unless this function will actually modify `name` and `version`.

Also, include some API for iteration so that the list could be extracted and publish by some device support code.

modules/libcom/src/iocsh/Makefile
lucmaga: The registry one was added by mistake.

However I think we can have this anyway, The way it is right now the modules ¿could? select to add this or not.

src/tools/generate_module_version.py
```diff
@@ -0,0 +1,37 @@
...
+/* THIS IS A GENERATED FILE. DO NOT EDIT! */
+#ifndef INC_{filename.stem.upper()}_H
+#define INC_{filename.stem.upper()}_H
+
+extern "C" {{
+#include <registerModules.h>
+#include <compilerDependencies.h>
+}}
+
+#define ___MODULE_NAME__ "{args.name}"
```
mdavidsaver, line 25: I think that C macros names starting with `__` are reserved for use by the preprocessor.  Since your script is emitting a complete C file, there would not much risk of collision with eg. `EPICS_MODULE_VERSION`.

modules/libcom/src/iocsh/Makefile
lucmaga: But  I guess the best solution would be to do what you and Michael suggested in person, that is to have it enable by default and just use the version from genVersionHeader.pl

modules/ca/src/client/Makefile
mdavidsaver: Use (or at least consider) `$(SHRLIB_VERSION)`, which is also used in shared library names (eg. SONAME).

configure/RULES_BUILD
lucmaga: Add the genVersionHeader.pl as a fallback.

modules/libcom/src/iocsh/registerModules.h
lucmaga: Done

modules/libcom/src/iocsh/libComRegister.c
lucmaga: Done

src/tools/generate_anchor_symbols.py
lucmaga: spaces are valid characters for names, as well as any other symbol besides '/' AFAIK.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,57 @@
...
+typedef struct {
+  ELLNODE node;
+  char *name;
+  char *version;
+} moduleInfo;
+
+static ELLLIST moduleRegistryList;
+static int initRegisterModule = 0;
+
+static int initModuleRegistry() {}
```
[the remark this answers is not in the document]
lucmaga: Fixed

lucmaga:
I'm not sure why this specific build is failing. Also I can't test on RTEMS, so I'm not sure how it works on those systems.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+  char *name;
+  char *version;
+} moduleInfo;
+
+static ELLLIST moduleRegistryList;
+static int initRegisterModule = 0;
+
+static moduleInfo *allocateModule(char *name, char *version) {
+  moduleInfo *module = callocMustSucceed(1, sizeof(moduleInfo),
+                                         "Fail to allocate module structure");
```
mdavidsaver: When sizing an allocation for a pointer type, prefer `sizeof(*module)`.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+  module->name = callocMustSucceed(1, strlen(name) * sizeof(char),
+                                   "Fail to allocate module name");
+  module->version = callocMustSucceed(1, strlen(version) * sizeof(char),
+                                      "Fail to allocate module version");
```
mdavidsaver, line 25: In C `sizeof(char)` will always be `1`, so this multiplier can be omitted.  `strlen()` returns the number of characters, excluding the null terminator.  When allocating, a `+1` is necessary to store a null.

Also, this could all be done with one allocation of `sizeof(*module) + strlen(name) + strlen(version) + 2`.

```c
size_t namelen = strlen(name), versionlen = strlen(version);
moduleInfo * module = callocMustSucceed(sizeof(*module) + namelen + versionlen + 2);
module->name = (char*)(module+1);
module->version = module->name + namelen + 1;
memcpy(module->name, name, namelen + 1);
memcpy(module->version, version, versionlen+1);
```

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+
+  strcpy(module->name, name);
+  strcpy(module->version, version);
+  return module;
+}
+
+INLINE void initLinkedList() {
```
mdavidsaver, line 32: This function should be `static`.  This permits the compiler to make the optimizations you probably intend with `INLINE`.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+#include <valgrind/valgrind.h>
+
+typedef struct {
+  ELLNODE node;
+  char *name;
+  char *version;
+} moduleInfo;
+
+static ELLLIST moduleRegistryList;
+static int initRegisterModule = 0;
```
mdavidsaver, line 17: Use `epicsThreadOnce()` for thread-safe singleton initialization, and an `epicsMutex` to guard the global list.  This should not much boilerplate, and will make the whole API safe.

Alternately, take some action to ensure `registerModule()` is only called before `iocInit()`.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+  moduleInfo *nextModule;
+  initLinkedList();
+  printf("Module list:\n");
+  if (!ellCount(&moduleRegistryList)) return;
+  nextModule = (moduleInfo *)ellFirst((&moduleRegistryList));
+  do{
+    printf("%s: %s\n", nextModule->name, nextModule->version);
+    nextModule = (moduleInfo *)ellFirst(nextModule);
+  }while(nextModule != NULL);
+}
```
mdavidsaver, line 49: Iteration of ellLib looks a bit more conventional with a for-loop.

```c
  ELLNODE *cur;
  initLinkedList();
  printf("Module list:\n");
  for(cur = ellFirst(&moduleRegistryList); cur ; cur = ellNext(cur)) {
    moduleInfo *nextModule = CONTAINER(cur, moduleInfo, node);
    printf("  %s: %s\n", nextModule->name, nextModule->version);
  }
```

I suggest indenting the individual module lines.  Also, `CONTAINER` is defined in `dbDefs.h`.

modules/libcom/src/iocsh/registerModules.h
```diff
@@ -0,0 +1,9 @@
+#ifndef INC_registerModules_H
+#define INC_registerModules_H
+
+#include "libComAPI.h"
+
+LIBCOM_API void registerPrintModules();
+LIBCOM_API int registerModule(char *name, char *version);
```
mdavidsaver: Please make the arguments `const char*`.

src/tools/genVersionHeader.pl
```diff
@@ -210,6 +216,7 @@ sub HELP_MESSAGE {
         -d         - Dry-run
         -i         - Question mode
         -q         - Quiet
+	-p	   - Only print the version and exit
```
mdavidsaver: Fix indentation.

mdavidsaver (changes requested):

modules/ca/src/client/Makefile
mdavidsaver: Looking again, please remove these new `*_VERSION` entries.  Having these redundant version numbers scattered around the tree invites us to forget about them, then they will become stale.

modules/libcom/src/iocsh/registerModules.h
mdavidsaver: @lucmaga While we have not really established a convention for this, I prefer to let the original commenter decide whether a topic has been resolved.

In this case, you have acted on one of my three points.

mdavidsaver:
> I'm not sure why this specific build is failing.

I do not see any build failing.  I guess that whatever was going wrong has been resolved?

mdavidsaver:
Overall, I like the idea of capturing support module library version information in a consistent way.  Something like an automatic software-bill-of-materials.  I would find this useful when eg. asked to support someone else's IOC.

Actually, this idea could be extended to the executable as well.

configure/CONFIG_SITE
```diff
@@ -178,5 +178,9 @@ LINKER_USE_RPATH = YES
 # All root directories are considered to be the same.
 LINKER_ORIGIN_ROOT = $(INSTALL_LOCATION)
 
+# Enables the epicsModuleVersions command. Enables the generation
```
anjohnson, line 181: This comment line calls the command `epicsModuleVersions` which is correct English for a series of module versions. Elsewhere though (including in the actual IOCSH command name) you use `epicsModulesVersion` and similar, which I would read as a single version number defining a set of modules. Please fix.

configure/CONFIG_BASE
```diff
@@ -53,9 +53,14 @@ PODTOHTML        = $(PERL) $(PODTOHTML_pl)
...
 FULLPATHNAME     = $(PERL) $(TOOLS)/fullPathName.pl
-GENVERSIONHEADER = $(PERL) $(TOOLS)/genVersionHeader.pl $(QUIET_FLAG) $(QUESTION_FLAG)
+GENVERSIONHEADER_pl = $(TOOLS)/genVersionHeader.pl
+GENVERSIONHEADER = $(PERL) $(GENVERSIONHEADER_pl) $(QUIET_FLAG) $(QUESTION_FLAG)
 
 MAKERPATH = $(PYTHON) $(TOOLS)/makeRPath.py
+REGISTERMODULE_py = $(TOOLS)/generate_module_version.py
+REGISTERMODULE = $(PYTHON) $(REGISTERMODULE_py)
+REGISTEANCHORSYMBOLS_py = $(TOOLS)/generate_anchor_symbols.py
+REGISTERANCHORSYMBOLS = $(PYTHON) $(REGISTERANCHORSYMBOLS_py)
```
anjohnson, line 63: One of these variable names is not like the others, I'm surprised that this passes the CI builds; is `generate_anchor_symbols.py` not needed to build Base? `REGISTEANCHORSYMBOLS_py` is missing an `R` which is present in the name used in the line below, so I'm not sure how the recipes running `$(REGISTERANCHORSYMBOLS)` can be working.

Can you make the variable names more closely match the program names they point to or run please? If I want to find a rule that uses `generate_anchor_symbols.py` I would expect it to use a variable like `GENERATEANCHORSYMBOLS` or maybe `GENANCHORSYMS`.

Most newer tool program names in EPICS use camelCase or flatcase; a few older ones used snake_case or PascalCase, and we have one that uses kebab-case but that was imported from elsewhere.

modules/libcom/src/iocsh/registerModules.h
lucmaga: No worries. Every community has a different convention on who/when to resolve conversations. I'll leave the topics open for you to resolve then.

lucmaga:
> Overall, I like the idea of capturing support module library version information in a consistent way. Something like an automatic software-bill-of-materials. I would find this useful when eg. asked to support someone else's IOC.

This is definitely useful, at ESS we are working on a project to add this information to a database, so we can monitor IOCs running on old versions.  We currently keep a similar implementation on require, that we could remove if this is merged.

modules/libcom/src/iocsh/registerModules.c
lucmaga: Why?

modules/libcom/src/iocsh/registerModules.c
lucmaga: I try to follow the explicit better than implicit rule most of the time. For the `sizeof(char*)` it is solved in build time with the `*1` but improves readability IMHO.
For the multiple allocations I see that it can be a small optimization, but in this case I would also prefer readability. This code only runs fore each module registration on the startup, the impact of this is negligible.

modules/libcom/src/iocsh/registerModules.c
lucmaga: the definition of `INLINE` is 
```
#ifndef INLINE_defs_EPICS
#define INLINE_defs_EPICS
#   ifndef __cplusplus
#       if defined (__GNUC__)
#           define INLINE static __inline__
#       elif defined (_MSC_VER)
#           define INLINE __inline
#       elif defined (_SUNPRO_C)
#           pragma error_messages (off, E_EXTERN_PRIOR_REDECL_STATIC)
#           define INLINE static
#       else
#           define INLINE static
#       endif
#   endif /* ifndef __cplusplus */
#endif /* ifdef INLINE_defs_EPICS */
```

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done.

modules/libcom/src/iocsh/registerModules.h
lucmaga: Done.

src/tools/genVersionHeader.pl
lucmaga: Done.

configure/RULES_BUILD
```diff
@@ -204,6 +204,32 @@ ifneq ($(DELAY_INSTALL_LIBS),YES)
 $(TESTPRODNAME) $(PRODNAME): | $(INSTALL_LIBS) $(INSTALL_DLLSTUB_LIBS)
 endif
 
+# Module register facility
+# Creates the registerModule.cpp targets for each LIBRARY and a _weakAnchor
+# for each PROD. The anchor is necessary for static builds otherwise the
+# linker will garbage collect the symbols.
+ifeq ($(REGISTER_LINKED_MODULES),YES)
```
anjohnson, line 211: Could you move this new block of rules downwards in the file to around line 450, where there are other similar divided sections with a header comment. That's also where `$(GENVERSIONHEADER)` is currently used.

configure/RULES_BUILD
anjohnson: You resolved this comment, please explain why all of these `$(strip )` calls are necessary.

modules/libcom/src/iocsh/libComRegister.c
```diff
@@ -488,6 +489,13 @@ static void installLastResortEventProviderCallFunc(const iocshArgBuf *args)
     iocshSetError(installLastResortEventProvider());
 }
 
+static const iocshFuncDef epicsModulesVersionFuncDef = {"epicsShowModulesVersion", 0, NULL,
+                                                        "Shows all the modules loaded on the ioc"};
+static void epicsModulesVersionCallFunc(const iocshArgBuf *args)
+{
+    registerPrintModules();
```
anjohnson, line 496: The IOCSH command name needs to match the routine that the command calls. On VxWorks IOCs (which EPICS 7.0.x still supports) the VxWorks target shell doesn't use the CallFunc at all. It looks up the commands in the application's symbol table and calls them directly, so using different names like this will confuse any VxWorks users that try to run the command.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+#include <freeList.h>
+#include <shareLib.h>
+#include <stdio.h>
+#include <string.h>
+#include <valgrind/valgrind.h>
+
+typedef struct {
+  ELLNODE node;
+  char *name;
+  char *version;
```
anjohnson, line 13: The general EPICS indentation standard for new code is to use 4 spaces in C/C++ files. We don't have that in the `.editorconfig` file because older source files had different standards, and we don't want to force PRs to reformat existing code to pass the CI checks (although some Core group members do sometimes reformat files).

src/tools/generate_anchor_symbols.py
anjohnson: The EPICS build system doesn't generally work with filenames that include spaces, because GNU Make uses space as a separator in lists of filenames. I don't think you need to check for that, but don't include spaces in any filenames that you generate.

modules/libcom/src/iocsh/registerModules.c
anjohnson: Please use `epicsStrDup()` from epicsString.h instead all of these allocations and `strcpy()` calls.

anjohnson (changes requested):
I would have liked to see an issue explaining the purpose of this project to allow for some discussion before the implementation was developed, but I think I understand it now. This PR will need an entry in the Release Notes explaining that and the new command.

I don't have time now to comment on everything here, I might have some more questions and comments later.

configure/RULES_BUILD
lucmaga: My mistake. I will remove it.

configure/CONFIG_BASE
lucmaga: I added a bug during my naming refactor. It was broken for static links because of this, its fixed now. Thanks.

configure/RULES_BUILD
lucmaga: That does not work because I change `LIBRARY_OBJS` that is used to define the `$(LIBRARY)` target on line 324. And `PROD_OBJS` used in lines 254 and 256.

configure/RULES_BUILD
lucmaga: Wait, I got confused actually. I remove the strips for the `$(strip $(1))` @simon-ess asked for. 

What I haven't remove is the on the foreach loop. I just did it.

configure/RULES_BUILD
```diff
@@ -204,6 +204,32 @@ ifneq ($(DELAY_INSTALL_LIBS),YES)
...
+$(foreach library, $(LIBRARY), $(eval $(call MODULE_REGISTER_TEMPLATE,$(library))))
+
+ifeq ($(strip $(STATIC_BUILD)), YES)
+define ANCHOR_TEMPLATE
+PRODUCT_OBJS += $(1)_weakAnchor.o
+$(1)_weakAnchor.cpp: $(REGISTERANCHORSYMBOLS_py)
+	$(REGISTERANCHORSYMBOLS) -p $(1) $$($(1)_LIBS)
+.PRECIOUS: $(1)_weakAnchor.cpp
+endef
+$(foreach prod, $(PROD_IOC), $(if $(eq $(strip $(prod)_LIBS),),,$(eval $(call ANCHOR_TEMPLATE,$(prod)))))
```
simon-ess: This `$(strip )` is wrong: you're doing
```make
$(strip $(prod)_LIBS)
```
which will only strip some whitespace at the beginning of `$(prod)_LIBS`. I'm assuming that you want instead
```make
$(strip $($(prod)_LIBS))
```
where you are trying to strip the whitespace around the _value_ of `$(prod)_LIBS`.

modules/libcom/src/iocsh/libComRegister.c
lucmaga: Done.

configure/CONFIG_SITE
lucmaga: It reads for me that modules can have more that one version with `epicsModuleVersions`. I've change to `epicsShowModuleVersion`.

configure/RULES_BUILD
lucmaga: Done.

configure/CONFIG_SITE
anjohnson: Without the 's' neither Module nor Version is a plural, so that says there is one module with one version number. English rules can be complicated and I never had to learn formal grammar myself (until I took Latin) so I asked Google's AI to explain the rule:

> "Module versions" describes a list showing different module names and their version numbers.
> Noun adjuncts: When you use a noun to describe another noun (like "module" describing "version"), the first noun acts as a modifier.
> Singular form rule: In English, a noun acting as a modifier stays in its singular form, even if it refers to many items.

modules/libcom/src/iocsh/registerModules.h
```diff
@@ -0,0 +1,9 @@
+#ifndef INC_registerModules_H
+#define INC_registerModules_H
+
+#include "libComAPI.h"
+
+LIBCOM_API void registerPrintModules();
+LIBCOM_API int registerModule(const char *name, const char *version);
+
+#endif /* INC_registerModules_H */
```
mdavidsaver, line 9: I would like to see some API for iterating modules programmatically.  eg. based either on callbacks or iterative calls.  eg.

```c
typedef struct {
    const char *name, *version;
} ModInfo;

// iteration alternative
// pass NULL for first.  Returns NULL after last
const ModInfo* moduleNext(const ModInfo* prev);

// callback alternative
typedef void (*ModNotify)(void *, const ModInfo*);
void moduleVisit(ModNotify fn, void *pvt);
```

I find the iteration option easier to use than the callback option, but iteration would require that new modules be appended, as opposed to inserted in some sorted order.

modules/libcom/src/iocsh/registerModules.c
mdavidsaver: > Why?

Typo protection.  Consider the following

```c
struct A;
struct B;
...
A* a = malloc(sizeof(*a));
A* a = malloc(sizeof(A));

# now change the variable name and type.

B* b = malloc(sizeof(*a)); // compile error
B* b = malloc(sizeof(A));  // oops! compiles, but allocates based on the wrong size.
```

modules/libcom/src/iocsh/registerModules.c
mdavidsaver: I understand this.  The `inline` is not giving the compiler any extra information, and it will anyway likely be ignored.  As it should be imo.  This is a case where it seems better to let the optimizer choose whether to inline or not.

configure/CONFIG_SITE
mdavidsaver: > I never had to learn formal grammar myself
> so I asked Google's AI to explain the rule:

Andrew, please either assert your own authority as EPICS Name Czar, or cite a real source.

modules/libcom/src/iocsh/registerModules.h
simon-ess: I think that the callback option would proably be useful for us as well, thought either could work.

lucmaga: I'll be glad to add it. But I would argue if this shouldn't be done in a following PR. IMHO, this change already looks big enough for a feature addition.

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done.

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done.

modules/libcom/src/iocsh/registerModules.c
```diff
@@ -0,0 +1,56 @@
...
+    moduleInfo *nextModule;
+    initLinkedList();
+    printf("Module list:\n");
+    if (!ellCount(&moduleRegistryList)) return;
+    nextModule = (moduleInfo *)ellFirst((&moduleRegistryList));
+    do{
+        printf("%s: %s\n", nextModule->name, nextModule->version);
+        nextModule = (moduleInfo *)ellFirst(nextModule);
+    }while(nextModule != NULL);
+}
```
anjohnson: Your list-traversing code here is more complicated that it needs to be, and using `ellFirst()` where I expected to see `ellNext()` makes it even more confusing. I wouldn't have commented if it weren't for that though since there are many examples of code similar to this in dbStaticLib.c.

Please look at `dbStateShowAll()` in modules/database/src/ioc/db/dbState.c for an example that is doing the same thing as this, but with no visible casts at all.

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done.

modules/libcom/src/iocsh/registerModules.c
lucmaga: Implemented @mdavidsaver suggestion.

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done

modules/libcom/src/iocsh/registerModules.c
lucmaga: Done

modules/ca/src/client/Makefile
lucmaga: Done.

src/tools/generate_anchor_symbols.py
lucmaga: Done

modules/database/src/std/Makefile
```diff
@@ -20,6 +20,7 @@ API_HEADER = dbRecStdAPI.h
 dbRecStd_API = dbRecStd
 
 LIBRARY_IOC += dbRecStd
+dbRecStd_VERSION = 7.0.10
```
mdavidsaver, line 23: Please remove all of these new _VERSION .

lucmaga: Done

Issue #939, opened 2026-07-23
Inconsistent yacc error handling


mdavidsaver:
The various yacc generated parsers in epics-base are inconsistent about signaling fatal error states.  Only `dbLoadTemplate.y` actions use `YYABORT` to immediately interrupt parsing.  In `asLib.y` actions calls `yyerror()`, which sets a special `yyFailed` global, then continues.  That global is checked after `yyparse()` has returned for some other reason.  `dbYacc.y` does something similar, although with both a `yyFailed` and `yyAbort` flags.  The `yyAbort` flag will cause `db_yyinput` to return as if EoF was encountered.

I think the correct way for an action in one of these parsers to signal an unrecoverable error to print an error, then `YYABORT` (which expands to a `goto`).

As a note, along with `YYABORT`, there is `YYERROR`.  The second triggers error recovery, which none of our parsers attempt.

anjohnson:
I don't agree that consistently aborting on first error is necessarily a desirable outcome.

The responses to errors are different because for some kinds of files it was desirable to continue loading and hence syntax-checking the file, instead of immediately stopping. Some IOCs can take quite a while to start, so only finding out about one error per boot can be tedious and slow.

If we provided a way to check all these file types at build time that might be less necessary, but we only have `ascheck` available for that at the moment. The behavior of asLib might have been added to support that `ascheck` program.

C & C++ compilers tend to continue after reporting the first error, javac is similar, while the Go compiler stops after seeing 10 errors. Before developing any changes we should agree on what we actually want to happen in each case.

simon-ess:
It would be nice if this were configurable - I take @anjohnson 's point that it is not _necessarily_ a desireable outcome, but there are times where it might be a good outcome.

anjohnson:
> WBNIF configurable

In some ways that feels like the `EPICS_ABORT_ON_ASSERT` environment parameter that was added not too long ago. I guess it's not quite the same though, and we probably wouldn't want a separate parameter for each yacc parser either.

I'm not saying that all our parsers currently do the most appropriate thing on error, but I don't agree with a blanket statement that they should all error out immediately.

mdavidsaver:
> C & C++ compilers tend to continue after reporting the first error

[...]

PR #960, opened 2026-09-05
dbChannel sanity check

mdavidsaver:
Apply some sanity checks for newly parsed `dbFldDes` (contents of `.dbd` file), and newly created `dbChannel` (taking into `rset::cvt_dbaddr()`).

Statements which I think are always true...

* Excluding `DBF_NOACCESS`, any `dbChannel` must have non-zero field_size _and_ max. element count.
* A scalar `DBF_STRING` (max. element count of 1) must have field_size sufficient for a null and one other character.
* For an array `DBF_STRING` (max. element count > 1) must have field_size exactly 40.
* For all field types except NOACCESS and STRING, field_size must be exactly as expected (as defined by `dbValueSize()`).

Several existing "error" check in the .db parser are made fatal, and one is extended.  I see these as "semantic" errors, to be signaled by the `yyerror()` (but not immediately `yyerrorAbort()`).  cf. https://github.com/epics-base/epics-base/issues/939#issuecomment-5260070530

The errors in `dbChannelCreate()` and friends are signaled by a new error code `S_dbLib_badDbAddr`.

The iocsh `dbgf` and friends are modified to print a more specific message than "not found".  This looks like:

```
    epics> dbgf TEST:SCO.PAQ
    PV 'TEST:SCO.PAQ' Field Not Found (0x2000009)
    epics> dbgf TEST:SC
    PV 'TEST:SC' Record Not Found (0x2000005)
```

### DIFF
--- a/modules/database/src/ioc/db/dbAccess.c
+++ b/modules/database/src/ioc/db/dbAccess.c
@@ -627,6 +627,7 @@ all_done:
 
 long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
 {
+    long ret = 0;
     dbFldDes *pflddes = pdbentry->pflddes;
     short dbfType = pflddes->field_type;
 
@@ -644,10 +645,34 @@ long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
 
         /* Let record type modify paddr */
         if (prset && prset->cvt_dbaddr) {
-            return prset->cvt_dbaddr(paddr);
+            ret = prset->cvt_dbaddr(paddr);
+        }
+        long size = dbValueSize(paddr->field_type);
+        if(ret) {
+            // oops
+
+        } else if(paddr->field_type == DBF_NOACCESS) {
+            // not directly usable, so anything goes...
+
+        } else if(paddr->no_elements == 0 || paddr->field_size == 0) {
+            // max. element count, or element stroage, of zero is unusable
+            ret = S_dbLib_badDbAddr;
+
+        } else if(paddr->field_type == DBF_STRING
+                  && (paddr->field_size <= 1
+                      || (paddr->no_elements > 1
+                          && paddr->field_size != MAX_STRING_SIZE)))
+        {
+            // A useful string needs at least 2 bytes of storage.
+            // Array of strings must provide exactly 40 bytes per element.
+            ret = S_dbLib_badDbAddr;
+
+        } else if(paddr->field_type > DBF_STRING && paddr->field_size != size) {
+            // for primative types other than STRING, provided element storage must be exact
+            ret = S_dbLib_badDbAddr;
         }
     }
-    return 0;
+    return ret;
 }
 
 /*
--- a/modules/database/src/ioc/db/dbTest.c
+++ b/modules/database/src/ioc/db/dbTest.c
@@ -22,6 +22,7 @@
 #include "epicsStdlib.h"
 #include "epicsString.h"
 #include "errlog.h"
+#include "errSymTbl.h"
 
 #include "callback.h"
 #include "dbAccessDefs.h"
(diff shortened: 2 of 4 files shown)


modules/database/src/ioc/db/dbTest.c
```diff
@@ -789,7 +790,7 @@ static long nameToAddr(const char *pname, DBADDR *paddr)
     long status = dbNameToAddr(pname, paddr);
 
     if (status) {
-        printf("PV '%s' not found\n", pname);
+        printf("PV '%s' %s (0x%lx)\n", pname, errSymMsg(status), status);
```
anjohnson, line 793: Please use `errSymLookup()` instead here, it handles showing the numeric status if no message can be found (but you do have to provide a buffer for it).

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -646,6 +646,27 @@ long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
         if (prset && prset->cvt_dbaddr) {
             return prset->cvt_dbaddr(paddr);
```
anjohnson: Would this be a good opportunity to check what `prset->cvt_dbaddr(paddr)` returned too? Not sure if doing that could have caught the scalcRecord code though.

anjohnson (commented):
Seems worthwhile and reasonable, a couple of minor details here.

modules/database/src/ioc/db/dbTest.c
mdavidsaver: > (but you do have to provide a buffer for it).

How should this buffer be sized?  Can we do better than a magic number like `128` or `256` bytes?

Looking at the generated `errSymTbl.c` shows that the longest is `S_dev_multDevice` at 63 bytes including the null.

modules/database/src/ioc/db/dbAccess.c
mdavidsaver: Good point.

modules/database/src/ioc/db/dbAccess.c
```diff
@@ -644,10 +645,34 @@ long dbEntryToAddr(const DBENTRY *pdbentry, DBADDR *paddr)
...
+                      || (paddr->no_elements > 1
+                          && paddr->field_size != MAX_STRING_SIZE)))
+        {
+            // A useful string needs at least 2 bytes of storage.
+            // Array of strings must provide exactly 40 bytes per element.
+            ret = S_dbLib_badDbAddr;
+
+        } else if(paddr->field_type > DBF_STRING && paddr->field_size != size) {
+            // for primative types other than STRING, provided element storage must be exact
+            ret = S_dbLib_badDbAddr;
```
lucmaga: This are all different logic issues that return the same error code. It would be better to improve, how the user knows what is the actual problem.

modules/database/src/ioc/db/dbTest.c
anjohnson: In `modules/libcom/src/error/errSymTbl.h` add
```C
#define ERRSYM_BUFLEN 80
```
I wondered about eventually turning that into a function which returns the size needed for the largest currently-registered symbol, but there is controversy about using VLAs in C and we'd need to talk about `__STDC_NO_VLA__`, so maybe not.

Another later project could have the routine `errSymbolAdd()` warn about any symbols that will be truncated when printed (accounting for the number if/when we add that to the normal output). They shouldn't be rejected just for length though.
The `makeStatTbl.pl` script should also warn about long messages at build time, but the runtime warning is still useful since external code can register its own status messages too.

Both `errSymDump()` and `errSymTestPrint(long errNum)` could be candidates for new iocsh commands, they were occasionally useful on VxWorks.

dirk-zimoch: I looked into changing `errSymMsg()` to always print the error number, as we discussed. Because this function returns a static string, this can only be done for existing messages (by adding the number to the message string before storing it in the lookup table). But it cannot be done in a re-entrant way if the code is unknown. Thus, I won't create a Codeathon issue for it.
`errSymLookup()` on the other hand could easily be modified to always print the error code in addition to the message. The only question is how to deal with too short buffers? Rather truncate the message or the error code?
Alas, everything would be much easier returning a C++ `std::string`. Add a third interface for C++? (That won't fix the two existing interfaces, though.)
Add an interfaces that directly prints into a `FILE*` instead of dealing with buffers? (Won't help much when printing with `errlogPrintf`.)

Issue #962, opened 2026-09-10
mbbiDirect and mbboDirect records incorrectly refer to read_mbbi and write_mbbo functions


pheest:
In  mbbiDirectRecord.dbd.pod and mbboDirectRecord.dbd.pod the function calls should refer to read_mbbiDirect and write_mbboDirect respectively.
This is a copy'n'paste mistake from mbbiRecord and mbboRecord.

It means that when the Direct record types are used, applications will actually return their 'normal' counterparts.
I have corrected the issue locally for my use needs.

anjohnson:
Hi Peter, I agree that the names used could be confusing and probably should be changed, but there is no chance that the code has actually been calling the wrong routines.

The name `read_mbbi` declared in the mbbiDirectRecord.dbd.pod file is a function pointer member of the `struct mbbidirectdset`. Each device support implements their own routines as static functions and puts pointers to them into an instance of that structure called its DSET (device support entry table). The record type calls the appropriate device routines through those function pointers – the function name isn't even visible outside of the device support implementation.

pheest:
Hi Andrew,

I can assure you that the code **was** calling the wrong routine in my tests (on Windows and using VS2022 with DLL build) that were using the record.

I detected this because I was seeking to use and monitor the B<n> fields with caget/camonitor and got no-go.
I did instead get e.g. ZRVL from the mbbi record type.

Cheers, Peter.

pheest:
It is because devMbbiDirectSoftCallback.c has already the name:
    static long read_mbbiDirect(mbbiDirectRecord *prec)

ralphlange:
Static names are in file scope.
I don't see how they could ever be called by name from outside the same source file.
Each Device Support puts them in its record type specific jump table. Their (local) static names are not relevant.

A screwed-up correlation between Device Support name and jump table name (i.e., a wrong DBD file) could lead to using a wrong jump table. But single routines?!

anjohnson:
> I was seeking to use and monitor the B fields with caget/camonitor and got no-go.

Please be explicit, what command were you actually running there, and what do you mean by "got no-go"? What error messages were you seeing (copy/paste is best here if you can reproduce that).

pheest:
Hi Andrew & Ralph.
>> what do you mean by "got no-go"
I was using caget and cainfo to query the field(s) while the IOC was running.
[...]

PR #963, opened 2026-09-10
Corrected copy'paste mistake in mbbiDirectRecord.dbd.pod and mbboDire…

pheest:
…ctRecord.dbd.pod which refer to the 'normal' mbbi and mbbo access routines.

This addresses https://github.com/epics-base/epics-base/issues/962


simon-ess:
As discussed in #962 this is not _strictly_ necessary, but it is still an improvement. LGTM.

PR #974, opened 2026-09-21
add lsi/lso mention in stringin/stringout documentation

iarimbola-pro:
Hello,

Following this [issue](https://github.com/epics-docs/epics-docs/issues/199), I added a mention towards lsi and lso inside the string input and string output documentation. 

Thank you for reviewing this pull request.


modules/database/src/std/rec/stringinRecord.dbd.pod
```diff
@@ -10,7 +10,7 @@
 =title String Input Record (stringin)
 
 The string input record retrieves an arbitrary ASCII string of up to 40
-characters. Several device support routines are available, all of which are soft
+characters (which can be bypassed by using L<Long String Input|lsiRecord/long-string-input-record-lsi>). Several device support routines are available, all of which are soft
```
anjohnson, line 13: Long lines like these are hard to read in the sources.

modules/database/src/std/rec/stringoutRecord.dbd.pod
```diff
@@ -10,7 +10,7 @@
 =title String Output Record (stringout)
 
 The stringout record is used to write an arbitrary ASCII string of up to 40
-characters to other records or software variables.
+characters(which can be bypassed by using L<Long String Output|lsoRecord/long-string-output-record-lso>) to other records or software variables.
```
anjohnson, line 13: Ditto.

anjohnson (changes requested):
Please word-wrap these links, both `L<…>` entries should have a newline before and after them (but not within).

Thanks!

modules/database/src/std/rec/stringoutRecord.dbd.pod
anjohnson: This also needs a space before the open-parenthesis `(`.

ralphlange:
Please rather rebase and force-push Pull Requests instead of merging back the target branch. GitHub's web interface can deal with rebases just fine, and the branch history stays a lot cleaner and more linear that way.

PR #976, opened 2026-09-22
Fix tiny typo

simon-ess:
This is not an interesting PR.


lucmaga:
LGTM

mdavidsaver:
As tiny and as uninteresting as it may be, I would like to see any PR remain open for at least 24 hours to allow all time zones an opportunity to comment.

simon-ess:
> As tiny and as uninteresting as it may be, I would like to see any PR remain open for at least 24 hours to allow all time zones an opportunity to comment.

Ah, fair enough.

